PR #92, opened 2017-10-08
Move target identification code to separate Makefile

g-oikonomou:
For some cross-platform examples, we need to tweak things based on the target platform. For this reason, target identification needs to happen early in the build process, before inclusion of the top-level `Makefile.include`.

Up to now, we had simply been replicating the target-identification code in example Makefiles, leading to code-replication, inconsistency, and on occasion bugs (e.g. ignoring `Makefile.target`).

This pull proposes moving target identification to a separate top-level `Makefile.identify-target`. Example Makefiles can include this "as a library" (e.g. the way the sensniff Makefile changes this pull). The top-level `Makefile.include` can just include this file now.

In the longer term, this will support easier creation of cross-platform examples (border routers etc).

Just a discussion for the time being.


g-oikonomou:
Closing for now

g-oikonomou:
Resumed in #122

PR #99, opened 2017-10-14
One main to rule them all...

g-oikonomou:
This pull implements a common main() routine provided by the Contiki-NG core. The routine is modular, so that platform maintainers can hook-in platform-specific hardware initialisation.

Currently, it is impossible for a platform to override this with a platform-specific main. I am not sure whether this is something we want to support. I am inclined to say "no" at this stage. This also means that before all platforms have been ported, this will probably not even compile. 

To answer the questions in #6: Yes, this will work for tickless platforms. Three stages should be OK for all platforms that I am familiar with, but nothing stopping us from adding more if it turns out it's required. I also need a closer look into NAT64 platforms and the Cooja platform.

I will adapt all platforms as soon as I can, but help for platforms that you folks are familiar with would be appreciated.

Obviously this will need testing on actual hardware:

* [x] CC13xx/CC26xx: @g-oikonomou
* [x] Zoul Remote: @simonduq
* [ ] Other CC2538: @g-oikonomou, @joakimeriksson ?
* [x] Sky: ?
* [ ] Nrf52dk: ?
* [x] JN516x: @simonduq
* [x] Cooja: @simonduq if Travis is happy then this works
* [x] Native: ?


atiselsts:
Hi George,
nice, I tested `hello-world` on Zolertia Firefly and NRF52DK. I see some discussions here about network testing - what that would imply for NRF52, and do I need a second NRF52 for that, or can it talk with some other Contiki hardware? (It uses the `ble_ipsp_mac_driver`)

P.S. Flashing NRF52DK seems to broken at the moment, as `erase.jlink` and `flash.jlink`are missing in Contiki-NG - were they removed on purpose?

g-oikonomou:
Thanks Atis, any network test would be enough, just need to make sure the stack is initialised properly.

I'll let someone else comment on the jlink question.

simonduq:
what are these? targets, tools? if removed, was not intentional!

atiselsts:
Simon, they're files. See here: https://github.com/contiki-os/contiki/tree/master/cpu/nrf52832
In Contiki-NG should be under `arch/cpu/nrf52832/flash.jlink` - `make erase` amd `make XXX.flash` are looking for them there.

simonduq:
Could you add them?

simonduq:
Please check https://github.com/contiki-ng/contiki-ng/pull/124
Other than that are we doing more tests or merging now? I'm leaning towards merging because we want it in anyway, and the sooner the more testing it gets pre-release.

atiselsts:
Thanks Simon - was just about to add the files today, but you were faster.

g-oikonomou:
Tested thoroughly in a testbed combining zoul/CC1350LP/CC2650LP and fixed a couple of things in the CC26xx startup sequence. Overall I'm pretty happy with it.

simonduq:
Let's git this in, thanks a billion!

PR #121, opened 2017-10-28
Fix RPL configuration under different multicast scenarios

g-oikonomou:
In the current master the SMRF and ESMRF engines are broken. `uip-mcast6.h` is used to auto-configure RPL MOP depending on engine. Currently, `rpl-conf.h` does not include this header and, as a result, under SMRF and ESMRF we still operate in MOP2. Both engines require MOP3.

This pull adds the missing include and also adds two new travis tests: Now we test all three engines.


simonduq:
Great 👍

PR #122, opened 2017-10-28
 Move target identification code to separate Makefile

g-oikonomou:
For some cross-platform examples, we need to tweak things based on the target platform. For this reason, target identification needs to happen early in the build process, before inclusion of the top-level `Makefile.include`.

Up to now, we had simply been replicating the target-identification code in example Makefiles, leading to code-replication, inconsistency, and on occasion bugs (e.g. ignoring Makefile.target or environment variables).

This pull proposes moving target identification to a separate top-level `Makefile.identify-target`. Example Makefiles can include this early on "as a library" (e.g. the way the sensniff Makefile changes in this pull). If the example Makefile does not need target identification, the top-level `Makefile.include` will always include `Makefile.identify-target`.

In the longer term, this will support easier creation of cross-platform examples that feature a large platform-independent code base and that can be easily extended with platform-dependent specifics.

This is now ready for consideration.

This is a sequel to the discussion in #92

### DIFF
--- a/Makefile.identify-target
+++ b/Makefile.identify-target
@@ -1,0 +1,9 @@
+ifeq ($(TARGET),)
+  -include Makefile.target
+  ifeq ($(TARGET),)
+    ${info TARGET not defined, using target 'native'}
+    TARGET=native
+  else
+    ${info using saved target '$(TARGET)'}
+  endif
+endif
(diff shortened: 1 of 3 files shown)


simonduq:
This is nice. There seem to be a couple more example that could use this (at least `coap-example`, `json-ws`,  `slip-radio`). Maybe they should also be covered before merging?

simonduq:
Giving it 👍 regardless, as this enables https://github.com/contiki-ng/contiki-ng/pull/123

g-oikonomou:
CoAP example probably, yes. Slip-radio I am already working on and will be in a separate pull. JSON-WS is history unless I hear otherwise :)

PR #124, opened 2017-10-29
nrf52832: adding missing jlink files

simonduq:
This adds two files that were removed unintentionally when moving the `cpu` folder to `arch/cpu`, due to `*.jlink` being in `.gitignore`.


atiselsts:
Oops, looks like this one was accidentally dropped somewhere along the way? The current develop branch is still missing the `.jlink` files.

PR #123, opened 2017-10-28
Tidy-up the examples directory

g-oikonomou:
This pull starts a tidy-up of the examples directory.

* Some obsolete examples and associated tests have been removed
  * rpl-simple: As discussed the other day, no longer serves a purpose
  * json-ws: Was not building for any of the platforms I tried
  * extended-rf-api: No longer serves a purpose
* All examples under ipv6/ have been moved to the top-level `examples/` dir. Associated tests and simulations updated.
* New platform-independent example: UDP echo server (was under cc2538dk with no good reason)
* The mess of a webserver config for native-border-router and rpl-border-router has been improved.

Sits on top of #121

This is still WiP, but I thought I'd open to discussion. Still to come:
* [ ] Cross-platform MQTT client example (will need #122)
* [x] Fixes for better cross-platform support in slip-radio (done in #131 and #133) and rpl-border-router (done in #132) (will need #122). 
* [x] Finalise to rpl-udp and udp-ipv6
* [ ] rpl-tsch-sixtop: Change .csc to use Cooja motes instead of Z1s. Could someone help @nfi @joakimeriksson @simonduq?

PR #131, opened 2017-10-29
Tidy-up slip_arch_init and all its side-effects

g-oikonomou:
In order to initialise SLIP, the border-router and slip-radio examples call `slip_arch_init`. 

Theory: This function expects the baudrate as argument, only it expects it in the shape of a platform-specific magic number. To get this number, platforms had to provide a `BAUD2UBR` macro which performs the conversion. For whatever reason, the examples assumed that this macro was defined in a file called `uart1.h` (probably copied over from the early ports where it actually was). As a result of that, in order to not touch the example code, many many ports ended up providing a stub `uart1.h` merely in order for the examples to build, but which otherwise does nothing other than `#define BAUD2UBR(b) (b)`.
 
Practice: All of our supported platforms ignore the argument to `slip_arch_init`. This means we don't need to do the baud-rate-to-magic-number conversion, which means we don't need to define the `BAUD2UBR(b)` macro, which means examples don't need to force-include `uart1.h`, which means we don't need to provide a stub for it!

So, this pull:

* Changes the prototype of `slip_arch_init` to no longer accept an argument
* Changes all its invocations to no longer pass an argument
* Removes all definitions of `BAUD2UBR()`, since it's no longer need as an argument.
* Changes slip-radio and border-router to no longer force-include `uart1.h`, since it's no longer needed for the definition of the macro...
* Deletes all those stub headers, since they no longer need included...

Sits on top of #123


g-oikonomou:
From my PoV this is ready for merging as soon as #123 has been approved

simonduq:
Hey this looks good, very cool PR description :) The diff shows trillions of commits though, not sure what à merge would look like. Better rebase?

g-oikonomou:
All nicely rebased and ready to merge from my PoV

PR #132, opened 2017-10-30
Cross-platform and tidied-up border-router

g-oikonomou:
This pull changes the border-router example:

* The example starts by calling `NETSTACK_MAC.off();`. This is done so the BR does not join a separate RPL network while it's waiting for a prefix. This call causes problems with the CC26xx IEEE RF driver, which turns RX off very aggressively when it can. Once a prefix has been received, we now call `NETSTACK_MAC.on();` and the example works for this platform just fine.
* The example now uses the logging module
* Code style fixes

Sits on top of #131


g-oikonomou:
There are multiple, platform-specific border-router examples for jn516x. I tried to borrow some ideas from them, but they are not even identical between themselves. I cannot determine whether they actually need to be different, any advice welcome.

I don't think we should block waiting till they have been tidied up, but I think we should be merging them into here in the longer term.

simonduq:
They have nothing very specific. Were separate because of different UART (I believe you fix this) and because of TSCH (but not we can already enable TSCH with just one line in makefile).

Can be basically removed, standard BR should work just as well now.

joakimeriksson:
This sounds great - the fewer specific examples the better!

simonduq:
If we move both BRs inside a single directory, do you think we could easily have a common makefile, such as native used the NBR code and any other platform uses the embedded version?

g-oikonomou:
Possibly, I'll have a look as part of #132

simonduq:
yeah which is the PR this comment is already in ;);)

simonduq:
Good. Comments:
* How about, instead of the left-over `printf`, use `LOG_INFO` and set the log level to `LOG_LEVEL_INFO` by default? This way, printouts looks consistent (the prefix thing)
* Which platforms was this tested on?

g-oikonomou:
As discussed in #132 ;) I have updated this pull.

We now have a single `rpl-border-router` example dir for all BR flavours. We invoke the build from the top-level dir. If TARGET is native, we build what was previously our `native-border-router`. Otherwise we get what was previously our `rpl-border-router`.

Code duplication reduced by a fair bit (for example all flavours use the same `httpd-simple` code, which was previously 99.9% identical across the different examples).

Some travis tests added to build this for all platforms and I have also addressed your comment on using logging.

simonduq:
Great. Already tried on zoul and with tutorial updated https://github.com/contiki-ng/contiki-ng/wiki/Tutorial:-RPL-border-router

Thanks!

PR #133, opened 2017-10-30
Cross-platform slip-radio example

g-oikonomou:
This pull applies a number of changes to slip radio example. This pull:

* Changes to example to use early target identification from #122 and, based on that, allows platform-specific extensions/configuration
* Moves sky-specific extensions/configuration to a subdir. Some limiting configuration defines now only apply to sky builds (e.g. limiting the uIP buffer size). Some extensions (e.g. custom commands) are built conditionally.
* Removes some code artifacts that have to do with legacy radios/platforms (e.g. mc13xx)
* Changes some functions to static
* Changes the example to use the logging module
* Adds some compile tests for the example to travis
* Fixes code style

Sits on top of #131, requires #122

PR #135, opened 2017-10-30
Remove obsolete SLIP_foo_CONF_NO_PUTCHAR

g-oikonomou:
slip-radio and rpl-border-router have this example-specific putchar which is used to do mux/demux when SLIP is used over the same interface as debug char i/o. As far as I can tell, among our current platforms only sky uses it, so I am moving it to the platform-specific part of the respective examples in #132 and #133. Once those two have been merged, those macros will be obsolete, and this pull removes them.

In terms of the diff, this pull here is orthogonal to the two mentioned above, however there is no point merging this one before merging the above two.

### DIFF
--- a/arch/platform/cc2538dk/contiki-conf.h
+++ b/arch/platform/cc2538dk/contiki-conf.h
@@ -178,10 +178,6 @@ typedef uint32_t rtimer_clock_t;
                                            the uart1_* API */
 #endif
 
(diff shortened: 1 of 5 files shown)


simonduq:
rebase needed here

g-oikonomou:
Rebased now, this should be OK

PR #138, opened 2017-10-31
simplified the timer example

joakimeriksson:
The all-timers example was a bit complex as it used multiple processes even if rtimer and ctimer do typically not need processes.


g-oikonomou:
This is part of #123 now. Closing

simonduq:
Looks good! :)

Issue #108, opened 2017-10-20
Energest: add deep sleep mode


atiselsts:
Just an idea for discussion: we have `ENERGEST_TYPE_CPU` and  `ENERGEST_TYPE_LPM` already. Should add something like `ENERGEST_TYPE_DEEP_LPM`. 

For example, on our CC2650 board, the average current values are:
- CPU active: 2.7 mA
- Sleep: 1.3 mA
- Deep sleep: ~0.01 mA (~10 uA)

The difference between deep sleep and sleep is two orders of magnitude. Surely that deserves a different energest class?

g-oikonomou:
Definitely a desirable feature.

PR #139, opened 2017-10-31
New energest type for deep sleep plus configurable type for Energest time source

nfi:
* Adds new energest type `ENERGEST_TYPE_DEEP_LPM` as discussed in #108.

* Adds function to get the total energest time: `ENERGEST_GET_TOTAL_TIME()`. This is essentially the sum of the times in different CPU modes (CPU + LPM + DEEP_LPM).

* Adds configurable type `ENERGEST_TIME_T` to support for clocks with wrapping time as time source for energest.

* Removes include of `energest.h` from `contiki.h` to avoid circular dependencies.

* Fixes some typos.


simonduq:
👍 if Travis is also happy ;)

PR #145, opened 2017-11-01
Border-router: move border-router.c to top level for cleaner compilation

simonduq:
This has two benefits:
* enables compiling the BR in Cooja -- Cooja requires a `.c` project file sitting in the same dir as the makefile and the firmware
* makes for cleaner compilation, now with all artifacts in the common dir, no more need to `cp`


g-oikonomou:
I see what you're proposing here, I like this 👍 Let's get this in when travis passes.

PR #147, opened 2017-11-02
New Energest example

nfi:
* Added an example for Energest
* Added the energest example to regression compile tests
* Show Energest configuration in `viewconf`


g-oikonomou:
Useful addition 👍

PR #148, opened 2017-11-02
Platform native: Avoid trying to use the tun when the tun is not available.

nfi:
This allows a native platform application to run without network.


g-oikonomou:
> This PR should solve this issue.

"Should"? :D

nfi:
>> This PR should solve this issue.
>
> "Should"? :D

"will" ;-)

simonduq:
good catch!

simonduq:
also tried on my linux, both with and without networking

PR #149, opened 2017-11-02
Added webhook for Gitter notifications on Travis jobs

simonduq:
don't merge, will need to change the hook

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,20 @@
 # See https://github.com/travis-ci/travis-ci/issues/6928#issuecomment-264227708
 group: deprecated
 
+# Webhook for Gitter notifications
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


simonduq:
don't merge, will need to change the hook

simonduq:
now with new hook, let's see if it works as intended

simonduq:
Good, this works. Ready from my PoV

simonduq:
Note this only notifies "Developers". If we also want to notify "Lobby" then we need to create another hook and add the new URL to `.travis.ml`.  Might play with this

simonduq:
OK, I don't actually I like this -- we get notifications whenever this `.travis.ml`, that is, potentially for every push to other github forks etc. Closing for now

PR #151, opened 2017-11-02
Renode example in rpl-udp demo

PiotrZierhoffer:
This PR adds Renode scripts and integration via Makefile, as discussed with Joakim.

This is specific to rpl-udp example on CC2538DK, but may be used for other binaries and boards as well, with little adjustments.

To run Renode in this setup run `make renode TARGET=cc2538dk`.


simonduq:
Thanks a lot!

PR #152, opened 2017-11-03
Update submodule Cooja to latest version

nfi:
Improves the warning message when MSPSim is not found to suggest a recursive submodule update. A non-recursive submodule update is not sufficient when Cooja itself is used as a submodule.


joakimeriksson:
👍

simonduq:
do we also pull this on branch release-4.0 ?

PR #154, opened 2017-11-04
Gitignore doxygen output files in their new location

g-oikonomou:
We moved doxygen documentation, so build artifacts are no longer being ignored. This commit sets .gitignore all documentation output in its new location.

This should IMHO be included in this release.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,8 +12,7 @@
 *~
 obj_*
 Makefile.target
-doc/html
-doc/latex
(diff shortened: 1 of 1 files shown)


simonduq:
I believe the latex is no longer build by the doxygen makefile. If that's indeed the case, we should probably only keep only the html rule

g-oikonomou:
Updated

simonduq:
Thanks!

PR #155, opened 2017-11-05
sixtop: bugfix adding a missing 'static' modifier

yatch:
I found a bug in `example/6tisch/sixtop/node-sixtop.c`, which will be resolved by this commit.


simonduq:
thanks!

PR #160, opened 2017-11-07
Avoid tsch crash when losing timesource

e120guru:
This PR resolves a particularly annoying bug that happens under some occations, for instance a lost sync just before trying to send KA. (This has been seen in cooja on repeated occasions)

In cooja this results in segfault, and on taret it results in data abort exception.

The reason is the fail-to-check-for-NULL-pointer in the code that sends the KA. I don't know if there might be a race condition since this apparently happens when tsch_is_associated is set, but there still is no timesouce. However, since the tsch_queue_get_time_source explicitly returns NULL in some cases, the calling code should check for NULL.

### DIFF
--- a/os/net/mac/tsch/tsch.c
+++ b/os/net/mac/tsch/tsch.c
@@ -313,7 +313,7 @@ keepalive_send(void *ptr)
         LOG_INFO_LLADDR(&n->addr);
         LOG_INFO_("\n");
     } else {
-        LOG_INFO("no timesource - KA not sent\n");
+        LOG_ERR("no timesource - KA not sent\n");
     }
   }
 }
--- a/os/net/mac/tsch/tsch.c
+++ b/os/net/mac/tsch/tsch.c
@@ -304,13 +304,17 @@ keepalive_send(void *ptr)
 {
   if(tsch_is_associated) {
     struct tsch_neighbor *n = tsch_queue_get_time_source();
-    /* Simply send an empty packet */
-    packetbuf_clear();
-    packetbuf_set_addr(PACKETBUF_ADDR_RECEIVER, &n->addr);
-    NETSTACK_MAC.send(keepalive_packet_sent, NULL);
-    LOG_INFO("sending KA to ");
-    LOG_INFO_LLADDR(&n->addr);
-    LOG_INFO_("\n");
+    if(n != NULL) {
+        /* Simply send an empty packet */
+        packetbuf_clear();
+        packetbuf_set_addr(PACKETBUF_ADDR_RECEIVER, &n->addr);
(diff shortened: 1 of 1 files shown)


yatch:
@e120guru Firstly, I agree with you, need to check the return value from `tsch_queue_get_time_source` before accessing it, `&n->addr`.

Secondly, I also think there might be a race condition... Although I may miss something, the timer for `keepalive_send()` is stopped when the node loses its time source, at [tsch-queue.c:L167](https://github.com/contiki-ng/contiki-ng/blob/0adc6bfbad6342073eeddc07f2ca50a674f5c8ec/os/net/mac/tsch/tsch-queue.c#L167). Then, it'd be better to replace `LOG_INFO` with `LOG_ERR` in the following line:

```C
LOG_INFO("no timesource - KA not sent\n");
```

I'm trying to reproduce the issue with a simple topology in which there are only one coordinator and one node; this doesn't work well so far...

Anyway, that's a great catch 😄

e120guru:
We first run into it when porting to a 1Mbps radio platform that changed the slot timings, and when the coojamote had not yet been updated to match that, we started so see click drifts, and then it started happening every now and then (not always though).

Maybe there are other reasons than a mote just have loosed sync that it might want to send a KA and this function returns NULL...? 

Anyhow, backtracking from the callstack after Cooja crashed it could be seen that it dereferenced the null pointer here...
We fixed it and got rid of the stability issue (that did happen on one or two occasions AFK as well on target) - well aware of that we might just hid the root cause.


Also, in regards to this PR - I was a bit tired yesterday, I now see that I made the slightly ugly 
If not X then Y else Z, construct instead of If X then Z else Y.
What does the Contiki code style say about this?

simonduq:
Thanks a lot for the bug report and fix! I agree with @yatch on switching from `LOG_INFO` to `LOG_ERR`.

Time-source handling was made more robust in NG -- did you witness the problem on Contiki-OS or Contiki-NG? Even if not in NG I like the fix as defensive programming.

As for code style, we don't have a rule regarding orders of the `if` blocks.

simonduq:
For instance the `ctimer_stop` @yatch is talking about was missing before NG.

e120guru:
We've seen it in Contiki-OS, we have not yet brought our code base up to NG yet...

Hopefully we will be able to do it very soon though... I just cought that the check-for-null was missing in NG too, even though perhaps the root cause might be resolved.... I'll update the PR to use LOG_ERR

simonduq:
Let me try to fix it directly for you using the github interface.

simonduq:
Seems to work. Will merge if Travis is green! :)

e120guru:
Thanks

PR #165, opened 2017-11-08
Rework CI test suite

simonduq:
Update: this PR also incorporates the full rework of the CI suite, which was initially part of https://github.com/contiki-ng/contiki-ng/pull/183
Simplifies and makes things more consistent, with more readable test output.

This PR adds the ability to run test simulations with multiple seeds, to increase coverage. The previous version also had the notion of multiple seeds but the test would stop at the first success, i.e., one success out of N was considered a success. With this modification, one can choose to run N seeds, and success will reported only if all seeds succeed.

This is intended for:
* possible coverage increase in the future (but costs Travis time)
* offline testing with increased coverage when preparing releases, e.g., run all tests on 100 seeds and make sure all is green


g-oikonomou:
We want this in. Shall we rebase?

simonduq:
rebased!

PR #166, opened 2017-11-08
Make RPL random rearrangement tests non-stochastic

simonduq:
The `07-rpl-random-rearrangement.csc` simulations current use `Math.random` in javascript, providing no control on the seed. This PR modifies the scripts to use a simple random generator instead, allowing to set the seed. The seed is set to the Cooja random seed, i.e. it can be configured freely when calling Cooja.

This makes the simulation deterministic, and should avoid cases where a test fails, is re-run and succeeds. As far as I can tell, the whole test suite should now be deterministic.

For increased test coverage of this test and all others, see https://github.com/contiki-ng/contiki-ng/pull/165


joakimeriksson:
Looks good - but what stops us from just picking the random number from the simulator itself? (e.g. Cooja)? Do not remember, but possibly sim.getRandomGenerator().nextRandom() or something like that?

simonduq:
Interesting, did not realize this was an option. Will try this out!

simonduq:
On a related note: when using Cooja motes, do we have a guarantee that Cooja consumes the random numbers deterministically? As motes are processes, cannot the execution lead to different orders in which, e..g, a node transmits and Cooja must cast a dice to decide if the Tx goes through?

joakimeriksson:
They should be consumed in the same order all the time and as long as all code is using the right generators it should be deterministic.

simonduq:
After more discussion with @joakimeriksson we do not have 100% certainty that Cooja motes result in deterministic execution (unlike MSPSim motes). It could be the case that the execution speed of processes affects the result.

Until we get full certainty, I propose to keep this PR as it is. This way, the network topology rearrangement, which have a major outcome on the result, remain deterministic.

joakimeriksson:
Yes, agree!

simonduq:
OK, based on proposal by @nfi , this is a simpler fix that uses the Java random number generator. The script creates its own instance so as not to be affected by the rest of the simulation.

PR #167, opened 2017-11-09
Updated CoAP and LWM2M with added DTLS support

nfi:
This is an implementation of CoAP and OMA LWM2M that has been modified to be less dependent on specific Contiki features. This means that there are significant changes in both CoAP and LWM2M (some processes removed and APIs changed). There is also a completely new example that shows how to make use of the LWM2M implementation in a standalone mode on top of Posix API's. These are the main new features:

* Added refactored version of tinyDTLS as submodule and support for DTLS in LWM2M and CoAP
* Improved LWM2M so that it supports block transfers, readout of complete objects both as TLV and JSON, added support for dynamic LWM2M object instances, and many more fixes
* Improved RD client in LWM2M that supports bootstrap and registration with both coap and secure coap. 
* Refactored CoAP to enable use outside Contiki, and fixed several bugs.

This is currently a work-in-progress PR so it is meant to be commented on but please do not pull it immediately. Most of the work is based on https://github.com/sics-iot/lwm2m-contiki where more information can be found on the wiki.


simonduq:
This is really a major contrib, much awaited, looking forward to having it in!!
What would you think of splitting it in two for easier review? One for CoAP(s) and one for LWM2M?

Also, I believe we need the following before merging:
* doc for CoAP(s) and LWM2M (maybe what is under `examples/coap` is a good starting point?)
* tutorial for CoAP(s) and LWM2M
* CI test CoAP(s) and LWM2M

Doxygen comments would also have been nice ;)

joakimeriksson:
Splitting is not so easy as we have developed them closely together in a separate Contiki fork for more than a year by now which means that it is a significant effort to port all old LWM2M code if we bring only CoAP in. Or completely remove LWM2M support for a while. Not sure if that is wanted?

Documentation will be moved into the wiki and we can add more readme documentation in examples/coap and examples/lwm2m directories.

We will also address the other comments!

simonduq:
Right, definitely not worth splitting in two PRs then!

simonduq:
I took a closer look, more detailed comments below:
* if we chose to keep TinyDTLS as a submodule (my vote goes for not, I'd rather have it directly in the tree), then it needs to move to contiki-ng org
* I would move tinydtls (regardless of whether it is a submodule or not) into os/net
* Configuration: now `makefile.coap` deals with `MAKE_WITH_DTLS` for inclusion of DTLS modules. How about having it fully module-based, i.e. just add the DTLS module, which will configure itself via its own makefile? Somehow feels strange that CoAP is just module inclusion but DTLS is not.
* lwm2m/standalone example: why so much code in there, can't some of it move to `os`? Ideally you'd want the lwm2m example to just be hello-world with the right modules and project-conf flags and that's it, so anyone can enable it as well very easily. Also what is the standalone thing? Why does it copy code form `os`? I got confused
* Have you compiled Doxygen and checked the modules are structured as they should? (right parent module)

joakimeriksson:
The reason for code in standalone is that all that code - e.g. coap-ipv4 and coap -hex, etc are 100% Contiki-NG independent and the copying is for really showing how to bring the files out from the tree to build independent of Contiki. Therefore the name standalone. We will also provide a more clean cut IPSO-objects example that is the "hello-world" example with LWM2M (and in that case also some ipso objects). This is already in the codebase but I think we can add more documentation and possibly put it in an lwm2m floder also. To show that we have two examples (where one is the standalone - independent one - and the other is the "hello-world" style example).

simonduq:
All existing examples show simple usages of Contiki-NG. Why should this one cover things such as standalone running? What means 100% Contiki-NG independent? What does it build on instead?

joakimeriksson:
It builds a complex DTLS + CoAP + LWM2M example that runs on top of Posix IPv4 API (linux / OS-X). By doing this we allow very easy integration of these protocol implementation into other OS:es and even for developers that use no-OS at all. It also builds a complete serial implementation where the IP-stack is in Java and all the data from DTLS + CoAP + LWM2M is assumed to be over serial-link (e.g. Bluetooth, or actual serial-line).

simonduq:
OK, based on offline discussion, what we probably want then is to clearly move out the standalone examples at top-level `examples-netstack-standalone` or even to own repository. These will showcase how parts of the Contiki-NG stack can be taken out for use on other systems.

joakimeriksson:
I discussed TinyDTLS in the repo with Niclas and I am a bit concerned about the license - it is an Eclipse license so not MIT, apache or BSD that are quite similar but a bit more complex. I think I vote for a repo at contiki-ng/ but maybe with some Contiki-style makefiles in to make integration as easy as possible.

joakimeriksson:
Added a basic IPSO object / lwm2m test that runs the leshan server and register a lwm2m client to it on native platform. There are 8-CoAP requests that are run in a python-unit-test framework from within leshan and if it prints OK the test is "green".

nfi:
Rebased on top of develop to resolve conflicts. Moved tinyDTLS submodule into os/net and updated CoAP/LWM2M to use the logging module.

simonduq:
Great, progress here :) What is left from my PoV:
* move standalone example to own repo/top-level dir
* document licensing issues, probably both README and wiki and maybe even a C flag for boot printout?
* check the compiled doxygen has the intended structure

simonduq:
Great! Ready for review or still wip?

nfi:
This PR should be ready for review!

joakimeriksson:
There is now a basic tutorial on the sandbox wiki:
https://github.com/contiki-ng/wiki-sandbox/wiki/LWM2M-and-IPSO-Objects-for-Contiki-NG

simonduq:
@g-oikonomou , any thoughts?

g-oikonomou:
Hey folks, I just had a super quick look at this one. It is very unlikely I'll have time to look at it thoroughly, but then again I could not immediately find anything wrong :D I'm merging develop back into this and once travis is green I am good 👍 to merge!

PR #169, opened 2017-11-10
Add a button HAL and corresponding drivers

g-oikonomou:
This pull proposes a platform-independent button API (and corresponding example) that has the following key features:

* Provides a uniform way to determine button-related events (which are no longer a `sensors_event`).
  * On button press (`button_hal_press_event`)
  * On button release (`button_hal_release_event`)
  * On continuous button press, generated with a period of 1 second (`button_hal_periodic_event`). (I did consider making the period configurable, but I think it's over-design.).

  This allows us to do something on button press, something else on continuous press for 5 seconds, something yet different on release.

* Supports an arbitrary number of buttons.
* Handles software debounce internally, on press as well as on release.
* The PR assumes that a platform already supports the new GPIO HAL. Platform developers/maintainers only need to define their platform's buttons, using a single macro call. Something as simple as:

    ```
    BUTTON_HAL_BUTTON(key_left, "Key Left", BOARD_IOID_KEY_LEFT, \
                      GPIO_HAL_PIN_CFG_PULL_UP, BOARD_BUTTON_HAL_INDEX_KEY_LEFT, \
                      true);

   BUTTON_HAL_BUTTON(key_right, "Key Right", BOARD_IOID_KEY_RIGHT, \
                      GPIO_HAL_PIN_CFG_PULL_UP, BOARD_BUTTON_HAL_INDEX_KEY_RIGHT, \
                      true);
    ```
This design brings a number of benefits:
* We expose a ~simple~ single button API to examples
* A button is no longer a sensor and therefore not bound to the requirements of the sensors API.
* We no longer make assumptions in `os/` as to how many buttons a platform will have, nor of what the symbol names will be.
* We have a single implementation of a s/w debounce, instead of dozens of instances of it. Platform/product developers don't need to spend time understanding/implementing these things.
* Minimizes code duplication: Most current platform button drivers are simply a different platform's driver pasted across.
* All platform buttons support the same features (whereas previously some supported press duration events, some did not).

Platforms currently supported:
* All CC13xx/CC26xx
* All CC2538


simonduq:
We had a discussion on an earlier version of this branch didn't we? Was trying to find it so we can link to it and followup.

g-oikonomou:
Closing this till further notice

g-oikonomou:
I have done a major rework of this pull and I am re-opening it.

Now uses the GPIO HAL, so basically if a device supports the GPIO HAL, it automatically also supports buttons.

I am editing the OP to reflect recent changes.

Tested with Launchpad/CC2650. More testing to come. Help with testing welcome.

g-oikonomou:
After a chat with @nfi last night:

* The interrupt handler now just polls a process. Debounce management is scheduled by the process.
* Added a way to retrieve the count of buttons on a device.
* Added a way to retrieve a button by index (0, 1, ... button_hal_button_count - 1).

Tested on launchpad/cc2650 and zoul/remote-reva

nfi:
I have tested the ipso-object example on Zolertia Firefly with button CoAP notifications up to Leshan and I think everything looks good.

g-oikonomou:
Can we like merge this? :)

joakimeriksson:
Yes, let's get it in! I updated branch now - can do the merge after that.

PR #170, opened 2017-11-10
Remove obsolete sensors from the native platform

g-oikonomou:
The native platform code has a number of sensor drivers, all of which appear to be stubs and/or unused. This commit removes them.

Even though the platform doesn't have LEDs, we leave this one intact so that LED-using examples can still build.


joakimeriksson:
Looks good - can not see any need for the sensors and even less for the beeper 👍

simonduq:
👍

PR #172, opened 2017-11-10
Removing multi-threading support

simonduq:
This is a proposal to remove multi-threading. Although this is in principle a nice feature, it ended up nearly never used. We have no examples nor tutorial and no of no use case that require it. Removing it has the benefit to ease maintenance and make new ports easier to write.

Note that some MT code is left in platform Cooja, at this seems needed for some reason with `cooja_mt_yield`. Thoughts?


yatch:
Indeed `cooja_mt_yield`  is necessary to give CPU back to `cooja` so that `cooja` can advances the clock in each cooja mote. For example, `cooja_mt_yield` shuold be called within a busy wait like this: https://github.com/contiki-ng/contiki-ng/blob/196d64aa66832b8bbedd433127806fd0fd7ab7ce/os/net/mac/tsch/tsch-slot-operation.c#L213.

Anyway, the generic `mt` APIs themselves can be removed, I think.

simonduq:
Is there anything more that could be removed (as part of `mt` in cooja motes)?

yatch:
After `grep`ing in the source tree, I believe yes, there is more we can remove. For instance, I found `cooja_mt_remove()` is not called from anywhere.

simonduq:
Now removed left-over functions!

PR #176, opened 2017-11-12
RPL non-storing: ensure routing links are stored for at least their advertised lifetime

simonduq:
In `rpl_ns_periodic` routing links had their lifetime decremented first, and then expired links were removed. The decrement period is 60s. This cause expiry within [period-60s;period[. With this fix, we first remove expired links and second, decrement the lifetimes. This results in an expiry delay within [period;period+60s[


simonduq:
will do a merge and rewrite of the loop before we consider merging

simonduq:
Ready now, with @joakimeriksson 's fix

joakimeriksson:
👍

PR #179, opened 2017-11-13
Bugfix in HAL for LEDs (leds.c)

fontikos:
This pull request fixes a bug in the HAL for the LEDs (leds.c). 

The static variable `leds` did not get updated resulting to misbehaviour, such as `leds_toggle()` not working (led always stays on) and `leds_on()` turning off other leds.

The fix removes the static variable `leds` and, instead, uses `leds_arch_get()` to get the state of the leds.

### DIFF
--- a/os/dev/leds.c
+++ b/os/dev/leds.c
@@ -33,22 +33,11 @@
 #include "dev/leds.h"
 #include "sys/clock.h"
 
-static unsigned char leds;
-/*---------------------------------------------------------------------------*/
-static inline void
-show_leds(unsigned char new_leds)
-{
-
-  leds = new_leds;
-
-  leds_arch_set(new_leds);
-}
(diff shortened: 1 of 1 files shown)


atiselsts:
This fixes a real bug, which was introduced in the changes between Contiki and Contiki-NG.

However, I think `show_leds()` should be removed, and its calls replaced with calls to `leds_arch_set()`. There is no good reason for `show_leds()` function to exist in the current code anymore.

fontikos:
Yes, I was considering getting rid of that inline function as well. I'll come back with an update.

simonduq:
cool! Is this ready @fontikos ?

atiselsts:
:+1:

fontikos:
Yes, @atiselsts' suggestion is included.

PR #181, opened 2017-11-15
Use ENERGEST_TYPE_DEEP_LPM energest state for CC2538 and CC26xx

atiselsts:
Pull request #139 introduces the deep sleep enegest state, but does not actually use it. This PR adds code that tracks this new deep sleep state for CC26xx and CC2538.


g-oikonomou:
👍

Issue #106, opened 2017-10-20
Allow to set a different number of MAC retransmissions for each packet


atiselsts:
Enables QoS classes at the network stack & application level. For example, monitoring traffic could be given fewer retransmissions that data collection traffic.

atiselsts:
Been working this today. This is implemented for SPHERE, but in a way that's not general enough, so I thought to redo it properly.

Current ideas are:
* add a new field to `struct uip_conn` and `struct uip_udp_conn`; encourage to set it from application code:
    `uint8_t max_mac_transmissions; /**< Number of max MAC-layer transmissions. */`
* define a new constant `UIP_DEFAULT_MAX_MAC_TRANSMISSIONS`, equal to 0. Meaning: the limit of transmissions is chosen by MAC layer instead - i.e. the current behavior. Use this whenever the application does not tell to do otherwise;
* use the Traffic Class field in IPv6 header to propagate the limit between hosts. Unfortunately, due to the way  6LoWPAN compression works, this means that having a non-zero value in `max_mac_transmissions` field will cause 6LoWPAN headers to grow by 1 byte. An alternative is to only care about the transmission limit at the originating node, and always use the default values when forwarding packets. Perhaps user-configurable choice?

simonduq:
Sounds good. Maybe for generality and for cross-technology contexts, we should not carry the #tx as traffic class, but rather use traffic classes such as 0, 1, 2, that are then mapped to link-layer policies (e.g., setting #tx to some value) at every node.

atiselsts:
Speaking about documentation, I feel this should be documented, but don't immediately see a good place in the Wiki. Suggestions?

simonduq:
Maybe the IPv6 page (which is currently almost empty with just links to some modules, but we should aim for much more content there)

Or under "programming Contiki-NG" somewhere

joakimeriksson:
Check #215 also! I do agree with the model of having traffic class 0,1,2,3,4... etc and then convert that to something matching some kind of QoS per node. But I guess it is fine for now with a bit of a "hack-ish" model (which is what I kept in the PR I did - but moved things up to the IP-layer)?

simonduq:
Wasn't this issue closed by https://github.com/contiki-ng/contiki-ng/pull/182 ?

PR #182, opened 2017-11-15
Allow to set a different number of MAC retransmissions for different classes of packets

atiselsts:
Solves #106.

This patch adds support for setting transmission limits for UDP and TCP connnections. I'm open to also doing if for ICMP and/or RPL packets, but then we need to discuss the API (some sort of `#define` for RPL, which is then used as a parameter for ICMP functions called by RPL?)

**Edit:** needs conditional compilation with `UIP_CONF_WITH_VARIABLE_RETRANSMISSIONS` set to nonzero.

Usage example for a UDP application:

```
  struct simple_udp_connection simple_udp_conn;
  simple_udp_register(&simple_udp_conn, UDP_CLIENT_PORT, NULL,
                      UDP_SERVER_PORT, udp_rx_callback);
  /* Set the MAC max transmission limit to 13 */
  uip_set_max_mac_transmissions(simple_udp_conn.udp_conn, 13);
```

Usage example for a TCP application:

```
  struct tcp_socket socket;
  tcp_socket_connect(&socket, &address, port);
  if(socket.c != NULL) {
    /* Set the MAC max transmission limit to 13 */
    uip_set_max_mac_transmissions(socket.c, 13);
  }
```


os/net/mac/csma/csma-output.c
```diff
@@ -85,9 +85,9 @@
 
 /* macMaxFrameRetries: Maximum number of re-transmissions attampts. Range 0--7 */
 #ifdef CSMA_CONF_MAX_FRAME_RETRIES
-#define CSMA_MAX_MAX_FRAME_RETRIES CSMA_CONF_MAX_FRAME_RETRIES
+#define CSMA_MAC_MAX_FRAME_RETRIES CSMA_CONF_MAX_FRAME_RETRIES
 #else
-#define CSMA_MAX_MAX_FRAME_RETRIES 7
+#define CSMA_MAC_MAX_FRAME_RETRIES 7
```
simonduq: maybe `CSMA_MAX_FRAME_RETRIES`?

os/net/ipv6/uipopt.h
```diff
@@ -518,6 +518,35 @@ void uip_log(char *msg);
...
 #define UIP_DEFAULT_PREFIX_LEN 64
 
+/**
+ * Enables selection of maximal MAC-layer transmission count at application layer
+ */
+#ifdef UIP_CONF_WITH_VARIABLE_RETRANSMISSIONS
+#define UIP_WITH_VARIABLE_RETRANSMISSIONS UIP_CONF_WITH_VARIABLE_RETRANSMISSIONS
+#else
+#define UIP_WITH_VARIABLE_RETRANSMISSIONS 0
+#endif
```
simonduq: do we really need to make this optional?

atiselsts: `.rom` overflow on Sky otherwise.

os/net/mac/csma/csma-output.c
atiselsts: I changed this to the current name to be consistent with TSCH. Tried grepping `MAX_MAC` in CSMA code, found nothing. With your suggest name I would still find nothing...

atiselsts:
Thanks for the review Simon. Please suggest API for ICMP if you want to this feature. One way would be to add an argument to `uip_icmp6_send`.

Actually using that API from RPL better would fit in a separate PR.

simonduq:
Yes, also works for me in a separate PR. New arg to `uip_icmp6_send` sounds good

os/net/mac/csma/csma-output.c
simonduq: OK I don't really mind about the name but let's keep it consistent with the `_CONF_` version of it

os/net/ipv6/uipopt.h
simonduq: argh, that is going to be annoying soon enough :s

os/net/mac/csma/csma-output.c
simonduq: Hmm I didn't notice there was already an inconsistency :p would be good to fix anyway, right?

PR #185, opened 2017-11-18
I2C API for Contiki-NG - WIP

joakimeriksson:
This is the I2C API we have been using at Yanzi for a while. This PR is intended to provide initial suggestion for discussions rather than being the final PR. We also need to find a few drivers to port.

I2C - We have ports for a few sensors at Yanzi - but no sensors yet on any other CC2538 based platform so it is not tested on a specific sensor at the moment. A port to TMote sky might be the easiest way to test.


fontikos:
Hi Joakim. I had a look at the SPI API and it looks great.

One quick suggestion is that we probably need an `spi_dev_arch_close()`. The cc26xx-cc13xx CPUs, for example, need to power off the respective power domain.

I also want to suggest to have the spi configuration (miso, mosi, clk, speed, mode) dynamic rather than #define'd. Some use case scenarios:
- Two peripherals that operate in the same spi bus at different speeds or spi modes
- Two peripherals that share the same spi controller but are connected in different pins

joakimeriksson:
Hi!

Yes, from my perspective - if we think it looks good - we should add what you need and then get it in and start working on this together! Thank for looking at it!

I'll try to get your comments fixed in the PR!

fontikos:
Sounds good. 

Here's my take on the SPI driver for cc26xx, if you wanna have a look:
https://github.com/fontikos/contiki-ng/commit/7fac61ae08446a07b03e712561c4475081fb6096

Would it make sense to separate the PR of the I2C and SPI HALs? Perhaps, we can make a clean fork/branch for the SPI HAL and give write access to whoever wants to help?

g-oikonomou:
I think it's a good idea to separate the two. Some API doc would also come in handy once we have agreed on how we want to do things.

nfi:
Yes, I also agree it might be a good idea to separate this into two PR. Both SPI and I2C use the same file `bus-status.h` for status codes but I think git should be able to handle it without conflicts. We can try to do this later today.

fontikos:
I may have missed something, but I am not sure I understand your reasoning for having SPI and I2C sharing the same status type.

E.g. `BUS_STATUS_ADDRESS_NACK` or `BUS_STATUS_DATA_NACK` appear irrelevant in an SPI context.

joakimeriksson:
Sure, they (e.g. bus-status) can also be separated as SPI is more basic and have no explicit addressing, etc.

nfi:
We have separated this PR into I2C API only and added a new PR for SPI #206. We have also separated the bus_status_t into separate status for I2C and SPI. We will address the other comments also.

g-oikonomou:
This will now happen via #571

PR #190, opened 2017-11-22
Various SLIP-related improvements

g-oikonomou:
This pull recommends the following changes to SLIP-related code:

* We have three places in the tree where the SLIP framing code is duplicated:
  * `slip_write()`: Outputs over SLIP an arbitrary number of bytes from a memory location. This is the one we keep.
  * `slip_send()`: Outputs the L3 contents of `uip_buf`. Now simply calls .`slip_write(&uip_buf[UIP_LLH_LEN], uip_len);`.
  * `slip_send_packet`: Provided by the slip-radio example. This is being removed and all calls now call `slip_write` instead.
* All calls to `slip_send` and `slip_write` were ignoring the return value, so the return type for both functions has been changed to `void`.
* SLIP statistics are now possible to turn on/off through a configuration define (were previously hard-coded to off).
* There was some dead code wrapped inside `#ifdef SLIP_CONF_MICROSOFT_CHAT` and `#ifdef SLIP_CONF_ANSWER_MAC_REQUEST.
`. These things don't seem to exist any more, so the respective blocks have been removed.
* Some variables changed to static.
* Code style fixes.
* Some documentation improvements.

### DIFF
--- a/arch/platform/jn516x/lib/slip.c
+++ b/arch/platform/jn516x/lib/slip.c
@@ -167,7 +167,7 @@ slip_write_char(uint8_t c)
   slip_arch_writeb(c);
 }
 /*---------------------------------------------------------------------------*/
-uint8_t
+void
 slip_write(const void *_ptr, int len)
 {
   const uint8_t *ptr = _ptr;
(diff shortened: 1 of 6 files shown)


g-oikonomou:
Forgot to say, there is yet another implementation of the entire SLIP logic under `arch/platform/jn516x/lib/slip.c`. This includes many platform-specific things, such as interrupt manipulation. It also has some logic about software flow control and perhaps more that I missed. I have no intention of touching it, but if whoever wrote it wishes to contribute back to the slip logic under `os` that would be great!

joakimeriksson:
Looks good - nice cleanup! 👍

simonduq:
This is a very welcome cleanup indeed! I merged from the github interface, shout if you would rather rebase before we pull in!

g-oikonomou:
Rebasing now

Issue #104, opened 2017-10-20
Stack guard


atiselsts:
See the discussion here: https://github.com/contiki-os/contiki/pull/2089

I can adapt it for Contiki-NG (some work required) and resubmit here. 

The one thing remaining to discuss here: do you have objections against it being enabled by default?
As I stated in the "old" Contiki PR discussion, I believe it would be much more useful if enabled. "Release" builds can always explicitly remove it, by simply adding a compilation flag.

g-oikonomou:
Closed in #192

PR #192, opened 2017-11-22
Add stack check library and example

atiselsts:
Adds stack filler and checker, as discussed in #104

Based on my previous pull request for Contiki: https://github.com/contiki-os/contiki/pull/2089
The differences are:
 * this one is enabled by default on most platforms;
 * this one does periodic checks (once 10 seconds) on most platforms;
 * it uses Simon's new logging module (using `LOG_LEVEL_MAIN` - hope that's ok?)

Platforms tested: nRF52dk, JN516x, Sky (in Cooja), Zolertia Zoul, srf06-cc26xx
Platforms not supported: Cooja, native.

Platforms not enabled by default: Sky (because of `.rom` limitations).


g-oikonomou:
Very desirable feature here.

examples/libs/stack-check/Makefile
```diff
@@ -0,0 +1,9 @@
+CONTIKI=../../..
+CONTIKI_PROJECT = main
+CFLAGS += -DPROJECT_CONF_H=\"project-conf.h\"
+
+CONTIKI_WITH_IPV6 = 1
```
g-oikonomou: I think we need `MAKE_NET_IPV6` here

os/sys/stack-check.c
```diff
@@ -0,0 +1,163 @@
...
+/*---------------------------------------------------------------------------*/
+void
+stack_check_init(void)
+{
+  uint8_t *p;
+
+  stack_origin = &p;
+  p = &_stack;
+
+  /* printf("eoh=%p eos=%p stack=%p\n", &_stack, GET_STACK_ORIGIN(), p); */
```
g-oikonomou: Maybe get rid of that?

os/sys/stack-check.c
```diff
@@ -0,0 +1,163 @@
...
+#include "sys/stack-check.h"
+#include "dev/watchdog.h"
+#include <string.h>
+
+#include "sys/log.h"
+#define LOG_MODULE "Stack"
+#define LOG_LEVEL LOG_LEVEL_MAIN
+
+/*---------------------------------------------------------------------------*/
+#if STACK_CHECK_ENABLED
```
g-oikonomou: Do we _really_ need this? Won't the linker do the job for us? Would it be an idea to initialise the library conditionally in `main()` and then always compile the code here?

arch/cpu/msp430/Makefile.msp430
```diff
@@ -182,6 +182,9 @@ endif # SMALL
 
 endif # IAR
 
+# Define the `_stack` symbol used by the stack check library to be equal to `_end`
+LDFLAGS += -Wl,--defsym=_stack=_end
```
g-oikonomou: Not sure, but I somehow think it's possible to achieve this inside the linker script?

atiselsts: The msp430 linker scripts are currently not shipped with Contiki. I don't think that adding a modified version of the linker script to the OS is preferable to one-line change in Makefile.

examples/libs/stack-check/Makefile
atiselsts: ok, I'll check - was copy paste from Contiki...

os/sys/stack-check.c
atiselsts: Yes, that would work. Do you prefer having the `#if` guards in `main()`, or have something else in mind?

os/sys/stack-check.c
atiselsts: ok

os/sys/stack-check.c
g-oikonomou: No problem with `#if` guards in `main()`, we use them for orchestra and the shell too.

arch/cpu/msp430/Makefile.msp430
g-oikonomou: Gocha OK, how about the linker scripts for other CPUs though?

atiselsts: Already done for cc2538 and cc26xx; can do that for nrf52. For the other platforms (JN516x and msp430) Makefiles.

os/sys/stack-check.c
atiselsts: Done, I like that there's no need for multiple "versions" of the functions in the stack check library code now!

arch/platform/jn516x/platform-conf.h
```diff
@@ -314,4 +314,8 @@ DR1174+DR1175:
 #define LEDS_GP2      64
 #define LEDS_GP3      128
 #define LEDS_CONF_ALL 255
+
+/* Platform-specific define for the end of the stack region */
+#define STACK_CONF_ORIGIN     ((void *)0x4008000)
```
simonduq: do these need to be hard-coded in `.h` file, aren't there ways to ask the linker and then pass the flag at compile time (CLFAGS)?

arch/cpu/msp430/Makefile.msp430
atiselsts: You ok with the change now?

atiselsts:
Simon, I added a regression test (for Sky). Agree it's a good idea.

arch/platform/jn516x/platform-conf.h
atiselsts: Not sure what exactly / technically you have mind here?

arch/cpu/msp430/Makefile.msp430
g-oikonomou: All good 👍

arch/platform/jn516x/platform-conf.h
simonduq: I was just wondering if there isn't a systematic way to find the end of stack from linkerscript, and use it as define here, instead of hardcoding it in the .h file. This way, if linker / linkscript changes, this doesn't break

os/sys/stack-check.c
```diff
@@ -0,0 +1,158 @@
...
+
+  PROCESS_BEGIN();
+
+  etimer_set(&et, STACK_CHECK_PERIOD);
+
+  while(1) {
+    PROCESS_WAIT_EVENT_UNTIL(etimer_expired(&et));
+
+    uint16_t actual = stack_check_get_usage();
+    uint16_t allowed = stack_check_get_reserved_size();
```
simonduq: code style: I was always assuming we don't accept variables declared in the middle of blocks, but I'm not sure now. Opinions?

atiselsts: I'll change for consistency.

arch/platform/jn516x/platform-conf.h
atiselsts: Technically it looks like it's possible to do this in the linker script like this.
`_stack_origin = ORIGIN(RAM) + LENGTH(RAM);`

Thanks. I will do it for the platforms that I can (not for msp430). For JN516x I can do it in Makefile - there is a symbol called `_ram_top` of which I'll make an alias.

simonduq: sounds great!

atiselsts:
Now with the `_stack_origin` symbol in linker scripts.

@g-oikonomou can you check if CC2538 linker script makes sense? For some reason when using this value in the checker I get huge stack "usage", some 5000 bytes more than expected. It's still much less than the maximum acceptable stack size, so that's not a problem in terms of functionality, just looks strange.

g-oikonomou:
I'm afraid this is beyond me :(

simonduq:
@atiselsts I see that there is still an open issue here, shall add work-in-progress label?

atiselsts:
Sorry, did not notice Travis failing - pushed an update new. Turns out that with the new change I only had updated 2 out of 3 total linker scripts for nrf52dk. Other than that, it's ready from my point of views.

simonduq:
What about the cc2538 pb with 5000 bytes off?

atiselsts:
Not sure if there's an any action to be done about that?

simonduq:
I don't know.. is it not fixable? If we leave it with a not fully expected behavior, it would good to at least document this somewhere as comment.

atiselsts:
A comment is not problem. I was thinking of three options:
1) revert back to the old behavior for CC2538 in particular and try to guess the stack start by taking an address of a stack-local variable soon after the start of the program;
2) wait until we get more information about this behavior, which may or may not lead to changes in how `_stack_origin` is defined in the CC2538 linker script;
3) document this somewhere and move along. Looking at the wiki, I don't immediately see a place for this, so a comment is probably the way to go for now.

simonduq:
options 1 or 2 sound best to me

atiselsts:
I think I found the problem in the cc2538 linker script and changed the `_stack_origin` location now. Don't have any cc2538 devices at hand, will check on monday.

simonduq:
Please give a heads up when this is tested!

atiselsts:
Successfully retested on Zoul and nrf52dk today.

simonduq:
Great, so this is ready isn't?

atiselsts:
Unless @g-oikonomou has further comments, yes!

g-oikonomou:
All my comments have been addressed! :)

PR #194, opened 2017-11-22
Master interrupt manipulation API

g-oikonomou:
This pull recommends an API for master interrupt manipulation. Nothing hugely fancy: retrieve current state, disable, enable.

Implemented for CC2538, CC13xx/CC26xx and msp430. Help needed for JN516x and NRF.

The ultimate goal is to provide mutexes (but that's a different pull).


atiselsts:
Long missing feature...

Would it make sense to provide a convenience function "restore previous state"?
```
static inline int_master_restore(bool was_enabled) {
  if(was_enabled) {
   int_master_enable();
  }
}
```

g-oikonomou:
> Would it make sense to provide a convenience function "restore previous state"?

Yeah, I had knocked together a macro for this, but I somehow thought it was not worth it. Can you send a commit my way?

g-oikonomou:
I've applied Atis' recommendation and rebased.

Who can help with the last two platforms?

joakimeriksson:
So - I guess we need to test on H/W also - I have the NRFs now - so maybe I could do a quick test?

g-oikonomou:
If you have a little bit of time available to test something, please could you spend it on #204? This one here will get updated a lot after #204 gets merged.

joakimeriksson:
I will try the NRFs with a focus on #204 - probably tomorrow or during the weekend.

PR #196, opened 2017-11-23
Netstack initializion fix

simonduq:
This reworks netstack initialization on cc2538 platforms. Initialize linkaddr in phase 2 but set RF parameters only in phase 3, i.e., after `netstack_init` is called. Without this fix, Zoul fails to run with cc1200.

### DIFF
--- a/arch/platform/cc2538dk/platform.c
+++ b/arch/platform/cc2538dk/platform.c
@@ -103,9 +103,6 @@ set_rf_params(void)
   short_addr = ext_addr[7];
   short_addr |= ext_addr[6] << 8;
 
-  /* Populate linkaddr_node_addr. Maintain endianness */
-  memcpy(&linkaddr_node_addr, &ext_addr[8 - LINKADDR_SIZE], LINKADDR_SIZE);
-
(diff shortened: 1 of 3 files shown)


g-oikonomou:
It seems that we are assuming that `linkaddr_node_addr` is always 8 bytes / `LINKADDR_SIZE == 8`. Are you sure about that?

simonduq:
I'm not sure, do you think we need to preserve the ` &ext_addr[8 - LINKADDR_SIZE]` logic?
I actually copied from platform `cc26xx-cc13xx`

g-oikonomou:
>  I actually copied from platform cc26xx-cc13xx

😊 

I think 2-byte LL address were used mainly by Rime, right? In that case perhaps we can think about removing support for that in the future?

Happy to merge this one here as is.

joakimeriksson:
I do think we might be interested in 2-bytes addresses also for IPv6 based communication in a distant future. Will give smaller packet sizes.

simonduq:
I agree with that, that's what Thread does and 6TiSCH etc. The downside is that you need to guarantee uniqueness through some registration procedure, but saves bytes.

simonduq:
What we'll need is actually more than hardcoded 2-or-8. Existing solutions start on a long address, register to get a short and then use it.

PR #198, opened 2017-11-23
CC26xx radio IEEE mode driver: limit the time spend in waiting for transmissions to finish

atiselsts:
This change significantly improved the stability of our CC26xx nodes, in terms of number of reboots per device, especially under heavy RF traffic.


simonduq:
Looks good to me! (even though I'm not familiar with this code)

PR #199, opened 2017-11-23
Improvements in BLE Advertisements for cc26xx-cc13xx

fontikos:
This pull enhances the arch-specific RF BLE driver for the cc26xx-cc13xx family.

There are four basic changes:
1. BLE MAC Address: The driver loads the actual hard-coded BLE MAC address of the device from the flash. The user can also specify a custom BLE MAC address by defining `BLE_ADDRESS_PTR`.
2. Adv Tx Power: The tx power for the advertisements is not hard-coded to +5 dBm anymore, but  configurable by the user via `rf_ble_set_tx_power()`.
3. BLE Adv Channels: The user can select any arbitrary subset of the three BLE advertisement channels to transmit the BLE advertisement packet.
4. Send Single BLE Advertisement: The user can transmit a single BLE advertisement with dynamic payload using `rf_ble_beacon_single()`.

### DIFF
--- a/arch/cpu/cc26xx-cc13xx/rf-core/rf-ble.c
+++ b/arch/cpu/cc26xx-cc13xx/rf-core/rf-ble.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2015, Texas Instruments Incorporated - http://www.ti.com/
+ * Copyright (c) 2017, University of Bristol - http://www.bristol.ac.uk/
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


simonduq:
Looks good. Only suggestion would be to use the logging module instead of `PRINTF` :) But this can get in as is AFAIC

fontikos:
Thanks Simon. Yes, using the logging module would be great, but at the moment all the other modules in `arch/cpu/cc26xx-cc13xx/rf-core/` are based on `PRINTF`. I don't have a strong opinion, but I'd probably opt for keeping the `PRINTF` for now for consistency, and upgrade the whole radio library to the logging module with a following PR.

simonduq:
Yep that sounds good

g-oikonomou:
This looks good

fontikos:
Minor printing error fixed.

PR #204, opened 2017-11-26
Provide common build infrastructure for CM3- and CM4-based systems

g-oikonomou:
Currently we have plenty of CM3-based CPUs (CC2538 / CC13xx-CC26xx) and platforms.

This pull is part of a sequence of pulls that aim to unify all CM3-related code under `arch/cpu/arm/cm3`. 

We start with the build system in this pull. We create a common `Makefile.cm3` that CM3-based CPUs can simply include from their own `Makefile.<cpu>`instead of duplicating code.

This common Makefile:
* Provides a single point of configuration of the toolchain (`CC`, `LD` etc)
* Provides common `CFLAGS`, `LDFLAGS` etc
* Configures `MODULES` used by all respective platforms/SoCs
* Provides common Make rules where possible

This pull also adopts this Makefile for all CC2538 and CC13xx/CC26xx platforms.

At this stage I am asking for your comments and ideas. It appears as though it should be pretty easy to use this approach for `nrf52dk` too.

Sits on top of #203


simonduq:
Nice. For `nrf52dk` you mean a new `arch/cpu/arm/cm4` folder?

g-oikonomou:
>Nice. For nrf52dk you mean a new arch/cpu/arm/cm4 folder?

Ugh is it CM4? I thought it was CM3.

What I have done in the past (sensniff unreleased work) was adopt a structure like so:

```
arch/arm/cm3/cc2538/zoul/remote
arch/arm/cm3/cc2538/zoul/firefly
arch/arm/cm3/cc2538/cc2538dk
arch/arm/cm3/cc13xx-cc26xx/cc13xx-cc26xx/launchpad/cc1350
arch/arm/cm3/cc13xx-cc26xx/cc13xx-cc26xx/sensortag/cc2650
[...]
```

Each directory could host sources and had a Makefile. So the idea is you started at the leaf and each Makefile was automatically including the one from its parent dir. This was working pretty darn well in terms of avoiding duplication.

I am not suggesting adopting all of that, but in this pull here I was thinking along similar lines. Perhaps we need another `Makefile` under `arch/cpu/arm`, covering all CPU versions, and then keep cm3 and cm4 in separate dirs. The differences will be very small at this stage (I think only `CFLAGS` and `LDFLAGS` at this stage, but should help moving forward.

simonduq:
Sounds all good to me. We should aim for 0 duplication :)

g-oikonomou:
Successfully tested on CC2538 and CC26xx. As far as I am concerned this is ready to merge unless @joakimeriksson discovers problems with nrf52dk.

joakimeriksson:
Ok, then I will test a few of the NRF examples and pull if it seems to work.

joakimeriksson:
Ok, the hello-blink and timer tests runs fine on the NRF52. But not 100% sure of CoAP examples. Will try them in master also - to see if there would be any difference.

PR #205, opened 2017-11-26
Minor tidy-up of Makefile.include

g-oikonomou:
This pull:

* Removes the useless make variable `CONTIKI_ARCH_DIRS`
* Tidies up the format of the definition of the `SOURCEDIRS` make variable


g-oikonomou:
Looks like `CONTIKI_ARCH_DIRS` was necessary after all 🤣

simonduq:
right, used as base for include path it seems

PR #206, opened 2017-11-28
SPI API for Contiki-NG - WIP

nfi:
This is the SPI API we have been using at Yanzi for a while. This PR is intended to provide initial suggestion for discussions rather than being the final PR. We also need to find a few drivers to port to show it is intended to be used.


fontikos:
Thanks. I'll prepare the `spi_dev_arch_*` implementations for cc26xx-cc13xx.

fontikos:
I believe that the first priority should be to finalise the HAL SPI API, defined in `spi-dev.h`, as mentioned by @g-oikonomou in #185 .

Based on this PR, my suggestion is the following:
https://github.com/fontikos/contiki-ng/commit/fd71e9d77ad19643a1558eae26de15641eb1d930

I added documentation according to my understanding, which might be wrong, so please do correct me! :)

In brief:
- Locking is controlled by `spi_dev_arch_lock()` and `spi_dev_arch_unlock()`. With the existing abstraction, it makes sense to hide the implementation of the locks from the SPI HAL.
- The MISO, MOSI, SCK pins, the CS pin, the bit rate, and the SPI mode are architecture-independent configurations of an SPI bus. I added them in the struct `spi_device_t`. All pins are declared as `uint32_t`. This can be upgraded with a GPIO HAL.
- I do not really understand the concept of an SPI timeout, as SPI does not really wait for anything. Does it?
- `spi_dev_arch_close()` added in the API to close and power down the interface.
- `spi_dev_arch_flush()` added in the API to flush any remaining bytes in receive buffer.
- `spi_dev_arch_select()` and `spi_dev_arch_deselect()` also added in the API to control the CS pin.

What do you think guys?

fontikos:
I had a lengthy chat with @g-oikonomou about this and I saving here some notes for future reference:
- To avoid code replication, we should move locking in the HAL. All `arch*` functions can assume that they have the lock.
- We can avoid flush in the API by calling it internally after each SPI transfer. This is relevant only to SPI controllers that have a FIFO read buffer.
- We can assume that as long as a device has one SPI controller, this will have the id 0. All platform independent examples will be based on SPI 0.
- We can couple locking/unlocking with opening/closing the SPI interface. At the cost of a few, perhaps uneccessary, open/close, this would ensure that the SPI interface is not forgotten open.

g-oikonomou:
> To avoid code replication, we should move locking in the HAL. All arch* functions can assume that they have the lock.

I'm still not 100% sure about that. While this makes sense, it does mean that all platform-specific SPI drivers will have to go through the HAL. We will have to be explicit that only the HAL is allowed to call `_arch` functions. Do we want to do this?

fontikos:
Why would one choose not to go through the HAL? Is there a use case?

g-oikonomou:
Platform-specific code might want to call the `_arch` functions directly / bypass the HAL to reduce code size. Or a developer might want to bypass it because they are developing something specific for a product that they don't plan to share, so they don't care about it being platform-independent.

joakimeriksson:
The _arch functions are intended to be whatever is specific per implementation - e.g. if there is a good locking mechanism on the cc2538-libraries - it would be used. Lock / unlock is also intended to do H/W initialisation for the specific SPI bus configured for that devices. So there can not really be a HAL lock unless there is only one single SPI bus on that H/W. 

Quick question - what is the select() / deselect intended for that can not be handled by the callbacks in the dev struct? The dev have a callback for doing select / deselect and can be either only a single GPIO or something more complex (including both CS and power-on / multiplexing or something like that).

joakimeriksson:
Also - the SPI Bus struct represents a specific SPI controller with the configuration and the SPI-dev have a pointer to it for it to be associated with it. Each device have the specific configuration including timeouts, speeds, etc. The bus itself know who is "locking" the bus, and what the H/W config is, etc. We can bring in the configuration for CC1200 SPI on the CC2538 (as we use it for that at Yanzi) and show how it is intended.

fontikos:
@g-oikonomou Good point about code size. Happy to keep it as it is, with locking inside the arch. I will soon provide a full SPI HAL proposal to have a look.

@joakimeriksson I noticed the function pointer for selecting/deselecting. It can be done like that; I just think it looks better if the SPI HAL is self-sufficient and implements everything part of the SPI protocol. I think that ideally, selecting/deselecting should be done in the HAL, but we need a GPIO HAL for this first.

joakimeriksson:
@fontikos Yes - lets move the select/deselect into the arch functions then! As long as we send in the device struct pointer we can do any type of super-complex multiplexing if needed for a specific platform.

fontikos:
Thanks Joakim for the edit above. I got your approach now.

Interestingly, I can spot some differences between cc2638 and cc26xx-xx13xx: 
- The mode is in phase/polarity in one and MODE0 ... MODE4 in the other
- The speed is in MHz in one and in Hz in the other

I think it'd be really good if we come we a configuration struct that is architecture-independent and put it in the HAL, so that the same SPI configuration (like the one you quoted above) works in all platforms!

joakimeriksson:
I do think most of the config is very independent while some specifics are more platform /cpu dependent. But the good thing is - even if the config struct needs to be slightly different it will be exactly the same drivers with 99% of the code being same and just some minor SPI related config and arch implementation different. Which means that less of the cc1200-arch.c and such specific files are needed (well maybe something for interrupts still - but not for SPI).

fontikos:
Agreed, but we can still try to harmonise it 100%, if possible. I already adopted your pha/pol, which I like it much more than #defines of TI. We can also harmonise the speed to either MHz or bitrate and convert it internally.

fontikos:
If there is still interest in an SPI HAL, I am considering making this my Christmas project and work on it during holidays.

One direction would be to prepare a new PR that builds on this one (based on this work-in-progress: https://github.com/fontikos/contiki-ng/commits/wip/spi-cc26xx-cc13xx). This new PR would provide a full implementation for `cc26xx-cc13xx`, including examples on how to use it based on the SPI peripherals of the `cc26xx-cc13xx` platforms. It will also be fully documented, so that others can provide the `spi_arch_*` implementation for other platforms.

I am also happy to provide the `spi_arch_*` implementation for the `cc26xx-cc13xx` platforms based on the HAL proposed in this PR. Yet, to do so, I'd first need full documentation first, so that I do not make wrong assumptions.

Let me know how do you wish to proceed.

simonduq:
Isn't this PR outdated by https://github.com/contiki-ng/contiki-ng/pull/264 ? Shall we close this one?

Issue #207, opened 2017-11-28
CoAP not parsing options


AntonioLangiu:
I noticed that the CoAP implementation is not parsing the received options.

After some debugging, I found out that during the refactor of er-coap [this line](https://github.com/contiki-ng/contiki-ng/blob/develop/os/net/app-layer/coap/coap.c#L547) was added in the`coap_parse_message` function. Removing that line fixes the problem.

joakimeriksson:
Thanks for the report. That certainly looks like a bug. We are currently working on a big PR with CoAP, LWM2M and DTLS and we will ensure that this bug is fixed when that gets in (or even before as a separate fix-PR).

AntonioLangiu:
Thank you! 👍

simonduq:
addressed in https://github.com/contiki-ng/contiki-ng/pull/210

joakimeriksson:
and now #210 is in - @AntonioLangiu: I guess this can be closed ?

AntonioLangiu:
Yes, it can be closed! Thank you for the fast merge!

PR #183, opened 2017-11-16
Contiker: Contiki-NG's Docker image

simonduq:
This PR brings Docker support for the NG toolchain and development environment (Dockerfile under `tools/docker`). It also uses the Docker image in Travis tests, so that one can exactly reproduce the Travis environment locally. It builds on https://github.com/contiki-ng/contiki-ng/pull/165 for cleaner CI tests, and reworks the CI framework further. More consistent scripts across the tests, and more readable test logs.

I have tested this on a Linux 64-bit host. The Docker image builds from a 32-bit Ubuntu, as required for Cooja "Contiki motes". 

To get started, you can either build the image or download it from Docker Hub. To build it:
```
cd contiki-ng/tools/docker
docker build -t contiki-ng .
```

To download it:
```
docker pull simonduq/contiki-ng
```

For convenience, one can use an alias such as (tested only on Linux):
```
export CNG_PATH='<ABS_PATH_TO_YOUT_CONTIKI_NG>'
alias contiker="docker run --privileged -v $CNG_PATH:/home/user/contiki-ng -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix -ti simonduq/contiki-ng"
```

The container uses a *bind mount* to directly access the Contiki-NG repo form the host. This is what Travis uses (it git clones on the host, then starts a container with direct access to the cloned git). This is also handy as a development environment, as it leaves the Contiki-NG git clone no the host, i.e. accessible by editors installed in the host.

We also provide as an alternative a standalone image, with Contiki-NG and pre-compiled Cooja included. The image is also on Docker Hub, just pull `simonduq/contiki-ng:standalone`.

With the alias above, one can then start a bash:
```
contiker bash
```
or even just:
```
contiker
```

Run a particular command:
```
contiker bash -c "make -C examples/hello-world TARGET=zoul"
```

Or even run Cooja directly, provided one has a X server running on the host, and `DISPLAY` points to it:
```
contiker cooja
```

There is one particular 6top test that does not pass. @yatch , any clue what the problem might be?
https://travis-ci.org/simonduq/contiki-ng/jobs/303194585

To reproduce:
```bash
docker pull simonduq/contiki-ng
contiker bash -c "make -C tests/16-6tisch 02-cooja-test-sixp-pkt.testlog"
```

Or `contiker cooja`, open and run the `02-cooja-test-sixp-pkt.csc`.

PR #210, opened 2017-11-29
CoAP option parsing fix

simonduq:
This PR addresses https://github.com/contiki-ng/contiki-ng/issues/207 , and adds a CI test for CoAP (runs CoAP server on target native, and requests it from linux host with `coap-cli`). Builds on https://github.com/contiki-ng/contiki-ng/pull/183 as it installs coap-cli in Docker.

It turns out CoAP was basically not working before. Shall we consider the fix (not the whole PR) for inclusion in `master´?


joakimeriksson:
Looks good. 👍

PR #212, opened 2017-11-30
Extend the logging module to support custom module prefix

nfi:
Adds a configuration option `LOG_CONF_OUTPUT_PREFIX` to specify a custom output function to output the log level and module name in the beginning of each log line. This makes it possible to add a timestamp or use a different log module prefix format than the default.

### DIFF
--- a/os/sys/log-conf.h
+++ b/os/sys/log-conf.h
@@ -81,7 +81,16 @@
 #define LOG_OUTPUT(...) printf(__VA_ARGS__)
 #endif /* LOG_CONF_OUTPUT */
 
-/* Custom log prefix output function -- default is LOG_OUTPUT */
+/*
+ * Custom output function to prefix logs with level and module.
+ *
+ * This will only be called when LOG_CONF_WITH_MODULE_PREFIX is enabled and
+ * all implementations should be based on LOG_OUTPUT.
+ *
+ * \param level     The log level
+ * \param levelstr  The log level as string
+ * \param module    The module string descriptor
+ */
 #ifdef LOG_CONF_OUTPUT_PREFIX
 #define LOG_OUTPUT_PREFIX(level, levelstr, module) LOG_CONF_OUTPUT_PREFIX(level, levelstr, module)
 #else /* LOG_CONF_OUTPUT_PREFIX */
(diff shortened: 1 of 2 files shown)


simonduq:
Great! Question: will this play nicely in the `LOG_WITH_LOC` cases, which uses `LOG_OUTPUT` directly?

nfi:
The module prefix is done before the LOC is generated so I do not think it will be any different from before. Here is an example with custom module prefix and LOC enabled:

```
2017-11-30 20:00:12.981 [coap-ipv4] DBG  - [coap-ipv4.c: 259] parsing endpoint coap://127.0.0.1 => endpoint at 127.0.0.1:5683
2017-11-30 20:00:12.981 [coap-ipv4] INFO - [coap-ipv4.c: 426] CoAP server listening on port 5683
```

But perhaps we should make also the LOC format configurable in the same way as the module prefix by adding  `LOG_CONF_OUTPUT_LOC(file, line)` which defaults to use `LOG_OUTPUT` as today?

simonduq:
I'm mostly concerned about the fact that `LOG_CONF_OUTPUT_PREFIX` might choose not to even rely on `LOG_OUTPUT` at all, while `LOG_WITH_LOC` is always `LOG_OUTPUT`-based.

Could either add `LOG_CONF_OUTPUT_LOC` as you suggest, or make sure `LOG_CONF_OUTPUT_PREFIX` always calls `LOG_OUTPUT` (but defines its own printf format)

nfi:
The log data also rely on `LOG_OUTPUT` so regardless of `LOG_WITH_LOC`, things will be strange if `LOG_CONF_OUTPUT_PREFIX` would not be based on `LOG_OUTPUT`. I simply assumed that any implementation of `LOG_CONF_OUTPUT_PREFIX` would be based on `LOG_OUTPUT`  like the rest of the log functions.

simonduq:
Yes right now I'm with you. Was somehow assuming one of the goals was to have `LOG_CONF_OUTPUT_PREFIX` possibly bypass `LOG_OUTPUT`.

Works for me. Is this ready for merging?

nfi:
Should I add `LOG_CONF_OUTPUT_LOC` also?

simonduq:
I'd say this is overkill. Maybe as a comment to `LOG_CONF_OUTPUT_PREFIX` say the def is expected to use `LOG_OUTPUT`?

nfi:
Ok, I added a comment that implementations of `LOG_CONF_OUTPUT_PREFIX` should be based on `LOG_OUTPUT`.

simonduq:
Ah that is great :)

PR #215, opened 2017-12-01
Uipbuf attributes and moving in transmission number into uipbuf attribute.

joakimeriksson:
This PR moves in the IP level transmission number into the uipbuf module and make use of uipbuf attributes for this. It also moves out all modifications of the 6LoWPAN adaptation layer and add the traffic class "tagging" into the IP level (tcpip.c). This PR also adds a very basic example of its usage in udp-rpl in udp-client.

```
      /* Set the number of transmissions to use for this packet -
           this can be used to create more reliable transmissions or
           less reliable than the default. Works end-to-end if
           UIP_CONF_TAG_TC_WITH_VARIABLE_RETRANSMISSIONS is set to 1.
         */
        uipbuf_set_attr(UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS, 1 + count % 5);
        simple_udp_sendto(&udp_conn, &count, sizeof(count), &dag->dag_id);

```

Question to all users of the QoS feature (e.g. end-to-end) - how do you like to use it?

Should it be a part of the udp_conn or not?

I think this feature requires a test - so I will try to set one up...


os/net/ipv6/uipopt.h
```diff
@@ -547,6 +531,19 @@ void uip_log(char *msg);
  */
 #define UIP_TC_MAC_TRANSMISSION_COUNTER_MASK 0x3F
 
+#ifdef UIP_CONF_TAG_TC_WITH_VARIABLE_RETRANSMISSIONS
+#define UIP_TAG_TC_WITH_VARIABLE_RETRANSMISSIONS UIP_CONF_TAG_TC_WITH_VARIABLE_RETRANSMISSIONS
+#else
+#define UIP_TAG_TC_WITH_VARIABLE_RETRANSMISSIONS 1
```
joakimeriksson: Maybe this should in fact be zero. As this feature will change traffic class as soon as we control the transmission count from uipbuf.

joakimeriksson:
Ok, that is probably another reason to turn things off by default (Sky overflow again...).

os/net/ipv6/tcpip.c
```diff
@@ -121,6 +121,19 @@ uint8_t
 tcpip_output(const uip_lladdr_t *a)
 {
   int ret;
+
+  /* Tag Traffic Class if we are using TC for variable retrans */
+#if UIP_TAG_TC_WITH_VARIABLE_RETRANSMISSIONS
+  if(uipbuf_get_attr(UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS) !=
+     UIP_MAX_MAC_TRANSMISSIONS_UNDEFINED) {
```
simonduq: this seems to check something that is never set by default (only happens to be 0)

os/net/ipv6/uipopt.h
simonduq: I also vote for disabling by default, because this is anyway unused by default, is a bit experimental, and not really a standard feature

os/net/ipv6/tcpip.c
joakimeriksson: uipopt.h?

joakimeriksson: uipopt.h?

simonduq: I mean if it is intended as 0 by default, we might want to explicitly set it as `UIP_MAX_MAC_TRANSMISSIONS_UNDEFINED`, not just count on the fact that `UIP_MAX_MAC_TRANSMISSIONS_UNDEFINED` is 0 and leave the attribute uninitialized

joakimeriksson:
Just shaved some bytes form Contiki-main.c to make it fit Sky. Now everything is green ;-)

os/net/ipv6/uipopt.h
```diff
@@ -547,6 +531,19 @@ void uip_log(char *msg);
...
+#else
+#define UIP_TAG_TC_WITH_VARIABLE_RETRANSMISSIONS 0
+#endif
+
+/**
+ * This is the default value of MAC-layer transmissons for uIPv6
+ *
+ * It means that the limit is selected by the MAC protocol instead of uIPv6.
+ */
+#define UIP_MAX_MAC_TRANSMISSIONS_UNDEFINED 0
```
simonduq: Let me rephrase my concern here. If we define this macro to a different value, things will break, because there is no `uipbuf_set_attr(UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS, UIP_MAX_MAC_TRANSMISSIONS_UNDEFINED)`.

simonduq:
Other than the comment I still have this is ready IMO

simonduq:
BTW, which is the sky test that now needs byte-optimization?

simonduq:
Hmm I see it is the border router. We should maybe disable the Web server on sky for these tests then

os/net/ipv6/uipopt.h
joakimeriksson: Ok, we can move this into the uip clearbuf so that it is set to this value there?

simonduq: works for me!

PR #218, opened 2017-12-03
Tidy-up doxygen docs

g-oikonomou:
This pull tidies up our doxygen documentation:

* Some doxygen groups for platforms were previously defined in `tools/doxygen/modules.txt`. Those doxygen groups did not always get deleted when the respective platform got removed. So for example we currently still have a doxygen group for exp5438.
  * This pull removes these empty groups.
  * We no longer create doxygen groups for platforms in this file. Platform-specific groups are created in a file called `doxygen-group.txt`, placed in the platform's respective directory. If a platform gets deleted in the future, so will the doxygen group.
* Some doxygen groups for CPUs were previously defined in `os/cpu/arm/arm.txt`. Those doxygen groups did not always get deleted when the respective CPU got removed. So for example we currently still have a doxygen group for AT91SAM7S.
  * This pull removes these empty groups.
  * CPU-specific groups are created in a file called `doxygen-group.txt`, placed in the CPU's respective directory. If a CPU gets deleted in the future, so will the doxygen group.
* Documentation for CPUs (stuff in `arch/cpu`) and device driver API (stuff in `os/dev` and `arch/dev`) were previously bundled in the same group. This pull separates them into two groups.
* We move CC2538 and CC13xx/CC26xx CPU documentation to the CPU group.
* We tidy-up CMSIS documentation.
* We re-structure the hierarchy for platform documentation (for example we put all CC2538-based platforms in the same group).
* We add a missing copyright header.


simonduq:
Great! We'll need to think eventually of auto-uploading this to https://contiki-ng.github.io/ngdoc/modules, perhaps through the existing Doxygen test, pushing to github.io for each `develop` build?

PR #219, opened 2017-12-03
Separate platform config from CPU config

g-oikonomou:
Currently, our various `contiki-conf.h` files contain:

* Platform-related configuration:
  * That the user is allowed to modify (e.g. whether to use the CC1200 or the CC2538 radio on the zoul).
  * That the user is _not_ allowed to modify (e.g. the pins connected to the CC2538 ROM bootloader).
* CPU-related configuration:
  * That the user is also allowed to modify (e.g. ROM bootloader enable/disable).
  * That the user may _not_ modify (e.g. TSCH timings, the value of `CLOCK_CONF_SECOND` or the data type for `clock_time_t`).

This leads to a number of problems:
* It is not immediately obvious to new users which defines they are / are not allowed to touch
* While porting, one will typically copy a `contiki-conf.h` from a different platform and apply some modifications to it.
* Changing something across the board requires messing about with multiple `contiki-conf.h` files.

In a nutshell, this pull tries to break down those configuration directives into separate files. This is by no means 100% done here, there are still some things in `-conf.h` files that users should not be touching, but it's a step forward.

This pull requires #204

# Non-configurable, CPU specifics

Non-configurable, CPU specifics are moved to header files suffixed by `-def.h` and are placed under various CPU dirs. For example, all our arm-based CPUs use `typedef uint32_t clock_time_t;`. Instead of having this replicated in every single `contiki-conf.h`, we move this to `arm/arm-def.h`. This reduces code duplication and also does not give users the false impression that it's OK to just change.

The way this works is, we establish a `-def.h` header for each CPU (e.g. `cc2538-def.h`). The platform's `contiki-conf.h` simply includes this file, which may in turn include an architecture-specific `-def.h` file.

For example:

* For all CC2538 platforms, `contiki-conf.h`  includes `cc2538-def.h`. This defines CC2538-specifics, such as `RTIMER_ARCH_SECOND` and TSCH timings for the CC2538 radio.
* `cc2538-def.h` includes `arm/cortex-m/cm3/cm3-def.h`, which in turn includes `arm/arm-def.h`. The latter provides values for `CLOCK_CONF_SECOND` and the data type for `clock_time_t`.

# Platform vs CPU configuration

This pull also establishes CPU-specific `-conf.h` header files, containing all the CPU configuration that the user _can_ modify.

The platform's `contiki-conf.h` provides platform defaults and then includes the CPU's configuration header at the bottom. The latter provides CPU-specific defaults, for everything that the example/platform has not provided a configuration for.


simonduq:
Nice!

I believe some platforms used to have a `contiki-conf.h` vs `platform-conf.h` with this intent (but only platform, not cpu). Sky still has a `platform-conf.h` for instance. I also see one for `nrf52dk` which now simply includes the `-def.h` instead. Any reason to keep any of these `platform-conf.h` files?

g-oikonomou:
No particular reason, no. I just didn't get round to updating all current platforms. Do you think it should happen here?

simonduq:
Would feel natural to me at least, yes :) (it almost just moving files, or is there more work involved I missed?)

g-oikonomou:
let me have a look

g-oikonomou:
@atiselsts @simonduq Are the following defines something we'd normally want to allow users to tweak?

* `TSCH_CONF_BASE_DRIFT_PPM`
* `TSCH_CONF_CHANNEL_SCAN_DURATION`
* `TSCH_CONF_RX_WAIT`

simonduq:
The first is H/W dependent and fixed.

The two others are more like config parameters that a user might want to tweak.

g-oikonomou:
Thanks

I'm looking into sky/msp430 and nrf52dk, but if everything else here is acceptable, don't let me delay you from merging this one ;)

g-oikonomou:
I have updated this pull to harmonise configuration file naming convention for nrf52dk and jn516x. I have also applied the recommended changes to Sky/msp430.

I'll fix travis problems if any and then this is ready from my end.

simonduq:
Great, I really like this! Factored out TSCH defs in cc2538 for instance etc.
Hard to assess by just looking at the code.

What HW was this tested on? I'm fine with taking the risk to break a platform or two here, we can fix later.

g-oikonomou:
I tested zoul and launchpad/cc2650. Joakim tested this indirectly by testing #204 for nrf52dk

simonduq:
Let's get this in then!

PR #221, opened 2017-12-03
Memory barriers, critical sections and mutexes

g-oikonomou:
This pull provides APIs and (some) implementations for:

* Memory barriers: Implemented for CC2538, CC13xx/CC26xx and Sky/msp430. nrf52dk provides its own copy of CMSIS, so it should be straightforward to provide an implementation for this platform too in the future. It's just a case of making `memory_barrier()` emit the CM3 `DMB` instruction.
* Critical sections: Platform-independent functions to enter/exit a critical section. Automatically apply memory barriers too if an implementation has been provided.
* Mutexes:
  * We provide an API and platform-independent implementations of mutex locking/unlocking. These rely on turning off the master interrupt.
  * We allow the platform code to override the generic implementations by providing CPU-specific functions. We provide such for CC2538 and CC13xx/CC26xx.

Requires #194


g-oikonomou:
Tested on Zoul and Launchpad CC2650, both with the Cortex-M specific implementation as well as the generic library.

simonduq:
Awesome work!

joakimeriksson:
I have tested on the NFR and it seems to work as fine as in "master" (or develop). So from NFR side it should be good to go!

PR #222, opened 2017-12-05
Clarify sensniff doc: It is also distributed as part of Contiki-NG

g-oikonomou:
Just a small doc improvement for Sensniff


simonduq:
Thanks!

PR #223, opened 2017-12-08
Disable RPL BR Web server on Sky

simonduq:
Currently the BR compiles on Sky leaving just a few bytes of ROM free. This disables the Web server (and in turn TCP) so save some kBs and leave room for future developments.


examples/rpl-border-router/Makefile
```diff
@@ -13,9 +13,17 @@ else
...
 include $(SOURCES_DIR)/Makefile
+# Enable Web server by default
+WITH_WEBSERVER ?= 1
 
 PROJECTDIRS += $(SOURCES_DIR)
-PROJECT_SOURCEFILES += httpd-simple.c webserver.c border-router-common.c
+
+ifeq ($(WITH_WEBSERVER),WITH_WEBSERVER)
+httpd-simple.c webserver.c
+endif
```
g-oikonomou: There is no reason to compile the web server conditionally

g-oikonomou: (that I can immediately think of, at least...)

g-oikonomou:
In fact I would rather the entire thing happened through headers. It should be possible to just change `project-conf.h` to make the line below conditional:

```
#define BORDER_ROUTER_CONF_WEBSERVER 1
```

and then define as 0 in sky's `target-conf.h`

examples/rpl-border-router/Makefile
simonduq: The reason was https://github.com/contiki-ng/contiki-ng/blob/develop/examples/rpl-border-router/common/httpd-simple.c#L46

But maybe we just remove this check so we can configure everything from headers again?

simonduq:
Yep I agree, see above

simonduq:
Pushed a fix, now fully header-based configuration

g-oikonomou:
Yep that's much much cleaner! Just need to change `project-conf.h` so it only defines conditionally.

Issue #24, opened 2017-05-30
Generic routing protocol API


simonduq:
Define routing drivers, and make RPL replaceable with e.g. RIP-NG, LoadNG or anything

simonduq:
Done via https://github.com/contiki-ng/contiki-ng/pull/224

PR #224, opened 2017-12-09
RPL as just a module: generic routing API

simonduq:
This PR adds `NETSTACK_ROUTING`, for routing protocol selection. The current options are "RPL Lite", "RPL Classic" and "Null Routing". Addresses https://github.com/contiki-ng/contiki-ng/issues/24

The API is still very much based on what RPL needs but can obviosly be extented for new protocols.

One side effect is that the source routing state management is now a module of uip, refactored from both RPL implementations. This was needed for external modules to be able to print the routing state with no knowledge of RPL. In the future I can imagine part of the SRH (IPv6 source routing header) creation and parsing also move from RPL to uip.


simonduq:
This one is now ready :)

simonduq:
This one is now ready :)

simonduq:
Yeah, successfully rebased and resolved 10s of conflicts! Anybody for a review? :p

PR #226, opened 2017-12-10
Add libraries and example for common data structures

### LINKED ISSUE #208: Libraries for common data structures
It would be nice to provide libraries with implementations of common data structures. In the first instance, I am thinking doubly linked lists, circular lists, queues, stacks. Naturally, there is nothing preventing us from providing libraries for more advanced data structures.

g-oikonomou:
This pull adds libraries for:

* Stacks
* Queues
* Circular, singly-linked lists
* Doubly-linked lists
* Circular, doubly-linked lists

The libraries for stacks and queues are merely wrappers around our existing list library, since it already provides all necessary functions.

The remaining ones are implemented from scratch and have no such dependency.

This pull also adds an example demonstrating / testing the above.

Closes #208

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,3 +31,4 @@ env:
   - TEST_NAME='compile-nxp-ports'
   - TEST_NAME='doxygen'
   - TEST_NAME='compile-tools'
+  - TEST_NAME='native-runs'
--- a/examples/libs/data-structures/Makefile
+++ b/examples/libs/data-structures/Makefile
@@ -1,0 +1,7 @@
+CONTIKI_PROJECT = data-structures
+
+all: $(CONTIKI_PROJECT)
+
+CONTIKI = ../../..
+
+include $(CONTIKI)/Makefile.include
--- a/examples/libs/data-structures/data-structures.c
+++ b/examples/libs/data-structures/data-structures.c
@@ -1,0 +1,383 @@
+/*
+ * Copyright (c) 2017, George Oikonomou - http://www.spd.gr
+ * Copyright (c) 2017, James Pope
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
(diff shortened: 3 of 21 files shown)


g-oikonomou:
I've added doubly-linked lists. This is ready from my PoV

simonduq:
What would you think about adding a test, e.g. by declining the example with the `unit-test` module, e.g as @yatch did in: https://github.com/contiki-ng/contiki-ng/blob/develop/tests/13-ieee802154/code-6tisch/test-sixp.c

g-oikonomou:
I had not noticed those. Are they relying on platform cooja or can they also be sky simulations? Can they be executed on native?

simonduq:
The 6top tests are on platform Cooja but for sure Sky would work just as well (albeit slower). This is the .js that checks the output:
https://github.com/contiki-ng/contiki-ng/blob/develop/tests/13-ieee802154/js/sixtop-test.js

Native also sounds like a great option, you could simply use `Makefile.script-test`, compile the example, run it and grep for failure :)

g-oikonomou:
I've added unit tests (sky sims) for everything except circular, doubly-linked lists. Have a look if you want. I'll probably add native tests too.

@yatch I am absolutely in love with this unit test lib 😍

g-oikonomou:
Now testing all libraries in cooja/sky + in platform native. AFAIK this is ready for review.

As a side-effect, I've added a new travis job (`08-native-runs`), where we can chuck other tests using platform native in the future.

yatch:
@g-oikonomou I love `unit-test`, too. Thank you, @nvt, for this cool stuff :wink:

g-oikonomou:
Now requires #246

nvt:
It's nice to extend the library with some fundamental data structures, and I'm glad that to see that the unit test module is being used more. 👍

simonduq:
Ah sorry for missing that earlier, but one last thing would be to add the new example as compile test: I believe we have all examples covered by tests on at least one platform, would be a pity to break this rule!

Other than that this looks ready to me, documented and tested :)

g-oikonomou:
Added for native, sky, zoul, nrf52dk

PR #227, opened 2017-12-10
uipbuf: added some documentation

joakimeriksson:
Added some documentation - some of which indicate feature that can be added (and is implemented in our code-base).

### DIFF
--- a/os/net/ipv6/uipbuf.h
+++ b/os/net/ipv6/uipbuf.h
@@ -35,33 +35,106 @@
 
 #include "contiki.h"
 
-/* Get the next header given the buffer - start indicates that this is
-   start of the IPv6 header - needs to be set to 0 when in an ext hdr */
-
+/**
+ * \brief          Get the next IPv6 header.
+ * \param buffer   A pointer to the buffer holding the IPv6 packet
+ * \param size     The size of the data in the buffer
+ * \param protocol A pointer to a variable where the protocol of the header will be stored
+ * \param start    A flag that indicates if this is expected to be the IPv6 packet header or a later header (Extension header)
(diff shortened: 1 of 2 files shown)


os/net/ipv6/uipbuf.h
```diff
@@ -35,33 +35,98 @@
 
 #include "contiki.h"
 
-/* Get the next header given the buffer - start indicates that this is
-   start of the IPv6 header - needs to be set to 0 when in an ext hdr */
-
+/* \brief          Get the next IPv6 header.
```
g-oikonomou: I think you need double-star here for doxygen to generate HTML for this

os/net/ipv6/uipbuf.h
```diff
@@ -35,33 +35,98 @@
...
+ *
+ */
 enum {
-  UIPBUF_ATTR_LLSEC_LEVEL,
-  UIPBUF_ATTR_LLSEC_KEY_ID,
-  UIPBUF_ATTR_INTERFACE_ID,
-  UIPBUF_ATTR_PHYSICAL_NETWORK_ID,
-  UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS,
-  UIPBUF_ATTR_FLAGS,
+  UIPBUF_ATTR_LLSEC_LEVEL,      /* Control link layer security level. */
```
g-oikonomou: I think (didn't test) you need `/**<` here for doxygen to generate HTML

g-oikonomou:
Thanks :)

os/net/ipv6/uipbuf.h
joakimeriksson: Maybe I should try adding that then!

PR #229, opened 2017-12-11
Sicslowpan UDP decompression fails when UIP_LLH_LEN != 0

laurentderu:
Sicslowpan places the UDP header at the wrong place when decompressing a packet when UIP_LLH_LEN is non-zero.

This is due to IPPAYLOAD_BUF macro adding a second time UIP_LLH_LEN : The buffer received by uncompress_hdr_iphc() is already offset by UIP_LLH_LEN as buffer is assigned to UIP_IP_BUF, which starts at UIP_LLH_LEN.


simonduq:
Seems right to me! Would like a confirmation from @joakimeriksson

PR #231, opened 2017-12-11
Native monotonic timer

laurentderu:
Currently the native platform uses gettimeofday() to get the time of all the timers, however gettimeofday() is not an absolute monotonic time source. When the system clock is updated (either because of drift, DST or user configuration) the time returned is shifted. This breaks completely the Contiki-NG timers, resulting in timers already expired before even starting or times scheduled billions of seconds later in time...

With this PR, the native platform uses clock_gettime(CLOCK_MONOTONIC). This function is guaranteed to always return a monotonic time, independent of any clock change on the platform.


g-oikonomou:
I don't particularly like this one. Other than linux and OSX, did you have any other OSs in mind?

g-oikonomou:
`clock_gettime(CLOCK_MONOTONIC, ...);` is in fact available in OSX, but only from 10.12 onwards. Unless I am misreading something, this PR uses it for any OSX version, right? If so, under what conditions will this PR use the alternative `gettimeofday()` option?

This is what I am proposing: https://github.com/g-oikonomou/contiki-ng/blob/21eeb7b0f63ce68547d5a9017c71227e66ae17a7/arch/platform/native/clock.c

I think that the conditional should in fact be

```
#ifdef __MAC_OS_X_VERSION_MIN_REQUIRED
#if __MAC_OS_X_VERSION_MIN_REQUIRED < 101200
```

laurentderu:
clock_gettime() is POSIX compliant, but indeed it took some time to be implemented in OSX.  

For OSX < 10.12 I have an implementation using mach_timebase_info() instead, but I removed it from this PR to not make the code too complex :) I can re-add it and use the OSX version guard to enable it or not. 

Many BSD flavours seem to have it, I honestly did not check Windows...

g-oikonomou:
Yep gocha.

So the way this PR currently stands, we use `clock_gettime()` under all circumstances, under some of which (old OSX) we will fail. We never use the `gettimeofday()` alternative (unless obscure OS!).

We can use either your monotonic implementation for < 10.12, or `gettimeofday()`. I am really not bothered, as long as we can actually compile in this OSX version.

arch/platform/native/Makefile.native
```diff
@@ -19,6 +19,10 @@ else
 CONTIKI_TARGET_SOURCEFILES += tun6-net.c
 endif
 
+ifeq ($(HOST_OS),Linux)
+TARGET_LIBFILES += -lrt
```
g-oikonomou: Do we really need this? Travis seems to be passing even without?

laurentderu: Yes, but only for glibc versions < 2.17 :)

g-oikonomou: Aha! :^)

simonduq: Any particular reason for using an old glibc?

laurentderu: Some OpenWRT targets are not using the most recent glibc if I remember correctly.

laurentderu:
Updated to check the MACOS version and keep gettimeofday() for version lower than 10.12

arch/platform/native/Makefile.native
simonduq: Ah ok, fair enough!

g-oikonomou:
Still think it'd be cleaner to have the #if conditional in one place and then use a single function (two flavours) to populate our own time representation struct. Then `clock_time()` and `clock_seconds()` won't need to be conditional.

laurentderu:
Indeed, I overlooked that part. I updated the PR accordingly and also replaced the hardcoded 1000 factor by CLOCK_CONF_SECOND both in clock.c and rtimer_arch.c

laurentderu:
Done.

PR #232, opened 2017-12-11
Update protocol field in remove_ext_hdr()

laurentderu:
The current remove_ext_hdr() do remove the extension header, but keep the protocol field of the IPv6 header unchanged. This force the calling function to manually set the protocol field afterwards.

With this PR, remove_ext_hdr() updates also the protocol field which is much safer.


joakimeriksson:
Prototype field = Protocol field (or NextHeader Type).

laurentderu:
Yes, I blame the autocorrecter :)

PR #233, opened 2017-12-11
Additional cleanup when becoming DODAG Root (rpl-classic)

laurentderu:
When a node has joined a DODAG as an intermediate router then later becomes DODAG Root, additional cleanup tasks are needed :

* The default route must be removed
* Any DAO retransmission must be cancelled
* Incoming DAO Ack should be ignored

Additionally, when doing a local repair, any pending DAO retransmission must be cancelled too, as there is no more a preferred parent to send it to.


joakimeriksson:
Looks ok from my point of view.

PR #235, opened 2017-12-11
Always try to find the best dag

laurentderu:
Currently select_dag() search first in the current DODAG, then and only if there are no suitable parent, it will search in the other DODAGS. This means that a node will always select the current dag, even if it's rank is worse than the rank in the other DODAGs it belongs to.
With this fix, all the DODAGs are explored and the the best parent across all is then selected.


simonduq:
Looks good. Please take a look at indentation, looks messed up

laurentderu:
Indeed, I cherry-picked while ignoring whitespace diffs, fixed now

PR #236, opened 2017-12-11
Do RPL probing on all the known DAGs (rpl-classic)

laurentderu:
The current RPL probing function does probing only in the current dag. With this PR, all the known dodag are probed in a round-robin manner. This improves a lot DODAG switching.

Note: In the orignal PR on Contiki, @simonduq suggested to move the select_dag() call into the probing function, if this is still relevant I will update the code accordingly :)


os/net/rpl-classic/rpl-timers.c
```diff
@@ -457,11 +457,29 @@ get_probing_target(rpl_dag_t *dag)
...
+  int new_dag = instance->last_dag;
+  do {
+    new_dag++;
+    if(new_dag >= RPL_MAX_DAG_PER_INSTANCE) {
+      new_dag = 0;
+    }
+    if(instance->dag_table[new_dag].used) {
+      dag = &instance->dag_table[new_dag];
+    }
+  } while(new_dag != instance->last_dag && dag == NULL);
```
simonduq: wouldn't it be a lot easier to read with a for loop over all DAGs, and a break when found?

laurentderu: We have to find the next dag to probe, with a simple loop we will always return the first dag. Adding the wrap around and the check for unused dags will make the code as complex as this one (unless I'm missing something)

simonduq: right, realized what I had in mind would actually result in more iterations, so, not ideal :p Let's stick to what you have!

PR #237, opened 2017-12-11
Don't autoconfigure address on RPL Root (rpl-classic)

laurentderu:
Currently a DODAG Root with a manually configured address will lose this address when rpl_set_prefix() is called and/or will receive an auto-configured address based on the IID and the prefix.

A DODAG Root should always be administratively configured and not rely on auto-configuration.


g-oikonomou:
👍

PR #239, opened 2017-12-11
rpl_ext_header_remove() should only remove RPL Hop-by-Hop extension header

laurentderu:
Currently, rpl_ext_header_remove() remove any existing Hop-by-Hop extension header, even if it does not contains RPL options in it. This conflict with other protocols which relies on HBH to add extra informations, like MLD.
With this PR, only RPL HBH header are removed, the others are kept untouched.


simonduq:
Two comments:
* I'm now confused with the `UIP_PROTO_DESTO`: does it really mean to return? What would you think about an explicit `return`, before the switch `default` statement?
* Doesn't rpl-lite also need the same fix?

laurentderu:
Actually the comment of the case contradict the use of return, instead the code should break to parse the next  header (if any). However this is hard to get as I could not find any usage of the Destination Option in the code.

And RPL-Lite does indeed need the same fix !

simonduq:
Agree, looks like we want a break in the `UIP_PROTO_DESTO` case. Would be great if you could add it and also apply the fix to RPL Lite in this PR!

laurentderu:
Done.

PR #240, opened 2017-12-11
RDNSS and DNSSL do not rely on O Flag

laurentderu:
According to NDP RFC, the 'O' flag in Router Advertisement messages indicates that some or all the network configuration is provided by another configuration protocol (usually DHCP). However, in Contiki-NG the meaning of the flag has been inverted for RDNSS and DNSSL options.

This PR fixes this by setting the 'O' flag to 0, and always processing the incoming RDNSS option.


laurentderu:
Just to clarify a bit, in RFC 4861, the 'O' flag means :

>       O              1-bit "Other configuration" flag.  When set, it
>                      indicates that other configuration information is
>                      available via DHCPv6.  Examples of such information
>                      are DNS-related information or information on other
>                      servers within the network.

The current code sets the value of the 'O' flag if the RA message contains a RDNSS or a DNSSL field, which is the opposite of the RFC which require to set it only if the information is not in the RA message.

The second part of the PR is the handling of the incoming RDNSS and DNSSL options in RA message. First, checking that the incoming 'O' flag is set to handle the option is plain wrong, as explained above.
Also, RFC-6106, which describes the RDNSS option, clearly indicates that a host should handle configuration from both RA messages and DHCPv6 messages, see [section-5.3.1](https://tools.ietf.org/html/rfc6106#section-5.3.1) last paragraph.

simonduq:
OK, I think I agree. But the PR contains a `PRIx32` macro, we need to decided whether we want these or not.

laurentderu:
Replaced the PRIx32 with an explicit cast, jn any case it will be cleaner to do the eventual switch to PRI in a dedicated PR.

PR #241, opened 2017-12-11
Fix compilation error when not in Router mode

laurentderu:
When compiling in non Router mode (UIP_CONF_ROUTER set to 0), the compilation aborts due to three warnings in uip_nd6.c : Unused variable and incompatible format.

To solve the incompatible format, I used PRI macros to be future safe, otherwise this error would pop up again when compiling for 64bits platform (or 16bits too ;) )


g-oikonomou:
That's standard C99 is it not? Confident that it won't break other possible toolchains? Is it not safer to cast to unsigned long and keep the %lu format?

joakimeriksson:
I do think they are C99 but not C89. See http://www.delorie.com/djgpp/doc/libc/libc_623.html

laurentderu:
They are C99, but it took some time to ave all the C libraries to implement them, although they are the recommended way to safely print values.
Multicast example suffer from the same problem, when building using llvm on 64 bits, the compilation fails too.

joakimeriksson:
Maybe we should check the IAR and some other compilers?

g-oikonomou:
I remember for certain that SDCC didn't have them. Irrelevant now of course, but once upon a time we were saying we wanted to be C90. Do we want to change this to C99? If so, how long before we start passing structs by value as arguments? ;)

joakimeriksson:
Yes, we should in fact take a strict policy decision on this one. To use or not to use PRI macros.

laurentderu:
https://github.com/contiki-ng/contiki-ng/wiki/Contributing requires already C99 :)

simonduq:
shouldn't `uip_ntohl` return a long and `uip_htons` a short instead of fix-width integers? Would basically resolve the issue

g-oikonomou:
> https://github.com/contiki-ng/contiki-ng/wiki/Contributing requires already C99 :)

Yeah I'd noticed that, but decided to not comment :D We do want inline functions, so we can no longer say we are C90.

> shouldn't uip_ntohl return a long and uip_htons a short instead of fix-width integers? Would basically resolve the issue

The libc variants accept and return fixed-width integers. We should stick to that.

On the topic of cross-platform printing of fixed-width, I am only aware of two options: Either explicitly promote / typecast (this is definitely portable across versions of that standard as well as toolchains) or use the macros recommended here (which I originally thought were not ANSI C, but Laurent said they were I checked again). Since they are standard C, I am happy for us to adopt them widely. Good point about IAR, perhaps someone can check?

joakimeriksson:
I had someone to check and it seems not supported in the IAR compiler. So I guess we might need to avoid PRIu32, etc. to avoid issues with other compilers.

laurentderu:
@joakimeriksson That's surprising, I checked the doc of IAR for ARM, and it is specified that DLIB contains inttypes.h :

> inttypes.h
> This include file defines the formatters for all types defined in stdint.h to be used by
> the functions printf , scanf , and all their variants.

And this comes from a rather old version of IAR (5.2). Maybe the C99 mode should be explicitly enabled or a macro needs to be set ?

A possible workaround is to define the macros if they don't exist.

joakimeriksson:
Yes, probably correct! I think he (who tested) might have missed the include!

simonduq:
So, shall we make a decision regarding use of `PRI` macros? Personally, I have no opinion long term, but am a bit reluctant in having them introduced here and there via new PRs. If we accept these macros, I'd like to see a PR that tries to adopt them generally (probably not too hard to spot, it's conversions to `int`, `unsigned`, `long`, `unsigned long` as `PRINTF` and `LOG_OUTPUT` arguments).

laurentderu:
Replaced the PRIu32 with an explicit cast, jn any case it will be cleaner to do the eventual switch to PRI in a dedicated PR.

simonduq:
Great, thanks!

PR #242, opened 2017-12-11
Add missing UIP_LLH_LEN

laurentderu:
At a few places, UIP_BUFSIZE is used without removing UIP_LLH_LEN, which could lead to buffer overflow.
Most of the time the macros used to calculate header or packet size already take into account UIP_LLH_LEN, these are the only places where UIP_BUF_SIZE is used directly and UIP_LLH_LEN is not subtracted.


joakimeriksson:
An interesting question from my perspective is if there is a way to add a new define for UIP_BUFSIZE that include the constant LLHLEN so that it cannot be missed. When do we process IP-packets and make use of the LLH? Maybe we just should subtract it once and for all as we always are LLHLEN up in uip_buf and have LLHLEN less than UIP_BUFSIZE - the code would look nicer i think?!

laurentderu:
It would be much safer like that, and also refactor all the definition of UIP_BUF, UIP_IP_BUF, UIP_UDP_BUF, ... into one header file would help avoiding missing to update one of them.

simonduq:
Doesn't this fix also apply to rpl-lite?

joakimeriksson:
So, let's put this refactoring of UIP_BUF related defines to the roadmap. It should also likely go into the uipbuf.h since we now have a separate file for the buffer?

laurentderu:
@simonduq indeed, this also needs to be applied to rpl-lite. Another reason to do a refactoring of UIP_BUFSIZE, UIP_BUF, ... macros.

mbaddeley:
I was literally having this issue in Contiki (old) yesterday, it's pretty easy to forget about the LLHLEN.

simonduq:
OK the PR looks good to me, ready to merge if extended to also cover RPL Lite

laurentderu:
Done.

PR #255, opened 2017-12-16
Fix list_insert bug

### LINKED ISSUE #254: List: list_insert bug
The function `list_insert` in `list.c` is buggy: when `previtem` is null, it pushes the new element (which (1) removes any old instance and then (2) inserts the new element). But when `previtem` is non-null, it just adds the new item without removing any old instance. Could in duplicate elements in the latter case.

Only reporting as `bug/low` because the function is currently not used in the codebase.

(report by Nikolai Kosmatov)

g-oikonomou:
This PR

* Fixes #254 
* Adds lists to the data structure unit test
* Tidies-up `list.c` a little bit

### DIFF
--- a/os/lib/list.c
+++ b/os/lib/list.c
@@ -302,7 +302,7 @@ list_insert(list_t list, void *previtem, void *newitem)
   if(previtem == NULL) {
     list_push(list, newitem);
   } else {
-  
+    list_remove(list, newitem);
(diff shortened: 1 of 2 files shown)


simonduq:
When I saw the PR title I thought "the fix needs a new CI tests" and now see you already have it :)
👍 
I'd wait for one more pair of eyes to approve, this is a sensitive module.

PR #251, opened 2017-12-15
BLEach a fully open-source IPv6-over-BLE stack for the TI CC26xx platform

spoerk:
This pull request implements **BLEach**, a fully open source IPv6-over-BLE stack for the Texas Instruments CC26xx hardware platform. BLEach in Contiki NG currently implements support for IPv6-over-BLE node devices according to the [RFC 7668](https://tools.ietf.org/html/rfc7668) specification. Devices running this IPv6-over-BLE stack are interoperable with any RFC 7668 compliant border router (e.g. Raspberry Pi, smartphone, laptops, ...).

Features:
- IPv6-over-BLE node implementation compliant to RFC 7668
- BLE link layer support according to the [Bluetooth Specification v4.1](https://www.bluetooth.com/specifications/bluetooth-core-specification/legacy-specifications) for _BLE advertisement_, _BLE connection slave_.

The website [www.iti.tugraz.at/BLEach](http://spoerk.github.io/contiki/?utm_source=contiki-pull-request) provides a more detailed description of the features and the code structure of BLEach.
The code documentation for BLEach can be found in this [README.md](https://github.com/spoerk/contiki/blob/master/cpu/cc26xx-cc13xx/net/README.md).
For a detailed description and evaluation of BLEach, see our paper [BLEach: Exploiting the Full Potential of IPv6 over BLE in Constrained Embedded IoT Devices](https://wwwpub.zih.tu-dresden.de/~mzimmerl/pubs/spoerk17bleach.pdf), ACM SenSys'17.

This implementation includes new implementations of the _radio_, and _MAC_ layer of Contiki that are tailored to support IPv6-over-BLE communication. Therefore, BLEach is agnostic to the application running in Contiki; developers may run the same application using IPv6 over IEEE 802.15.4 or IPv6 over BLE by simply changing the project's configuration file at compile time.

This pull request provides one example project: a simple UDP client that exchanges UDP packets over IPv6 over BLE with a server.

PR #256, opened 2017-12-16
Split radio APIs for different radio types

g-oikonomou:
Following discussion in #251, this is what a new radio API might look like.

* We would have two types of radio driver: 
  * `radio_802154_driver_t`
  * `radio_ble_driver_t`
* We would have three header files:
  * `radio-802154.h` with specifics to IEEE 802.15.4 radios
  * `radio-ble.h` with specifics to BLE radios
  * `radio.h` would only contain common macros, data types etc. This file would no longer be included directly, code will have to include one of the two headers above, depending what radio type it needs to manipulate.
* We would extend netstack. We would have:
  * `NETSTACK_RADIO_802154`. This is what to use in code that will only work with .15.4 radios (e.g. in CSMA and TSCH)
  * `NETSTACK_RADIO_BLE`. This is what to use in code that will only work with BLE radios
  * `NESTACK_RADIO` would now map to one of the above and should be used by code that will work with either. An example is `netstack_init()`, which can still simply do `NETSTACK_RADIO.init()`
* We could move everything to `dev/radio/`
* We would introduce a BLE nullradio driver.

By no means complete, just throwing it out there for comments. Most likely more common things could stay in `radio.h` and some things in `radio.h` would have to be moved to the tech-specific header.

Many of the functions are common (`init`, `on`, `off`, `get_`, `set_`). Maybe it would also be nice to come up with a clean way for the 15.4 or BLE driver to pull those in from somewhere else.


simonduq:
@spoerk I'd like to hear your thoughts about this. To what extent does is the current radio interface inadequate for BLE? Is it just that the cc2650 BLE only needs a subset of the API? Are the semantics of the API any different? Would the API need new primitives for BLE specifically?

spoerk:
I think this is a great idea. During the development of #251, I focused on being compliant to the existing architecture of Contiki and therefore used several workarounds so that a BLE radio can be supported. Having a dedicated BLE radio interface would make further BLE development easier and the existing BLE radio code better to understand.

The main problem with BLE is that the BLE radio interface is fundamentally different to the IEEE 802.15.4 interface. The BLE radio is usually a dedicated radio chip that autonomously handles timing of transmissions, packet (re-)transmissions, and buffer handling. The application core has no immediate feedback on scheduled or ongoing communication and can only configure the BLE communication by providing a set of BLE parameters to the BLE radio core.
To provide the parameters to the radio core and get decoupled feedback, the BLE specification defines a standardized interface between application and radio core, which is called **Host Controller Interface (HCI)**.

I suggest that we change the `radio-ble.h` to be compliant to the HCI specification.
This would make it easier to port Contiki to new BLE platforms. 
Having this standardized BLE radio interface, it would also make it easier to extend the current BLE functionality of Contiki (e.g., use the BLE advertising and scanning to exchange large data packets).

PR #257, opened 2017-12-16
Add GPIO Hardware Abstraction Layer

g-oikonomou:
This pull introduces a GPIO HAL. Fundamentally we do two things:

* We define a GPIO API:
This provides platform-independent ways to configure/set/read/write/clear/toggle one or multiple pins, enable/disable pin interrupts etc.

  I have decided to be as flexible as possible here. For each function, platform developers have the option to provide a function-like macro, or a `static inline` implementation, or a proper global symbol. The reasoning behind this is: CC2538 provides function-like macros that do direct register access: I want to keep this without losing efficiency. CCxxwares provide some `static inline` functions, I want to be able to keep those too without losing efficiency. So I'm delegating the choice to the port developer.

  I am also providing fallback, software based implementations for the toggle functions if the hardware does not have anything better to offer.

* We move GPIO event handling to `os/`:
Instead of each port having to define a callback mechanism for GPIO event handlers, we do this in a platform-independent fashion. Platforms now simply need to call `gpio_hal_event_handler()` from inside GPIO ISRs, passing as argument a pin mask of the pins that have their corresponding interrupt flag asserted. For each of those pins, this function calls an event handler previously registered through `gpio_hal_register_handler()`.

Implemented for all CC13xx/CC26xx and CC2538 platforms/boards. 

Sky and nrf52dk if/when I can be bothered 😆 . jn516x is not going to happen 😉 Hence the `help wanted` label.

This is untested WIP but I thought I'd throw it out there for feedback.


g-oikonomou:
Implemented for CC2538 and CC13xx/CC26xx. Tested with Zoul/Firefly and Launchpad/cc2650.

Now with example and travis tests! 😃 

I won't be doing any serious development for the next couple of weeks, so porting for other platforms will have to wait.

I'm removing the WIP label, AFAIK this is ready for review/pull.

g-oikonomou:
This is now ready for review/merge

arch/platform/zoul/firefly-reva/board.h
```diff
@@ -99,9 +99,18 @@
...
-#define LEDS_BLUE     (1 << 3) /**< LED2 (Blue)  -> PD3 */
-#define LEDS_RED      (1 << 5) /**< LED3 (Red)   -> PD5 */
+#define LEDS_GREEN_PIN      4
+#define LEDS_GREEN    (1 << LEDS_GREEN_PIN) /**< LED1 (Green) -> PD4 */
+
+#define LEDS_BLUE_PIN       3
+#define LEDS_BLUE     (1 << LEDS_BLUE_PIN)  /**< LED2 (Blue)  -> PD3 */
+
+#define LEDS_RED_PIN        5
+#define LEDS_RED      (1 << LEDS_RED_PIN)   /**< LED3 (Red)   -> PD5 */
```
simonduq: should this use `gpio_hal_pin_to_mask`?

g-oikonomou: I'd rather not at this stage:
* Semantically those `LEDS_RED_PIN` are not of type `gpio_hal_pin_t` but a magic number used internally by the port's LED driver.
* Plus, this is fundamentally a configuration file that is meant to help configure `gpio-hal.h` and I don't want to mess about with circular include situations now that this is basically ready.
* Plus I've got something else in mind for LEDs... :)

simonduq: ok sure!

PR #262, opened 2017-12-17
Link without using an archive file

g-oikonomou:
For reasons discussed long ago and that have to do with how `ld` handles weak symbols, it was decided that it was preferable to link using an explicit list of object files, instead of adding object files to an archive first and then linking using this archive. This was applied for the CC2538, but not for CC13xx/CC26xx.

This pull applies this link logic to all CM3s. For this to work with the Contiki-NG build system, we need `.SECONDEXPANSION:`, so this pull moves that to the common CM3 Makefile.

### DIFF
--- a/arch/cpu/arm/cortex-m/cm3/Makefile.cm3
+++ b/arch/cpu/arm/cortex-m/cm3/Makefile.cm3
@@ -17,7 +17,9 @@ CPU_STARTFILES = ${addprefix $(OBJECTDIR)/,${call oname, $(CPU_START_SOURCEFILES
 ### Compilation rules
 CUSTOM_RULE_LINK = 1
 
-%.elf: $(CPU_STARTFILES) %.co $(PROJECT_OBJECTFILES) $(PROJECT_LIBRARIES) contiki-$(TARGET).a $(LDSCRIPT)
+.SECONDEXPANSION:
+
+%.elf: $(CPU_STARTFILES) $$(CONTIKI_OBJECTFILES) %.co $(PROJECT_OBJECTFILES) $(PROJECT_LIBRARIES) $(LDSCRIPT)
 	$(TRACE_LD)
 	$(Q)$(LD) $(LDFLAGS) ${filter-out $(LDSCRIPT) %.a,$^} ${filter %.a,$^} $(TARGET_LIBFILES) -lm -o $@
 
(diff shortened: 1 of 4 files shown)


simonduq:
Am curious: why does CM3 need its own link rule?

g-oikonomou:
It's not strictly a CM3 thing, it would affect any port using weak symbols. See below the relevant excerpt, full details in contiki-os/contiki#1102.

> Another issue with the Contiki target library is that it may alter the apparent behavior of the weak vs. strong symbols, because of the way `ld` handles archives, which may make it discard archive object files containing strong versions of referenced symbols:
>
> *  If a symbol has a weak and a strong version in this library, both inside the same object file, then the linker uses the strong definition.
> * If a weak symbol in this library has a strong counterpart in an object file outside, then the linker uses the strong definition.
> * If a strong symbol in this library is inside an object file containing other referenced symbols, and has a weak counterpart anywhere, then the linker uses the strong definition.
> * If a strong symbol in this library is the only symbol referenced in its object file, and has a weak counterpart in an object file outside, then the linker uses the strong definition if this library is linked first, and the weak one otherwise.
> * If a strong symbol in this library is the only symbol referenced in its object file, and has a weak counterpart in another object file in this library, then the linker uses the definition from the first of these objects added when creating this archive.
> * If a symbol has a weak and a strong version, one in this library, and the other in another library, then the rules are the same as if both were in the Contiki target library.

Now consider this:
```
void GPIOIntHandler(void) __attribute__ ((weak, alias("IntDefaultHandler")));
```

This bullet is of particular relevance here:
> If a strong symbol in this library is the only symbol referenced in its object file, and has a weak counterpart in an object file outside, then the linker uses the strong definition if this library is linked first, and the weak one otherwise.

Now imagine:

* A file that is _not_ inside the library and that defines a weak symbol for `GPIOIntHandler` as above.
* A second file that defines a strong symbol for `GPIOIntHandler` handler and also some other function.

If the latter file is a member of the library and the other function is referenced, then the linker will use this object file. But if the `GPIOIntHandler` symbol is the only symbol referenced, then the linker will choose whatever it encounters first. Because we link the library last, the linker will in this case choose the weak symbol defined in the former object above, resulting in a firmware that contains the weak symbol.

simonduq:
But I believe this explains why not to use an archive, right?
My question was more why the platform happens to (even before this PR) have its own linking rule.

simonduq:
Noticing now that most platforms actually have their own link.

simonduq:
This is ready, right?

g-oikonomou:
That's what this explains yes.

Both platforms had always had their own link rule. Firstly because I wanted to specify objects that need to appear at the start of the list of things we pass to `ld`. Secondly because for CC2538 we are generating the linker script as part of the build, and the link rule needed to have the linker script generation as a pre-req.

This is ready, yes. I'm not sure if nrf52dk uses weak symbols too. I'll look at making this more generic in the future.

simonduq:
Right, now I started see why the custom rules are needed. Thanks!

PR #263, opened 2017-12-18
CI tests: cleaner compilation artifacts

simonduq:
This cleans up CI tests artifacts, i.e., uses more consistent extensions (which are gitignored)

### DIFF
--- a/tests/00-doxygen/Makefile
+++ b/tests/00-doxygen/Makefile
@@ -30,15 +30,15 @@ DOCDIR=../../tools/doxygen
 all: clean summary
 
 doxygen:
-	@make -C $(DOCDIR) 2> doxygen.runerr > doxygen.runlog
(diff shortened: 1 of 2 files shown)


tests/00-doxygen/Makefile
```diff
@@ -30,15 +30,15 @@ DOCDIR=../../tools/doxygen
...
+	@make -C $(DOCDIR) 2> doxygen.err > /dev/null
 
 summary: doxygen
 	@( \
 		1> summary; \
-		if [ -s doxygen.runerr ] ; then \
+		if [ -s doxygen.err ] ; then \
 			echo "Doxygen: TEST FAIL" | tee summary; \
 			echo "Errors:"; \
-			cat doxygen.runerr; \
```
g-oikonomou: Which is why using Make variables is often a good idea 😄

PR #266, opened 2017-12-20
Various native platform updates

laurentderu:
This PR contains bunch of small improvements for the native platform :

* Allow to change the duration of the select timeout (useful on low performance routers)
* Make the stdin select callback configurable (otherwise keeping stdin open prevent putting the process in background)
* Allow to change UIP_CONF_BYTE_ORDER
* Change uip_stats_t from short to int (memory is not a constaint on native, and 16bits counter is too short)


laurentderu:
ok, will do.

laurentderu:
and done :)

PR #270, opened 2017-12-21
Making RPL Border Router an OS module

simonduq:
This PR reworks the build system and moves the BR to `os/services`. To run as a border router, one now just has to add `MODULES += os/services/rpl-border-router` to the project makefile. The idea is that it becomes define a border router for any existing project, not necessarily starting from `examples/rpl-border-router`. The border example remains: it is simply an inclusion of the BR module, along with the good old example webserver.

Build system changes:
* Modules can now provide configuration macros through the file `module-macros.h`. This is a `.h` file only intended for definition of constants, independent from external constants. All `module-macros.h` files are included at compilation with `-imacros`.
* Projects can include modules with relative path, using `MODULES_REL`, instead of `PROJECTDIRS`. These will exactly behave as normal modules: files automatically compiled, makefile and macro file included.


g-oikonomou:
This is a very welcome change in principle. I see that the PR changes things so we always assume `AUTOSTART_ENABLE`, am I misreading this? Hence no longer the need of a separate make rule for the example's main `.c` file?

Did you check if automatic dependency generation lists files passed through `-imacros` into the `.d`?

simonduq:
Yes I figured we no longer need `AUTOSTART_ENABLE`, this seems like a relic of old days where Contiki was more APP-oriented. And I never liked the `.co` thing. Totally open to opposite views.

Yes the `.d` generation seems to handle `-imacros` properly :)

examples/rpl-border-router/native/Makefile.native
```diff
@@ -3,5 +3,6 @@ MODULES += os/services/slip-cmd
 MAKE_MAC = MAKE_MAC_OTHER
 MAKE_NET = MAKE_NET_IPV6
 
+PREFIX ?= fd00::1/64
```
laurentderu: Just a pedantic remark: a /64 prefix should be fd00::/64, without a trailing 1

simonduq: Right. But the address fd00::1 is also used by tunslip6, right? Then, would you rename the variable `PREFIX`? What is a suitable name?

os/services/rpl-border-router/Makefile.rpl-border-router
```diff
@@ -0,0 +1,10 @@
+CFLAGS += -DBUILD_WITH_RPL_BORDER_ROUTER=1
```
g-oikonomou: I don't see the need for this. It strikes me as though this could be controlled by the pre-processor through `-conf.h`

os/services/rpl-border-router/Makefile.rpl-border-router
```diff
@@ -0,0 +1,10 @@
...
+
+include $(CONTIKI)/Makefile.identify-target
+
+# Include 'embedded' module for non-native platforms
+ifneq ($(TARGET),native)
+	MODULES += os/services/rpl-border-router/embedded
```
g-oikonomou: What about native?

examples/rpl-border-router/native/Makefile.native
laurentderu: Following the code, PREFIX will become the IP Address of the tun interface (in rpl-native-router and tunslip6) and also be used to create the DODAG Prefix, so both are wrong :)

It could be something like :
TUN_ITF_IP?=fd00::1

os/services/rpl-border-router/Makefile.rpl-border-router
```diff
@@ -0,0 +1,10 @@
...
+endif
+# Include optional target-specific module
+MODULES += os/services/rpl-border-router/$(TARGET)
```
g-oikonomou: I am not very convinced about this one I'm afraid. We are going through a lot of effort to abstract various parts of `os` including peripheral HALs and all that. Are we sure we want to open this precedent of platform-specific code under `os`? Perhaps we can come up with a clean way of pulling this from within `arch/platform`?

g-oikonomou:
It feels like those `module-macros.h` files are basically configuration that users are not meant to modify. Would it make sense to merely include them through `project-conf.`h when the respective module is activated? The build system can then merely pass a `-D` to let us know that a module is active.

os/services/rpl-border-router/Makefile.rpl-border-router
g-oikonomou: Scrap the comment above.

g-oikonomou:
Just thinking aloud here. I am a little concerned about how the order of inclusion of things through `imacros` is not immediately obvious by looking at the source code, that's all.

examples/rpl-border-router/native/Makefile.native
simonduq: But the prefix part is also needed no?

os/services/rpl-border-router/Makefile.rpl-border-router
simonduq: No I think you were right weren't you?

os/services/rpl-border-router/Makefile.rpl-border-router
simonduq: Native is included through `MODULES += os/services/rpl-border-router/$(TARGET)`

os/services/rpl-border-router/Makefile.rpl-border-router
simonduq: Are you bothered only with the `sky` folder here or also `native`?

simonduq:
Right but then you have to add manually both something in make and in `project-conf`. Or maybe we do it form `contiki.h`, like we call init functions in `contki-main`?

As we use `-imacros` not `-includes`, all what `modules-macros.h` can really do is define constants. But we need to document that it may not rely on external defs.

If you think about it, we already have per-module makefiles, which one could already misuse with order-dependent definitions.

os/services/rpl-border-router/Makefile.rpl-border-router
g-oikonomou: OK now I see what you've done. Hmmm, yeah native is platform-dependent too isn't it. Not sure, I'll sleep over it! :D

simonduq: Yep no worries, nothing urgent, see this really as an initial proposal. I would really want the BR moved form `examples` to `os`, but still quite open how we actually do this. The sky code is so little code that I believe it would be easy to move out. Native is really different, as this is embedded BR vs NBR.

os/services/rpl-border-router/Makefile.rpl-border-router
g-oikonomou: My use-case here was an example where a user may wish to enable/disable the module depending on circumstances. It would be nice to be able to do this through header files to take advantage of dependency generation. This would mean that the example's makefile would always include the module in the build, only we'd get rid of it at link time.

This of course cannot work unless we know the module can always compile.

simonduq:
OK, new update: all really platform-specific code left was putchar for SLIP on sky -- it is now moved to arch, just like this already was the case for other platforms.

Only thing dependent on platform under `os/services/rpl-border-router` now is the `embedded` vs `naitve` folders. These share almost no code now with this PR -- maybe we should rather separate them again, as two distinct modules?

g-oikonomou:
> maybe we should rather separate them again, as two distinct modules?

Those things are strictly BR stuff, right? Since I cannot imagine a use-case where some other part of the code would want to include them as modules independently, I think best if we left them where we are merely as subdirs.

Let's get this one merged in as soon as it's ready 👍

simonduq:
It's ready from my PoV :)

PR #272, opened 2017-12-29
Update soc-rtc.c for use with Sensor Controller

necropaz:
ti_lib_aon_rtc_combined_event_config() should only use event AON_RTC_CH0 & 1. AON_RTC_CH2 is used for the CC2XXX sensor controller. If someone want to use the sensor controller on the cpu, AON_RTC_CH2 can't be used by the soc_rtc.c.

### DIFF
--- a/arch/cpu/cc26xx-cc13xx/dev/soc-rtc.c
+++ b/arch/cpu/cc26xx-cc13xx/dev/soc-rtc.c
@@ -86,7 +86,7 @@ soc_rtc_init(void)
   ti_lib_aon_event_mcu_wake_up_set(AON_EVENT_MCU_WU0, AON_EVENT_RTC_CH0);
   ti_lib_aon_event_mcu_wake_up_set(AON_EVENT_MCU_WU1, AON_EVENT_RTC_CH1);
   ti_lib_aon_event_mcu_wake_up_set(AON_EVENT_MCU_WU2, AON_EVENT_RTC_CH2);
-  ti_lib_aon_rtc_combined_event_config(AON_RTC_CH0 | AON_RTC_CH1 | AON_RTC_CH2);
+  ti_lib_aon_rtc_combined_event_config(AON_RTC_CH0 | AON_RTC_CH1);
 
   HWREG(AON_RTC_BASE + AON_RTC_O_SEC) = SOC_RTC_START_TICK_COUNT;
 


g-oikonomou:
The LPM module uses channel 2 to schedule one-shot events, near here: https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/cc26xx-cc13xx/lpm.c#L350

I am not very clear as to how this will keep working after your recommended changes here.

necropaz:
So that would mean that the sensor controller from the chip can't be used with contiki? Can we make an option flag or something that we can use to enable this?

g-oikonomou:
Adding sensor controller support has been in my wish list for quite a while now, but I think things are a bit more complicated.

necropaz:
so for what we need the CC2650_FAST_RADIO_STARTUP? Can't we use an other RTC? Because if i use my pull request i can use the sensor controller and it works perfectly....

g-oikonomou:
In the project where you have tested sensor controller operation, are you using TSCH? And if so, have you been testing networking?

In a nutshell: In order to take full advantage of the chip's low-power characteristics, we turn off as many things as we can when dropping to low power mode. The negative side effect of this is that waking up takes longer. In order to meet the time synchronisation requirements of TSCH, we wake up a bit earlier than we normally would and give the radio a little more time to become ready. Scheduling this earlier wakeup is what we use this RTC CH2 for. I am not sure if this can be done using CH1 instead, perhaps Atis @atiselsts can comment.

necropaz:
No in my project I use CSMA Mode. I have to different Sensor types, one has an analog output and the other has an digital ouput. I use the sensor controller (ADC and TDC Module) to read the sensors 10 times (~about 10 seconds) while the main processor is sleeping. After I collect 10 measurments I send this values over CoAP Protocoll to a server.

atiselsts:
To move the pre-wakeup functionality from CH2 to CH1 we would probably need to add a couple more global variables and complicate the logic.

There are some reports that a workaround is possible (see the last post):
https://www.bountysource.com/issues/40366981-write-program-for-sensor-controller-cc26xx
Is anyone fully understanding how exactly the sensor controller code would interact with the current contiki-ng code?

g-oikonomou:
Thanks Atis I thought so. Depends on your definition of "fully" :)

@necropaz if you try with TSCH, you may find that things will stop working

necropaz:
Okay so we can't use CH2 because of the TSCH. 

The other workaround from https://www.bountysource.com/issues/40366981-write-program-for-sensor-controller-cc26xx will maybe also not working correctly. Because if Sensor Controller is used with RTC CH2 then we normaly adjust the tickrate of the RTC CH2. I think with TSCH the RTC CH2 can't have an tickrate of 1 second.

I found an other possibel solution for this. In the Sensor Controller Studio if found this:

- If irregular wake-up intervals are required, use the Timer Event Trigger resource instead of RTC-Based Execution Scheduling, and implement the Event Handler Code instead of the Execution Code 

**Description** 
**Use timer to trigger the Event Handler Code** 
The resource uses AUX Timer 1 to trigger the Event Handler Code task code block after a configurable period of time. The timer runs off 4 kHz ticks generated by the RTC. 
This can typically be used to: 

- Run a Sensor Controller task at irregular intervals, without unnecessary wake-ups from the RTC. 

- Follow up on actions performed by the Execution Code, for reading results from and powering down an external sensor device after a period of time. 

- Timer-based Event Handler Code triggering can be set up from within the Initialization Code, Execution Code or Event Handler Code by calling evhSetupTimerTrigger() with specified timeout. The trigger is one-shot. 

Note the following: 

- There can only be one active trigger at a time, and setting up a new COMPB, timer or GPIO trigger will cancel any previously configured trigger. 

- Only one task per project can implement the Event Handler Code task code block. 

- Only one resource per project can use AUX Timer 1. 

- Enabling this resource will cause the AUX domain to run at 32 kHz (instead of having no clock) when the Sensor Controller and System CPU are in standby mode. The resulting AUX domain clock switching increases slightly the time to enter and exit standby mode for both Sensor Controller and System CPU. 
Trigger Timing 

The timer delay can be up to one 4 kHz period shorter than specified due to alignment with the 4 kHz signal from the RTC. Also, the 4 kHz signal can jitter by +/- 1 x 32 kHz clock period when the RTC is calibrated to compensate for RCOSC LF frequency error (when used). 
When the timer event trigger occurs, the wake-up is delayed by approximately 400 us if both Sensor Controller and System CPU are in standby mode. This delay includes: 

- 4 kHz signal synchronization (at 32 kHz in standby) 

- Vector fetch + running two instructions to trigger wake-up (at 32 kHz in standby) 

- Wake-up trigger signal propagation (at 32 kHz in standby) 

- AUX domain wake-up 

This is not a very beautiful workaround but maybe the only way, that Sensor Controller works with Contiki NG and TSCH.

I will try this workaround next week.

g-oikonomou:
@necropaz that's some useful stuff.

Can I just ask, from what you are writing I understand that the sensor controller doesn't strictly need RTC CH2 for fundamental operation, you just need it to schedule something periodic, right?

I'm trying to understand the use-case here. Are you trying to perform operations in the sensor controller without waking up the main MCU?

g-oikonomou:
I am closing this since it did not gain too much traction.

PR #275, opened 2018-01-11
Include strings.h to access strncasecmp

superjudge:
This patch fixes a missing include of `strings.h` in two files that make use of the `strncasecmp` function in the cc26xx-web-demo; the missing includes makes the build of the demo fail.

### DIFF
--- a/examples/platform-specific/cc26xx/cc26xx-web-demo/cc26xx-web-demo.c
+++ b/examples/platform-specific/cc26xx/cc26xx-web-demo/cc26xx-web-demo.c
@@ -55,6 +55,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
 
 #include "ti-lib.h"
 /*---------------------------------------------------------------------------*/
--- a/examples/platform-specific/cc26xx/cc26xx-web-demo/net-uart.c
+++ b/examples/platform-specific/cc26xx/cc26xx-web-demo/net-uart.c
@@ -71,6 +71,7 @@
 
 #include <stdint.h>
 #include <string.h>
+#include <strings.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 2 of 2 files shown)


simonduq:
Thanks. Could you add a build of the example to the CI suite, to ensure non-regression in the future? Just add a line in https://github.com/contiki-ng/contiki-ng/blob/develop/tests/02-compile-arm-ports-01/Makefile

g-oikonomou:
This I believe varies across the different versions of arm-gcc (for reasons that I am uncertain about).

simonduq:
Meaning we won't be able to trigger this in CI? Then let's get this in as is?

superjudge:
I had a look, and it seems all the variations of the cc26xx-web-demo are already included in the above mentioned `Makefile`, i.e. the correct combinations of builds for sensortag or launchpad, and cc26xx and cc1310.

Whether this will correctly trigger potential regressions or not, I have not idea. However, it seems POSIXly correct to include `strings.h`, as that seems to be where `strncasecmp` is defined.

Also note that `strings.h` was also already included in `httpd-simple.c` and `mqtt-client.c` (in the same example, of course), but it was missing from the two files included in this patch, so it seems somewhat correct to make the fix from that perspective.

For reference, I'm building on Ubuntu 16.04.3 LTS, using the following Linaro ARM GCC distribution:

```
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2017-q4-major) 7.2.1 20170904 (release) [ARM/embedded-7-branch revision 255204]
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

simonduq:
OK, thanks!

PR #277, opened 2018-01-12
NullNet examples: check paylaod size before reading it

simonduq:
Adds sanity check to input callback. Without it, the function might be called even with empty payload, e.g. when receiving a TSCH keep-alive.


g-oikonomou:
I am likely being very thick here, but I cannot see how the if statement fixes anything

simonduq:
All it does is ensure we do not read anything out bounds of the `data` array.

g-oikonomou:
OK I'd missed the part where the other device is sending `unsigned` :D

PR #282, opened 2018-01-26
Fixed alignment bug for msp430

wirstrom:


nfi:
👍 Good catch! Looks good.

PR #283, opened 2018-01-26
fixed slip-radio config, added shell to NBR and added parsing to no-framer

joakimeriksson:
This is a set of fixes to get NBR to work at all and to get shell active on serial input (for listing routes, etc). The problem was that the no-framer did not parse any address or seq-no from the packets which caused the CSMA duplicate filtering to go crazy on the packets.

Additional fixes:
- generic mechanism to set channel and PAN ID in all serial radios (rather than only tmote sky).
- improved printouts of incoming packets when turning on debug
- removed unused codebase for Tmote Sky to get rid of "half-dead" code.

Will also soon add a CI test for this based on the current rpl-border-router test.


simonduq:
Great. I would suggest removing the sky-slip-radio test once a NBR+SR test is added. sky-slip-radio is realyl a hack with hardcoded inlined packetbuf attributes etc. No point keeping this IMO.

In the long run we should move all BR commands to the shell I think!

joakimeriksson:
Pushed a CI test now - but seems a bit unstable - I think it is related to loss over serial rather than anything else. Probably a good idea to test with something else than sky - maybe cooja motes.

joakimeriksson:
Great - now it is green! Many thanks for the CI help!

PR #289, opened 2018-02-01
Renode: fix vector table offset in the script.

PiotrZierhoffer:
The cc2538-node.resc script was vulnerable to changing the position of
the vector table.
It also had a wrong hardcoded value. Now it will automatically verify
the position of the `vectors` symbol and set the vtor position
accordingly.


g-oikonomou:
Let's merge this one then!

PR #295, opened 2018-02-16
Allow Shell operaction for CC13xx/CC26xx

### LINKED ISSUE #187: Serial input on CC1310 Launchpad
Running through the NG Shell tutorial  on a C1310 Launchpad serial input doesn't work. It needs the line 
`  cc26xx_uart_set_input(serial_line_input_byte);`
added after
`  serial_line_unit();` in platform_init_stage_two in platform.c

The terminal program also needs to send CR+LF at the end of the line

g-oikonomou:
As discussed in #187, the CC13xx/CC26xx platform code goes to sleep very aggressively: Unless we have explicitly requested a peripheral to stay on, the code will turn it off when entering a low power mode.

Leaving the UART on during low power increases energy consumption by quite a lot, so the default platform code allows it to go to sleep. This basically means that we lose UART RX in low power, in other words the shell won't work.

This pull implements a temporary workaround: If we are building with Shell support, the platform code will automatically request UART RX while in low-power mode. This makes all affected platforms behave as per the Shell tutorial (which we should update, really, to state that enabling the Shell will increase energy consumption for some platforms).

Fixes #187

### DIFF
--- a/arch/platform/srf06-cc26xx/platform.c
+++ b/arch/platform/srf06-cc26xx/platform.c
@@ -55,6 +55,7 @@
 #include "vims.h"
 #include "dev/cc26xx-uart.h"
 #include "dev/soc-rtc.h"
+#include "dev/serial-line.h"
 #include "rf-core/rf-core.h"
 #include "sys_ctrl.h"
 #include "uart.h"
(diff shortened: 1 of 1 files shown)


g-oikonomou:
I forgot to ping @mike-forster here, Mike can you confirm whether this fixes things in your setup? Cheers.

mike-forster:
Yes, adding the line as described works. I haven't looked into the power implication of adding it though.

AnasIbrahim:
I guess the serial operation itself still won't work, it will only work if you compiled the shell module. But in case you need the serial you don't want the shell module then !!.

PR #297, opened 2018-02-18
Use PRI macros for safe printf and logs

simonduq:
Questions:
* I used PRIu64. Is this safe @nfi ?
* How do we deal with `size_t`, `clock_time_t`, `rtimer_time_t`?

The PR does not have 100% coverage, as I think this is very difficult to achieve. I believe this to be a good start.

### DIFF
--- a/arch/platform/cc2538dk/contiki-conf.h
+++ b/arch/platform/cc2538dk/contiki-conf.h
@@ -41,6 +41,7 @@
 
 #include <stdint.h>
 #include <string.h>
+#include <inttypes.h>
 /*---------------------------------------------------------------------------*/
 /* Include Project Specific conf */
 #ifdef PROJECT_CONF_PATH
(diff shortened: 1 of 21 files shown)


g-oikonomou:
> How do we deal with size_t, clock_time_t, rtimer_time_t?

I would just explicitly cast to unsigned long and print `%lu`

g-oikonomou:
I should add, I don't see it's worth the effort to apply this to `arch` code other than native.

simonduq:
Both good points, txs, will fix

nfi:
@simonduq Not sure about the platforms in Contiki-NG but I have used several platforms that compiles fine with PRIu64 but crashes at first print. I think the safest choice would be to avoid printing 8 byte values. Otherwise we should document that a Contiki-NG platform must support printing 8 byte values and add platform tests that verifies that it does not cause issues. Not nice if a debug print causes a device to crash.

simonduq:
Closed that one by accident, sorry.

Now ready, I have addressed what we discussed above.

g-oikonomou:
Is this good to go?

simonduq:
Yes, although it's not comprehensive, i.e. I'm sure there are some more places where PRI macros could be used. Hard to spot them all. But I believe it's a start :)

PR #303, opened 2018-02-21
Add next-pointer to coap_periodic_resource_s

superjudge:
Items of type `coap_periodic_resource_s` might end up as parts of a linked
list. However, the struct was missing a next pointer field, which meant
that other fields of the struct was overwritten when the item was put in
a list using `list_add`. See `coap_resource_s` for reference.

### DIFF
--- a/os/net/app-layer/coap/coap-engine.h
+++ b/os/net/app-layer/coap/coap-engine.h
@@ -114,6 +114,7 @@ struct coap_resource_s {
 };
 
 struct coap_periodic_resource_s {
+  coap_periodic_resource_t *next;
   uint32_t period;
   coap_timer_t periodic_timer;
   const coap_resource_periodic_handler_t periodic_handler;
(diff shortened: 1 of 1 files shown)


nfi:
Good catch! This is clearly a bug. However I think a better solution might be to simply remove the obsolete list of periodic resources `coap_resource_periodic_services` in `coap-engine.c`. I think this list only remains in the code by mistake.

superjudge:
Aha, yes, I see, the `coap_resource_periodic_services` list is not used, presumably because the resource is added as user data to the `periodic_timer`, right?

I'll fix that. Should I modify this pull request with those changes, or should I create a new one, so we get a cleaner history?

joakimeriksson:
You could just do a forced push to this PR with a cleaned up history.

superjudge:
I opted for closing this patch and opening a new one (#304), as the actual fix also involve changes to the notification part of the observer pattern.

PR #307, opened 2018-02-23
TSCH schedule printout

simonduq:
This PR adds `LOG_PRINT`, a series of macros that forces logging regardless of the log level. It also cleans up `tsch_schedule_print`.


g-oikonomou:
Thinking aloud, I am not 100% sure about the idea of the logging module forcing printing. Maybe when a user says `LOG_LEVEL_NONE` for a module, they actually really do want the logging level to silence entirely.

g-oikonomou:
I think it would make sense if the entire `tsch_schedule_print()` function were something we only ever used at debugging level?

simonduq:
As the function is called `tsch_schedule_print` I'd find it confusing if a call might result in nothing printed out depending on the TSCH log level. Instead, I'd say it's the caller's responsibility to call at suitable debug levels. Note that the function is not used in the codebase, just provided as a tool.

PR #308, opened 2018-02-23
TSCH: fix a bug in slot scheduling

atiselsts:
The scheduling of a new TSCH timeslot at the end of `tsch_slot_operation()` will enter a count-to-infinity loop if called less or equal-to `RTIMER_GUARD` ticks before the start of the next TSCH timeslot (i.e. 1 or 2 ticks on most platforms on which `RTIMER_GUARD==2`). This patch fixes this behavior by busywaiting until the start of the next slot in that case.

In order to avoid double busywaiting, the busywait code is now executed in `TSCH_SCHEDULE_AND_YIELD` if and only if it was not done in `tsch_schedule_slot_operation()`.

This code was developed as part of the SPHERE project (http://irc-sphere.ac.uk/)

Previously discussed in https://github.com/contiki-os/contiki/pull/2140

Without this patch, TSCH is very unstable on CC26xx under heavy traffic, since the nodes frequently reboot with watchdog timer timeouts.


os/net/mac/tsch/tsch-slot-operation.c
```diff
@@ -1010,12 +1011,12 @@ PT_THREAD(tsch_slot_operation(struct rtimer *t, void *ptr))
         TSCH_ASN_INC(tsch_current_asn, timeslot_diff);
         /* Time to next wake up */
         time_to_next_active_slot = timeslot_diff * tsch_timing[tsch_ts_timeslot_length] + drift_correction;
+        time_to_next_active_slot += tsch_timesync_adaptive_compensate(time_to_next_active_slot);
```
simonduq: why do we need to move this line?

atiselsts: previously `tsch_schedule_slot_operation` was called with the value of `time_to_next_active_slot` before the compensation was applied.

simonduq: oh right, I somehow was `time_to_next_active_slot` as `current_slot_start` ..
got it!

simonduq:
This looks good. Have you tried on other platforms than cc2650?

atiselsts:
The couple zoul nodes I have did not seem to be terribly broken after this.

simonduq:
OK I'll try to gather a bit extra empirical evidence things remain rock solid on at least cc2538 :)

simonduq:
Tested on our local testbed, both cc2538 and cc1200. Good to go!

simonduq:
(except Travis is green but Gibhub still shows a yellow flag, not sure why)

Issue #312, opened 2018-02-24
rpl-lite preforms worse than contiki 3.x


zsrkmyn:
We use `examples/rpl-udp/udp-{server,cilent}.c` to test RPL on cooja mote.

In simulation, we add 100 clients and 1 server. The `csc` file can be found at https://pastebin.com/93WcJT2u

We use rpl-lite as the routing protocol, and set `RPL_NS_CONF_LINK_NUM` to 120 in `project-conf.h`.

When setting `RPL_CONF_PROBING_SEND_FUNC` to `rpl_icmp6_dio_output`, after 1 hours in simulation time, only several nodes around the root can join the RPL instance. Setting `RPL_CONF_PROBING_SEND_FUNC` to `rpl_icmp6_dis_output` makes it little better.

However, when we run `examples/simple-udp-rpl` in the contiki 3.x with the same distribution of these motes, it takes only 4 minutes in simulation time for the root to recieve the first message from the farest nodes.

We also dig out the log of RPL modules, but there are so many unreasonable packet transmissions and parent switchs that we are hard to describle them here. And the log file is larger than 100MiB, so we'll upload it if you need.

Feel free to ask us for more details! XD

simonduq:
Thanks for reporting. This is a challenging simulation environment, large and very sparse.

The problem seems to be with CSMA in Cooja Motes. In Cooja motes, transmission timing is coarse-grained, with only 1ms precision. The current CSMA timings result in re-transmissions with very low backoff exponent. Combine the two above and you get cascaded collisions of nodes colliding and re-trying every 1ms.

With some work on CSMA I got your simulation file work reliably. What I did:
* Set `CSMA_MIN_BE` to 3 and `CSMA_MAX_BE` to 5. This is what the IEEE standard stipulates anyway. I believe the reason we've been using different values were mostly ContikiMAC related, and I think we should adopt the standard default instead.
* `backoff_period`: increased by 20 (or more). This is more specific to Cooja motes. Helps spread out re-transmissions and in turn decrease contention.

BTW, note that `RPL_NS_CONF_LINK_NUM` is now obsolete (will update the doc), please use `NETSTACK_MAX_ROUTE_ENTRIES` instead.

simonduq:
New defaults proposed in https://github.com/contiki-ng/contiki-ng/pull/315

zsrkmyn:
The PR improves the performance, but the network is still not as good as the 3.x version.

[...]

PR #315, opened 2018-02-24
CSMA timing

simonduq:
This does two things:
* adopt standard default values for `CSMA_MIN_BE` and `CSMA_MAX_BE`
* increase backoff period for Cooja motes

This is discussed in https://github.com/contiki-ng/contiki-ng/issues/312


simonduq:
Yes that might be a good idea! How much would that affect performance?

PR #319, opened 2018-02-26
Add a generic LED HAL

g-oikonomou:
This pull proposes (and makes a step towards) the replacement of the traditional LEDs API with a generic, platform-independent LED HAL.

The assumption is that each LED is connected to a GPIO pin using either positive or negative logic. LEDs on a device are numbered incrementally, starting from 0 and counting upwards. Thus, if a device has 3 LEDs they will be numbered 0, 1 and 2.

The API provides functions to manipulate a single LED at a time `led_hal_led_...` as well as functions to manipulate multiple LEDs at a time `led_hal_leds_...`.

Sitting on top of the GPIO HAL, the LED HAL implements GPIO pin manipulation internally. Therefore, instead of implementing GPIO manipulation in an arch-specific driver, platform developers now only need to do two things to add LED support for their device:

* Define an array (`led_hal_arch_leds[]`) of variables of type `led_hal_led_t` to represent their platform's LEDs. Each such variable corresponds to a LED and specifies the respective GPIO pin and also whether the HAL should use negative logic for this LED.
* Specify the number of LEDs on their device by defining `LED_HAL_CONF_COUNT`.

The LED HAL sits on top of the GPIO HAL and will therefore only work on platforms that support the latter. For that reason, this pull attempts to maintain co-existance of the two APIs in the code under `os/` and `examples/`.

Shameless ping also to @fontikos for any thoughts.


fontikos:
Nice pull George. The only thing that I can see missing is a way for the developer to select a particular colour (let's say a platform-independent alerting app that needs to use the RED led). This can be addressed with a `led_hal_get_red()` API that returns the id number of the red LED or NULL if the platform has no RED led.

g-oikonomou:
Thanks for sharing thoughts.

I've been thinking about the colors business and I don't think we should be doing this at run time, since LEDs are always known at build time. Perhaps we can use defines that map color to led number, but I struggle to make up my mind as to what defaults to specify. Should we assume that a board will always have Red and Green? (I can immediately think of many that don't have blue)? Or we can default to a value that signifies such color is not present. Something like

```
#define LED_HAL_LED_NOT_PRESENT 0xFF
#define LED_HAL_LED_RED LED_HAL_LED_NOT_PRESENT
```

Should allow platform-specific examples to specify what LEDs to manipulate using the color scheme. Should also allow cross-platform examples to at least compile, but the call to manipulate the pink LED will do nothing...

simonduq:
Yep that sounds good to me. Am still not sure about the naming though, Couldn't `LED_HAL_LED*` and `led_hal_led*` be replaced with `LED_*` and `led_*`?

g-oikonomou:
Yep, I'll have a think about the nomenclature. I think probbers `led_on()` and `led_multi_on()` and so on could work.

Any other thoughts?!

g-oikonomou:
I have reworked this entirely.

Instead of introducing new HAL files, I am changing the existing `leds.[ch]`.

The two files combine the two APIs, `LEDS_CONF_LEGACY_API` can be used to select the old one.

The new API retains `leds_on()`, `leds_off()`, `leds_toggle()`, `leds_init()`, `leds_set()`, using the exact same semantics. The only differences are:
* The new API sits on the GPIO HAL and no longer requires arch-specific implementations.
* The argument datatypes for the old functions are changed to match the new ones. So for the overlapping functions we have the same prototypes.

The new API introduces functions to manipulate a single LED at a time, e.g. `leds_single_on`.

The semantics of colours are also retained unchanged. Colours are an ORd bitmask of LEDs. There are some differences though:

* Firstly, we now define 5 LED colours: Red, Green, Blue, Yellow, Orange. These are sufficient to cover all currently-supported platforms.
* Secondly, unless a platform specifies that a LED of a specific colour exists, the HAL will assume that it does not.
* Trying to manipulate a non-existent LED will not cause build errors, but will not cause any changes to LED state either.
* We no longer map non-existing LED colours to existing ones.

The new API has been implemented for all CC2538 and CC13xx/CC26xx boards. The old API is still in use for the remaining platforms.

The pull also introduces a platform-independent driver for the RGB part on Zolertia devices.

Example for the HAL and the RGB to follow.

I have also done some minor tidying up of the IPSO and CoAP examples.

g-oikonomou:
Now with examples and tests. Tested on launchpad/cc2650, zoul/remote-reva and zoul/firefly.

See previous post on how I addressed the various comments. Ready for review and merging.

simonduq:
Looks good!

PR #311, opened 2018-02-23
A script to build all examples

atiselsts:
This PR introduces the first step towards automated regression testing of whether all Contiki-NG examples can be compiled for appropriate platforms: https://github.com/contiki-ng/contiki-ng/issues/217

It adds `PLATFORMS_ONLY` and `PLATFORMS_EXCLUDE` Makefile variables to restrict examples to specific platforms and uses them to clearly mark platform-specific examples as such. (It's also used for the Sky platform to avoid building examples that lead to the notorious Sky memory overflow.)

PR #327, opened 2018-02-27
Various Zoul-related fixes

g-oikonomou:
This pull fixes a bunch of compile errors for the Zoul platform:

* Firstly, we fix `platform.c` when the RTCC is enabled
* We fix various problems under `examples/platform-specific/zoul`
* We add travis tests for all Zoul-specific examples

Discovered thanks to work in #311!


g-oikonomou:
For examples that are only meant to be used for specific Zoul boards, we update `Makefile.target` to specify them.

examples/platform-specific/zoul/orion/client/Makefile.target
```diff
@@ -1 +1,2 @@
 TARGET = zoul
+BOARD = orion
```
simonduq: I'm not sure the client assumes Orion

examples/platform-specific/zoul/orion/ip64-router/Makefile.target
```diff
@@ -1 +1,2 @@
 TARGET = zoul
+BOARD = orion
```
simonduq: yes but we also want BOARDS_ONLY, right?

atiselsts:
What about the `LPM_CONF_MAX_PM 1` change - should we do it? I don't see it here, so I'm confused.

PR #330, opened 2018-02-28
CC26xx IEEE mode radio driver: allow reception of packets shorter than 5 bytes

### LINKED ISSUE #322: Maybe a bug in cc26xx-cc13xx radio driver
Hi, we are using commit abd18639b to test our work, but find it's hard for sensors to form a network using RPL-lite.

We dig into the logs and codes, and find something goes wrong [here](https://github.com/contiki-ng/contiki-ng/blob/develop/os/net/mac/csma/csma-output.c#L229). It's seems that `NETSTACK_RADIO.pending_packet()` always returns a number bigger than 0, but the length returned by `NETSTACK_RADIO.read(ackbuf, CSMA_ACK_LEN)` in the following line is always 0. (It's not reasonable, we have packet pending but read a zero-length thing.) So the result of `send_one_packet()` can always be`MAC_TX_COLLISION`, and thus the link status cannot be updated by `link_stats_packet_sent` in `link-stats.c`.

We doubt that there are something wrong with the radio driver, and by reseting `arch/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c` to commit 38987352c, all things go well!

We are using TI SensorTag.

atiselsts:
Aims to fix #322


arch/cpu/cc26xx-cc13xx/rf-core/ieee-mode.c
```diff
@@ -263,6 +263,11 @@ static uint8_t rx_buf_1[RX_BUF_SIZE] CC_ALIGN(4);
 static uint8_t rx_buf_2[RX_BUF_SIZE] CC_ALIGN(4);
 static uint8_t rx_buf_3[RX_BUF_SIZE] CC_ALIGN(4);
 
+/* The size of the metadata (excluding the packet length field) */
+#define RX_BUF_METADATA_SIZE sizeof(rfc_dataEntry_t)
```
g-oikonomou: This is also 8 bytes, but only by coincidence.

Near L621

```
  cmd->rxConfig.bAutoFlushCrc = 1;
  cmd->rxConfig.bAutoFlushIgn = 0;
  cmd->rxConfig.bIncludePhyHdr = 0;
  cmd->rxConfig.bIncludeCrc = 1; /* This keeps the CRC: 2 bytes */
  cmd->rxConfig.bAppendRssi = 1; /* This adds 1 byte */
  cmd->rxConfig.bAppendCorrCrc = 1; /* This adds 1 byte */
  cmd->rxConfig.bAppendSrcInd = 0;
  cmd->rxConfig.bAppendTimestamp = 1; /* This adds 4 bytes */
```

g-oikonomou: Maybe what we need is something along the lines

```
#define IEEE_MODE_INCLUDE_CRC 1

cmd->rxConfig.bIncludeCrc = IEEE_MODE_INCLUDE_CRC;

#define METADATA_SIZE ((2 * IEEE_MODE_INCLUDE_CRC) + (4 * IEEE_MODE_INCLUDE_TIMESTAMP))
```

atiselsts:
I see, of course. I've updated the PR.

g-oikonomou:
Looks great, did you get a chance to test?

@zsrkmyn does this fix your setup?

atiselsts:
Tested briefly, TSCH still works on SPHERE boards.

g-oikonomou:
Fixes #322

zsrkmyn:
Sorry for replying such late. A brief test shows that it works fine on my CC2650 SensorTag. Thanks for fixing it :-)

PR #339, opened 2018-03-05
Add cfs support for srf06-cc26xx

dongdongbh:
We add cfs support for srf06-cc26xx's ext-flash to contiki-ng, and test it on sensortag, the test project in `example/storage/cfs-coffee`, and all test passed.


g-oikonomou:
Cool!

Can you update the example's README to say that it has been tested on Sensortag? I'd also dare state that it is expected to also work on Launchpad CC13xx/CC26xx

dongdongbh:
OK, I just tested it on Launchpad CC26xx, all tests passed, and I also add the tested platforms to the example's README.

PR #345, opened 2018-03-11
Only print RADIO_PARAM_PAN_ID at startup if the radio supports it

g-oikonomou:
In prop mode, the CC13xx/CC26xx radio does not do any h/w frame filtering. For this reason, the respective RF driver does not support `RADIO_PARAM_PAN_ID` (and a number of other things, such as `RADIO_PARAM_16BIT_ADDR`).

As a result, all images that use prop mode will print junk at startup. Basically, the line below in `platform_init_stage_three()` uses variable `pan` uninitialised.

```
  LOG_INFO(" RF: Channel %d, PANID 0x%04X\n", chan, pan);
```

This pull tidies up the function, initialises variables and makes sure to only print PAN ID if `NETSTACK_RADIO.get_value()` succeeds.

Reported by @m-calvo (Can you test this please?)


arch/platform/srf06-cc26xx/platform.c
```diff
@@ -198,7 +198,13 @@ platform_init_stage_three()
           ti_lib_chipinfo_chip_family_is_cc13xx() == true ? "Yes" : "No",
           ti_lib_chipinfo_supports_ble() == true ? "Yes" : "No",
           ti_lib_chipinfo_supports_proprietary() == true ? "Yes" : "No");
-  LOG_INFO(" RF: Channel %d, PANID 0x%04X\n", chan, pan);
+  LOG_INFO(" RF: Channel %d", chan);
+
+  if(NETSTACK_RADIO.get_value(RADIO_PARAM_PAN_ID, &pan) == RADIO_RESULT_OK) {
+    LOG_INFO(", PANID 0x%04X\n", pan);
```
simonduq: you probably want `LOG_INFO_` here (not to prepend the log prefix)

arch/platform/srf06-cc26xx/platform.c
```diff
@@ -198,7 +198,13 @@ platform_init_stage_three()
...
+  }
+  LOG_INFO("\n");
```
simonduq: you probably want `LOG_INFO_` here (not to prepend the log prefix)

simonduq: shouldn't this go to `else` to avoid printing `\n` twice?

g-oikonomou: Argh I never meant for the PANID line to have `\n` will fix in a moment

g-oikonomou: That's done now and rebased

arch/platform/srf06-cc26xx/platform.c
```diff
@@ -209,7 +209,13 @@ platform_init_stage_three()
           ti_lib_chipinfo_chip_family_is_cc13xx() == true ? "Yes" : "No",
           ti_lib_chipinfo_supports_ble() == true ? "Yes" : "No",
           ti_lib_chipinfo_supports_proprietary() == true ? "Yes" : "No");
-  LOG_INFO(" RF: Channel %d, PANID 0x%04X\n", chan, pan);
+  LOG_INFO(" RF: Channel %d", chan);
+
+  if(NETSTACK_RADIO.get_value(RADIO_PARAM_PAN_ID, &pan) == RADIO_RESULT_OK) {
+    LOG_INFO(", PANID 0x%04X", pan);
```
simonduq: I still think you want `LOG_INFO_` here instead (note the final `_`)

g-oikonomou: Still not used to this... So I need `INFO_` here as well as in the final `\n`

simonduq: Yes, and actually also the "RF: channel one". You need it whenever continuing a previous line of log. Else a prefix will added every time

g-oikonomou: Yep. The "RF: Channel" is at the start of a line.

PR #347, opened 2018-03-13
Fix build error and tidy-up BLE example

g-oikonomou:
This pull:

* Adds missing includes to fix a compile error
* Moves the BLE example to the platform-specific directory
* Adjusts the example's Makefile to match current practice
* ~Modes~ Moves the BLE readme from the driver directory to the example directory.


atiselsts:
:+1:

g-oikonomou:
So I've added a test for the example itself and another one for nullnet under this platform.

PR #355, opened 2018-03-18
CC26xx Sensorcontroller Demo

Rasfumi:
This Pull Request shows a way to use the CC26xx Sensorcontroller.
It includes the Sensorcontroller Studio Project and the generated files.
Because Contiki is using all available RTC Channels the Sensorcontroller task is triggered by a Software Event of the main CPU. The Sensorcontroller then triggers a callback function when the task is finished. 

The code is wrapped into a Contiki-Sensor structure that can be used like any other contiki sensor. For a usage example see here:
[very-sleepy-demo.zip](https://github.com/contiki-ng/contiki-ng/files/1822663/very-sleepy-demo.zip)


g-oikonomou:
Cool, this has been desirable for quite a while.

Why do we need to distribute SCS files? Can we just not distribute what we need to distribute and create a wiki page on how to use Contiki-NG with the SCS?

Rasfumi:
Sure, the .scp and .lst files are not actually needed, they are just there to show what the AUX RAM Image is doing.
The Code was just briefely tested and found to be working before rebasing on the current develop branch. I will do some more testing and power consumption measurements when i find time for it. My plan is to construct a capacitive soil humidity sensor that actually works on a Coin Cell for a long time (6+ months). At the moment my avaliable time is very limited so any input is very welcome

g-oikonomou:
I would like to understand the workflow. My understanding is that the process goes somewhere along these lines:

* Use SCS to write programme for sensor controller
* Click a button, SCS exports a bunch of files
* Then what?

I believe (please correct me if I am wrong) that among the output files is a sensor controller firmware image baked into a C code file. The OS needs to somehow place that image into the sensor controllers RAM from where it gets executed. What else does the OS need to do? What other files does SCS export and how do those other files need use?

In a perfect world what we'd have would be "native" (for lack of a better word) Contiki-NG support for the sensor controller. Contiki-NG firmware would know how to upload an image to the sensor controller and execute it, it would also know how to retrieve events and output from the sensor controller.

g-oikonomou:
I'll label this as work in progress since this is a feature we definitely want. This is not in a state that I would consider ready for merging though, and I am afraid I cannot provide any further feedback than what I have already written. Not until I've understood the inner workings better.

g-oikonomou:
I am closing this. As I wrote above, this is a very desirable feature but this PR here is not how I would approach things and it did not get any traction.

PR #362, opened 2018-03-22
TSCH Keepalive loop fix

### LINKED ISSUE #361: TSCH keepalive issue
In tsch.c we have:
```
/* Tx callback for keepalive messages */
static void
keepalive_packet_sent(void *ptr, int status, int transmissions)
{
  [...]
  /* We got no ack, try to recover by switching to the last neighbor we received an EB from */
  if(status != MAC_TX_OK) {
    if(linkaddr_cmp(&last_eb_nbr_addr, &linkaddr_null)) {
      [...]
    } else {
      [...]
      /* Try to get in sync ASAP */
      tsch_schedule_keepalive_immediately();
      return;
    }
  }
[...]
}
```
which results in a crazy loop if:
1) last_eb_nbr_addr is not null,
2) the queue is full (or any other reason causing status=MAC_TX_ERR I guess)

because this callback is called immediately if there is an error.

All the attempts to fill the queue with keepalive messages flood the network with error messages, see the attached screenshot. In my case, after a few minutes, one keepalive is successfully sent and the device goes back to its normal behavior. but I am not sure this is always the case.

screenshot

I was ready to open a PR with a quick fix : changing status =! MAC_TX_OK to status==MAC_TX_NOACK in order to be aligned with the comment "/*We got no ack, [...]". However, I am not sure what the actual behavior of TSCH should be in this case, so I prefer to open issue.

Tim024:
Fixes #361

### DIFF
--- a/os/net/mac/tsch/tsch.c
+++ b/os/net/mac/tsch/tsch.c
@@ -275,7 +275,7 @@ keepalive_packet_sent(void *ptr, int status, int transmissions)
   LOG_INFO_(", st %d-%d\n", status, transmissions);
 
   /* We got no ack, try to recover by switching to the last neighbor we received an EB from */
-  if(status != MAC_TX_OK) {
+  if(status == MAC_TX_NOACK) {
     if(linkaddr_cmp(&last_eb_nbr_addr, &linkaddr_null)) {
       LOG_WARN("not able to re-synchronize, received no EB from other neighbors\n");
       if(sync_count == 0) {


os/net/mac/tsch/tsch.c
```diff
@@ -274,8 +274,8 @@ keepalive_packet_sent(void *ptr, int status, int transmissions)
   LOG_INFO_LLADDR(packetbuf_addr(PACKETBUF_ADDR_RECEIVER));
   LOG_INFO_(", st %d-%d\n", status, transmissions);
 
-  /* We got no ack, try to recover by switching to the last neighbor we received an EB from */
-  if(status != MAC_TX_OK) {
+  /* We got no ack or collision, try to recover by switching to the last neighbor we received an EB from */
+  if(status == MAC_TX_NOACK || status == MAC_TX_COLLISION) {
```
simonduq: Ah no I meant the other way around. Collision are like errors: when they occur, we do not want to reschedule immediately. The test should be `if(status == MAC_TX_NOACK)`

Tim024: Indeed, my bad. It is corrected now

simonduq:
could you squash the commits for a clean history?

PR #366, opened 2018-03-23
Cross-platform MQTT client example

g-oikonomou:
This pull proposes a cross-platform MQTT client example and deletes the respective example under `cc2538-common`.

The example will work on all platforms that support the button HAL. With a little hacking it also works with native.

Easiest way to test is to run a mosquitto broker on a linux/OSX device and by setting `MQTT_CLIENT_CONF_BROKER_IP_ADDR` to the v6 address of said device.

Still somewhat of a WiP / need to test for various boards, but I thought I'd throw it out here for comment.

Requires #365

### DIFF
--- a/examples/mqtt-client/Makefile
+++ b/examples/mqtt-client/Makefile
@@ -1,0 +1,13 @@
+CONTIKI_PROJECT = mqtt-client
+all: $(CONTIKI_PROJECT)
+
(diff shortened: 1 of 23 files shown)


g-oikonomou:
Tested on Firefly, Remote Rev A and launchpad/cc2650.

Ready for consideration this one

Issue #128, opened 2017-10-29
RPL-lite: Slow rejoin if root resets


g-oikonomou:
1. Root starts
2. Mesh node starts and joins the network smoothly
3. Root restarts.
4. Mesh node thinks its link to the root is fresh and no longer sends DAOs. Node thinks it has joined the network, only the root has no path to it.

In the attached PCAP, the root restarts at around frame 52. See how the mesh node does not start probing with unicast DIOs until quite a few minutes later. There is still no down path to the node about 20' after the root restart.

[slow-rejoin.pcapng.zip](https://github.com/contiki-ng/contiki-ng/files/1425431/slow-rejoin.pcapng.zip)

simonduq:
Thanks for the detailed report.

The natural way to handle this is with DAG version or DTSN increments, maybe periodically or under some condition. Nodes will notice the reboot because these are lollipop, and will local repair immediately and re-send DAO.

A problem is that this will increases also in the reboot-free cases. An alternative, less portable, is to store the last version in persistent memory and increment upon reboot. Need to think more about it and measure overhead, probably for next release.

simonduq:
OK I've been giving this more thought, let me describe below the different options and their trade-offs. There are at least two dimensions in the design space: how and when to fix the topology.

How:
* No action: with no action, nodes will eventually re-register their route with a new DAO. With the current defaults in RPL Lite, this happens at a 30-minute period.
* DTSN increment: the root can increment the DTSN it sends in its DIO. When noticing the increment from their direct parent, nodes will re-register their DAO, and increment their own DTSN, asking their sub-tree to also re-register. Summary: has the whole network resend DAOs, at the pace of the current DIO Trickles, i.e., there is Trickle reset involved.
* Trickle reset on DTSN increment: we could decide, as an implementation policy, to also reset the Trickle timer whenever seeing a DTSN increment. This will result in much faster re-registration than above. But will also increase control traffic noticeably for the next tens of minutes, until Trickle converges again.
* Global repair: the root can increment the DODAG version. In reaction, nodes will perform a local repair (reset all RPL state and timers including Trickle) and propagate the new version. This results in the topology getting rebuilt from the ground up. A downside is that during the repair, a node may become temporarily parentless, as it has to attach nodes from the DAG, and may be unable to find an acceptable one (acceptable and fresh link estimates etc).

When:
* When the network operator decides. This is already available: an administrator can trigger a repair at the root via the shell.
[...]

PR #369, opened 2018-03-24
RPL Lite: Increment DTSN when no path to destination is found

simonduq:
When not finding a path towards a node that should be in our DODAG, reset DTSN. This will force the whole to network to re-register their route.

This should mitigate #128. Not as well as with NVM storage of the version prior to reboot. But will accelerate node re-join.

### DIFF
--- a/os/net/routing/rpl-lite/rpl-dag.c
+++ b/os/net/routing/rpl-lite/rpl-dag.c
@@ -212,6 +212,19 @@ global_repair_non_root(rpl_dio_t *dio)
 }
 /*---------------------------------------------------------------------------*/
 void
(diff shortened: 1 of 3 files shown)


simonduq:
So, I do have doubts about the mechanism in this PR, see https://github.com/contiki-ng/contiki-ng/issues/128

This PR will basically increase control traffic whenever a packet targets an unknown node in the DODAG. This might result in prohibitive overhead, and is an attack vector for actors willing to degrade network quality.

Thoughts?

simonduq:
Closing, as this solution seems appears to be suboptimal.

Issue #377, opened 2018-03-26
TSCH: write Doxygen


simonduq:

simonduq:
Promoting to 4.1

simonduq:
Closed by https://github.com/contiki-ng/contiki-ng/pull/421

PR #379, opened 2018-03-26
Cleanup TSCH .h files

simonduq:
This re-arranges TSCH .h files to make it easier to find constants, types, and defines. This is done following the same structure as RPL Lite. Naming of many macros is also simplified as a result.

Most constants, configs and types are now moved to `tsch-const.h`, `tsch-conf.h` and `tsch-types.h`. This was not done for TSCH modules that are really standalone, such as `tsch-security`, `tsch-asn` or `tsch-log`.

An obvious next step will be to add Doxygen. Added to the roadmap: #377


atiselsts:
I'm afraid that this doesn't make me too excited, seeing that TSCH users will have to change their `project-conf.h` files everywhere once again. Thought that migrating my applications from Contiki to Contiki-NG was already hard enough ;)

simonduq:
What about just doing the re-arrangement to `tsch-conf`, `tsch-const` and `tsch-types` but preserving constant names for backwards compatibility?

atiselsts:
Personally I have no problems with the rearrangements and like the new `tsch-types.h` file.

simonduq:
OK will remove the name changes then. Thanks for the feedback :)

simonduq:
Done

PR #380, opened 2018-03-26
TSCH burst mode

### LINKED ISSUE #168: Orchestra and fragmented packets
It would be really handy with a good way of scheduling a listening (and transmitting) slot immediately if the IPv6 packet was fragmented by 6lowpan layer and there is more fragments to send... This really would make a huge difference in our low duty cycle leaf nodes.

@simonduq and @joakimeriksson mentioned that this might be up for consideration in a standardisation work of Orchestra... Is, and in that case when, would that be of for grabs?

simonduq:
This implements the TSCH burst mode, described in the IETF I-D ASF (https://tools.ietf.org/html/draft-duquennoy-6tisch-asf-01), and considered for inclusion in 6TiSCH. The idea is simple: for slots that support burst mode, whenever there are more packets in queue for a given neighbor, we schedule more slots on-the-fly as long as we get ACKs.

Marking as work-in-progress because I would rather get this merged after 4.1, so this gets more testing.

Builds on https://github.com/contiki-ng/contiki-ng/pull/379

Closes #168


simonduq:
We just discovered that this is actually part of IEEE 802.15.4-2015!
Page 152, Section 7.2.1.3 Frame Pending field:
*"When operating in TSCH mode, the frame pending bit can be set to one to indicate that the recipient should stay on in the next timeslot and on the same channel if there is no link scheduled."*

There is a slight difference with the current PR though, where we stick to the same *channel offset*, not *channel*. Will update to match the standard behavior.

atiselsts:
Keeping the same channel is a bad idea for reliability reasons. Compliance with standard is important, but that seems like an oversight in the standard? Perhaps you can make a compile time flag that would allow the user to choose between these two behaviors.

simonduq:
Well I was also unsure at first but what IEEE proposes now seems superior IMO. If a channel was idle for the first packet, why not stick to it for the subsequent ones? Here we're not in the "we failed to transmit on channel X, let's retry on Y" case.

atiselsts:
Actually, quantifying what is the best option here could be a research problem!

simonduq:
Now updated to match the standard, i.e., stick to the same physical channel. Did not keep the old behavior, as switching to the new actually involved significant changes. Below is a screenshot illustrating what a burst looks like in Cooja:
tsch-burst

os/net/mac/tsch/tsch-packet.c
```diff
@@ -459,4 +459,18 @@ tsch_packet_parse_eb(const uint8_t *buf, int buf_size,
   return curr_len;
 }
 /*---------------------------------------------------------------------------*/
+/* Set frame pending bit in a packet (whose header was already build) */
+void
+tsch_packet_set_frame_pending(uint8_t *buf, int buf_size)
+{
+  buf[0] |= (1 << 4);
```
atiselsts: Might be better to use a `#define` here instead of the constant `4`

os/net/mac/tsch/tsch-packet.c
```diff
@@ -459,4 +459,18 @@ tsch_packet_parse_eb(const uint8_t *buf, int buf_size,
...
+}
+/*---------------------------------------------------------------------------*/
+/* Get frame pending bit from a packet */
+int
+tsch_packet_get_frame_pending(uint8_t *buf, int buf_size)
+{
+  return (buf[0] >> 4) & 1;
```
atiselsts: Same here

os/net/mac/tsch/tsch-slot-operation.c
```diff
@@ -167,6 +167,13 @@ static struct tsch_link *backup_link = NULL;
 static struct tsch_packet *current_packet = NULL;
 static struct tsch_neighbor *current_neighbor = NULL;
 
+/* Indicates whether an extra link is needed to handle the current burst */
+static int burst_link_scheduled = 0;
+/* Counts the length of the current burst */
+int tsch_current_burst_count = 0;
+/* The physical channel of the current burst */
+static uint16_t burst_tsch_current_channel = 0;
```
atiselsts: I don't immediately see the need for this new variable. Why is `tsch_current_channel` not enough? While in burst mode, that variable should not be updated.

os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-ns.c
```diff
@@ -104,7 +104,7 @@ init(uint16_t sf_handle)
   /* Add a Tx link at each available timeslot. Make the link Rx at our own timeslot. */
   for(i = 0; i < ORCHESTRA_UNICAST_PERIOD; i++) {
     tsch_schedule_add_link(sf_unicast,
-        LINK_OPTION_SHARED | LINK_OPTION_TX | ( i == rx_timeslot ? LINK_OPTION_RX : 0 ),
+        LINK_OPTION_SHARED | LINK_OPTION_TX | ( i == rx_timeslot ? LINK_OPTION_RX : 0 ) ,
```
atiselsts: This is probably an accidental change. Same in `orchestra-rule-unicast-per-neighbor-rpl-storing.c`

os/net/mac/tsch/tsch-packet.c
simonduq: agree

os/net/mac/tsch/tsch-slot-operation.c
simonduq: agree

simonduq:
Comments addressed

PR #385, opened 2018-03-27
Bugfix on Address Resolution by Neighbor Discovery Protocol

yatch:
# Summary

A neighbor cache entry cannot be updated with the link-layer address in a received Neighbor Advertisement. This causes communication failure, which happens only if Neighbor Discovery Protocol (NDP) is used for address resolution. With the default configuration of `contiki-ng`, this bug has no effect since NDP is disabled (`UIP_CONF_ND6_AUTOFILL_NBR_CACHE` is set with 1).

# Tests
Eight tests are introduced under `tests/09-ipv6`.

In each test, two nodes are created. One of them sends a ping packet to the other. As a result, if the ping initiator has a neighbor cache entry having a correct MAC address of the responder, the test is considered success. 

Differences among them are MAC (TSCH or CSMA), address scope (link-local or global), and routing configuration (with RPL or without RPL). When RPL is disabled, NDP is enabled instead in `os/contiki-default-conf.h`.

### without the fix
```shell
05-ping-lla-csma-wo-rpl                  TEST FAIL    0/1 -- failed seeds: 1
06-ping-ula-csma-wo-rpl                  TEST FAIL    0/1 -- failed seeds: 1
07-ping-lla-tsch-wo-rpl                  TEST FAIL    0/1 -- failed seeds: 1
08-ping-ula-tsch-wo-rpl                  TEST FAIL    0/1 -- failed seeds: 1
```

### with the fix
```shell
05-ping-lla-csma-wo-rpl                  TEST OK      1/1
06-ping-ula-csma-wo-rpl                  TEST OK      1/1
07-ping-lla-tsch-wo-rpl                  TEST OK      1/1
08-ping-ula-tsch-wo-rpl                  TEST OK      1/1
```

# Note
This PR is a clean and simple version of https://github.com/contiki-os/contiki/pull/1984.


joakimeriksson:
Just a few questions for my understanding:

1. What happens to the rest of the information in the nbr-table ? Will it be there still associated with the  old ll-address?
2. What happens when it is changing back to this ll-address if there is already an entry int the nbr-table with that ll-address?

yatch:
@joakimeriksson 

1. Usually yes, but It depends. `uip_ds6_nbr_update()` allocates a new `nbr` by `uip_ds6_nbr_add()`, which could remove the least used entry in `nbr_table_allocate()` if `NBR_TABLE_FIND_REMOVABLE` is enabled.
2. If the new ll-address is associated with another neighbor cache when `uip_ds6_nbr_update_ll()`, nothing happens and `-1` is returned.

Typically, a neighbor cache is associated with `linkaddr_null` at the time a NS is sent. When `uip_ds6_nbr_update_ll()` is called as a result of receiving a NA, `linkaddr_null` is the old ll-addr and the ll_addr in the NA is the new ll-addr.

tests/09-ipv6/js/ping-test.js
```diff
@@ -0,0 +1,45 @@
+TIMEOUT(10000, log.testFailed());
+
+dst_ip_addr = "fe80::202:2:2:2";
```
yatch: Oops..., this script uses only the link-local address of node-2. I'll fix this soon.

simonduq:
> 1. Usually yes, but It depends. 

This confused me. Looks like you're removing the entry, no?

simonduq:
Except for that confusion, this PR looks good to me

joakimeriksson:
@simonduq - I think the removal is only of the DS6-ND part of the neighbortable - so there might (if we have another than the null-lladdr) something left even if the ds6-nd part is gone - I guess. Probably ok as I guess other things are also removed when the ds6-nd entry is gone (but not link-stats, etc).

simonduq:
Oh yes, I see. Unless we're missing something this is totally safe. Taken care of by the `nbr-table` module.

simonduq:
I approved but would like another approve, e.g. by @joakimeriksson on this one as this is sensitive and complex code.

yatch:
@simonduq @joakimeriksson 

I'm sorry for this confusion.

> I think the removal is only of the DS6-ND part of the neighbortable - so there might (if we have another than the null-lladdr) something left even if the ds6-nd part is gone - I guess. Probably ok as I guess other things are also removed when the ds6-nd entry is gone (but not link-stats, etc).

Thanks! This explains well. I try to clarify it more.

As @joakimeriksson said, the removal is only for the entry of `ds6_neighbors` for `linkaddr_null` in this case. This is done by `uip_ds6_nbr_rm()`, which actually unsets "used" bit in `ds6_neighbors`. Other table entries associated with `linkaddr_null` are not affected by this removal.

> What happens to the rest of the information in the nbr-table ? Will it be there still associated with the old ll-address?

Then, the right answer to the latter question is "yes".

The newly introduced `uip_ds6_nbr_update_ll()` calls `uip_ds6_nbr_add()` after the removal, which could remove an existing `nbr_table` entry when there is no available memory for a new `nbr_table` and `NBR_TABLE_FIND_REMOVABLE` is enabled. You can find the code in `nbr_table_allocate()`.

> What happens to the rest of the information in the nbr-table ? Will it be there still associated with the old ll-address?

In this sense, the answer to the former question is "it depends".

simonduq:
I'm leaning towards postponing this one to 4.2, the timing is getting tight for such sensitive changes. Opinions?

yatch:
@simonduq I agree... Definitely, it's a safer decision. Then, I'd propose to tag 4.2 to https://github.com/contiki-ng/contiki-ng/pull/401 as well.

simonduq:
Now that we have the feature-frozen branch `release-4.1`, I propose to merge in `develop`. This will get 6 months testing until 4.2.

Issue #248, opened 2017-12-13
Replace activation-by-macro of some features by conditional code


laurentderu:
Many features in Contiki-NG are only compiled when setting a macro to 1. 6LBR tries to provides runtime configuration of many features and this conflict sometime with the activation-by-macro.

Compiling both Storing and Non-Storing mode in RPL-Classic is not a problem as both can be enabled using the right macro and then the actual mode is selected by the MOP.
On the other hand, features like DAO-Ack or Multicast can not be easily activated at run-time due to how the macro are used.

The proposal is to replace some features activation-by-macro with conditional code and rely on the dead-code removal of the compiler to not increase the code size. For example, the following pseudo code :

```
#if RPL_WITH_DAO_ACK
   // DAO-Ack specific code
#else
  // No DAO-Ack
#endif
``` 

is replaced by :

```
if (RPL_WITH_DAO_ACK) {
   // DAO-Ack specific code
} else {
  // No DAO-Ack
}
``` 

I did the exercice for DAO-Ack, and I got the following figures for Native, Sky and CC2538DK with RPL_CONF_WITH_DAO_ACK set to 0 :

*Native* :
Code : +152 bytes
Data : +0 bytes
BSS : + 32 bytes

*Sky* :
Code : +2 bytes
Data : +2 bytes
BSS : + 20 bytes

*CC2538DK* :
Code : +0 bytes
Data : +4 bytes
BSS : + 32 bytes

(The BSS increase is due to the dao_retransmission_timer in the DAG struct)

When RPL_CONF_WITH_DAO_ACK is set to 1, the code and data size is unchanged.

The full code can be seen here : https://github.com/cetic/contiki-ng/tree/test-dao-ack-compilation

simonduq:
What would you think about doing it like `RPL_IS_STORING`? Gets the best of both worlds I believe: enables run-time selection, but still makes it possible to compile the feature out and save memory.

laurentderu:
I have to look more closely, For example, in https://github.com/contiki-ng/contiki-ng/blob/9d740ca8e6452ad038f3f7934e2f0810ce54e427/os/net/rpl-classic/rpl-icmp6.c#L1090

it's an #if ... #else ... #endif

You should keep the #if ... #else ... #endif and inside the first block, add a if (RPL_IS_DAO_ACK()) and duplicate the code of the #else branch.

simonduq:
[...]

PR #393, opened 2018-03-30
Queue Mode implementation in LwM2M

carlosgp143:
This Pull Request implements the Queue Mode inside the already present LwM2M implementation. This Queue Mode is useful for saving energy in the LwM2M client, sleeping for long periods of time. The idea is that the client sleeps for a defined period of time, then it wakes up, sends a registration update message, and after that it can receive requests.

A new object is added in the resource model, the Queue Mode object (`lwm2m-qmode-object.[ch]`) for managing the parameters of this mode: awake and sleep times, and a flag to enable the dynamic adaptation that has been implemented for the awake time.

All the inserted code and files are only compiled if the flag `LWM2M_Q_MODE_ENABLED`  is activated, to maintain the same memory footprint if it is not used. Also, the flag `LWM2M_Q_MODE_INCLUDE_DYNAMIC_ADAPTATION` is used to include or not the necessary code for the implemented dynamic adaptation of the awake time.

I've also added compilation tests. In a close future I will add a test using Leshan server and python, similar to the ones  in `tests/18-coap-lwm2m`.


os/services/lwm2m/lwm2m-engine.c
```diff
@@ -80,10 +80,24 @@
 #define USE_RD_CLIENT 1
 #endif /* LWM2M_ENGINE_CONF_USE_RD_CLIENT */
 
+
+#if LWM2M_Q_MODE_ENABLED
```
g-oikonomou: Any particular reason why not just compile everything, always? I would just get rid of _all_ those conditionals in this file. The linker can garbage collect objects where possible

os/services/lwm2m/lwm2m-engine.h
```diff
@@ -114,6 +116,14 @@ void lwm2m_notify_object_observers(lwm2m_object_instance_t *obj,
 
 void lwm2m_engine_set_opaque_callback(lwm2m_context_t *ctx, lwm2m_write_opaque_callback cb);
 
+#if LWM2M_Q_MODE_ENABLED
```
g-oikonomou: Ditto here.

I think those prototypes should be present here all the time, I cannot see a reason why not.

os/services/lwm2m/lwm2m-engine.h
```diff
@@ -114,6 +116,14 @@ void lwm2m_notify_object_observers(lwm2m_object_instance_t *obj,
...
+uint8_t lwm2m_engine_is_waked_up_by_notification();
```
g-oikonomou: For API functions, I'd really like to see some meaningful doxygen documentation. And yes I know the state of documentation in this file is not great, but I think we should try to improve instead of maintain.

os/services/lwm2m/lwm2m-notification-queue.c
```diff
@@ -0,0 +1,239 @@
+/*
+ * Copyright (c) 2017, RISE SICS AB.
```
g-oikonomou: Are we assuming that you are a member of the SICS team? If not, then this perhaps needs to change

os/services/lwm2m/lwm2m-notification-queue.c
```diff
@@ -0,0 +1,239 @@
...
+/**
+ * \file
+ *         Implementation of functions to manage the queue of notifications
+ * \author
+ *         Carlos Gonzalo Peces <carlosgp143@gmail.com>
+ */
+/*---------------------------------------------------------------------------*/
+#include "lwm2m-notification-queue.h"
+
+#if LWM2M_Q_MODE_ENABLED
```
g-oikonomou: Here too

os/services/lwm2m/lwm2m-qmode-conf.h
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * \file
+ *         Queue Mode Configuration Parameters
+ * \author
+ *         Carlos Gonzalo Peces <carlosgp143@gmail.com>
+ */
+
+#ifndef __QMODE_CONF_H__
+#define __QMODE_CONF_H__
```
g-oikonomou: We typically do _not_ use double underscores for include guards.

os/services/lwm2m/lwm2m-engine.c
carlosgp143: My main goal was to not increase the memory footprint when people wants to use LwM2M but not operate with the Queue Mode at all. Maybe it's not so critical and everything should be compiled... I guess we need to reach an agreement here. @joakimeriksson and @nfi what do you think about it?

os/services/lwm2m/lwm2m-engine.h
carlosgp143: The same, we need an agreement.

os/services/lwm2m/lwm2m-notification-queue.c
carlosgp143: Yes, I've been working in the NES group as a Master student since September.

os/services/lwm2m/lwm2m-engine.h
carlosgp143: Should I do the documentation for the API functions of Queue Mode in this PR? Or should we all improve the documentation of lwm2m API functions in a separate PR?

os/services/lwm2m/lwm2m-engine.c
g-oikonomou: OK I won't insist, happy with whatever you lot decide

os/services/lwm2m/lwm2m-engine.h
g-oikonomou: Again, up to you + team

os/services/lwm2m/lwm2m-notification-queue.c
g-oikonomou: 👍

os/services/lwm2m/lwm2m-engine.h
joakimeriksson: Lets add that soon but in a separate PR. Niclas and I should do the work on adding the doxygen on most of the APIs as we did the initial lwm2m PR.

os/services/lwm2m/lwm2m-engine.c
joakimeriksson: I agree that we should try to avoid conditionals in the C-files as much as possible in the long run. Possibly by separating some parts into different files or even starting to accept larger code-base as typical chips are growing and we are kind of giving up on fitting things on the Tmote Sky now. Niclas and I (and Carlos) can try to make a PR with some refactorisation towards avoiding most of the conditionals in the LWM2M code.

simonduq: Related: https://github.com/contiki-ng/contiki-ng/issues/248

carlosgp143: If we consider that the memory footprint is not critical and all the Queue Mode code can just be compiled always with LwM2M, then there is no point on keeping the preprocesor conditionals. I think is a good solution to keep the code as it is for now, and make another PR to remove them.

Also, as @joakimeriksson says, we can move all the Q-Mode functions defined in `lwm2m-engine.[hc]` and `lwm2m-rd-client.[hc]` to other files if you think that would improve the code, and just keep one big conditional in those files, as in `lwm2m-qmode-object` or `lwm2m-notification-queue`. I think that is fast and I can even do it for this PR, or maybe we can just take care of everything in a future PR... what do you think?

nvt:
@carlosgp143 You have made some commits since the review, so please let the merge team know whether this PR is ready to be examined again, or if there is some more work planned.

carlosgp143:
Hi! Yes, everything is done from my side and from what @joakimeriksson, @nfi and me have been talking in offline discussions. 

@g-oikonomou you request some changes at the very beginning, but I think we have already addressed all of them... Is there anything else that I should consider?

carlosgp143:
Hi! I got some internal feedback so I'm gonna do some changes in the code in the next few days. Basically I will simplify the `notification-queue` code to make it more simple and also separate the Queue Mode implementation and the Queue Mode object that we added, as it is not part of the specification. 

So the WIP label can be added again :-)

g-oikonomou:
I'm happy with this. You may want to adjust those include guards.

PR #396, opened 2018-03-30
Various build system improvements

g-oikonomou:
This PR proposes a number of improvements to our build system

* Harmonises the pattern of the`.map` filename.
All map files are now called `example-platform.map` (e.g. `hello-world-native.map`). This is an improvement because previously some platforms were creating e.g. `contiki-sky.map`. For example directories that generate multiple firmwares, the `.map` file generated second would overwrite the one generated first. With this change in place, we generate a separate `.map` file per corresponding firmware.
* Platform native generates a `.map` file on OS X
* We extend the output of `make usage` and we add a `make help` alternative


g-oikonomou:
I jumped the gun on this one hitting "Open" before I was done with the OP. Edited now.

PR #398, opened 2018-03-30
Implement the new HALs for platform native

g-oikonomou:
This PR implements the following enhancements for platform native:

* We implement the GPIO HAL. This is just a driver that emulates a device with 32 pins.
* With the GPIO HAL in place, we can also:
  * Transition the platform to the new LED HAL. Native becomes a platform with zero LEDs, but LED-using examples will now compile and link perfectly. Requires #405 
  * Transition the platform to the new Button HAL. We specify a platform with zero buttons, but examples using the new button HAL will also link and work correctly. Requires #400 

With the above changes in place, we can now use the MQTT client example as-is (Requires #397). This pull enables this example for platform native and adds a respective travis test.

MQTT client on native has been tested to work with mosquitto.

Requires #397, #400, #405


g-oikonomou:
This PR now has a totally larger scope. I have edited the title and OP to reflect this and re-opening.

tests/01-compile-base/Makefile
```diff
@@ -22,6 +22,7 @@ rpl-border-router/sky \
 slip-radio/sky \
 ipv6-hooks/sky \
 nullnet/native \
+mqtt-client/native \
```
simonduq: why `mqtt-client`? wasn't `dev/gpio-hal/` intended instead?

g-oikonomou: This is as intended. Basically this pull enables 3 examples for native, but I only added this one to the tests. Do you see a point in adding the other two as well?

simonduq: Oh right, I see. Probably no point in adding the other two.

PR #399, opened 2018-03-30
Use os/lib/dbg-io for platform jn516x

g-oikonomou:
This PR changes platform jn516x so it uses the generic dbg-io in os/lib/dbg-io.

Needs tested on hardware.

Requires #371 and #394


simonduq:
I'll try this on hardware before merging

simonduq:
Tested on hardware, just works

g-oikonomou:
All nicely rebased now

PR #405, opened 2018-04-01
Allow the LED HAL to work on platforms without LEDs

g-oikonomou:
This pull improves the LED HAL by allowing it to "just work" for platforms that do not feature any LEDs. 

The LED HAL will compile just fine under all circumstances. If the port supports the GPIO HAL, the LED HAL will also link without errors. No special defines required to signal that a platform has no LEDs (the HAL defaults to assuming zero LEDs).


os/dev/leds.c
```diff
@@ -97,7 +97,11 @@ leds_toggle(leds_mask_t ledv)
 /*---------------------------------------------------------------------------*/
 #else /* LEDS_LEGACY_API */
 /*---------------------------------------------------------------------------*/
+#if LEDS_COUNT
```
simonduq: for more clarity I would suggest `#if LEDS_COUNT > 0` instread

PR #411, opened 2018-04-05
CoAP: enable calling coap_get_payload with a NULL pointer

simonduq:
Mostly a defensive programming feature. Also useful for just getting the payload len without having to provision a pointer.


os/net/app-layer/coap/coap.c
```diff
@@ -1116,7 +1116,9 @@ int
 coap_get_payload(coap_message_t *coap_pkt, const uint8_t **payload)
 {
   if(coap_pkt->payload) {
-    *payload = coap_pkt->payload;
+    if(payload != NULL) {
+      *payload = coap_pkt->payload;
+    }
     return coap_pkt->payload_len;
   } else {
     *payload = NULL;
```
nfi: Checking for NULL should be done in both places the argument is used. 👍 from me if this is fixed.

simonduq:
fixed, thanks!

PR #415, opened 2018-04-05
CoAP: auto-initialize module

simonduq:
This PR proposes to auto-initialize the CoAP module whenever it is included, so that one does not have to call `coap_engine_init()` in addition to adding the module.


g-oikonomou:
I've no idea why travis is kicking up dirt here, but it doesn't seem transient.

simonduq:
fixed

simonduq:
Done!

PR #416, opened 2018-04-06
6LoWPAN fragment generation cleanup

simonduq:
This PR reworks the 6LoWPAN fragment generation logic to minimize the number of fragments and fix fragment count estimation. It also reworks logging overall, and removes duplicate code (fragment copy and transmission).

On `develop`, with default settings, we have 104 bytes left for 6LoWPAN payload in each 802.15.4 packet.  The first fragment has a 5-byte header, subsequent fragments a 4-byte header. Because fragment offsets are in 8 bytes increments, all but the last fragment must have an IPv6 payload multiple of 8. Currently, Contiki-NG also truncates the last fragment, unnecessarily. Below is an example ping with 451 bytes payload, resulting in 6 fragments:
```
01:00.593	ID:1	[INFO: 6LoWPAN   ] output: sending packet len 507
01:00.593	ID:1	[INFO: 6LoWPAN   ] uip_len: 507, fragments: 6, free bufs: 7
01:00.593	ID:1	[INFO: 6LoWPAN   ] output: 1st fragment (len 72, tag 0)
01:00.593	ID:1	[INFO: 6LoWPAN   ] output: fragment (offset 15, len 96, tag 0)
01:00.593	ID:1	[INFO: 6LoWPAN   ] output: fragment (offset 27, len 96, tag 0)
01:00.593	ID:1	[INFO: 6LoWPAN   ] output: fragment (offset 39, len 96, tag 0)
01:00.593	ID:1	[INFO: 6LoWPAN   ] output: fragment (offset 51, len 96, tag 0)
01:00.593	ID:1	[INFO: 6LoWPAN   ] output: fragment (offset 63, len 3, tag 0)
```

With this PR, we save one fragment whenever the last has enough space left to accommodate the last bytes:
```
01:34.960	ID:1	[INFO: 6LoWPAN   ] output: sending IPv6 packet with len 507
01:34.960	ID:1	[INFO: 6LoWPAN   ] output: header len 48 -> 24, total len 507 -> 483, MAC max payload 104, frag_needed 1
01:34.960	ID:1	[INFO: 6LoWPAN   ] output: fragmentation needed, fragments: 5, free queuebufs: 7
01:34.960	ID:1	[INFO: 6LoWPAN   ] output: fragment 1/5 (tag 0, payload 72)
01:34.960	ID:1	[INFO: 6LoWPAN   ] output: fragment 2/5 (tag 0, payload 96, offset 120)
[...]

PR #426, opened 2018-04-10
6LoWPAN avoid buffer overflow in case of IPv6 header larger than packetbuf

simonduq:
The current compression routine keeps on filling packetbuf without ever checking the buffer capacity. Now that we have source-routing, there actually are cases where large IPv6 headers are produced, causing an overflow. With this PR, such packet will be dropped.

In the long run, we probably want to be able to fragment packets with large headers, i.e., compress the header in a larger buffer than packetbuf.

Requires https://github.com/contiki-ng/contiki-ng/pull/416


simonduq:
OK, just realized this is a hotfix, will prepare a version for 4.1 that does not require https://github.com/contiki-ng/contiki-ng/pull/416

PR #427, opened 2018-04-10
6LoWPAN avoid buffer overflow in case of IPv6 header larger than packetbuf

simonduq:
The current compression routine keeps on filling packetbuf without ever checking the buffer capacity. Now that we have source-routing, there actually are cases where large IPv6 headers are produced, causing an overflow. With this PR, such packet will be dropped.

In the long run, we probably want to be able to fragment packets with large headers, i.e., compress the header in a larger buffer than packetbuf.

**This is sensitive code and it is not tested as much as I wish it were. The bug is severe enough though that I believe we want this in 4.1. Very careful reviews much appreciated here.**


os/net/ipv6/sicslowpan.c
```diff
@@ -910,11 +937,13 @@ compress_hdr_iphc(linkaddr_t *link_destaddr)
            NHC byte extra - before the next header here. This is due to
            next not being elided in that case. */
         if(!IS_COMPRESSABLE_PROTO(*next_hdr)) {
+          CHECK_BUFFER_SPACE(1);
```
joakimeriksson: I guess this it the first critically needed check as we have gotten into the loop here?

joakimeriksson:
I would go further in the future - with a more integrated approach so that we instead of both a CHECK and a "write" have buffer write/read function that will do checks automatically. But that is not for now - this is a fix (which make me think it should be safe to not have it before the while as we can assume that if anyone is using IPv6 they should at least fit a header in packetbuf?).

simonduq:
I pushed a compromise: for the IP header, replaced many checks with a single check of the worst-case, which is 38 bytes. This is followed by the ext and transport header loop, which does fine-grained checking.

Issue #425, opened 2018-04-10
[Potential] Security Vulnerabilities within Erbium


bsmelo:
Hello developers of Contiki-ng/Erbium,

My name is Bruno, and I'm an MSc. student in Brazil within the Institute of Computing from the University of Campinas.
As part of my research on the application of fuzzing techniques for robustness and security black-box testing of CoAP implementations, I've tested your library. The sample used in my research was compiled from distribution/commit 9777ac4 @ 2018-02-03. The application used to test it was `examples/coap/plugtest-server.native`.

I'm contacting you because the application mentioned above was one of the samples for which our tool was able to detect robustness and/or security issues. In a broad sense, every failure we found can actually be classified as a security vulnerability, because they impact availability --- the application either aborts or needs forceful restart in order to restore servicing CoAP requests. However, we didn't go as far as performing a thorough root-cause analysis for those failures, since it would be unfeasible for us (more than 100 failures were detected across 25 samples, each one using a different CoAP library, spanning 8 programming languages) and thus out-of-scope of this particular research.

We think that one of our main contributions is the opportunity to make a real-world impact on IoT security by reporting those failures to CoAP libraries' maintainers, with a comprehensible and easy way to replicate them so developers can further investigate and fix those failures. So, in order to follow up with a responsible disclosure process, we ask for a proper e-mail address (or any other form of contact) so we can send you:
- A script to reproduce the failures;
- A pcap file used by the script, containing the packets causing the failures;
- A logfile with the stacktraces we got for each reported failure.

We expect a reply anytime soon.
Please let us know if which form of contact should we use --- or if it's ok to use this channel.

Thanks & Regards,
Bruno Melo.

simonduq:
Hi,
This is the right place to report such bugs. Please do share the script, pcap and logs, as well as complete instructions how to reproduce. We will investigate.
Thanks a lot for reporting!

bsmelo:
Hello,

We found 1 failure on this sample. You can find the script [1], the pcapng file [2] and the logfile [3] in this message.

The script uses Scapy [4] to read the packets from the pcapng file as well as sending them, which unfortunately requires sudo to send packets in the network. Although sudo for this is ugly (imho), you can still easily inspect the very small and simple script.

Please let us know if you were able to reproduce it and/or if further support is needed for that.

[...]

PR #430, opened 2018-04-12
Added check for block offset overflow in CoAP block2 requests.

nfi:
The block offset is stored in a signed variable in calls to CoAP handlers and too large block offsets will overflow into negative values.

The fixes the issue reported in #425.

### DIFF
--- a/os/net/app-layer/coap/coap-engine.c
+++ b/os/net/app-layer/coap/coap-engine.c
@@ -199,10 +199,18 @@ coap_receive(const coap_endpoint_t *src,
           new_offset = block_offset;
         }
 
-        /* call CoAP framework and check if found and allowed */
-        status = call_service(message, response,
-                              transaction->message + COAP_MAX_HEADER_SIZE,
-                              block_size, &new_offset);
+        if(new_offset < 0) {
+          LOG_DBG("Blockwise: block request offset overflow\n");
+          coap_status_code = BAD_OPTION_4_02;
+          coap_error_message = "BlockOutOfScope";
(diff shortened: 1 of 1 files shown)


g-oikonomou:
@nfi Should we include this in the upcoming release?

Also, I did not look at this in detail, but I was getting the impression when I looked at #425 that this might have been caused by packets with unknown option values, was I mistaken?

simonduq:
+1 on including in 4.1!
@nfi does this fix all the issues in #425 ?

nfi:
@g-oikonomou I would vote for including it in 4.1 also. It is possible to reproduce this issue using only known legal option values together with a CoAP resource that does not validate the block offset such as the example plugfest large-update resource.

@simonduq It fixes the crash described in #425. But testing with #425 also showed that the current CoAP implementation is too forgiving when parsing CoAP headers. Several options in #425 should make CoAP simply respond with Bad Request without processing the payload.

nfi:
To clarify: the issue is that the example CoAP resource plugfest large-update does not check the block offset. If all block2 capable CoAP resources would check the offset instead of simply trusting it, this would not be an issue.

Adding a check in CoAP engine as done in this PR will make sure it will not be an issue even if a CoAP resource does not check the block offset but also adds a limit of the max block offset. Perhaps a better approach would be to update the example CoAP resources after all and document that all block2 capable CoAP resources need to check this. What do you think?

For the long term, I think changing the CoAP handler API to not use signed variables would completely remove the issue.

joakimeriksson:
From what it looks in the RFC it is not a significant limitation to cut down to 31 bits.

PR #432, opened 2018-04-13
Automatically resolve realpath of CONTIKI

lafka:
When using relative path (ie `./contiki-ng`) some Makefiles, like cortex m3, fails
to filter out ldscripts. Ensure that contikis path is normalized.

The relevant code is in `arch/cpu/arm/cortex-m/cm3/Makefile.cm3`:

```
%.elf: $(CPU_STARTFILES) $$(CONTIKI_OBJECTFILES) %.o $(PROJECT_OBJECTFILES) $(PROJECT_LIBRARIES) $(LDSCRIPT)
	$(TRACE_LD)
	$(Q)$(LD) $(LDFLAGS) ${filter-out $(LDSCRIPT) %.a,$^} ${filter %.a,$^} $(TARGET_LIBFILES) -lm -o $@
```

My understanding of makefiles are limited but it seems like here `LDSCRIPT` path is not expanded `filter-out` trying to remove ./contiki-ng/arch/cpu/cc26xx-cc13xx/cc26xx.ld` but the received pattern is actually `contiki-ng/arch/cpu/cc26xx-cc13xx/cc26xx.ld`.

The issue is trivial and easy to fix by using `CONTIKI = contiki-ng`, or setting `CONTIKI` to constant and using `realpath`.


g-oikonomou:
I do not understand this PR. Can you provide an example that I can use to reproduce this? I need to understand expected vs observed behaviour

g-oikonomou:
Thanks for the analysis! So basically what we need to do here is make sure we can build correctly from an out-of-tree location, which is something that we want to be able to do but (AFAIK at least) is not extensively tested. Let me look at it over the next few days.

Thinking aloud: Perhaps what we should also do in a separate PR is add a travis test to build out of tree.

lafka:
I'm not sure how others normally setup their environemnt, for me it's important that i can track different contiki revisions (all the platform code is stored in the contiki tree) so out of tree is required. It's been working fine for me for `srf06-cc26xx`, `native` and our custom platform code.

Having travis do a second run with everything out of tree seems like a good angle to start from.
Having said that, in this specific case it may be difficult to catch these errors since the binary looks fine until its on the actual hardware (although if the linker had died on the warnings it would have been easier to figure out the issue).

Regarding this PR I think it would be better to act defensively and fail with a meaningful error on ambiguous input than trying to guess what the user meant.

Let me know how it goes!

simonduq:
+1 on testing out-of-tree build.
+1 on treating warnings as errors like we already do for compilation. Let's basically add `--fatal-warnings`?

lafka:
Adding fatal warnings works as expected:

```diff
diff --git a/Makefile.include b/Makefile.include
index 984521c2b..e8ce224e5 100644
--- a/Makefile.include
+++ b/Makefile.include
@@ -40,6 +40,7 @@ LOWERCASE = -abcdefghijklmnopqrstuvwxyz/
 UPPERCASE = _ABCDEFGHIJKLMNOPQRSTUVWXYZ_
 TARGET_UPPERCASE := ${strip ${shell echo $(TARGET) | sed y!$(LOWERCASE)!$(UPPERCASE)!}}
 CFLAGS += -DCONTIKI=1 -DCONTIKI_TARGET_$(TARGET_UPPERCASE)=1
+LDFLAGS += -Wl,--fatal-warnings
 ifneq ($(BOARD),)
 TARGET_BOARD_UPPERCASE := ${strip ${shell echo $(BOARD) | sed y!$(LOWERCASE)!$(UPPERCASE)!}}
 CFLAGS += -DCONTIKI_BOARD_$(TARGET_BOARD_UPPERCASE)=1
```

And the result:

```
[user@no-mesh tiny6pan-contiki-ng]$ make TARGET=srf06-cc26xx BOARD=launchpad/cc1310
  CC        ./contiki-ng//arch/cpu/cc26xx-cc13xx/lib/cc13xxware/startup_files/ccfg.c
  CC        ./co.....
/usr/lib/gcc/arm-none-eabi/7.1.0/../../../../arm-none-eabi/bin/ld:contiki-ng//arch/cpu/cc26xx-cc13xx/cc26xx.ld:51: warning: redeclaration of memory region `GPRAM'
make: *** [contiki-ng//arch/cpu/arm/cortex-m/cm3/Makefile.cm3:24: tiny6pan.elf] Error 1
```


A quick check and a warning would be nice to so users know where to start digging:

```diff
diff --git a/Makefile.include b/Makefile.include
index 984521c2b..169bedf9d 100644
--- a/Makefile.include
+++ b/Makefile.include
@@ -4,6 +4,18 @@ ifndef CONTIKI
   ${error CONTIKI not defined! You must specify where Contiki resides}
 endif
 
+REALCONTIKI = $(realpath $(CONTIKI))
+
+# The directory of the initial makefile
+MAKEFILEDIR := $(realpath $(dir $(abspath $(firstword $(MAKEFILE_LIST)))))
+# The expanded path, relative to initial makefile call
+REALCONTIKI := $(REALCONTIKI:$(MAKEFILEDIR)/%=%)
+
+ifneq ($(CONTIKI),$(REALCONTIKI))
+  ${warning CONTIKI = "$(CONTIKI)" - includes extra characters which may unexpected errors in compiled binary.}
+  ${warning Set CONTIKI = "$(REALCONTIKI)" in your Makefile to fix this}
+endif
+
 # Enable Werror by default. To disable from command line, use make WERROR=0.
 # Setting this option is also important for tests on Cooja motes to check for warnings.
 WERROR ?= 1
```


Should i revert original change and add this instead?

simonduq:
I was trying to reproduce locally, but unable to, probably due to a different compiler version (I use what we have in docker, 4.9.3.

I was wondering if we couldn't find a fix on the link command itself, e.g. by using `realpath` there. Am not too comfortable with the warning you are proposing above, as it will show up consistently for relative paths.

Very much in favor of disallowing any linker warning.

lafka:
I'm using GNU Make 4.2.1.

Adding linker warnings resolves the issue completely.

The warnings should only emit when the path can be written in a shorter form. Unfortunately it has the unintended side effect of warning on absolute paths that are better expressed as relative (specifically sub directories). I agree it's not perfect.

When i come to think of it, even the `$(realpath $(CONTIKI))` might have issues if there are any spaces in the path which may break some setups. Might be enough to just add the linker warnings.

simonduq:
Yes, at least, it won't let you build a broken firmware. I'd say, let's do only the link warning thing for now.

lafka:
Removed path conversion and added `LDFLAGS`

PR #443, opened 2018-04-18
Fixing cc26xx 6lbr client example

### LINKED ISSUE #442: cc26xx-web-demo/cetic-6lbr-client does not support RPL Classic
Does simply not compile, see https://github.com/contiki-ng/contiki-ng/issues/431#issuecomment-382214734

It is a problem because 6lbr still requires RPL Classic.

simonduq:
The cc26xx 6lbr client example was meant for RPL Classic.
This PR:
* Fixes compilation of the example for RPL Classic
* Only compiles the example when RPL Classic is enabled
* Adds a compile test for this example with RPL Classic

Fixes #442

### DIFF
--- a/examples/platform-specific/cc26xx/cc26xx-web-demo/Makefile
+++ b/examples/platform-specific/cc26xx/cc26xx-web-demo/Makefile
@@ -5,8 +5,13 @@ PLATFORMS_ONLY = srf06-cc26xx
 
 MODULES_REL += ./resources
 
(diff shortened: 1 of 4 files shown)


g-oikonomou:
I think when we get to a stage where 6lbr supports RPL lite then we need a nice cross-routing (so to speak) way of retrieving the preferred parent. This patch fixes compilation for RPL classic, but now the same file won't compile under lite (used to!). Even though we don't need it to do so at this stage, it would be nice to get rid of conditional compilation and let the linker handle things based on the values of macros.

But for now 👍

PR #451, opened 2018-04-24
Main: always log 802.15.4 default channel

simonduq:
This PR proposes to print the default 15.4 channel unconditionally (including when custom MAC implementations are used, such as in the native BR).


simonduq:
Wait or maybe we should show this only in the non-TSCH case? This would be consistent with `viewconf`. Or else, we update `viewconf` to always show the default channel.

g-oikonomou:
I cannot immediately see a reason why not show it for any MAC layer. I think I'd vote about adding it to viewconf unconditionally.

simonduq:
The reason for TSCH would be that, well, TSCH basically ignore the default channel altogether. Only uses own hopping sequence.

g-oikonomou:
OK I'm confused now. What is it that is undesirable about the current code in the first place?

simonduq:
It was that custom MAC, i.e. not CSMA nor TSCH, did not have the channel printed out. A custom MAC is currently in use in the RPL native BR.

g-oikonomou:
Hm OK I see. So if TSCH then hopping sequence, else channel?

simonduq:
Yes! :) I think so too.

simonduq:
done :)

PR #453, opened 2018-04-24
Added jool-start.sh script

simonduq:
For use in https://github.com/contiki-ng/contiki-ng/wiki/NAT64-for-Contiki‐NG


g-oikonomou:
Does this need a shebang? Presumably as per the wiki this is linux only? Perhaps we need to make this more explicit here and/or in the wiki?

Great you managed to get Jool to play nicely, I remember failing miserably a few years ago :D

simonduq:
Yes, should add the shebang. Jool now works nicely yes! It is Linux only though, unfortunately. Will make sure the doc is clear about this.

simonduq:
Shebang added

PR #461, opened 2018-04-26
Added a periodic timer to the hello-world example

nfi:
Programming a device with the hello-world example and then connect to the device via serial port to see the output normally shows nothing since the example only prints the message once after boot. This might be confusing for new users.

This PR adds a periodic timer to print the message periodically and also replaces the `printf()` with a call to the logging API.

### DIFF
--- a/examples/hello-world/hello-world.c
+++ b/examples/hello-world/hello-world.c
@@ -46,10 +46,21 @@ AUTOSTART_PROCESSES(&hello_world_process);
 /*---------------------------------------------------------------------------*/
 PROCESS_THREAD(hello_world_process, ev, data)
 {
+  static struct etimer timer;
+
   PROCESS_BEGIN();
 
-  printf("Hello, world\n");
-  
+  /* Setup a periodic timer that expires after 10 seconds. */
+  etimer_set(&timer, CLOCK_SECOND * 10);
+
+  while(1) {
+    printf("Hello, world\n");
+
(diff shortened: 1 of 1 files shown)


g-oikonomou:
I clicked approve, but do we really want to use logging here? Under what circumstances would a user want to disable this output?

examples/hello-world/hello-world.c
```diff
@@ -39,17 +39,32 @@
...
-  
+  /* Setup a periodic timer that expires after 10 seconds. */
+  etimer_set(&timer, CLOCK_SECOND * 10);
+
+  while(1) {
+    LOG_INFO("Hello, world\n");
+
+    /* Wait for the periodic timer to expire and then restart the timer. */
+    PROCESS_WAIT_EVENT_UNTIL(etimer_expired(&timer));
+    etimer_restart(&timer);
```
simonduq: maybe `etimer_reset` instead?

nfi: I have changed to `etimer_reset` now.

simonduq:
I vote for it because the output gives more of an OS feel. Classic printf almost gives the impression the process escaped from its host if you see what I mean ;) No strong opinion tho

nfi:
I do not have any strong opinion regarding the logging either. However it seems to break one of the regression tests so either we switch back to `printf` or I need to update the regression test. Either way is fine for me.

g-oikonomou:
Either sounds good to me Niclas

PR #462, opened 2018-04-26
Updated users of the SLIP driver to respect UIP_LLH_LEN

### LINKED ISSUE #191: SLIP-cmd ignores UIP_LLH_LEN
To send an IPv6 datagram over SLIP, what we should be doing (and have been, even before #190) is to start at `&uip_buf[UIP_LLH_LEN]`. However, we still send `uip_len` bytes, which is clearly wrong. The current code base works by fluke, because on all cases where SLIP is being used `UIP_LLH_LEN` is 0. Change `UIP_LLH_LEN` and we'll overrun `uip_buf`. An easy fix could be to send `uip_len - UIP_LLH_LEN`.

Secondly, the SLIP-cmd interface puts commands/responses into `uip_buf` starting at position 0 and then uses slip_send. Things only work because `UIP_LLH_LEN` is 0. An easy fix could be to start populating/parsing starting from position `UIP_LLH_LEN`

nfi:
The SLIP driver adds incoming data into `uipbuf[UIP_LLH_LEN]` and some users of the SLIP driver simply assumed that UIP_LLH_LEN is zero. This PR makes sure that listeners on the SLIP driver respect `UIP_LLH_LEN` and this fixes #191.


simonduq:
Was this tested on H/W?

nfi:
Tested with native rpl-border-router and Zolertia Firefly as slip-radio but so far only with UIP_LLH_LEN set to 0.

joakimeriksson:
I also tested now. Nothing breaks - it seems.

Issue #459, opened 2018-04-26
A pass on all tutorials prior to 4.1


simonduq:
Basics:
* [ ] Hello, World! 
* [ ] Logging
* [ ] NG shell
* [ ] RAM and ROM usage
* [ ] Energy monitoring
* [ ] Timers and events

Networking:
* [ ] IPv6 ping
* [ ] RPL basics
* [ ] RPL with border router
* [ ] TSCH and 6TiSCH
* [ ] CoAP
* [ ] LWM2M

Simulation:
* [ ] Cooja: getting started
* [ ] Cooja: simulating a RPL network
* [ ] Cooja: simulating a RPL network with a border router
* [ ] Running Contiki-NG in Renode

simonduq:
I propose the removal of *Introduction to Contiki-NG Programming* , whose content is redundant with *Processes and events*

g-oikonomou:
Renode tutorial on Mac OS:

```
$ make renode TARGET=cc2538dk SCRIPT=rpl-udp.resc 
  CC        ../../arch/cpu/cc2538/./ieee-addr.c
  CC        ../../arch/cpu/cc2538/cc2538.lds
  CC        ../../arch/cpu/cc2538/./startup-gcc.c
  CC        udp-client.c
  LD        udp-client.elf
arm-none-eabi-objcopy -O ihex udp-client.elf udp-client.i16hex
srec_cat udp-client.i16hex -intel -o udp-client.hex -intel
arm-none-eabi-objcopy -O binary --gap-fill 0xff udp-client.elf udp-client.bin
cp udp-client.elf udp-client.cc2538dk
  CC        udp-server.c
  LD        udp-server.elf
arm-none-eabi-objcopy -O ihex udp-server.elf udp-server.i16hex
srec_cat udp-server.i16hex -intel -o udp-server.hex -intel
arm-none-eabi-objcopy -O binary --gap-fill 0xff udp-server.elf udp-server.bin
cp udp-server.elf udp-server.cc2538dk
renode rpl-udp.resc
make: renode: No such file or directory
make: *** [renode] Error 1
rm udp-client.o udp-server.o udp-client.i16hex obj_cc2538dk/startup-gcc.o udp-server.i16hex
```

PiotrZierhoffer:
This was already reported to us by Joakim earlier today. On macOS we need to manually create a symlink or an alias, as DMGs are just mounted and can't influence the rest of the filesystem.

[...]

PR #474, opened 2018-04-28
Tidy-up renode for OS X and enable for all CC2538DK

g-oikonomou:
This pull:

* Enables `make renode` for OS X without having to use an alias. The build system invokes renode using the correct command for all supported OSs.
* Enables the `renode` target for all CC2538DK examples. To add support for more examples, all we need to do is to provide the corresponding `.resc` file.

If the user / Makefile does not specify a renode config `.resc` file through the `SCRIPT` variable, the build system will automatically try with `<example-dir-name>.resc` (e.g. `hello-world.resc`)

See discussion in #459


g-oikonomou:
I think it's fine on develop, I'm somewhat disinclined to sneak in last minute changes to the RC branch. Thanks for the review!

simonduq:
Do we want only cc2538dk or also all other cc2538 platforms?

joakimeriksson:
Some things will work on other platforms but I guess Zoul with the CC1200 need significant testing before we know it works. Maybe the OpenMote - if it is without a CC1200?

g-oikonomou:
The original code only supported CC2538DK

```
-ifneq ($(TARGET),cc2538dk)
-	$(error Only the cc2538dk TARGET is supported for Renode demo scripts)
-endif
```

If other platforms may work, then we need to test and update docs etc as to what is expect to work and what is not.

Issue #197, opened 2017-11-23
Semantics of the Radio API


g-oikonomou:
The radio API is not documented and this has caused confusion as to how functions are meant to behave. I think we need to document `radio.h` in terms of what each function is meant to do:

* `init`:
  * Is it meant to enable frame filtering and auto ACKs, if supported? Will TSCH then go and disable those things since they need done in software?
  * Is it meant to select a default channel, or will the channel get populated explicitly later?
  * Is it meant to put the radio in RX?
* `transmit`, `channel_clear`, `send`, `receiving_packet`:
  * If they are called when the radio is off, are they expected to turn the radio on automatically, or are they expected to fail?
  * If they are expected to turn the radio on automatically, what are they expected to do at the end of the operation? Restore the radio to its previous state? Leave the radio on?
  * Is the MAC layer perhaps guaranteed to explicitly turn the radio on before it tries to perform one of those operations (so the answer to the previous question would be to fail if off)?

simonduq:
We really need to fix this. A lot easier now with our more limited set of supported platforms.
One more:
* `RADIO_RX_MODE_ADDRESS_FILTER`: is this only about address filtering (as the name suggests) or is it about 802.15.4 frame parsing (as some of the radio drivers implement)? Shall we add `RADIO_RX_MODE_FRAME_FILTER`?

g-oikonomou:
Hi there all.

I have been looking at this extensively over the last few days. Our MAC layers are being a little inconsistent in terms of how they use the radio API. This could be a result of our MAC layer implementations second-guessing what the radio does, which in turn has resulted in radio drivers second-guessing what the MAC layer does.

## The behaviour of the `transmit()` function
The main inconsistency is in terms of the `transmit()` function:

* CSMA will call `on()` once at the start and it will then call `transmit()` whenever it has a frame to send. However, it will not explicitly put the radio back in RX after TX - it will expect this transition to happen automatically. Now this is fine for radios that automatically revert to RX after TX, but not all radios do (especially so sub-ghz radios). Assuming this behaviour, our drivers try to put radios back in RX after TX.
* TSCH behaves slightly differently:
  * If we use `TSCH_RADIO_ON_DURING_TIMESLOT` then the behaviour is similar to CSMA. However if `TSCH_RADIO_ON_DURING_TIMESLOT` is 0 then TSCH will explicitly call `on()` after TX.
[...]

PR #477, opened 2018-04-30
rx full IRQ Handling for cc13xx/cc26xx

MartinNPN:
Hello,

now the correct pull request.

After i testes these pull request from contiki https://github.com/contiki-os/contiki/pull/2161 and it gave at high traffic a huge improvement for me, it should be even changed at contiki-ng.

It only implements the rx full IRQ handling, even the points regarding the bypass of the rf on checks should not apply to it.

Please check the request.

Regards,
Martin


atiselsts:
Cool, this whole problem reminds me of the reasons why Linux NAPI was introduced: https://en.wikipedia.org/wiki/New_API

George will correct me if I'm wrong, but after reading the discussion in contiki#2161, this PR does not cover all possible scenarios? So, what needs to be done according to my read of the previous discussion:

-  Set a flag (`volatile bool is_rx_full`) in the interrupt handler
-  Clear the `is_rx_full` flag in the `read_frame` function after reading the frame and turning the radio back on, if needed (not sure if another flag is needed here to remember the desired Rx state?)
-  Don't go to deep sleep if radio is disabled because of rx buffers being full (`is_rx_full == true`)

`smartrf_settings_cmd_prop_rx_adv.status` is not a good flag to rely, because there can be race conditions with other functions being called before `read_frame` that might change its state, as discussed.

atiselsts:
It should be ok to read `smartrf_settings_cmd_prop_rx_adv.status` in the IRQ handler; what I meant it's not reliable anymore when the `read_frame` is executed, since it could be rewritten by that time. So it would be better to remember the status ion other ways.

I can give the implementation a shot, which you could then test, since you already have the high packet rate setup in place - but first let's check with @g-oikonomou on his input.

MartinNPN:
Ok, i get your point, the reading of `smartrf_settings_cmd_prop_rx_adv.status` should be at the IRQ.

atiselsts:
`read_frame` is not guaranteed to be called immediately after the IRQ returns. There can be race conditions with other radio operations: see "Scenario 3". 

I would go even further and say that even if we would be 100% sure that those race conditions are not realistically possible with the *current* Contki-NG code, it would still be better not to rely on the `smartrf_settings_cmd_prop_rx_adv.status` flag as the flow might change in the future.

g-oikonomou:
There are a number key problems here:

* As discussed here, the driver is relying too much on the value of `smartrf_settings_cmd_prop_rx_adv.status` to determine whether the RF is in RX mode. This needs fixed and what Atis is suggesting with a volatile variable makes sense. Polling the process when the RX buffer is full is a good idea.

* The second problem is that the buffer should not be filling up so easily in the first place. As I have commented elsewhere, it makes little sense that we have received 4 frames before processing the 1st one. I believe the problem to be that we are sitting way too much time in each CCA check.

* Lastly, this was mainly observed by ContikiMAC, really due to #197. The driver tries to restore the radio to its previous state after each time it has completed an operation (e.g. a CCA or a TX). This on occasion can break slip-radio. I have no expectation to fix this as part of this pull, but nice to have in mind for future sub-ghz TSCH and/or CSL!

MartinNPN:
Hello,

ok, i think you are right, cause the rx frame process is not called immediately it should be better to use a variable. 

The other two point may be the root of the problem, so they should be fixed too.

arch/cpu/cc26xx-cc13xx/rf-core/prop-mode.c
```diff
@@ -795,6 +796,12 @@ read_frame(void *buf, unsigned short buf_len)
     entry->status = DATA_ENTRY_STATUS_PENDING;
   }
 
+  if(rx_is_full) {
+    rx_is_full = false;
```
g-oikonomou: The comparison and setting to false should be inside a critical section, since the interrupt may fire after the comparison, but before setting the value to false. In this case, we may end up with the variable being set to false erroneously. Please use the functions in critical.h

arch/cpu/cc26xx-cc13xx/rf-core/prop-mode.c
```diff
@@ -795,6 +796,12 @@ read_frame(void *buf, unsigned short buf_len)
...
+    printf("RXQ was full, re-enabling radio!\n");
+    rx_on_prop();
```
g-oikonomou: We may have to think of a better logic here in the future, but since we are only using CSMA this will do for now I think

g-oikonomou:
@MartinNPN just to make sure we are on the same page, I am waiting for this to get rebased and the history cleaned up before merging.

simonduq:
Ping @MartinNPN

PR #479, opened 2018-05-03
Make 802.15.4 ACK's configurable for CSMA mac

lafka:
Make 802.15.4 ACK's configurable for CSMA mac
    
There was no way to control the ACK flag in 15.4 header making it impossible to send frames that should not be acknowledged.
    
Introduce new fields`CSMA_CONF_REQUEST_ACK` to allow some level of configurability. `CSMA_CONF_SEND_SOFT_ACK` still controls whetever the receiver will actually send an ACK from software.

This simplifies previously suggested logic by making it a system wide configuration. In the future, and if desirable, the upper layers can should have a way to toggle ACKs on a packet by packet basis


simonduq:
Maybe a solution is to extend `PACKETBUF_ATTR_MAC_METADATA` and make it a bitfield, with one new bit for "unacked unicast".

lafka:
Updated. Simplified by only having a system wide ack flag.

os/net/mac/csma/csma.h
```diff
@@ -48,9 +48,15 @@
 #ifdef CSMA_CONF_SEND_SOFT_ACK
 #define CSMA_SEND_SOFT_ACK CSMA_CONF_SEND_SOFT_ACK
 #else /* CSMA_CONF_SEND_SOFT_ACK */
-#define CSMA_SEND_SOFT_ACK 0
+#define CSMA_SEND_SOFT_ACK 1
```
simonduq: why this change?

lafka: ooops! separating out a lot of changes on my side here; seems like i staged the wrong change. should be `CSMA_REQUEST_ACK 1` to have same behaviour as before. Fixing.

lafka:
Updated, setting the correct value for `CSMA_SEND_SOFT_ACK` and `CSMA_REQUEST_ACK`.

os/net/mac/csma/csma-output.c
```diff
@@ -200,6 +200,8 @@ send_one_packet(void *ptr)
       case RADIO_TX_OK:
         if(is_broadcast) {
           ret = MAC_TX_OK;
+        } else if (!CSMA_REQUEST_ACK) {
```
simonduq: One more thing: I wonder if instead of this added condition, we shouldn't just replace the variable `is_broadcast` with a new `is_ack_requested`. Would make more sense w.r.t handling of packet pending prior to tx (code block above).

lafka: Makes perfect sense!

One thing though, I don't get the point of the second condition here:

```c
if(NETSTACK_RADIO.receiving_packet() ||
   (!is_broadcast && NETSTACK_RADIO.pending_packet())) {
```

We don't transmit if the radio just received a packet and the outgoing packet is unicast?  What's the logic behind it? Is the idea that if we received a unicast packet (and not done processing) it's considered a collision since we should be responding with an ACK first?



Where you thinking of something like this? 

```diff
diff --git a/os/net/mac/csma/csma-output.c b/os/net/mac/csma/csma-output.c
index 9c88d569d..9c70ea8c3 100644
--- a/os/net/mac/csma/csma-output.c
+++ b/os/net/mac/csma/csma-output.c
@@ -168,27 +168,28 @@ backoff_period(void)
 static int
 send_one_packet(void *ptr)
 {
+  int expect_ack;
   int ret;
   int last_sent_ok = 0;
 
+  expect_ack = !packetbuf_holds_broadcast() || CSMA_REQUEST_ACK;
+
   packetbuf_set_addr(PACKETBUF_ADDR_SENDER, &linkaddr_node_addr);
-  packetbuf_set_attr(PACKETBUF_ATTR_MAC_ACK, 1);
+  packetbuf_set_attr(PACKETBUF_ATTR_MAC_ACK, expect_ack);
 
   if(NETSTACK_FRAMER.create() < 0) {
     /* Failed to allocate space for headers */
     LOG_ERR("failed to create packet\n");
     ret = MAC_TX_ERR_FATAL;
   } else {
-    int is_broadcast;
     uint8_t dsn;
     dsn = ((uint8_t *)packetbuf_hdrptr())[2] & 0xff;
 
     NETSTACK_RADIO.prepare(packetbuf_hdrptr(), packetbuf_totlen());
 
-    is_broadcast = packetbuf_holds_broadcast();
 
     if(NETSTACK_RADIO.receiving_packet() ||
-       (!is_broadcast && NETSTACK_RADIO.pending_packet())) {
+       (expect_ack && NETSTACK_RADIO.pending_packet())) {
 
       /* Currently receiving a packet over air or the radio has
          already received a packet that needs to be read before
@@ -198,7 +199,7 @@ send_one_packet(void *ptr)
 
       switch(NETSTACK_RADIO.transmit(packetbuf_totlen())) {
       case RADIO_TX_OK:
-        if(is_broadcast) {
+        if(!expect_ack) {
           ret = MAC_TX_OK;
         } else {
           rtimer_clock_t wt;
```

simonduq: The idea I believe is that if there is something pending we might have the Rx fifo full and be not able to receive the ACK.

At first glance the patch above looks good yes!

simonduq:
Hi, any plan to update this PR with the discussed enhancement?

lafka:
> Hi, any plan to update this PR with the discussed enhancement?

Yes, i'm working on that part of stack again. Expect to finish PR within 2-3 weeks time.

lafka:
Rebased and simplified.

I don't see a use case anymore for tagging individual packets with ACK flag. I've changed the patch to either set ACK for all or no packets.

os/net/mac/csma/csma-output.c
```diff
@@ -168,11 +168,15 @@ backoff_period(void)
...
+  int expect_ack;
   int ret;
   int last_sent_ok = 0;
 
+  expect_ack = !packetbuf_holds_broadcast() && CSMA_REQUEST_ACK;
+
   packetbuf_set_addr(PACKETBUF_ADDR_SENDER, &linkaddr_node_addr);
-  packetbuf_set_attr(PACKETBUF_ATTR_MAC_ACK, 1);
+
+  packetbuf_set_attr(PACKETBUF_ATTR_MAC_ACK, expect_ack);
```
simonduq: ok we had a bug there, setting `PACKETBUF_ATTR_MAC_ACK` even for broadcast, but somehow didn't have any effect, as the framer didn't look at this attribute for broadcast, and left the ACK Requested frame field unset?

os/net/mac/csma/csma-output.c
```diff
@@ -205,7 +206,9 @@ send_one_packet(struct neighbor_queue *n, struct packet_queue *q)
 
       switch(NETSTACK_RADIO.transmit(packetbuf_totlen())) {
       case RADIO_TX_OK:
-        if(is_broadcast) {
+        if(!expect_ack) {
+          ret = MAC_TX_OK;
+        } else if (!expect_ack) {
```
simonduq: `if(!expect_ack) ..  else if (!expect_ack)`?

os/net/mac/csma/csma-output.c
```diff
@@ -214,6 +217,7 @@ send_one_packet(struct neighbor_queue *n, struct packet_queue *q)
           RTIMER_BUSYWAIT_UNTIL(NETSTACK_RADIO.pending_packet(), CSMA_ACK_WAIT_TIME);
 
           ret = MAC_TX_NOACK;
+
```
simonduq: a whitespace was added here

PR #491, opened 2018-05-04
Improved default file size in Antelope

nvt:
This PR makes it possible to run Antelope using the default configuration on some platforms with relatively small file system areas. Previously, the shell-db example would not work on Zoul Firefly without reducing DB_COFFEE_RESERVE_SIZE by a large amount. Thanks to @joakimeriksson for reporting the problem.


joakimeriksson:
Seems to work fine on Zoul now!

PR #493, opened 2018-05-06
Makefile.include: include nullnet only when using MAKE_NET_NULLNET

### LINKED ISSUE #465: Incorrect CFLAGS with MAKE_NET_OTHER?
When we build with `MAKE_NET=MAKE_NET_OTHER` the command line includes `-DNETSTACK_CONF_WITH_NULLNET=1 -DNETSTACK_CONF_WITH_OTHER=1`. This doesn't feel right.

simonduq:
Fixes #465


Makefile.include
```diff
@@ -172,7 +172,9 @@ endif
 ifeq ($(MAKE_NET),MAKE_NET_IPV6)
   CFLAGS += -DNETSTACK_CONF_WITH_IPV6=1
   MODULES += os/net/ipv6
-else
+endif
+
+ifeq ($(MAKE_NET),MAKE_NET_NULLNET)
   CFLAGS += -DNETSTACK_CONF_WITH_NULLNET=1
   MODULES += os/net/nullnet
```
nfi: I suggest `MODULES` is moved up to the other `MAKE_NET_NULLNET` check at line 170 and that this second `MAKE_NET_NULLNET` check is simply removed.

simonduq: Oh yes good catch! Will do

simonduq:
Fixed

PR #497, opened 2018-05-07
Shell improvements

mcr:
These three small patches add a second newline to the shell, make the prompt always start on a new line, and skip commands that are empty.

### DIFF
--- a/os/services/shell/shell.c
+++ b/os/services/shell/shell.c
@@ -106,7 +106,7 @@ shell_output_lladdr(shell_output_func output, const linkaddr_t *lladdr)
 static void
 output_prompt(shell_output_func output)
 {
-  SHELL_OUTPUT(output, "#");
+  SHELL_OUTPUT(output, "\n#");
   shell_output_lladdr(output, &linkaddr_node_addr);
   SHELL_OUTPUT(output, "> ");
 }
--- a/os/services/shell/shell.c
+++ b/os/services/shell/shell.c
@@ -130,20 +130,22 @@ PT_THREAD(shell_input(struct pt *pt, shell_output_func output, const char *cmd))
     args++;
   }
 
-  /* Lookup for command */
-  cmd_ptr = shell_commands;
-  while(cmd_ptr->name != NULL) {
-    if(strcmp(cmd, cmd_ptr->name) == 0) {
-      static struct pt cmd_pt;
-      PT_SPAWN(pt, &cmd_pt, cmd_ptr->func(&cmd_pt, output, args));
-      goto done;
+  /* only process lines with something on them (having removed leading space) */
+  if(strlen(cmd) > 0) { 
+    /* Lookup for command */
+    cmd_ptr = shell_commands;
+    while(cmd_ptr->name != NULL && strcmp(cmd, cmd_ptr->name) != 0) {
+      cmd_ptr++;
     }
-    cmd_ptr++;
+
+    if(cmd_ptr->name != NULL) {
+	static struct pt cmd_pt;
+	PT_SPAWN(pt, &cmd_pt, cmd_ptr->func(&cmd_pt, output, args));
+    } else {
+      SHELL_OUTPUT(output, "Command not found. Type 'help' for a list of commands\n");
+    }
   }
 
-  SHELL_OUTPUT(output, "Command not found. Type 'help' for a list of commands\n");
-
-done:
   output_prompt(output);
   PT_END(pt);
 }
--- a/os/services/shell/shell.c
+++ b/os/services/shell/shell.c
@@ -131,7 +131,7 @@ PT_THREAD(shell_input(struct pt *pt, shell_output_func output, const char *cmd))
   }
 
   /* only process lines with something on them (having removed leading space) */
-  if(strlen(cmd) > 0) { 
+  if(strlen(cmd) > 0) {
     /* Lookup for command */
     cmd_ptr = shell_commands;
     while(cmd_ptr->name != NULL && strcmp(cmd, cmd_ptr->name) != 0) {
@@ -139,8 +139,8 @@ PT_THREAD(shell_input(struct pt *pt, shell_output_func output, const char *cmd))
     }
 
     if(cmd_ptr->name != NULL) {
-	static struct pt cmd_pt;
-	PT_SPAWN(pt, &cmd_pt, cmd_ptr->func(&cmd_pt, output, args));
+      static struct pt cmd_pt;
+      PT_SPAWN(pt, &cmd_pt, cmd_ptr->func(&cmd_pt, output, args));
     } else {
       SHELL_OUTPUT(output, "Command not found. Type 'help' for a list of commands\n");
     }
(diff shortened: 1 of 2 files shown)


os/services/shell/shell.c
```diff
@@ -130,6 +130,9 @@ PT_THREAD(shell_input(struct pt *pt, shell_output_func output, const char *cmd))
     args++;
   }
 
+  /* ignore empty lines */
+  if(strlen(cmd)==0) goto done;
```
simonduq: Code style: even single-line blocks must be within `{}`
Goto: we generally try to have them only if it simplifies the flow significantly. In this case I'd personally find it neater to just put the command lookup in an `if` block.

os/services/shell/shell.c
```diff
@@ -106,7 +106,7 @@ shell_output_lladdr(shell_output_func output, const linkaddr_t *lladdr)
 static void
 output_prompt(shell_output_func output)
 {
-  SHELL_OUTPUT(output, "#");
+  SHELL_OUTPUT(output, "\n#");
```
simonduq: OK. Is this for the case of empty input?

mcr: This just ALWAYS starts a prompt on newline, in case the previous output was half-baked, etc.

os/services/shell/shell.c
mcr: Got rid of both uses of goto.

simonduq:
Tested and works for me. I'd like to hear about others preferences regarding the extra newline. No strong opinion on my side, so this is ok with me.

g-oikonomou:
I do not feel strongly about the extra newline. I can see the motivation.

I do feel somewhat strongly about spaces instead of tabs :)

mcr:
Simon Duquennoy <notifications@github.com> wrote:
    > Looks good now. Minor: looks like you've used TABs for indentation
    > instead of spaces. I'll test before approving.

Shouldn't be me!   My editor doesn't do that!

-- 
]               Never tell me the odds!                 | ipv6 mesh networks [ 
]   Michael Richardson, Sandelman Software Works        | network architect  [ 
]     mcr@sandelman.ca  http://www.sandelman.ca/        |   ruby on rails    [

g-oikonomou:
Just put the thing through `uncrustify-check-style`, see what it comes up with that falls within your proposed diff, and adjust accordingly.

mcr:
Maybe make uncrustify-check-style one of the travis tests. I'll see if I can add that.

g-oikonomou:
You may wish to look at `uncrustify-changed.sh`, but do keep in mind that among all the things uncrustify complains about, we only care about a sub-set. So uncrustify non-empty output on travis would not necessarily constitute style failure. Plus we'd also have to fix the format on existing files first, and that could end up being quite a diff .... ;)

mcr:
okay, ran uncrustify on the directory. There were many changes.
What is the subset that I should care about?  shouldn't uncrustify.cfg take care of that?

g-oikonomou:
Please see this page: https://github.com/contiki-ng/contiki-ng/wiki/Code-style

Unfortunately I have been unable to capture all style requirements in the cfg. For example, it is removing the spaces used to align the fields of that struct. This can be ignored.

You should not bother with anything outside the contributed diff, unless you are also planning to contribute a code style tidy-up

mcr:
I will revise this patch to just be my code.
Openssl has the same problem: "here use this tool to make your code look right"/"oh, but ignore this part"...  it's hard.  Do you have an emacs C-style definition, by any chance?

yatch:
@mcr I have the following lines in my config:

```
(setq-default indent-tabs-mode nil)
(setq c-basic-offset 2)
(setq c-default-style '((java-mode . "java")
                        (awk-mode . "awk")
                        (other . "bsd")))
(c-set-offset 'case-label '+)
```

firmwareguru:
@mcr I'd like to get backspace handling into the shell system.  If you'd like to incorporate this feature with your PR, here is the patch:

```
diff --git a/os/dev/serial-line.c b/os/dev/serial-line.c
index 80ff5510b..00d373869 100644
--- a/os/dev/serial-line.c
+++ b/os/dev/serial-line.c
@@ -53,6 +53,10 @@
 #define END 0x0a
 #endif

+#ifndef BS
+#define BS 0x08
+#endif
+
 static struct ringbuf rxbuf;
 static uint8_t rxbuf_data[BUFSIZE];

@@ -108,7 +112,13 @@ PROCESS_THREAD(serial_line_process, ev, data)
       PROCESS_YIELD();
     } else {
       if(c != END) {
-        if(ptr < BUFSIZE-1) {
+        /* handle backspace */
+        if(BS == c) {
+          if(ptr > 0) {
+            ptr--;
+          }
+        }
+        else if(ptr < BUFSIZE-1) {
           buf[ptr++] = (uint8_t)c;
         } else {
           /* Ignore character (wait for EOL) */
```

mcr:
Evan Ross <notifications@github.com> wrote:
    > @mcr I'd like to get backspace handling into the shell system.  If
    > you'd like to incorporate this feature with your PR, here is the patch:

okay.

os/services/shell/shell.c
simonduq: I tried this newline thing and I'm not a fan: doesn't help with the case of constantly coming log output anyway, and I IMO the empty lines takes away part of the shell look-and-feel

simonduq:
@mcr any plan to update this?

jedward1977:
Hi
Is the shell problem about CR and/or LF EOL going to be resolved in this commit?
I am trying to use the shell and it is not working even if using stty command.

make TARGET=srf06-cc26xx BOARD=launchpad/cc2650 login PORT=/dev/ttyACM0
rlwrap /home/john/projects/tools/contiki-ng/tools/serial-io/serialdump -b115200 /dev/ttyACM0
connecting to /dev/ttyACM0 [OK]
[INFO: Main      ] Starting Contiki-NG-develop/v4.2-28-g42436cc-dirty
[INFO: Main      ] - Routing: RPL Lite
[INFO: Main      ] - Net: sicslowpan
[INFO: Main      ] - MAC: CSMA
[INFO: Main      ] - 802.15.4 PANID: 0x0e0e
[INFO: Main      ] - 802.15.4 Default channel: 21
[INFO: Main      ] Node ID: 6531
[INFO: Main      ] Link-layer address: 0012.4b00.0f83.1983
[INFO: Main      ] Tentative link-local IPv6 address: fe80::212:4b00:f83:1983
[INFO: CC26xx/CC13xx] TI CC2650 LaunchPad
[INFO: CC26xx/CC13xx]  RF: Channel 21, PANID 0x0E0E
[INFO: App       ] Starting Erbium Example Server

xxxxx
asaaa

firmwareguru:
@jedward1977  The shell isn't turned on by default.  You need to include the module in a project's makefile, like this:

`MODULES += os/services/shell `  (which then sets `BUILD_WITH_SHELL=1`)

simonduq:
@jedward1977 if this works for you please would be great if you could cherry-pick the fix commit and open a new PR with it (because this PR seems inactive)

simonduq:
OK this has been pending for a very long time... now the shell in `develop` already skips empty lines. I cherry-picked the CR/LF commit into https://github.com/contiki-ng/contiki-ng/pull/1067

Closing this. Thanks for contributing!

PR #502, opened 2018-05-09
Release v4.1

g-oikonomou:
For consistency, please change the merge commit message to "Release v4.1"


g-oikonomou:
Branch master diverged from develop at the point of release of v4.0 (the merge commit of v4.0 never appeared in branch develop's history. We are now merging master into release-4.1 in order to resolve this divergence, and then release-4.1 back into master to execute release v4.1.

This will not be required for the next release in the future, since master will no longer be diverging.

Issue #499, opened 2018-05-09
RPL Lite (and Classic to a lesser extent) issue in congested environment


ianno:
I tried running the simulation in `examples/rpl-udp/rpl-udp-cooja.csc` using nodes with TSCH enabled.
In this scenario, nodes will periodically lose their connection to the network (recurrent 'Not reachable yet' messages).
I'm pasting the modified Makefile (`examples/rpl-udp/Makefile`):

```
CONTIKI_PROJECT` = udp-client udp-server
all: $(CONTIKI_PROJECT)

CONTIKI=../..

MAKE_MAC = MAKE_MAC_TSCH
include $(CONTIKI)/Makefile.include
```

ianno:
It seems the issue is with the RPL LITE driver.
Using `MAKE_ROUTING = MAKE_ROUTING_RPL_CLASSIC` in the Makefile makes the simulation working without problems.

simonduq:
Hey I believe the issue to be specific to Cooja motes, where there are known sync issues. But I see your added comment, definitely needs investigating more.

ianno:
Thanks for the super fast reply.
If it can help, this is what happens:

```
07:49.735	[WARN: TSCH      ] ! EB JP too high 36, leaving the network	ID:13
07:49.735	[INFO: TSCH-LOG  ] {asn 00.0000b757 link  0   7   0  0 ch 20} drift 0	ID:13
07:49.735	[INFO: TSCH-LOG  ] {asn 00.0000b757 link  0   7   0  0 ch 20} bc-0-0 rx LL-000c->LL-NULL, len  35, seq  60, edr   0, dr   0	ID:13
07:49.735	[WARN: TSCH      ] leaving the network, stats: tx 16, rx 13, sync 1	ID:13
07:49.735	[INFO: TSCH Queue] update time source: 000c.000c.000c.000c -> (NULL LL addr)	ID:13
07:49.735	[WARN: RPL       ] local repair (TSCH leaving)	ID:13
07:49.735	[INFO: RPL       ] reset MRHOF	ID:13
07:49.735	[INFO: RPL       ] removing all neighbors	ID:13
07:49.735	[INFO: RPL       ] parent switch: fe80::20c:c:c:c -> (NULL IP addr)	ID:13
07:49.735	[WARN: RPL       ] no parent, scheduling periodic DIS, will leave if no parent is found	ID:13
[...]

PR #503, opened 2018-05-09
RPL Lite: disable MRHOF ETX squaring by default

simonduq:
As discussed in https://github.com/contiki-ng/contiki-ng/issues/499 , and as reported by other users in the past, the added churn from squared ETX can cause the network to never converge to a stable state. This happens in scenarios with high congestion or poor links, where link estimates keep changing, and the changes are amplified by squaring.

This PR disabled the feature by default. A rule of thumbs is to always start without, and if the network operate stably, try enabling it to push the reliability further. But in more noisy/dense environments, stick to the default.

### DIFF
--- a/os/net/link-stats.c
+++ b/os/net/link-stats.c
@@ -64,7 +64,7 @@
 /* ETX fixed point divisor. 128 is the value used by RPL (RFC 6551 and RFC 6719) */
 #define ETX_DIVISOR                     LINK_STATS_ETX_DIVISOR
 /* Number of Tx used to update the ETX EWMA in case of no-ACK */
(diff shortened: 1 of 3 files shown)


simonduq:
Does not play well with the probing test, moving back to develop.

simonduq:
Increased `ETX_NOACK_PENALTY`, which was too low to trigger fast parent switch after no-ACK, in the case of RPL Lite without ETX squaring (higher hysteresis threshold than Classic).

simonduq:
Re-aiming for release. This fix will make bring a more conservative default RPL Lite config.

PR #504, opened 2018-05-09
RPL urgent probing fix

simonduq:
As discussed in https://github.com/contiki-ng/contiki-ng/issues/499, this prevents RPL from queuing several probing packets at any given time. In poor link scenarios, it would otherwise happen that an urgent probe was scheduled several times before even getting the link callback (executed only after reaching the maximum number of MAC tx attempts).


g-oikonomou:
This looks reasonable. Shall we merge?

PR #505, opened 2018-05-10
Fixed driver for DHT22 sensor in zoul platform 

ianno:
The DHT22 driver did not work.
The example in `examples/platform-specific/zoul/test-dht22.c` was not compiling because of wrong types.
However, also with fixed types the driver was not able to read the sensor properly. The output of the example above was just `Failed to read the sensor`.
The culprits were some parameters slightly off, and a misplaced pin reading. 
Now it is fixed.

(Tested with "ReMote rev b" and "Grove Temperature and Humidity Sensor Pro").

### DIFF
--- a/arch/platform/zoul/dev/dht22.c
+++ b/arch/platform/zoul/dev/dht22.c
@@ -72,7 +72,7 @@ dht22_read(void)
 {
   uint8_t i;
   uint8_t j = 0;
-  uint8_t last_state;
+  uint8_t last_state = 0xFF;
   uint8_t counter = 0;
   uint8_t checksum = 0;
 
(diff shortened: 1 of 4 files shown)


g-oikonomou:
Thanks @ianno 

The `Makefile` in this example is problematic, or travis would have picked this up.

```
CONTIKI_PROJECT += test-zonik test-dht22.c test-ac-dimmer.c
```

Those should be without the `.c` at the end. Can you fix and include in the pull? This will make sure everything here is tested correctly.

g-oikonomou:
Looks like the same applies for `test-servo`

ianno:
Sure. Just updated the Makefile.

g-oikonomou:
👍

Issue #220, opened 2017-12-03
Asynchronous RF operations


g-oikonomou:
Currently all our RF operations are synchronous.

Especially for TX, we block waiting for potentially a very long time. It would be nice to be able to return immediately at the start of the TX operation, with the radio notifying about the end of TX through a callback.

A similar technique could be applied to the `on()` operation, in the case of radios that have a very long startup time, e.g. due to a need to fire up a crystal.

arurke:
:+1: 
For what is worth, I remember @bkozak-scanimetrics implemented an async radio driver for cc2650 as part of https://github.com/contiki-os/contiki/pull/1549 . I never got around to testing it myself though.

yatch:
Interesting; this would need some adjustments on MAC implementations including TSCH, wouldn't it?

g-oikonomou:
Yep, MAC implementations as well as radio drivers to get those TX notifications.

yatch:
@g-oikonomou OK; thanks! 😄

Issue #101, opened 2017-10-17
Improved API for random number generation


joakimeriksson:
We need to add an API for filling a buffer with random values as this is often a type of functions used for various encryption and security mechanisms. 

Something like:

random_fill(uint8_t *buffer, size_t len);

joakimeriksson:
Implemented in CSPRNG.

PR #506, opened 2018-05-11
A Secured Version of Coordinated Sampled Listening (CSL)

kkrentz:
This pull request adds an implementation of the standardized MAC protocol "coordinated sampled listening (CSL)". This implementation has a couple of nice features, some of which go beyond what was standardized:
- Channel hopping à la MiCMAC
- Burst forwarding
- Compensation for clock drifts
- Strong freshness via wake-up counters
- On-the-fly rejection of unwanted frames
- Authentication and encryption using pairwise session keys
- Denial-of-sleep-resilient session key establishment

Other additions are:
- An asynchronous radio API, as suggested in #220 
- A cryptograhically-secure pseudo-random number generator
- Support for IEEE 802.15.4 short addresses

Some issues remain:
- Presently, this CSL implementation only runs on the CC2538 SoC
- Pairwise session keys are currently derived from a network-wide key

Feel free to try it out. I added a plain sender/receiver example in examples/akes/csl.


g-oikonomou:
So CSL in the title caught my attention here, but there's a lot more! 

I strongly feel this needs to be split up. I cannot see why e.g. the CSPRNG cannot be separate. Or the asynch radio API. Or the short addr support.

I had a quick look at the csprng work and if we are going to introduce it, I think it's time we started thinking about a proper RNG API too (#101)

g-oikonomou:
Or the Layer 2 pub-sub, the nbr table updates, the CC2538 lpm update, the rtimer updates, the updates to packetbuf, or the RADIO <-> MAC result conversion :)

For asynchronous mode, I see you have implemented a separate radio driver. Not opposed to that as such, but did you consider a possible scenario of a single driver that can operate in either mode? Too complicated (although mind you, in the mega-long term. it'd be nice to have asynch mode only...). I also see that `transmit` is not checking whether we are in TX mode already. Intentional? Guaranteed to never happen?

simonduq:
Lots of interesting stuff in there! Agree this needs slitting, one PR per feature. Looks like *many* potential PRs :)

Regarding CSL, is this a standard-compliant implem (+ optional features you mention) or more like a ContikiMAC v2? In other words, does it inter-operate with independent implementations?

Other general comments:
* For new features, we will ask for:
  * Doxygen annotations
  * CI tests
  * a Wiki documentation page
  * ideally, a Wiki tutorial
* The code is very low in comments. Please add a lot more comments, to make long-term maintenance possible.

kkrentz:
Okay, the next I will do is try to divide this into multiple PRs ...

@g-oikonomou Regarding the asynchronous radio API, it should be possible to integrate this functionality into the existing API. However, I am afraid that this would result in horribly complex radio drivers. Regarding the transmit function, such assertions are enabled while debugging, but, otherwise, I just assume that the MAC protocol knows what it does.

@simonduq IEEE 802.15.4-2015 specifies two versions of CSL - one where wake-up frames are transmitted as a continuous sequence and one with gaps in between. The version with gaps reminds me of X-MAC, while the other version shares ideas with BoXMAC-1. I opted to implement the version where wake-up frames are transmitted as a continuous sequence because that one leads to better denial-of-sleep resilience. That said, my implementation is not fully standard-compliant as far as security and channel hopping are concerned. In my opinion, the security part of CSL requires third-party additions, such as session key establishment, better replay protection, and denial-of-sleep defenses. The standardized channel hopping mechanism, on the other hand, ensues impractically long periodic wake ups.

simonduq:
Non-standard additions for things beyond the scope of the standard are absolutely fine. For instance, key exchange in 15.4, scheduling in TSCH, etc. These are done on top of the standard, without breaking compliance. You also mentioned drift compensation: this might not be described in the standard, but it is commonsense to have, and does not affect compliance -- it just emulates better clocks.

For the protocol itself, we will insist on having a standard mode. Extensions are fine if they provide significant advantage, but if we are to ship CSL with Contiki-NG, we need to be able to configure it for interoperability.

simonduq:
Adding WIP flag because of the ongoing split into sub-PRs

PR #507, opened 2018-05-11
Support for IEEE 802.15.4 short addresses

kkrentz:
These are some slight changes in order to support IEEE 802.15.4 short addresses.


arch/cpu/cc2538/ieee-addr.c
```diff
@@ -70,11 +69,16 @@ ieee_addr_cpy_to(uint8_t *dst, uint8_t len)
     if(((uint8_t *)IEEE_ADDR_LOCATION)[3] == oui_ti[0]
        && ((uint8_t *)IEEE_ADDR_LOCATION)[2] == oui_ti[1]
        && ((uint8_t *)IEEE_ADDR_LOCATION)[1] == oui_ti[2]) {
-      for(i = 0; i < len / 2; i++) {
-        dst[i] = ((uint8_t *)IEEE_ADDR_LOCATION)[len / 2 - 1 - i];
-      }
-      for(i = 0; i < len / 2; i++) {
-        dst[i + len / 2] = ((uint8_t *)IEEE_ADDR_LOCATION)[len - 1 - i];
+      if(len == 2) {
```
g-oikonomou: The original code block is meant to work for `len==2` as well, so the `if/else` should not be required. Have you spotted a bug here?

kkrentz:
There are some reasons for preferring short addresses over extended addresses:
- Short addresses reduce the length of IEEE 802.15.4 headers quite considerably
- 6LoWPAN provides better compression for interface identifiers that are derived from short addresses, although I do not know whether this is currently implemented in `sicslowpan.c`
- For some reason, CSL was only standardized for short addresses
- It simplifies the implementation of certain pairwise key predistribution schemes, such as Blom's scheme or the fully pairwise keys scheme.

os/net/linkaddr.h
```diff
@@ -110,6 +110,12 @@ int linkaddr_cmp(const linkaddr_t *addr1, const linkaddr_t *addr2);
  */
 void linkaddr_set_node_addr(linkaddr_t *addr);
 
+/**
+ * \brief      Derives an EUI-64 address from the given address
+ */
+void linkaddr_to_eui_64(uint8_t *dst, const linkaddr_t *addr);
+void linkaddr_from_eui_64(linkaddr_t *addr, const uint8_t *src);
```
simonduq: didn't look very closely, but shouldn't these function use (or be used by) `uip_ds6_set_addr_iid` and `uip_ds6_set_lladdr_from_iid`?

simonduq:
Short addresses for 6LoWPAN (I think it should already work in nullnet?) are wanted :)
What I would expect networks to do is boot on the long address and then obtain a unique address and switch to it at runtime. Not sure if starting directly from a short address is even allowed in 6LoWPAN?

os/net/linkaddr.h
kkrentz: Right. Anyway, the actual reason why I moved this conversion to `linkaddr.ch` has gone. I am now doing this conversion inside `uip-ds6.c`.

arch/cpu/cc2538/ieee-addr.c
simonduq: Wouldn't be nice to avoid the conditional, i.e., fix the for loops so they work on any address len?

kkrentz: I had just uploaded a version, where the loops are fixed ...

PR #512, opened 2018-05-12
Harmonize tools dir and the login, serialdump, serialview targets

g-oikonomou:
This pull:

* Removes platform-specific (sky, jn516x) serialdump variants
* Removes all serialdump pre-built binaries
* Provides a single, portable implementation of serialdump in `tools/serialdump.c`.
  * This implementation now uses `tools-utils.[ch]` for baudrate handling.
  * We now use a single filename for the serialdump binary, simply called `serialdump` irrespective of OS.
  * Handling of `O_SYNC` and `O_DIRECT` is tidied-up.
  * Code style is fixed.
* Changes the `Makefile` under `tools/` such that it can be used to build `tunslip6` as well as `serialdump`
* Enables the `login`, `serialdump`, `serialview` make targets for _all_ embedded platforms. This is achieved through a new `Makefile.embedded`, included by `Makefile.include` if we are building for anything other than `native` or `cooja`. Embedded target Makefiles now simply need to specify the baudrate and port.
* Harmonises the way make targets can specify that they require `serialdump` or `tunslip6`. This includes the `login` and friends targets, as well as example-specific targets (e.g. the border-router specific `connect-router` target). This is achieved by a new top-level `Makefile.tools`, which is included by `Makefile.include` unconditionally.

This is work in progress, for comment at the moment. We also need to add those things to the usage string.

Requires #510 #511


atiselsts:
Easier for new users to start using Contiki-NG is the main reason in my mind.

Not quite following the platform argument: Windows, Linux, Mac, aren't all versions working?

g-oikonomou:
I don't think those binaries will work on RPis and similar platforms with non-intel CPUs. I was also thinking 32 vs 64 bit etc.

atiselsts:
Ok, Raspberry PI and other non-x86 platforms is a fair point. Still, perhaps keep the windows version? Does it not work both for 32 and 64-bit Windows?

I like the fact that now `make login` builds the serialdump automatically on first use.
Still, I got an error because of not having `rlwrap` on my system - it would be nice if Contiki-NG was more helpful and suggested to install the rlwrap package at that point instead of just bailing out with an error.

Also, when trying to use the new serialdump I got `open: Invalid argument` error because of the new `O_DIRECT` flag.

g-oikonomou:
> Also, when trying to use the new serialdump I got open: Invalid argument error because of the new O_DIRECT flag.

This is actually very useful. I struggled to understand the logic behind that one, don't even know if we actually need it in the first place (even if available). I wonder if we should just get rid of it altogether.

Can you confirm that with current develop branch serialdump runs without  `O_DIRECT` in your system? It should be entering the 2nd if block here:

```
  fd = open(device, O_RDWR | O_NOCTTY | O_NDELAY | O_DIRECT | O_SYNC);
  /* Some systems do not support certain parameters (e.g. raspbian)
   * Just do some random testing. Not sure  whether there is a better way
   * of doing this. */
  if(fd < 0 && errno == EINVAL) {
    fd = open(device, O_RDWR | O_NOCTTY | O_NDELAY | O_SYNC);
  }
```

> Still, perhaps keep the windows version? Does it not work both for 32 and 64-bit Windows?

I've no idea. I have been assuming Cygwin here, not sure if I'm right in doing so.

> Still, I got an error because of not having rlwrap on my system - it would be nice if Contiki-NG was more helpful and suggested to install the rlwrap package at that point instead of just bailing out with an error.

Yep, agreed. Perhaps if rlwrap is unavailable, we can warn and run without it instead of failing?

Something like the last commit that I just pushed. Naturally will need tidied-up a bit because currently the message appears even when we run `make clean`, but what do you think about the approach?

g-oikonomou:
I've made the warning message conditional in the last commit, now only appears if any of `login`, `serialdump` or `serialview` are among the make goals.

(This is a rebase and force-push to also fix some trailing whitespaces)

simonduq:
I've been frustrated with this for a while, great to see this finally tackled :) Also in favor or removing all binaries. Also in windows. If there is demand, we can host the exe on github.io and put the link on the wiki. 

The old motivation I remember for having binaries was to enable running Contiki on systems that do not even have gcc. But really don't think this is an issue anymore. Every RaspberryPI distros are shipped with gcc nowadays.

g-oikonomou:
@simonduq do you know if by "windows" we mean cygwin here? @atiselsts Atis and I discussed earlier, and neither of us is certain... :)

simonduq:
I don't know. I believe that `wcapslip` can do both Cygwin and plain Windows, but no clue about `serialdump`. One more reason to ditch it. We'll find out the day somebody actually needs it!

g-oikonomou:
Note to self: Get rid of `timestamp` too and switch to using -T

g-oikonomou:
I have removed the `tools/timestamp` script. The `serialview` and `serialdump` make targets now use `serialdump -T` instead.

fontikos:
Test results on cygwin.

```
$ make
cc -Wall -Werror -O2 tunslip6.c tools-utils.c -o tunslip6
cc -Wall -Werror -O2 serialdump.c tools-utils.c -o serialdump
```

```
$ make TARGET=cc2538dk PORT=/dev/ttyS47 serialview
../../tools/serialdump -b115200 -T /dev/ttyS47
connecting to /dev/ttyS47 [OK]
[2018-05-15 18:01:02] [ext-flash] Flash test successful.
[2018-05-15 18:01:02] [INFO: Main      ] Starting Contiki-NG-develop/v4.0-305-g5570f81e9-dirty
[2018-05-15 18:01:02] [INFO: Main      ]  Net: sicslowpan
[2018-05-15 18:01:02] [INFO: Main      ]  MAC: CSMA
[2018-05-15 18:01:02] [INFO: Main      ] Link-layer address 0012.4b00.0e07.d5a5
[2018-05-15 18:01:02] [INFO: Main      ] Tentative link-local IPv6 address fe80::212:4b00:e07:d5a5
[2018-05-15 18:01:02] [INFO: CC26xx/CC13xx] TI CC1310 LaunchPad
[2018-05-15 18:01:02] [INFO: CC26xx/CC13xx]  RF: Channel 25, PANID 0x4B1200
[2018-05-15 18:01:02] [INFO: CC26xx/CC13xx]  Node ID: 54693
[2018-05-15 18:01:02] CC26XX demo
[2018-05-15 18:01:02] -----------------------------------------
[2018-05-15 18:01:02] Bat: Temp=22 C
[2018-05-15 18:01:02] Bat: Volt=3320 mV
```

g-oikonomou:
So we just checked this with @fontikos . That was with `O_DIRECT`, funnily enough, which also seems to work fine on Mac OS. Reportedly though. linux didn't like it...

g-oikonomou:
Add a commit that makes the timestamp format easily configurable:

```
$ make serialview
rlwrap ../../tools/serialdump -b115200 -T 
[...]
$ make SERIALDUMP_TIME_FMT=%s serialview
rlwrap ../../tools/serialdump -b115200 -T%s 
```

g-oikonomou:
Hi @atiselsts if you get a moment can you check whether it's `O_SYNC` or `O_DIRECT` causing you problems and also basically if the thing works properly without the troublesome flag?

atiselsts:
`O_DIRECT` is causing the invalid argument problem. Otherwise it seems to work. Yes, the develop branch works by entering the `if(fd < 0 && errno == EINVAL){` branch.

nfi:
`O_DIRECT` should not be needed and I think we can simply remove it. I have been running without it for a long time.

g-oikonomou:
I have done some testing on OSX without `O_DIRECT` and I'm not seeing any issues either, so I've removed it.

We can now pass `SERIALDUMP_TIME_FMT="format"` make argument. Better dependency handling means that if sources change, `make login` will automatically trigger a rebuild of `serialdump` (same for `tunslip6`).

This is ready for review from my side, so Im removing the WiP label.

atiselsts:
As we discussed offline, might be nice to keep at least the serialdump executable for Windows to simplify getting started with Contiki.

g-oikonomou:
Would be nice if someone could check whether the one we have is native Win vs Cygin

atiselsts:
It's Cygwin.

g-oikonomou:
OK so perhaps we can keep that one there, yes. Perhaps we can even provide a new binary based on the latest sources. We may have to rename it to just serialdump to align with the rest and then exclude it from the build system? Or shall I leave it in the build system and then users who want to do so can re-build? This could cause git to flag it as changed though.

Also, any idea if it's 32 v 64 bit?

The _one_ thing I am definitely not interested in doing is committing a new binary every time the source changes.

simonduq:
I'd remove all binaries and provide the useful ones via a link on the wiki.

atiselsts:
Let's keep the windows one (after rebuilding and renaming). Building the .exe from source is not as trivial as on linux, as the user needs to install Cygwin first. (To use the .exe, the user does not need to install anything.) Downloading via a link is still an extra step, the user need to find the link in the wiki first, and sometimes links become broken with time.

Repository size is not an issue, since the .exe itself is very small; cygwin1.dll is 1MB, but it's already there and will not change in reasonable future. I haven't heard any 32-bit vs. 64-bit complains, so imagine it's not an issue (don't have 32-bit system to test).

g-oikonomou:
Could I ask someone to do me a favor and try to run outside cygwin a win executable built from current source?

atiselsts:
Do we have plans to overwrite motelist for all platforms (including Windows) in Python?
If not, then we should absolutely add back `cygwin1.dll` to make the motelist work regardless of what we decide about serialdump.

g-oikonomou:
I've started fiddling with a pythonic motelist, but nowhere near there yet. Are you saying motelist is broken on win?

g-oikonomou:
I've thought a bit more about this:

In terms of motelist: We should harmonize those too in the near future, so I am not that bothered about the current windows executable. And it's not like people have been complaining all over the place that it's not working...

Distributing an executable + the dll might allow people to run those tools natively on windows, but I don't think anything else would work, plus this is not a setup that we want to maintain, right?

IMHO our approach for the Windows user would be to either use Docker (assuming that's possible) or to install Cygwin + make + toolchains (perhaps we should add a wiki page here). In both scenarios, I cannot see a reason to distribute neither a binary nor the `cygwin1.dll`, since building the tool would happen nicely from source.

IMHO we should stick with this pull as-is.

simonduq:
Works for me. I see that you have changes regarding `MOTE` and `PORT`. Does this affect usage in any way?

g-oikonomou:
The intention is for `PORT` to work across the board and for `MOTE` to map to `PORT` for platforms that supported the latter (not all did).

simonduq:
Sounds good. Making the transition to `PORT` smooth :)

simonduq:
With the new Makefiles at the top-level, I feel it would be a good idea to add a few words at the top of each to summarize the purpose of each file.

g-oikonomou:
I've added some 3-4 lines of comments at the top of each top-level Makefile.

PR #513, opened 2018-05-12
rpl-udp example: remove uipbuf tuning of MAC retries

simonduq:
I believe the `uipbuf_set_attr` example did cause confusion, by sending some unreliable packets. When not ACKed at MAC, can have bad consequences on link estimates, as ETX will be updated with `ETX_NOACK_PENALTY`. Also, the example is intended as a minimal RPL UDP example and seems to be the starting point to many users, so I think this is not the ideal place to document `uipbuf_set_attr`.

### DIFF
--- a/examples/libs/ipv6-uipbuf/Makefile
+++ b/examples/libs/ipv6-uipbuf/Makefile
@@ -1,0 +1,5 @@
+CONTIKI_PROJECT = udp-client udp-server
+all: $(CONTIKI_PROJECT)
+
+CONTIKI=../../..
+include $(CONTIKI)/Makefile.include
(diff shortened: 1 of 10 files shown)


atiselsts:
How about moving this API demonstration to another example this, then?

I see the problem with `ETX_NOACK_PENALTY`, but to be honest, it sounds like a bug, an unforeseen consequence we should have noticed when the transmission attribute was introduced.
It would be more logical to use `transmissions + X` as the transmission count for ETX when no ack is received instead of a fixed `ETX_NOACK_PENALTY`.

simonduq:
Agree with everything. Sorry for a PR that was not very constructive ;)
BTW what would be nice for such API is to have it documented on the wiki as well.

simonduq:
Done :)

atiselsts:
OK, I see, since `MAX_FRAME_RETRIES` is 7 both for TSCH and CSMA, the new value `12`  for `ETX_NOACK_PENALTY` keeps the old behavior exactly.
Nice to have one more example! I checked that it builds on all platforms.

simonduq:
Right, and I should add a text for at least one platform!

simonduq:
Test added (on Zoul)

Issue #514, opened 2018-05-13
./../os/net/ipv6/uip-ds6.c:94:22: error: ‘iid_prefix’ defined but not used


cuongdodinh:
Dear All,

I run in Raspi, How can I fix this error?

make TARGET=native all
  CC        ../../os/net/ipv6/uip-ds6.c
../../os/net/ipv6/uip-ds6.c:94:22: error: ‘iid_prefix’ defined but not used [-Werror=unused-const-variable=]
 static const uint8_t iid_prefix[] = { 0x00, 0x00 , 0x00 , 0xff , 0xfe , 0x00 };
                      ^~~~~~~~~~
cc1: all warnings being treated as errors
../../Makefile.include:317: recipe for target 'obj_native/uip-ds6.o' failed
make: *** [obj_native/uip-ds6.o] Error 1

@chenek  Would you like help me?

Thanks & B.R,

chenek:
@cuongdodinh I suppose you can remove claim iid_prefix in uip-ds6.c to fix the issue.

simonduq:
What concerns me here is that this wasn't caught by the CI suite. Which compiler are you on? It looks like almost all compiled (all in our test suite) do not issue unused variable warnings on `const`. Ideally we should first reproduce in Docker, add a test and then fix.

cuongdodinh:
Dear Simonduq,
I use Raspi 3. Downloaded Contili-NG on it. Then compiler in Raspi 3.

Thank you.

simonduq:
What's your compiler version?

cuongdodinh:
Dear SemonDuq,

Raspbian 6.3.0.0 -18+raspi1 + deb9u1 (20170516)

simonduq:
Please provide the full output, e.g.
```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/4.6/lto-wrapper
Target: arm-linux-gnueabihf
[...]

PR #515, opened 2018-05-14
Fix to uip-ds6.c

kkrentz:
Fixes a compilation warning ...

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -49,6 +49,7 @@ CFLAGS += -DCONTIKI_BOARD_$(TARGET_BOARD_UPPERCASE)=1
 CFLAGS += -DCONTIKI_BOARD_STRING=\"$(BOARD)\"
 endif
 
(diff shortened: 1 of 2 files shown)


simonduq:
I'm not comfortable with fixing this without ensuring non-regression. We need to either be able to trigger the warning in CI, or make sure the warning is always disabled.

From https://patchwork.ozlabs.org/patch/517212/ it appears that gcc-6 issues warnings on unused consts, something earlier versions didn't do. I couldn't find a way to add the warning as well on gcc-4, which we use in Docker. My suggestion for now would be to add `-Wno-unused-const-variable` in native builds, to at least make sure users don't hit this problem again.

This PR closes #514

simonduq:
Added https://github.com/contiki-ng/contiki-ng/issues/517 to make sure we don't leave this `-Wno` in there forever. Thanks!

Issue #518, opened 2018-05-14
Native fails to build on OS X


g-oikonomou:
#432 broke native builds on OSX: clang does not like `--fatal-warnings`

The corresponding flag on OSX can be `-fatal_warnings` (one dash, and an underscore).

g-oikonomou:
Fixed in #519

PR #519, opened 2018-05-14
Added a configurable linker flag to enable linker warnings as errors

nfi:
OSX requires a different linker flag to enable linker warnings as errors. The native platform will now set the correct linker flag when compiling under OSX. Also made it possible to disable this feature and allow linker warnings by building with `WERROR` set to 0.


g-oikonomou:
Fixes #518

PR #522, opened 2018-05-15
Cryptographically-Secure Pseudo-Random Number Generator (CSPRNG)

kkrentz:
This PR adds a CSPRNG, which, e.g., will come in handy when generating cryptographic keys or hash chains. For generating cryptographic random numbers, it uses AES-128 in OFB mode. For generating an initial seed, it uses an exchangeable `csprng_seeder`.  This PR also includes implementations of `csprng_seeder`s for CC2538-based platforms, such as the `cc2538_sram_seeder`, which extracts seeds from power-up SRAM states. For the time being, this CSPRNG does not provide forward and backward security, i.e., once its internal state leaks to an attacker, he can predict past and future outputs of the CSPRNG. This could be changed by calling the configured `csprng_seeder` not only at startup, but also at runtime once in a while.

### DIFF
--- a/arch/cpu/cc2538/dev/cc2538-rf.c
+++ b/arch/cpu/cc2538/dev/cc2538-rf.c
@@ -49,6 +49,7 @@
 #include "dev/sys-ctrl.h"
 #include "dev/udma.h"
 #include "reg.h"
+#include "lib/iq-seeder.h"
 
 #include <string.h>
 /*---------------------------------------------------------------------------*/
@@ -253,6 +254,42 @@ get_rssi(void)
   }
 
   return rssi;
+}
+/*---------------------------------------------------------------------------*/
+/**
+ * \brief       Reads the current I/Q data of the received signal.
+ * \param value The least significant bit (LSB) of the I coordinate and the LSB
+ *              of the Q coordinate are concatenated and stored here.
+ *
+ * If not done already, this function first enables the RX mode and waits for
+ * the RSSI_VALID bit to go high. Hence, this function should only be called
+ * at start up or by the MAC protocol to avoid conflicts.
+ */
+static void
+get_iq_lsbs(radio_value_t *value)
+{
+  uint8_t was_off = 0;
+
+  /* If we are off, turn on first */
+  if((REG(RFCORE_XREG_FSMSTAT0) & RFCORE_XREG_FSMSTAT0_FSM_FFCTRL_STATE) == 0) {
+    was_off = 1;
+    on();
+  }
+
+  /* Wait on RSSI_VALID */
+  while((REG(RFCORE_XREG_RSSISTAT) & RFCORE_XREG_RSSISTAT_RSSI_VALID) == 0);
+
+  /* Wait until the channel seems clear for better randomness */
+  while(!(REG(RFCORE_XREG_FSMSTAT1) & RFCORE_XREG_FSMSTAT1_CCA));
+
+  /* Read I/Q LSBs */
+  *value = REG(RFCORE_XREG_RFRND)
+      & (RFCORE_XREG_RFRND_IRND | RFCORE_XREG_RFRND_QRND);
+
+  /* If we were off, turn back off */
+  if(was_off) {
+    off();
+  }
 }
 /*---------------------------------------------------------------------------*/
 /* Returns the current CCA threshold in dBm */
@@ -595,6 +632,10 @@ init(void)
 
   set_poll_mode(poll_mode);
 
+#if CSPRNG_ENABLED
+  iq_seeder_seed();
+#endif /* CSPRNG_ENABLED */
+
   process_start(&cc2538_rf_process, NULL);
 
   rf_flags |= RF_ON;
@@ -913,6 +954,9 @@ get_value(radio_param_t param, radio_value_t *value)
     return RADIO_RESULT_OK;
   case RADIO_PARAM_LAST_LINK_QUALITY:
     *value = crc_corr & LQI_BIT_MASK;
+    return RADIO_RESULT_OK;
+  case RADIO_PARAM_IQ_LSBS:
+    get_iq_lsbs(value);
     return RADIO_RESULT_OK;
   case RADIO_CONST_CHANNEL_MIN:
     *value = CC2538_RF_CHANNEL_MIN;
--- a/arch/cpu/cc2538/dev/cc2538-sram-seeder.c
+++ b/arch/cpu/cc2538/dev/cc2538-sram-seeder.c
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2015, Hasso-Plattner-Institut.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * This file is part of the Contiki operating system.
+ *
+ */
+
+/**
+ * \addtogroup csprng
+ * @{
+ *
+ * \file
+ *         SRAM-based CSPRNG seeder.
+ * \author
+ *         Konrad Krentz <konrad.krentz@gmail.com>
+ */
+
+#include "dev/cc2538-sram-seeder.h"
+#include "sys/rtimer.h"
+#include "lpm.h"
+#include <string.h>
+
+#define SRAM_BYTES 1024
+
+/* in the middle of the second half of the SRAM */
+static const uint8_t *sram_pointer = (uint8_t *)0x20001FFF;
+
+/*---------------------------------------------------------------------------*/
+static void
+reset_sram(void)
+{
+  /*
+   * Although it should not, the system gets stuck when sleeping less than
+   * two seconds at this point. Sleeping for two seconds seems to be safe.
+   */
+  rtimer_arch_schedule(RTIMER_NOW() + RTIMER_ARCH_SECOND * 2);
+  lpm_enter();
+}
+/*---------------------------------------------------------------------------*/
+void
+cc2538_sram_seeder_seed(void)
+{
+  struct csprng_seed seed;
+  uint16_t byte_pos;
+  uint8_t bit_pos;
+  uint8_t sram_snapshot[SRAM_BYTES];
+  uint16_t i;
+  uint8_t j;
+  int bit1;
+  int bit2;
+
+  byte_pos = 0;
+  bit_pos = 0;
+
+  while(1) {
+    reset_sram();
+    memcpy(sram_snapshot, sram_pointer, SRAM_BYTES);
+    reset_sram();
+
+    /* von Neumann extractor */
+    for(i = 0; i < SRAM_BYTES; i++) {
+      for(j = 0; j < 8; j++) {
+        bit1 = (1 << j) & sram_snapshot[i];
+        bit2 = (1 << j) & sram_pointer[i];
+        if(bit1 < bit2) {
+          seed.u8[byte_pos] |= 1 << bit_pos++;
+        } else if(bit1 > bit2) {
+          seed.u8[byte_pos] &= ~(1 << bit_pos++);
+        }
+        if(bit_pos == 8) {
+          bit_pos = 0;
+          byte_pos++;
+        }
+        if(byte_pos == CSPRNG_SEED_LEN) {
+          csprng_feed(&seed);
+          return;
+        }
+      }
+    }
+  }
+}
+/*---------------------------------------------------------------------------*/
+
+/** @} */
--- a/arch/platform/cooja/dev/moteid.c
+++ b/arch/platform/cooja/dev/moteid.c
@@ -31,6 +31,8 @@
 #include "dev/moteid.h"
 #include "lib/simEnvChange.h"
 #include "lib/random.h"
+#include "lib/csprng.h"
+#include <string.h>
 
 const struct simInterface moteid_interface;
 
@@ -44,8 +46,14 @@ static void
 doInterfaceActionsBeforeTick(void)
 {
   if (simMoteIDChanged) {
+    struct csprng_seed csprng_seed = { 0 };
+
     simMoteIDChanged = 0;
-	random_init(simRandomSeed);
+    random_init(simRandomSeed);
+    memcpy(csprng_seed.u8,
+        &simRandomSeed,
+        MIN(sizeof(simRandomSeed), sizeof(csprng_seed.u8)));
+    csprng_feed(&csprng_seed);
   }
 }
 /*-----------------------------------------------------------------------------------*/
--- a/os/lib/csprng.c
+++ b/os/lib/csprng.c
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2013, Hasso-Plattner-Institut.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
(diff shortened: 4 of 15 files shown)


simonduq:
Interesting, and nicely executed.
Comments:
*  Maybe `csprng_init` should move to `main` instead of platform init functions
* Do we need an example with a compile test? All the files should already be compiled (not linked), right?

arch/cpu/cc2538/lpm.h
```diff
@@ -139,6 +139,11 @@ void lpm_init(void);
  */
 void lpm_enter(void);
 
+/**
+ * Ensure that an interrupt will wake up the device again.
+ */
+void lpm_enter_pm2(void);
```
g-oikonomou: This is something I have very strong reservations about.

You will see that the lpm module offers an `lpm_enter()` function, which will try to enter PM2 if i) it is enabled by the configuration and ii) if the system is overall in a good state to enter this PM. `lpm_enter()` makes sure to drop to a low-power mode gracefully, respecting the state of various on-chip peripherals as well as the state of the OS' scheduler at the point of dropping to low-power. As a very minor point, it also correctly handles the state of the ENERGEST module.

What this pull proposes is a new function that can be used to brute-forcefully enter PM2, merely for the sakes of resetting the state of the non-retention area of the CC2538's SRAM so a seed can be generated. There does not seem to be any consideration as to what other tasks may have been scheduled to happen, how long after the call to generate the seed. Call this function while the native USB driver is in use and you will immediately cause a USB device disconnect from the host (`lpm_enter()` will pick up the fact that the USB is active and it will not disrupt it).

What you want to be doing here is collect randomness from the state of the SRAM non-retention area opportunistically. The design I would recommend would be to add hooks to `lpm_enter()` and `lpm_exit()`. Upon entering a low power mode, check if it's gonna be PM2. If so, then upon wakeup inspect the state of the SRAM and gather randomness. Latch it and use it when required. Fail gracefully if sufficient randomness has not been generated.

Have a look at the newer and more advanced LPM implementation of the CC13x0/CC26x0. See how the LPM module notifies all interested parties before transition to a low-power state / upon wakeup. A similar approach would be recommended here.

arch/platform/cc2538dk/platform.c
```diff
@@ -154,6 +155,9 @@ platform_init_stage_two()
   crypto_init();
   crypto_disable();
 #endif
+#if CSPRNG_ENABLED
+  csprng_init();
```
g-oikonomou: Agree with @simonduq This should go to `main()`.  `cc2538-conf.h` should then `#define CSPRNG_CONF_ENABLED 1`.

g-oikonomou:
These are overall very desirable additions, but I do have concerns about the design and implementation of things - especially on the CC2538-specific parts - as well as about the fact that API documentation is somewhat on the thin side.

In other news, perhaps this could be an opportunity to think about modernised APIs for random number generation (#101) as well as for seeding. Perhaps (not as part of this particular pull), we could end up with drivers for random generators suitable for crypto and random generators for other stuff (e.g. backoffs).

os/lib/iq-seeder.c
```diff
@@ -0,0 +1,161 @@
...
+
+    /* append I/Q LSBs to accumulator */
+    accumulator[byte_pos] |= iq << bit_pos;
+    bit_pos += 2;
+    if(bit_pos == 8) {
+      bit_pos = 0;
+      byte_pos++;
+    }
+  }
+  NETSTACK_RADIO.off();
```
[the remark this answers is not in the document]
kkrentz: If I got you right, you are concerned that this may interfere with the MAC protocol. However, currently, seeding just happens once at start up before the network stack is initialized. For implementing reseeding at runtime, my plan would be to add a function to `csprng.c`, which gets called if a new seed is ready to be mixed with the current seed. In this way, the MAC protocol can, e.g., collect I/Q data in the background and notify the CSPRNG without any conflicts.

os/lib/iq-seeder.c
```diff
@@ -0,0 +1,161 @@
...
+  bit_pos = 0;
+  byte_pos = 0;
+  memset(accumulator, 0, COLUMN_COUNT);
+
+  NETSTACK_RADIO.on();
+  for(iq_count = 0; iq_count <= (COLUMN_COUNT * 8 / 2); iq_count++) {
+    NETSTACK_RADIO.get_value(RADIO_PARAM_IQ_LSBS, &iq);
+
+    if(!iq_count) {
+      /* drop first I/Q reading because it is always 0 for some reason */
```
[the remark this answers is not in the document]
kkrentz: I had just tried to reproduce this issue, but it seems to gone.

arch/cpu/cc2538/dev/cc2538-mix-seeder.c
```diff
@@ -0,0 +1,77 @@
...
+ *         Mixing CSPRNG seeder.
+ * \author
+ *         Konrad Krentz <konrad.krentz@gmail.com>
+ */
+
+#include "cc2538-mix-seeder.h"
+#include "cc2538-sram-seeder.h"
+#include "lib/iq-seeder.h"
+
+struct csprng_seed last_seed;
```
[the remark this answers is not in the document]
kkrentz: yes

arch/cpu/cc2538/dev/cc2538-mix-seeder.c
```diff
@@ -0,0 +1,77 @@
...
+    ((uint8_t *)result)[i] ^= ((uint8_t *)temp_result)[i];
+  }
+}
+/*---------------------------------------------------------------------------*/
+static void
+generate_seed(struct csprng_seed *result)
+{
+  struct csprng_seed temp_result;
+
+#if LPM_CONF_MAX_PM >= 2
```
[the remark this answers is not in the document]
kkrentz: I generalized the `mix-seeder` so that other platform can reuse it. Now, the platform configures the `mix-seeder` to use as many entropy sources as possible.

arch/cpu/cc2538/dev/cc2538-rf.c
```diff
@@ -257,6 +257,32 @@ get_rssi(void)
   return rssi;
 }
 /*---------------------------------------------------------------------------*/
+static radio_value_t
+get_iq_lsbs(void)
```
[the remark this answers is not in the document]
kkrentz: I looked at the user guide once again, but, in my version (swru319c.pdf), it is only written that the radio should be in RX mode and that the RSSI valid bit should be high. This used to work flawlessly.

arch/cpu/cc2538/dev/cc2538-sram-seeder.c
```diff
@@ -0,0 +1,111 @@
...
+ *         Konrad Krentz <konrad.krentz@gmail.com>
+ */
+
+#include "cc2538-sram-seeder.h"
+#include "sys/rtimer.h"
+#include "lpm.h"
+#include <string.h>
+
+#define SRAM_BYTES 1024
+#ifdef LPM_CONF_ENABLE
```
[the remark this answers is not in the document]
kkrentz: Moved this

arch/cpu/cc2538/lpm.h
kkrentz: I am now reusing `lpm_enter`, but could not avoid some suboptimal adaptations of `lpm.c`.

kkrentz:
Thanks for your feedback. I added several doxygen annotations and two compile tests.

arch/platform/cc2538dk/platform.c
kkrentz: This went to contikimac-main.c. I have not enabled CSPRNG_CONF_ENABLED by default because it may not be needed, but I have no objection to enabling it by default.

arch/cpu/cc2538/dev/cc2538-rf.c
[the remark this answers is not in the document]
g-oikonomou: RSSI valid being high: Absolutely. Let me look at the user guide again, cause it was a much earlier version when this driver was first written... :)

os/lib/iq-seeder.c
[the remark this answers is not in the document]
g-oikonomou: I'm with you now, yep. So this happens only once as you say, and that's before the stack gets initialised. All good.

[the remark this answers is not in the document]
simonduq: Maybe a comment in the code explaining that would be a good, for users to understand the limitations.

arch/cpu/cc2538/lpm.h
kkrentz: UPDATE: I am now getting away without modifications to `lpm.c` ...

os/lib/iq-seeder.c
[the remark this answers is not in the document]
kkrentz: I am now mentioning this limitation in the doxygen description of `iq_seeder_seed`

kkrentz:
UPDATE: Seeders now notify the CSPRNG of new seeds and the CSPRNG itself takes care of mixing new seeds with the current one. This way, we can do seeding at more appropriate places,e.g., reading power-up SRAM states earlier at start up and sampling I/Q data after the radio was initialized. Also, this approach is suitable for doing reseeding at runtime - simply call `csprng_feed` as a new seed becomes available.

kkrentz:
Rebased and removed compilation tests because if we enable the CSPRNG by default, the current tests compile the CSPRNG already. Let me know you require further changes.

os/contiki-main.c
```diff
@@ -90,6 +91,7 @@ main(void)
 
   netstack_init();
   node_id_init();
+  null_seeder_seed();
```
simonduq: What's the rationale for calling this even when there is a real seeder?

kkrentz:
* Added a usage example
* Disabled the seeding of the CSPRNG by default
* As far as I know, DTLS does not depend on cryptographic random numbers. That said, RFC 7925 suggests using "quality random numbers" and explicitly refers to RFC 4086, which I follow here. Thus, for compliance, it makes sense to change the submodule tinydtls to use this CSPRNG as a next step.
* Regarding the `null_seeder`, enabling and disabling would necessitate further configuration hassle ...

simonduq:
One last little thing on my side: as you add a new example, please a compile test for it on a least one example. We'll need @g-oikonomou 's approval here.

kkrentz:
According to the paper "Cryptographic Randomness on a CC2538: a Case Study", DTLS does require cryptographic random numbers unless a pre-shared key is used. However, I think I can not change the tinyDTLS to use my CSPRNG inside this PR because tinyDTLS is located in a different repository. Another interesting message of this paper is that I/Q readings may be biased under strong interference. As a remedy, I am now waiting for the channel to become clear when getting I/Q readings.

g-oikonomou:
Hi @kkrentz we just discussed this internally again. Just wondering whether you are still active/interested here? Should we revisit this? Do you have any time to do any further work if required?

kkrentz:
I can take care of further adjustments.

g-oikonomou:
That's great. Thanks. I'll have a good look at it again

arch/cpu/cc2538/dev/cc2538-sram-seeder.c
```diff
@@ -0,0 +1,110 @@
...
+
+    /* von Neumann extractor */
+    for(i = 0; i < SRAM_BYTES; i++) {
+      for(j = 0; j < 8; j++) {
+        bit1 = (1 << j) & sram_snapshot[i];
+        bit2 = (1 << j) & sram_pointer[i];
+        if(bit1 < bit2) {
+          seed.u8[byte_pos] |= 1 << bit_pos++;
+        } else if(bit1 > bit2) {
+          seed.u8[byte_pos] |= 0 << bit_pos++;
```
PeterKietzmann, line 94: If (for whatever reason) `seed.u8[byte_pos]` is not initialized properly and is not zero at that position, you will eventually not set  a `0`. How about  `&= !(1 << bit_pos++)`, or did I misunderstand the code?

kkrentz: As `seed` is zeroed when entering this function, I see no problem. However, I should replace `seed.u8[byte_pos] |= 0 << bit_pos++;` with `bit_pos++;` to avoid confusion.

kkrentz: I took on your suggestion.

kkrentz:
For our LifeSec project, I have ported uECC and libcoap to Contiki-NG, as well as improved the ECC implementation for CC2538 motes. This stuff all depends on a CSPRNG. Thus, if there was interest for those contributions, it would be necessary to merge this PR first ...

arurke:
@kkrentz I have plans to look at this some time after the summer, and hopefully we can get it merged. Libcoap seems like a

arurke:
> For our LifeSec project, I have ported uECC and libcoap to Contiki-NG, as well as improved the ECC implementation for CC2538 motes. This stuff all depends on a CSPRNG. Thus, if there was interest for those contributions, it would be necessary to merge this PR first ...

@kkrentz I have plans to look at this some time during or after the summer. Libcoap seems very promising imo., you have any idea on the code size and memory usage? Curious if it plays well with the class of devices Contiki-NG targets.

kkrentz:
@arurke I do not have exact numbers at hand. Libcoap runs on CC2538-based OpenMotes without complaints. My port makes use of the `heapmem` module, which I configured to use 4KB of RAM. Probably, you can go even lower if you do not need OSCORE.

arch/platform/cooja/dev/moteid.c
```diff
@@ -43,9 +45,13 @@ int simRandomSeed;
 static void
 doInterfaceActionsBeforeTick(void)
 {
+  struct csprng_seed csprng_seed;
```
pjonsson, line 48: Contiki-NG is compiled as C99 nowadays, so you can avoid the memset in line 53 by doing `struct csprng_seed csprng_seed = {0};`. Also push the variable declaration inside the if-statement since it isn't used anywhere else in this function.

arch/platform/cooja/dev/moteid.c
```diff
@@ -43,9 +45,13 @@ int simRandomSeed;
...
+
   if (simMoteIDChanged) {
     simMoteIDChanged = 0;
-	random_init(simRandomSeed);
+	  random_init(simRandomSeed);
```
pjonsson, line 52: Indentation level for the new code in this block looks wrong.

arurke:
I am not familiar with the cc2538 platform and had trouble reviewing (or testing) this PR with confidence. I was thus thinking to do a cc13xx/cc26xx seeder, but after looking at the task it seems to be overly coupled with csprng (platform developers need to familiarize themselves with csprng). I wonder if we can make the seeders generic, reduce the coupling between the CSPRNG and seeders, and in the process make supporting new platforms much clearer and simpler. Thinking aloud:

@kkrentz Could we consider the iq-seeder and sram-seeder as simply True RNGs? If so, we could make a common API for TRNGs (e.g. `trng-hal.h` with a `init()` and `read()`). I could easily implement it for cc13xx/cc26xx using the HW TRNG, and you have two for cc2538 in this PR (so we would need to consider how to handle, e.g. if one is preferred over the other). This will allow the init/feed of the csprng to be platform-generic and can be done in contiki-main.c using the TRNG API. It would thus be easy to support csprng on new platforms as they would only need to implement the TRNG API (solving my initial problem). Lastly it will make true random numbers available to other modules than the csprng (such as seeding `random.h`). Thoughts?

This is somewhat related to #101, but can probably be handled separately. [RIOT-OS HWRNG HAL](https://doc.riot-os.org/group__drivers__periph__hwrng.html) for reference (but does not guarantee true random).

kkrentz:
Currently, my seeders only produce a certain number of true random bits. They are not for general purpose. However, for seeding the normal RNG, you can use csprng_rand. That would be more resource-efficient.

Seeding at start up is insufficient for forward and backward security. Furthermore, there should be multiple seeders as a fallback solution if one is compromised.

arurke:
>Currently, my seeders only produce 256 true random bits. They are not for general purpose use.

I don't necessary see this as an issue. I would imagine a TRNG API to be described as something like "Expensive, with no guarantee on amount or at which time random data can be provided. Use with caution as security-related modules in the OS such as csprng may rely on it.".

> Seeding at start up would be insufficient to achieve forward and backward security. After all, some "seeders" may have to collect entropy and cannot run directly at start up.

My example with start-up was inspired by this PR which only does seeding at boot(?). There is of course no issue with asking the TRNG API for data at a later stage (if as you say entropy must be collected), or ask for more data to re-seed. I would imagine a simple csprng-seed-process could fill such a task (which might provide another argument for a platform-agnostic TRNG API as it would allow the process to also be made platform-agnostic and simply call the TRNG API (as opposed to being aware of cc2538-sram-seeder, iq-seeder, etc.)). But again, if I understand correctly this is something that is not yet solved/implemented for either this PR or a TRNG API approach.

> Ideally, there should be multiple seeders to have a fallback if one is compromised.

I don't see any issue with a TRNG API implementation using multiple sources. For example for cc2538 it could utilize both approaches in this PR (radio and sram), and have a strategy on how it uses these to supply callers with random data.

Please let me know if I am completely barking at the wrong tree here. Or if I am not conveying myself clearly, I can make a draft PR with some pseudo-implementation to illustrate my points.

kkrentz:
I see two problems with your idea of a "seeding process". First, if I, e.g., seed with I/Q LSBs at runtime, this must be coordinated with the MAC driver. Likewise, the SRAM seeder can only collect entropy at specific times, namely when waking up from LPM 2. Second, that seeding process would need to know which "seeders" exist. By contrast, my current design defaults to use all available seeders without any configuration effort.

A more general concern with your suggested TRNG API is that I see no use case for it. I have been happily using the CSPRNG in various protocols and never encountered the need for using true random numbers instead of `csprng_rand`. That is, generalizing the seeders to produce something else than CSPRNG seeds leads to unnecessary complexity in my opinion.

arurke:
Sorry for the late follow-up. I believe we understand each other viewpoints, and although I would still love to see some more abstraction I don't think we should stall the PR due to this - I can always propose some refactoring in a future PR. I also note that George and Simon looked through this without such concerns.

As soon as I find the time I will review the code, and hopefully we can finally get this merged :)

os/lib/csprng.c
```diff
@@ -0,0 +1,110 @@
...
+
+static struct csprng_seed seed;
+static unsigned read_state_bytes;
+static int seeded;
+
+/*---------------------------------------------------------------------------*/
+void
+csprng_feed(struct csprng_seed *new_seed)
+{
+  uint8_t i;
```
arurke, line 63: Just a heads up: We have moved from c89, so you are free to define the variable where you initialize. I personally prefer that style, but I don't believe we enforce either way, so feel free to make changes or leave as is.

os/lib/csprng.c
```diff
@@ -0,0 +1,110 @@
...
+#include <string.h>
+
+/* Log configuration */
+#include "sys/log.h"
+#define LOG_MODULE "CSPRNG"
+#define LOG_LEVEL LOG_LEVEL_NONE
+
+static struct csprng_seed seed;
+static unsigned read_state_bytes;
+static int seeded;
```
arurke, line 57: Rather use a bool for improved clarity?

arurke (changes requested):

os/lib/csprng.c
kkrentz: Putting declarations first complies with [Contiki's code style document ](https://docs.contiki-ng.org/en/develop/doc/project/Code-style.html) and much of the code in `os/` ...

arurke: I was not aware the style document actually had a "should always", I am fairly certain we don't enforce either way, so I will see if we won't change that wording. In any case, the current style in this PR is fine.

arurke (approved):

arurke:
Thx for the changes. I am reluctant to merge this alone as I don't usually deal with cc2538 or security - let's see if we can entice @g-oikonomou or some of the other @contiki-ng/maintainers

g-oikonomou (approved):
I see there have been many changes here since I first looked at it. The PR is much more self-contained now - I previously had concerns about some changes that felt off-topic but those appear to have been removed.

I am happy for this to go ahead, no objections.

arurke:
@kkrentz We want to include this in the upcoming release, are you able to rebase this PR?

PR #523, opened 2018-05-15
snprintf-based logging for IPv6 and RPL neighbor

simonduq:
This adds `uiplib_ipaddr_describe` for consistent printing of IPv6 addresses into a string, de-duplicating three code locations. 

Similarly, this adds `rpl_neighbor_describe` to print description of a RPL neighbor into a string. This enable a new shell command `rpl-status`, and might as well be used in the future by Web or CoAP servers to deliver RPL status.

### DIFF
--- a/os/sys/log.c
+++ b/os/sys/log.c
@@ -50,6 +50,7 @@
 
 #include "sys/log.h"
 #include "net/ipv6/ip64-addr.h"
+#include "net/ipv6/uiplib.h"
 
 int curr_log_level_rpl = LOG_CONF_LEVEL_RPL;
 int curr_log_level_tcpip = LOG_CONF_LEVEL_TCPIP;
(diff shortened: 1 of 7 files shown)


os/sys/log.c
```diff
@@ -84,35 +85,9 @@ struct log_module all_modules[] = {
...
-        if(f > 0) {
-          f = -1;
-        } else if(i > 0) {
-          LOG_OUTPUT(":");
-        }
-        LOG_OUTPUT("%x", a);
-      }
-    }
-  }
+  char buf[40];
```
g-oikonomou: Should we change 40 to some define that states max IPv6 address length in string format, specified in uiplib.h?

g-oikonomou:
Nothing wrong with a little code duplication! :D

PR #528, opened 2018-05-17
node killed at the end of 08-native-runs tests

Rehan-MALAK:
pgrep pattern shoud be smaller than 15 characters

Indeed `CODE=test-data-structures` can't be killed with `pgrep $CODE | xargs kill -` as we can see with 

```
user@81958babe9d8:~/contiki-ng$ make -C tests/08-native-runs/
make: Entering directory '/home/user/contiki-ng/tests/08-native-runs'
========== Running script test 01-test-data-structures.sh ==========
Starting native node
Closing native node
test-data-structures             TEST OK
========== Summary ==========
test-data-structures             TEST OK
make: Leaving directory '/home/user/contiki-ng/tests/08-native-runs'
```

and then 

```
user@81958babe9d8:~/contiki-ng$ top

top - 18:51:02 up 3 days, 11:45,  0 users,  load average: 0.70, 0.50, 0.52
Tasks:   4 total,   2 running,   2 sleeping,   0 stopped,   0 zombie
%Cpu(s):  8.2 us, 18.8 sy,  0.0 ni, 72.9 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem :  8109780 total,  2764576 free,  3602476 used,  1742728 buff/cache
KiB Swap:  8203260 total,  7881004 free,   322256 used.  4605484 avail Mem 

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                               
  578 user      20   0    2412   1452   1364 R  99.7  0.0   0:16.61 test-data-struc                       
    1 user      20   0    2372    568    524 S   0.0  0.0   0:00.01 sh                                    
    7 user      20   0    3764   2988   2672 S   0.0  0.0   0:00.00 bash                                  
  596 user      20   0    5876   3084   2664 R   0.0  0.0   0:00.00 top
```

### DIFF
--- a/tests/08-native-runs/01-test-data-structures.sh
+++ b/tests/08-native-runs/01-test-data-structures.sh
@@ -16,7 +16,7 @@ sleep 2
 
 echo "Closing native node"
 sleep 2
-pgrep $CODE | xargs kill -9
+pgrep ${CODE:0:15} | xargs kill -9
 
 if grep -q "=check-me= FAILED" $CODE.log ; then
   echo "==== make.log ====" ; cat make.log;
--- a/tests/08-native-runs/01-test-data-structures.sh
+++ b/tests/08-native-runs/01-test-data-structures.sh
@@ -16,7 +16,7 @@ sleep 2
 
 echo "Closing native node"
 sleep 2
-pgrep ${CODE:0:15} | xargs kill -9
+kill -9 ${CPID}
 
 if grep -q "=check-me= FAILED" $CODE.log ; then
   echo "==== make.log ====" ; cat make.log;
--- a/tests/08-native-runs/01-test-data-structures.sh
+++ b/tests/08-native-runs/01-test-data-structures.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+source ../utils.sh
 
 # Contiki directory
 CONTIKI=$1
@@ -16,7 +17,7 @@ sleep 2
 
 echo "Closing native node"
 sleep 2
-kill -9 ${CPID}
+kill_bg $CPID
 
 if grep -q "=check-me= FAILED" $CODE.log ; then
   echo "==== make.log ====" ; cat make.log;
--- a/tests/17-tun-rpl-br/04-border-router-traceroute.sh
+++ b/tests/17-tun-rpl-br/04-border-router-traceroute.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+source ../utils.sh
 
 # Contiki directory
 CONTIKI=$1
@@ -38,8 +39,8 @@ HOPS=`wc $BASENAME.scriptlog -l | cut -f 1 -d ' '`
 
 echo "Closing simulation and tunslip6"
 sleep 1
-kill -9 $JPID
-kill -9 $MPID
+kill_bg $JPID
+kill_bg $MPID
 sleep 1
 rm COOJA.testlog
 rm COOJA.log
--- a/tests/17-tun-rpl-br/05-native-ping.sh
+++ b/tests/17-tun-rpl-br/05-native-ping.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+source ../utils.sh
 
 # Contiki directory
 CONTIKI=$1
@@ -22,7 +23,7 @@ STATUS=${PIPESTATUS[0]}
 
 echo "Closing native node"
 sleep 2
-pgrep hello-world | sudo xargs kill -9
+kill_bg $CPID
 
 if [ $STATUS -eq 0 ] ; then
   cp $BASENAME.log $BASENAME.testlog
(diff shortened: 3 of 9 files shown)


g-oikonomou:
I cannot reproduce this problem with branch develop on Mac OS

Rehan-MALAK:
Ok, so let's wait for somebody else to test `make -C tests/08-native-runs/` and then `top` on the docker image on Linux and check if there is still the `test-data-struc` process running.

By the way, here is an extract of the man pages of pgrep/pkill on Debian : 

> NOTES
>        The process name used for matching is limited to the 15 characters present in the  output  of  /proc/pid/stat.
>        Use the -f option to match against the complete command line, /proc/pid/cmdline.
> 

Maybe the Mac Os X version behaves differently.

yatch:
I think I reproduced this with the docker image of `simonduq/contiki-ng`:
```
user@ef2a93144f27:~/contiki-ng$ git rev-parse HEAD
a2f0a949dcb1910fb3b271456b0b2b829333af7c

user@ef2a93144f27:~/contiki-ng$ ps u | grep test-data-structures
user        12  0.0  0.0   3024   796 pts/0    S+   22:04   0:00 grep test-data-structures

user@ef2a93144f27:~/contiki-ng$ make -C tests/08-native-runs
make: Entering directory '/home/user/contiki-ng/tests/08-native-runs'
========== Running script test 01-test-data-structures.sh ==========
Starting native node
Closing native node
test-data-structures             TEST OK
========== Summary ==========
test-data-structures             TEST OK
make: Leaving directory '/home/user/contiki-ng/tests/08-native-runs'

user@ef2a93144f27:~/contiki-ng$ ps u | grep test-data-structures
user        34  102  0.0   2272  1376 pts/0    R    22:05   0:06 ../../tests/07-simulation-base/code-data-structures//test-data-structures.native
user        53  0.0  0.0   2920   496 pts/0    R+   22:05   0:00 grep test-data-structures
```

g-oikonomou:
> Maybe the Mac Os X version behaves differently.

No such note on OS X. But this patch does not break OS X, so 👍 if it's a confirmed problem on Linux

tests/08-native-runs/01-test-data-structures.sh
```diff
@@ -16,7 +16,7 @@ sleep 2
 
 echo "Closing native node"
 sleep 2
-pgrep $CODE | xargs kill -9
+pgrep ${CODE:0:15} | xargs kill -9
```
yatch: Why don't we use `CPID`??? Is there any reason to avoid `kill -9 ${CPID}`?

g-oikonomou:
I think this may have to do with the design of parts of procfs, so would only be applicable of linux

Rehan-MALAK:
@yatch 
I think you are right... this `pgrep ... | sudo xargs kill -9` seems useful only for killing a `sudo command` because `$!` gives the `sudo command ` PID instead of the `command` PID. 
```bash
 #!/bin/bash

sudo -v
MYCMD=sleep
PARAM=60
sudo $MYCMD $PARAM &
CPID=$!
sleep 1
ps aux | grep -e $MYCMD
echo CPID  : $CPID
echo pgrep : $(pgrep $MYCMD)
```

which gives 

```
$ ./test.sh 
root        14  0.0  0.0   4972  3432 pts/0    S+   01:02   0:00 sudo sleep 60
root        16  0.0  0.0   2228   544 pts/0    S+   01:02   0:00 sleep 60
user        18  0.0  0.0   3032   880 pts/0    S+   01:02   0:00 grep -e sleep
CPID : 14
pgrep : 16
```

Here a simple `kill -9` seems fine.

simonduq:
I wonder how we could homogenize this. There other tests using `pgrep`, which may cause the same issue the day a change a process name to something larger than 15.

Two things:
* several tests to the `pgrep` / `xargs` trick on non-sudo commands -- unneeded
* wherever we need `pgrep`, would `-f` help make sure we never run into trouble?

simonduq:
BTW I couldn't reproduce on my system, I suspect that even with Docker, this is partly dependent on the host.

yatch:
> even with Docker, this is partly dependent on the host.

It's interesting...; I was using Docker for Mac yesterday night.

simonduq:
For the record, my host:
```
$ uname -a
Linux ubuntu 4.13.0-39-generic #44~16.04.1-Ubuntu SMP Thu Apr 5 16:43:10 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

Rehan-MALAK:
@simonduq 
As we have the PID of the background processes with $!, I think we should just forget about these pgrep subtleties (15 characters, -f option, platform dependency...). 
So in 39da0ae I suggest to use a new file **tests/utils.sh** with all our bash tricks and source it when needed at the beginning of our scripts... for example, here, I added a  kill_bg function taking the PID as the only parameter and differentiating the sudo case from the non-sudo case.

simonduq:
Sounds great, thanks! :)

PR #533, opened 2018-05-18
Simplify and homogenize node-id across all platforms

### LINKED ISSUE #195: The semantics of `node_id`
I never quite managed to get my head around this one. How do people use it? What is it needed for? I am thinking that we need to come up with a uniform way of handling it across all supported ports.

simonduq:
This homogenizes node-id, simply assigning it the last two bytes of the link-layer address. Removes the ability to store node-id on flash, which was only available on sky. Removes TinyOS nodeid handling.

In the long run, once we have runtime support for both long and short addresses, I believe we won't need node-id anymore. But until then, it's a useful 2-byte node identifier. Planning to contribute tools to uniquely assign IDs from a fixed set of MAC addresses, useful in testbeds or even in certain deployments.

Closes #195


g-oikonomou:
YES PLEASE!

simonduq:
That was fast ;)

g-oikonomou:
Furious too :P

PR #539, opened 2018-05-23
Feedback in CoAP send request 

carlosgp143:
Hi All!

Right now the function `coap_send_request` does not give any feedback in case there is no available transaction available to use (if `coap_new_transaction` returns NULL). Therefore, the applications on top cannot know if the request is actually been sent or not. 

Therefore, this small PR makes the `coap_send_request` return 1 if there is a transaction available and 0 otherwise, so that applications running on top can know if the request is actually sent.


simonduq:
Looks good. Would be nice to add doxygen for the new function prototype (although CoAP does not have any doxygen annotations yet)

carlosgp143:
I added the doxygen annotation for the prototype :-)

simonduq:
Thanks!

PR #542, opened 2018-05-25
Account for preamble+SFD time in the TSCH Rx offset calculation

simonduq:
The expected Rx time in TSCH is based on SFD timestamps, i.e., the time at the end of the reception of the SFD byte. But the PHY-layer frame starts earlier: in 802.15.4, there is a 4-byte preamble, and the SFD byte itself takes one byte of air time. And the radio must be already listening when these are on air.

This PR moves the listen period earlier, so as to center it on the expected start of the PHY frame.


simonduq:
Added WIP flag: we probably want to make this hw-dependent, as not all chips/drivers wait for all 4 bytes of preamble.

simonduq:
Documented the issue at https://github.com/contiki-ng/contiki-ng/issues/565

PR #547, opened 2018-05-25
Add support for documentation on readthedocs.io

### LINKED ISSUE #279: Doxgyen online auto-upload
Basically needs to be pushed at
https://github.com/contiki-ng/contiki-ng.github.io/tree/master/ngdoc

whenever a change is merged on `develop`. We could do it in Travis, using secret environment variables for the Github credentials https://docs.travis-ci.com/user/environment-variables/

g-oikonomou:
This pull configures Contiki-NG to work on https://readthedocs.org/.

We create a simple landing page which points to various parts of the API documentation. The API documentation itself is an automatically-triggered doxygen build. readthedocs allows us to flag branches/tags as active/inactive. For active ones, pushing triggers a new build which is what we want.

readthedocs will host multiple different versions of the documentation. Currently it only makes sense to show the docs corresponding to branch `develop`, but in the future we can enable this for `master` as well as for future releases. Currently the alias `latest` (default when a version is not specified) will point to docs for branch `develop`.

Example build: https://g-oikonomou-contiki-ng.readthedocs.io/en/latest/

It is also possible to build/test locally by running make under `tools/readthedocs`. The output can be found under `./_build` (again, doxygen build will be automatically triggered). For this to work, one needs to install `sphinx` and `sphinx_rtd_theme`.

Closes #279


g-oikonomou:
When building, doxygen generates a `html/dynsections.js`. This is used to collapse/expand things in the modules and files pages (probably elsewhere too). This is meant to get copied over the `_build/_static` dir, alongside all other doxygen-generated files. It does when running locally, but it doesn't when running on readthedocs and I've no idea why. If you open your browser's debugging tools, you will see an error 404 for this particular file and the expand/collapse functionality doesn't work.

Funnily enough, the same js file seems to also be missing from https://github.com/contiki-ng/contiki-ng.github.io and the API docs hosted there suffer from the same problem.

Ideas?

g-oikonomou:
Fixed by shipping our own dynsections.js for now at least

simonduq:
Nice! :)
I've noticed the js issue earlier but never really digged...
As for release versions, would we have `master`, or can we do per-tag, with one rev per release?

simonduq:
Do we need a CI test to check if this builds?

g-oikonomou:
The way it works is, the admin interface presents us with a list of all tags and branches. We enable/disable whatever we wish and then we make latest point to something. So we can have master as well as release/vX.Y.

We could add travis, but readthedocs users have been asking for status integration with github and I believe they are working on that. So perhaps we should wait. Plus, this will only break if someone touches anything under `tools/readthedocs/`, so we can check manually for now.

g-oikonomou:
Still not fully satisfied with this js business, so adding work-in-progress

g-oikonomou:
This is now ready for merge. Instead of getting sphinx to copy our doxygen output over to the readthedocs output dir, we now have a sphinx extension that does this. This allows us much more control in terms of what to copy to where.

Unfortunately, this doesn't solve the fundamental problem of missing .js files. For whatever reason, doxygen on my computer generates some files that doxygen on readthedocs does not, even though I even tested the same doxygen version. So for now, we still need to ship the missing .js

g-oikonomou:
Add testing on travis.

Requires #550 

Requires docker image update, as per Dockerfile changes here.

g-oikonomou:
As per the diff :)

tools/docker/Dockerfile
```diff
@@ -95,6 +95,10 @@ ENV PATH="${HOME}/renode:${PATH}"
 RUN git clone --recursive https://github.com/contiki-ng/contiki-ng.git ${CONTIKI_NG}
 RUN ant -q -f ${CONTIKI_NG}/tools/cooja/build.xml jar
 
+# Install sphinx and sphinx_rtd_theme, required for building and testing the
+# readthedocs API documentation
+RUN sudo pip install sphinx_rtd_theme sphinx
```
simonduq: Just tried, this requires installing ` python-pip` first

g-oikonomou:
pip is a nice thing to have anyway!

Shall we add it as part of this pull here?

simonduq:
I would say yes, and we also need https://github.com/contiki-ng/contiki-ng/pull/554. This way we get this PR green before merging.

g-oikonomou:
Sure, I'l push it. Presumably with apt?

I've looked at #554 but I don't feel qualified to review it, unfortunately

simonduq:
OK, next time is is rebased/merged on `develop`, Travis should build the new docker image and run the tests in it. But now there's a conflict in `.travis.yml`.

g-oikonomou:
Resolved

simonduq:
I thought we'd never see this one green! ;)

PR #550, opened 2018-05-27
Create the (failed) summary log even when sub-make fails

g-oikonomou:
The doxygen test on travis will produce a test summary if doxygen passes with warnings, but it will _not_ generate a test summary if doxygen fails with an error.

In this latter scenario, the travis job will fail - as it should - but by looking at the output we get no information as to what caused it to fail.

This is what we are seeing:

```
make: Entering directory '/home/user/contiki-ng/tests/00-doxygen'
make[1]: Entering directory '/home/user/contiki-ng/tools/doxygen'
> Cleaning Documentation
 done.
make[1]: Leaving directory '/home/user/contiki-ng/tools/doxygen'
Makefile:33: recipe for target 'doxygen' failed
make: *** [doxygen] Error 2
make: Leaving directory '/home/user/contiki-ng/tests/00-doxygen'
The command "docker run --privileged -v $CNG_HOST_PATH:/home/user/contiki-ng -ti $DOCKER_IMG bash --login -c "make -C tests/??-$TEST_NAME";" exited with 2.
$ $CNG_HOST_PATH/tests/check-test.sh $CNG_HOST_PATH/tests/??-$TEST_NAME; exit $?;
grep: /home/travis/build/g-oikonomou/contiki-ng/tests/00-doxygen/summary: No such file or directory
======== Test outcome =======
Test /home/travis/build/g-oikonomou/contiki-ng/tests/00-doxygen FAIL
```

The reason why this happens is, doxygen returns with non-zero, so the sub-make returns non-zero, so the top-level make exits with an error while processing rule `doxygen`, i.e. before getting to the `summary`

This is en example such failure: https://travis-ci.org/g-oikonomou/contiki-ng/jobs/384369894

This pull changes the Makefile such that the `doxygen` recipe ignores errors caused by the invocation of sub-make.

We also change `make`, to `$(MAKE)`, which is something we should do across the board, really... :)


simonduq:
I'm not too familiar with this makefile, I assume you have double-checked that tests still fail when Doxygen fails as well as when there are errors/warnings?

g-oikonomou:
Yes tested here: https://travis-ci.org/g-oikonomou/contiki-ng/builds/384840220

(doxygen succeed but generates a warning because of documentation problems)

Issue #186, opened 2017-11-21
cc2538 timer trigger insanely


yoyo890121:
Hi,
I try to add a process with timer to do something periodically in `border-router-embedded.c` and `coap-example-server.c`.
I use TSCH MAC layer and my device is cc2538dk.
I tried contiki and timer trigger insanely in beginning, and I set `LPM_CONF_ENABLE` to 0 it still occurred sometime, so I wanted to try contiki-ng.
It seemed to work well in beginning, but sometimes this situation occurred after 3 days in my test.

Thanks.

```
PROCESS(node_process, "Node Timer");

PROCESS_THREAD(node_process, ev, data)
{
  static struct etimer etaa;
  PROCESS_BEGIN();
  static int i=0;  
  etimer_set(&etaa, CLOCK_SECOND * 10);  
  while(1) {
    printf("10 seconds. \t i=%d \n", i);
    PROCESS_YIELD_UNTIL(etimer_expired(&etaa));
    etimer_reset(&etaa);
    i++;    
  }
  PROCESS_END();
}
```

simonduq:
Thanks for reporting. Would you mind sharing a full example code? (c/h and makefile).

yoyo890121:
https://gist.github.com/yoyo890121/37336870bc84d291c8f29222c4754079
https://gist.github.com/yoyo890121/6bacd2e9b1f527f2b8e76f130519d1c1

I just edited these files in `contiki-ng/examples`, and set `#define LOG_CONF_LEVEL_MAC  
  LOG_LEVEL_DBG` to print log for testing contiki-ng.

Thanks for helping.

simonduq:
OK thanks. Would be ideal if you could actually share a self-contained branch we can directly try out. BTW is the problem on the BR, the other node or both? Any steps how to reproduce and how the bug exactly manifests itself will help.

Until then, what I think you should try is a basic hello-world with your etimer loop and TSCH enabled. Please let us know how this goes!

yoyo890121:
https://github.com/yoyo890121/contiki-ng/tree/hong-testing
Hi, I forked contiki-ng and start a basic hello-world with etimer loop and TSCH enabled testing.
The problem both occurred on BR and node, and when the problem occurred, console printed 
`10 seconds.  i=(i)  ` insanely to show this situation.

Thanks.

simonduq:
Thanks, we will look into this when we find the time!

g-oikonomou:
[...]

PR #556, opened 2018-05-28
rtimer issues: redundant calls to rtimer_arch_schedule

kkrentz:
When calling `rtimer_set` from within the rtimer ISR, `rtimer_arch_schedule` gets called twice, even though there is code that tries to prevent this from happening. This PR provides a fix through small changes to `rtimer.c`.

### DIFF
--- a/os/sys/rtimer.c
+++ b/os/sys/rtimer.c
@@ -62,12 +62,10 @@ rtimer_set(struct rtimer *rtimer, rtimer_clock_t time,
 	   rtimer_clock_t duration,
 	   rtimer_callback_t func, void *ptr)
 {
-  int first = 0;
-
   PRINTF("rtimer_set time %d\n", time);
 
-  if(next_rtimer == NULL) {
-    first = 1;
+  if(next_rtimer) {
+    return RTIMER_ERR_ALREADY_SCHEDULED;
   }
 
   rtimer->func = func;
@@ -76,9 +74,7 @@ rtimer_set(struct rtimer *rtimer, rtimer_clock_t time,
   rtimer->time = time;
   next_rtimer = rtimer;
 
-  if(first == 1) {
-    rtimer_arch_schedule(time);
-  }
+  rtimer_arch_schedule(time);
   return RTIMER_OK;
 }
 /*---------------------------------------------------------------------------*/
@@ -92,10 +88,6 @@ rtimer_run_next(void)
   t = next_rtimer;
   next_rtimer = NULL;
   t->func(t, t->ptr);
-  if(next_rtimer != NULL) {
-    rtimer_arch_schedule(next_rtimer->time);
-  }
-  return;
 }
 /*---------------------------------------------------------------------------*/
 


simonduq:
I haven't looked closely yet, but is this related to https://github.com/contiki-ng/contiki-ng/issues/186 by any chance?

simonduq:
Not sure I fully got the issue. How do the new functions differ from `RTIMER_CLOCK_LT` and `RTIMER_CLOCK_DIFF`?

simonduq:
(except for the fact they are documented, unlike the existing macros :s)

kkrentz:
Maybe the naming of the functions was a bit confusing. I am not doing a plain comparison like done by `RTIMER_CLOCK_LT`. This will not work in boundary cases where one of the timestamps wrapped around zero. Instead, I added functionality to handle such cases.

simonduq:
But `RTIMER_CLOCK_DIFF` and `RTIMER_CLOCK_LT` do handle wrapping, by working on signed integers. A 
timer close to wrapping will be considered as *before* a timer that just wrapped (as long as the time delta is less than half the wrapping period).

kkrentz:
This is tricky - like it. Nevertheless, what retains to be useful of this PR are the fixes to `rtimer_set` and `rtimer_arch_schedule`. I retained these changes and cut the rest.

arch/cpu/cc2538/rtimer-arch.c
```diff
@@ -70,21 +70,17 @@ rtimer_arch_init(void)
...
-  now = RTIMER_NOW();
-
   /*
    * New value must be 5 ticks in the future. The ST may tick once while we're
    * writing the registers. We play it safe here and we add a bit of leeway
    */
-  if((int32_t)(t - now) < 7) {
-    t = now + 7;
+  if(!rtimer_is_schedulable(t)) {
+    t = RTIMER_NOW() + RTIMER_GUARD_TIME;
```
simonduq: There is something that feels unnatural to me here. If we abstract away the notion of schedulability, why do we still compute `t` manually here? What if `rtimer_is_schedulable` gets modified in the future with criteria different from just looking at the guard? So, if we are to add this `rtimer_is_schedulable`, maybe it also makes sense to also add a function that returns the next scheduleable time.

os/sys/rtimer.h
```diff
@@ -84,6 +85,7 @@ typedef uint64_t rtimer_clock_t;
 
 #define RTIMER_CLOCK_MAX           ((rtimer_clock_t)-1)
 #define RTIMER_CLOCK_LT(a, b)      (RTIMER_CLOCK_DIFF((a),(b)) < 0)
+#define RTIMER_CLOCK_LT_OR_EQ(a,b) (RTIMER_CLOCK_DIFF((a),(b)) <= 0)
```
simonduq: code style (and consistency with line above): missing space after comma in macro arguments

arch/cpu/cc2538/rtimer-arch.c
kkrentz: My motivation to expose `rtimer_is_schedulable` was that my current CSL implementation depends on it. Yet, I found a workaround for this function ... so I removed this function from this PR

os/sys/rtimer.h
kkrentz: That macro is no longer needed

simonduq:
OK. At first glance the rtimer changes look like a good simplification but I'm not 100% confident. For instance, is it safe not to call `rtimer_arch_schedule` in `rtimer_run_next` after `t->func(t, t->ptr);`? Also, the PR is changing the behavior of `rtimer_set` in the case where the timer is already set. Not sure about the side effects. Would like to hear more opinions here.

kkrentz:
I created a small [demo](https://github.com/kkrentz/contiki-ng/commit/f7a8ad7be056718117a21a02a29cfe99841850eb) to demonstrate that calling `rtimer_set` from within the rtimer ISR results in `rtimer_arch_schedule` being called twice. This is the example code:
```c
static void
on_timeout_2(struct rtimer *rt, void *ptr)
{
  printf("on_timeout_2\n");
}
static void
on_timeout_1(struct rtimer *rt, void *ptr)
{
  printf("on_timeout_1\n");
  rtimer_set(&timer, RTIMER_NOW() + RTIMER_SECOND, 1, on_timeout_2, NULL);
}
PROCESS_THREAD(rtimer_bug_process, ev, data)
{
  PROCESS_BEGIN();

  rtimer_set(&timer, RTIMER_NOW() + RTIMER_SECOND, 1, on_timeout_1, NULL);

  PROCESS_END();
}
```

I also inserted a ```printf``` into ```arch/cpu/cc2538/rtimer-arch.c```:
```c
void
rtimer_arch_schedule(rtimer_clock_t t)
{
  rtimer_clock_t now;

  printf("rtimer_arch_schedule\n");
```

Currently, the serial output will be:
```
rtimer_arch_schedule
on_timeout_1
rtimer_arch_schedule
rtimer_arch_schedule
on_timeout_2
```

With my fix, the serial output is as expected:
```
rtimer_arch_schedule
on_timeout_1
rtimer_arch_schedule
on_timeout_2
```

yanpads:
```C
void
rtimer_run_next(void)
{
  struct rtimer *t;
  if(next_rtimer == NULL) {
    return;
  }
  t = next_rtimer;
  next_rtimer = NULL;
  
  /* 1st call */
  t->func(t, t->ptr);  
  
  if(next_rtimer != NULL) {

    /* 2nd call */
    rtimer_arch_schedule(next_rtimer->time);
  }
  return;
}
```

1st call: `rtimer_set` is called via `tsch_slot_operation`.

2nd call: `next_rtimer` is always true because of `next_rtimer = rtimer` inside `rtimer_set` during 1st call. It will result then result to   `rtimer_arch_schedule(next_rtimer->time)`.

alexrayne:
This patch forces to rtime callback invoke rtimer_set to setup next timer. This is breaks native behaviour - where application can prvide a set of rtimers, that called by rtmer scheduler.
code like will not work, since timer2 not invokes : 

```
static void
on_timeout_2(struct rtimer *rt, void *ptr)
{
  printf("on_timeout_2\n");
}
static void
on_timeout_1(struct rtimer *rt, void *ptr)
{
  printf("on_timeout_1\n");
  rtimer_set(&timer, RTIMER_NOW() + RTIMER_SECOND, 1, on_timeout_2, NULL);
}
PROCESS_THREAD(rtimer_bug_process, ev, data)
{
  PROCESS_BEGIN();

  rtimer_set(&timer1, RTIMER_NOW() + RTIMER_SECOND, 1, on_timeout_1, NULL);
// this timer is ommited, cause scheduler bother only about timer1
  rtimer_set(&timer2, RTIMER_NOW() + RTIMER_SECOND*3, 1, on_timeout_2, NULL);

  PROCESS_END();
}

```

kkrentz:
Do we need to maintain this special behavior of the native platform?

simonduq:
Hi!

nfi (approved):

joakimeriksson:
@kkrentz - Can you do a rebase so that we get the test run again - and then we will merge this!

PR #558, opened 2018-05-29
Add back link layer security for CSMA

joakimeriksson:
Adding back link-layer security for CSMA as it was removed when removing LLSEC as a separate layer. This is yet work in progress but is here for starting discussions and for testing.


kkrentz:
Looks like this re-adds some of my old code. An issue with this old version is that it lacks support for reboots. In fact, if a node reboots, its frame counter will start over from zero, thereby ensuing reuses of CCM nonces. Another issue is that if I reboot, my neighbors will consider my frames as replayed (also because my frame counter starts over from zero). Most severely, since anti-replay data is not persisted across reboots, replay attacks become possible after reboots. One solution to all these problems is persisting all anti-replay data, as well as the current frame counter in flash memory, but this solutions neither seems to be energy-efficient nor scalable. As an alternative solution, we proposed a session key establishment scheme called AKES (and have ongoing work on a corresponding key revocation and rekeying protocol). Actually, I have ported AKES to Contiki-NG's CSMA-MAC already, but what is left to be done is securing acknowledgement frames. Anyway, if there is a desire to merge this, I can put together a PR.

os/net/mac/csma/csma-security.h
```diff
@@ -0,0 +1,74 @@
...
+ * \author
+ *         Olav Frengstad <olav@tiny-mesh.com>
+ */
+
+#ifndef CSMA_SECURITY_H_
+#define CSMA_SECURITY_H_
+
+
+#ifdef CSMA_CONF_LLSEC_DEFAULT_KEY0
+#define CSMA_LLSEC_DEFAULT_KEY0 CSMA_LLSEC_DEFAULT_KEY0
```
simonduq: missing `_CONF`

os/net/mac/csma/csma-security.h
```diff
@@ -0,0 +1,74 @@
...
+#else
+#define CSMA_LLSEC_DEFAULT_KEY0 {0x10, 0x11,0x12,0x13,0x14,0x15,0x16,0x17,0x18,0x19,0x1a,0x1b,0x1c,0x1d,0x1e,0x1f}
```
simonduq: inconsistent spacing after comma

os/net/mac/csma/csma-security.c
```diff
@@ -0,0 +1,267 @@
...
+int
+csma_security_create_frame(void)
+{
+  packetbuf_set_attr(PACKETBUF_ATTR_FRAME_TYPE, FRAME802154_DATAFRAME);
+  return NETSTACK_FRAMER.create();
+}
+int
+csma_security_parse_frame(void)
+{
+  return NETSTACK_FRAMER.parse();
```
simonduq: What happens if a secure frame lands here? Will we have a suitable warning in the framer?

os/net/mac/csma/csma-security.c
```diff
@@ -0,0 +1,267 @@
...
+}
```
lafka: Should the else block include a empty version of `csma_security_set_key/2`? It's exported in `csma.h` and is expected to be available.

os/net/mac/csma/csma-security.c
```diff
@@ -0,0 +1,267 @@
...
+#include <stdio.h>
+#include <string.h>
+#include "ccm-star-packetbuf.h"
+/* Log configuration */
+#include "sys/log.h"
+#define LOG_MODULE "CSMA"
+#define LOG_LEVEL LOG_LEVEL_MAC
+
+#if LLSEC802154_USES_AUX_HEADER && LLSEC802154_USES_FRAME_COUNTER
+
```
lafka: `MIC_LEN` is not defined, should add: 
`#define MIC_LEN LLSEC802154_MIC_LEN(CSMA_LLSEC_SECURITY_LEVEL)`

lafka:
In `framer_802154_setup_params()` on line 165 (framer-802154.c) LLSEC is conditionally enabled if `PACKETBUF_ATTR_SECURITY_LEVEL` is set.

This causes an issue with fragmentation as the attribute is never set before calling `framer_hdrlen = NETSTACK_FRAMER.length()` inn sicslowpan.c. This causes fragmentation to discard the first fragment as it can't fit it in packetbuf (potentially it might also causes packets with sizes close to `mac_max_payload`  to drop as well but that's unverified).

Setting the packetbuf attribute fixes the issue:

```diff
@@ -1594,6 +1595,16 @@ output(const linkaddr_t *localdest)
     packetbuf_set_attr(PACKETBUF_ATTR_MAC_ACK, PACKETBUF_ATTR_MAC_ACK_DISABLE);
   }
 
+  /* In case LLSEC is enabled for 802154, set flag here so framer_hdrlen can
+     calculate the correct frame length */
+#if LLSEC802154_CONF_ENABLED
+  packetbuf_set_attr(PACKETBUF_ATTR_SECURITY_LEVEL,
+    FRAME802154_SECURITY_LEVEL_NONE != CSMA_LLSEC_SECURITY_LEVEL);
+#endif /* LLSEC802154_CONF_ENABLED */
```
Including MAC specifics in sicslowpan is not the cleanest as it would break TSCH.

@joakimeriksson would you like me to open a PR directly to your fork with my changes?

joakimeriksson:
@lafka yes, please do!

os/net/mac/csma/csma-security.c
joakimeriksson: It is exactly the same thing as happens with CSMA today - frame parsed, security parameters ignored, rubbish sent in as packet payload (as it is not at all handling security if it is turned off).

joakimeriksson:
Removed the work-in-progress as I think it is ok. I think all requested modifications are fixed.

os/net/ipv6/sicslowpan.c
```diff
@@ -1615,6 +1619,11 @@ output(const linkaddr_t *localdest)
   }
 #endif /* SICSLOWPAN_COMPRESSION >= SICSLOWPAN_COMPRESSION_IPHC */
 
+#if MAC_CONF_WITH_CSMA && LLSEC802154_CONF_ENABLED
+    packetbuf_set_attr(PACKETBUF_ATTR_SECURITY_LEVEL,
```
simonduq: why not do this in the MAC output function? Would about the check for "CSMA" in sicslowpan

joakimeriksson: It is also set in the CSMA layer - but I think this is for getting a more correct expected size of the header length so that the fragmentation gets good input size for the header - but should it not be setting the correct level instead of 0/1?

simonduq: Having conditionals on the MAC just doesn't feel right. If we need info from the MAC here then let's add a function in the MAC API.

Agree we prob wan the level, not just 0/1

joakimeriksson: I totally agree - how is this handled with TSCH? Is security always on, or is it assuming always there security headers and therefore do not create "full" packets if security level happens to be 0? (if allowed). Optimally there would be no dependencies at all here - on any MAC level features except querying for header size - so maybe the framer calls should also be handled via the NETSTACK MAC - so that the MAC layer could add whatever expected config it needs for configuring the framer?

simonduq: Looks like we have a bug in the TSCH case; security is just not accounted for here :s

simonduq: Yep I think that's the most sensible solution here

kkrentz:
I just want to remind you that combining static keys with frame counters leads to various problems, such as:
- If a node reboots, its frame counter starts over. This causes neighbors to consider a rebooted node's frames as replayed and also ensues a nonce reuse in 802.15.4 security.
- Moreover, if a node reboots, its anti-replay data gets lost, thus exposing rebooted nodes to replay attacks.

Therefore, I would suggest to go for session keys and actually have an implementation of this lying around. If there is interest in such a solution, I can create a PR.

os/net/ipv6/sicslowpan.c
lafka: > Having conditionals on the MAC just doesn't feel right. If we need info from the MAC here then let's add a function in the MAC API.

Agreed. I put it there since I found it equally awkward to put it anywhere else, looking at it now it seems worse than before..

> Maybe sicslowpan.c should call the MAC for getting the header size for this packet - and then the MAC that happen to understand the details of its configuration would be able to calculate it together with the framer.

Or should the MAC rather export a function to perform its own initialization of `packetbuf`?

lafka:
> I just want to remind you that combining static keys with frame counters leads to various problems, such as:
> 
> * If a node reboots, its frame counter starts over. This causes neighbors to consider a rebooted node's frames as replayed and also ensues a nonce reuse in 802.15.4 security.
> * Moreover, if a node reboots, its anti-replay data gets lost, thus exposing rebooted nodes to replay attacks.
> 
> Therefore, I would suggest to go for session keys and actually have an implementation of this lying around. If there is interest in such a solution, I can create a PR.

A PR would be great, I'm definitely interested in adding session keys in the future!

os/net/ipv6/sicslowpan.c
simonduq: > Or should the MAC rather export a function to perform its own initialization of packetbuf?

Would work for me just as well!

joakimeriksson: I will try to do a PR where we add a MAC function for the length - I do think it is the main (possibly the only) thing 6LoWPAN will need to know - befor creating fragments.

kkrentz: The way I solved this (also in #506) was to configure an extra framer that "wraps" the actual framer.

simonduq:
@kkrentz I guess there are two parts (correct me if wrong) to what you have:
1. Mechanisms for dynamic key management in the CSMA sec implem. These I would say are definitely wanted
1. Session management, needed in principle, but non-standard. If you have the code ready please PR so we can consider for inclusion! I would say this should go as a `service`

joakimeriksson:
I am doing an attempt at 1 right now - considering to att a few shell commands to configure level and keys to have a "demo" also. PR will be updated soon...

joakimeriksson:
Added link-layer security level and key in uipbuf module so that anyone sending IP packets can control sec-level dynamically. Also added a few shell commands so that when having Shell enabled together with LLSEC there are some commands available (llsec-set-level and llsec-set-key).

os/net/ipv6/uipbuf.c
```diff
@@ -131,3 +138,15 @@ uipbuf_is_attr_flag(uint16_t flag)
...
 /*---------------------------------------------------------------------------*/
+void
+uipbuf_init(void)
+{
+  /* And initialize anything that should be initialized */
+  uipbuf_set_attr(UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS,
+                  UIP_MAX_MAC_TRANSMISSIONS_UNDEFINED);
+  /* set the not-set default value - this will cause the MAC layer to
+     configure its default */
+  uipbuf_set_attr(UIPBUF_ATTR_LLSEC_LEVEL, UIPBUF_ATTR_LLSEC_LEVEL_MAC_DEFAULT);
```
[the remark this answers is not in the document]
joakimeriksson: true - should be set as default - my bad.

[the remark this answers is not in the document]
joakimeriksson: true - should be default here.

os/services/shell/shell-commands.c
```diff
@@ -729,6 +729,75 @@ PT_THREAD(cmd_6top(struct pt *pt, shell_output_func output, char *args))
...
+    if(lv < 0 || lv > 7) {
+      SHELL_OUTPUT(output, "Illegal LLSEC Level %d\n", lv);
+      PT_EXIT(pt);
+    } else {
+      uipbuf_set_default_attr(UIPBUF_ATTR_LLSEC_LEVEL, lv);
+      uipbuf_clear_attr();
+      SHELL_OUTPUT(output, "LLSEC default level set %d\n", lv);
+    }
+  }
+  SHELL_ARGS_NEXT(args, next_args);
```
simonduq: this should be removed (or is it needed above, before accessing `args`?)

simonduq:
Because this affects general modules such as uipbuf and the MAC driver API, we decided to merge this in develop after the 4.2 feature-freeze, i.e., this will be part of release 4.3

simonduq:
@lafka probably not :)

joakimeriksson:
Fixed a few things that Niclas commented and also removed some unused stuff in the shell code and finally fixed so that security level and key-index is moving upwards from packetbuf into uipbuf attributes.

simonduq:
Agree. In TSCH we do it at join-time with `TSCH_CONF_JOIN_SECURED_ONLY`, not on a per-packet basis. Assumption is that if you joined a secure network then you'll accept all packets in this network, whether they are secured or not.

More per-packet configurability as you suggest would be useful, for both CSMA and TSCH.

examples/rpl-udp/udp-client.c
```diff
@@ -59,7 +59,8 @@ PROCESS_THREAD(udp_client_process, ev, data)
       LOG_INFO("Sending request %u to ", count);
       LOG_INFO_6ADDR(&dest_ipaddr);
       LOG_INFO_("\n");
-      snprintf(str, sizeof(str), "hello %d", count);
+      /* avoid the risk of not fitting the '\0' in the string */
+      snprintf(str, sizeof(str) - 1, "hello %d", count);
```
simonduq: I believe `snprintf` *always* writes the null terminating char

os/net/ipv6/sicslowpan.c
```diff
@@ -1588,13 +1568,23 @@ output(const linkaddr_t *localdest)
...
+  packetbuf_set_attr(PACKETBUF_ATTR_SECURITY_LEVEL,
+    uipbuf_get_attr(UIPBUF_ATTR_LLSEC_LEVEL));
+#if LLSEC802154_USES_EXPLICIT_KEYS
+  packetbuf_set_attr(PACKETBUF_ATTR_KEY_INDEX,
+    uipbuf_get_attr(UIPBUF_ATTR_LLSEC_KEY_ID));
+#endif /* LLSEC802154_USES_EXPLICIT_KEYS */
+#endif /*  LLSEC802154_USES_AUX_HEADER */
 
-  mac_max_payload = MAC_MAX_PAYLOAD - framer_hdrlen;
+  mac_max_payload = NETSTACK_MAC.max_payload();
```
simonduq: so this results in the exact same value as before, right?

joakimeriksson: That is the intended semantics of the max_payload() function yes!

examples/rpl-udp/udp-client.c
joakimeriksson: true - will fix. There might be some old Microsoft snprintf that did not.

simonduq: please give a heads up when fixed, then I'm ready to approve this PR!

joakimeriksson: this is fixed in the last commit.

os/services/shell/shell-commands.c
```diff
@@ -801,6 +869,10 @@ const struct shell_command_t builtin_shell_commands[] = {
 #if TSCH_WITH_SIXTOP
   { "6top",                 cmd_6top,                 "'> 6top help': Shows 6top command usage" },
 #endif /* TSCH_WITH_SIXTOP */
+#if LLSEC802154_ENABLED
+  { "llsec-set-level", cmd_llsec_setlv, "'> llsec-set-level <lv>': Set the level of link layer security (show if no lv argument)"},
+  { "llsec-set-key", cmd_llsec_setkey, "'> llsec-set-key <id> <key>': Set the key of link layer security (show if no id key argument)"},
```
nfi: The command does not show the key.
```suggestion
  { "llsec-set-key", cmd_llsec_setkey, "'> llsec-set-key <id> <key>': Set the key of link layer security"},
```

joakimeriksson: yes, should I add that or remove the comment? I am thinking that we should not show the key?!

PR #560, opened 2018-05-29
Leshan server qmode fix

Rehan-MALAK:
The leshan server with qmode enabled introduced in 1dfa62fc88 is downloaded in tests/18-coap-lwm2m/09-lwm2m-qmode-standalone-test.sh but it's not the one that is actually used in the test...
Update also the .gitignore


carlosgp143:
Hi! 
Sorry, my bad. I missed it in the standalone test. It should be as it is done right now :-)
Thank you for solving this issue!

PR #561, opened 2018-05-30
Add list_contains feature into list.h

carlosgp143:
Hi!
I think this is an useful feature that is missing right now in the list implementation. You can easily check if an item is present in a list.

### DIFF
--- a/os/lib/list.c
+++ b/os/lib/list.c
@@ -324,4 +324,26 @@ list_item_next(void *item)
   return item == NULL ? NULL : ((struct list *)item)->next;
 }
 /*---------------------------------------------------------------------------*/
+/**
+ * \brief      Check if the list contains an item
+ * \param list The list that is checked
+ * \param item An item to look for in the list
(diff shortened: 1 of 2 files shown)


simonduq:
Useful addition. Would be nice to add a couple extra checks to the following CI test: https://github.com/contiki-ng/contiki-ng/blob/develop/tests/07-simulation-base/code-data-structures/test-data-structures.c

carlosgp143:
Good idea! It is done now :-)

nfi:
Useful feature. Since the new function is a boolean function - what do you think about using the return type `bool` instead of `int` (see `stdbool.h`) and return `true`/`false` instead of `1`/`0` ?

carlosgp143:
Sure! I thought it was simple with 1/0 but of course it is more clear with true/false.

nfi:
Thanks for the update! It is more a code style thing. Using `bool` there is no question how to check for success or failure but there are many functions returning `int` where zero is success and non-zero an error code.

PR #562, opened 2018-05-31
Reworked busy-wait routines

simonduq:
This PR does a few things:
* Adds common routines for busy-waiting, in `rtimer.h`, thus de-duplicating code
* Makes the busy-wait routing fully portable, i.e. also works in Cooja motes without explicit addition of Cooja-mt-yield instructions
* Reworks CSMA busy-wait logic

### DIFF
--- a/os/sys/rtimer.h
+++ b/os/sys/rtimer.h
@@ -55,6 +55,7 @@
 
 #include "contiki.h"
 #include "dev/watchdog.h"
+#include <stdbool.h>
 
 /** \brief The rtimer size (in bytes) */
 #ifdef RTIMER_CONF_CLOCK_SIZE
@@ -189,18 +190,20 @@ void rtimer_arch_schedule(rtimer_clock_t t);
 
 /** \brief Busy-wait until a condition. Start time is t0, max wait time is max_time */
 #define RTIMER_BUSYWAIT_UNTIL_ABS(cond, t0, max_time) \
-  do { \
-    while(!(cond) && RTIMER_CLOCK_LT(RTIMER_NOW(), (t0) + (max_time))) { \
-      watchdog_periodic(); \
-    } \
-  } while(0)
+  ({                                                                \
+    bool c;                                                         \
+    while(!(c = cond) && RTIMER_CLOCK_LT(RTIMER_NOW(), (t0) + (max_time))) { \
+      watchdog_periodic();                                          \
+    }                                                               \
+    c;                                                              \
+  })
 
 /** \brief Busy-wait until a condition for at most max_time */
-#define RTIMER_BUSYWAIT_UNTIL(cond, max_time) \
-  do { \
-    rtimer_clock_t t0 = RTIMER_NOW(); \
-    RTIMER_BUSYWAIT_UNTIL_ABS(cond, t0, max_time); \
-  } while(0)
+#define RTIMER_BUSYWAIT_UNTIL(cond, max_time)       \
+  ({                                                \
+    rtimer_clock_t t0 = RTIMER_NOW();               \
+    RTIMER_BUSYWAIT_UNTIL_ABS(cond, t0, max_time);  \
+  })
 
 /** \brief Busy-wait for a fixed duration */
 #define RTIMER_BUSYWAIT(duration) RTIMER_BUSYWAIT_UNTIL(0, duration)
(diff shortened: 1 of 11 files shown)


kkrentz:
These are handy routines. I am just wondering whether there is any advantage of introducing these routines as macros rather than functions. If they were functions, the code size would probably shrink.

atiselsts:
Looks good. I rather don't mind the macro approach. With macros one can do nice things such as printing the `__LINE__` from which they were called, and not sure that the code size is an issue on any platform except sky.

One comment is that it would be nice to get a return value here. I suppose you can encapsulate the macro statements in an expression to do that.

```
/** \brief Busy-wait until a condition. Start time is t0, max wait time is max_time */
#define RTIMER_BUSYWAIT_UNTIL_ABS(cond, t0, max_time)               \
  ({                                                                \
    bool c = cond;                                                  \
    while(!c && RTIMER_CLOCK_LT(RTIMER_NOW(), (t0) + (max_time))) { \
      watchdog_periodic();                                          \
    }                                                               \
    c;                                                              \
  })

/** \brief Busy-wait until a condition for at most max_time */
#define RTIMER_BUSYWAIT_UNTIL(cond, max_time)       \
  ({                                                \
    rtimer_clock_t t0 = RTIMER_NOW();               \
    RTIMER_BUSYWAIT_UNTIL_ABS(cond, t0, max_time);  \
  })

/** \brief Busy-wait for a fixed duration */
#define RTIMER_BUSYWAIT(duration) RTIMER_BUSYWAIT_UNTIL(0, duration)

```

Now it possible to use the macros in conditions:
```
  if(RTIMER_BUSYWAIT_UNTIL(false, RTIMER_SECOND)) {
    printf("wait for `false` succeeded\n");
  } else {
    printf("wait for `false` failed\n");
  }
```

I think the most practical is to introduce another macro that logs an error message if the wait failed:
```
#define RTIMER_LOG_BUSYWAIT_UNTIL(cond, max_time, message)  \
  (RTIMER_BUSYWAIT_UNTIL(cond, max_time) || printf("%s:%d: %s\n", __FILE__, __LINE__, message))

  ...
  RTIMER_LOG_BUSYWAIT_UNTIL(false, RTIMER_SECOND, "wait for `false` failed");
```

simonduq:
Sounds like a good idea but are you sure the encapsulation thing works at all? I'm very unsure.

atiselsts:
It's a GCC extension.

simonduq:
Do you know where this is documented?
Also, if it's gcc-only, we might simply want to avoid. I think (hope) we're still compile-able on e.g. IAR or Clang.

atiselsts:
Hi Simon, fundamentally, I now think that it's not good to just continue as if nothing has happened after waiting for a condition and then failing to meet it. There should be some option to do something, at least for debug builds, e.g. to print the faulty line and file. The exact way how it's done don't matter that much.

W.r.t statement expression - it will work on almost any modern compiler, including clang. Don't know about IAR. If you're serious about supporting it, that's tough one. There are quite a lot `__attribute__` etc. GNUC code in Contiki already. There is one recent (2015) guide for using IAR, which just mentions "lot of errors": http://processors.wiki.ti.com/index.php/Contiki-IAR#Step-5.C2.A0:_Building_the_project_and_removing_the_errors. Fundamentally, I don't think it will be safe to use even if it can be built (e.g. there are macros that expand to `__attribute__((packed))` on GNUC and to nothing otherwise). One extreme is going back to supporting SDCC and every other old, limited compiler, which just leads to bad code in my opinion.

simonduq:
OK, I will give this a try. If we end up concerned with portability to other compilers we can always have a fallback on the classic version

simonduq:
Pushed new version with your proposal!

os/sys/rtimer.h
```diff
@@ -186,6 +187,26 @@ void rtimer_arch_schedule(rtimer_clock_t t);
 #define RTIMER_GUARD_TIME (RTIMER_ARCH_SECOND >> 14)
 #endif /* RTIMER_CONF_GUARD_TIME */
 
+/** \brief Busy-wait until a condition. Start time is t0, max wait time is max_time */
+#define RTIMER_BUSYWAIT_UNTIL_ABS(cond, t0, max_time) \
+  ({                                                                \
+    bool c;                                                         \
```
g-oikonomou: Suggest including stdbool.h here

simonduq: done

simonduq:
@atiselsts can we move forward with this PR? (I'll take care of conflict resolution after comments or approve)

PR #563, opened 2018-05-31
Make CSMA stop waiting when ACK is received

MartenBE:
This PR aims to address the following issues:

- When the ACK waiting time is large (for example `#define CSMA_CONF_ACK_WAIT_TIME (RTIMER_SECOND / 2)`, the watchdog kicks in before the ACK waiting time is over, resetting the node.

- CSMA uses a fixed period to wait and only then reads the received packet from the cc1200 driver. Therefore, the next scenario could occur: node A sends a message to node B and waits for `CSMA_CONF_ACK_WAIT_TIME` time. In this interval the ACK is send from node B to A, but soon followed with an answer message from node B to A before the `CSMA_CONF_ACK_WAIT_TIME` interval of node A has passed. The ACK comes first and occupies the driver's buffer, causing node A to discard the answer message from node B and thus triggering a retransmission from node B later on.

As CSMA is a widely used base feature of contiki-ng, I urge everyone to test this PR extensively. It helped me achieve better performance and reduced duty cycle, but I've only had limited means (Zolertia Remote with CC1200 chip) to try this out.

With kind regards,

Martijn

### DIFF
--- a/os/net/mac/csma/csma-output.c
+++ b/os/net/mac/csma/csma-output.c
@@ -207,6 +207,7 @@ send_one_packet(void *ptr)
           wt = RTIMER_NOW();
           watchdog_periodic();
           while(RTIMER_CLOCK_LT(RTIMER_NOW(), wt + CSMA_ACK_WAIT_TIME)) {
+            watchdog_periodic();
 #if CONTIKI_TARGET_COOJA
             simProcessRunValue = 1;
             cooja_mt_yield();
@@ -225,6 +226,7 @@ send_one_packet(void *ptr)
               watchdog_periodic();
               while(RTIMER_CLOCK_LT(RTIMER_NOW(),
                                     wt + CSMA_AFTER_ACK_DETECTED_WAIT_TIME)) {
+                watchdog_periodic();
 #if CONTIKI_TARGET_COOJA
                 simProcessRunValue = 1;
                 cooja_mt_yield();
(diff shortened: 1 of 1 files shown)


atiselsts:
Interesting... This pull request certainly cannot hurt. But under what conditions would you need so large waiting times for ACKs?

And in my opinion, the second `while` loop should also have the `!NETSTACK_RADIO.pending_packet()` condition added.

simonduq:
Hey, looks like we've been working on the exact same issue! Proposed https://github.com/contiki-ng/contiki-ng/pull/562/files just a couple hours earlier. The proposed CSMA enhancement include exactly what is in this PR (+what @atiselsts suggests). What are the odds...

MartenBE:
@atiselsts This was mainly caused by an abundance of logging, which causes an significant impact on processing time on the Zolertia (I needed it to figure some stuff out) and some overhead due to interrupts in the cc1200 driver.
At first I've copied the whole receiving/pending/channelclear condition from the if-statement that follows, but the CCA inserted some additional issues and only "pending" seemed enough for me as after receiving returns true, pending will follow. I've set the after ack wait time to 0 as I didn't saw its purpose, so It could be it could also use the extra condition.

@simonduq I ran into this issue last week doing some kind of stresstesting with CoAP and was looking for a clue in the cc1200 driver as to why some of my packets were not showing up on wireshark and some ACKs were 0.8 s late. Only the last days I've started to check the CSMA side. Feel free to discard this PR in favor of your own :)

simonduq:
So, now with https://github.com/contiki-ng/contiki-ng/pull/562 merged I believe this one can be closed :)
Thanks!

PR #569, opened 2018-06-05
CI: fail test in case building Docker image failed

simonduq:
Currently, when building the docker image fails, the normal test will continue anyway. This PR aborts the test early whenever the image failed to build.


g-oikonomou:
I think we're good!

PR #566, opened 2018-06-04
Travis Docker push fix

simonduq:
As it turns out, https://github.com/contiki-ng/contiki-ng/pull/554, failed to push the new Dockerfile to Dockerhub. No real harm, in that this does not break anything, but let's get this fixed.

The issue was in the test for changes in the `tools/docker` directory, done as follows:
```
git diff --name-only $TRAVIS_BRANCH...HEAD -- tools/docker | wc -l
```

This worked fine in PRs, but failed to detect the Dockerfile change on the merge CI run.

This PR fixes the issue by using this instead:
```
git diff --name-only $TRAVIS_COMMIT_RANGE -- tools/docker | wc -l
```

# Test

I tested this on my own repo, as follows:
* Do a PR that changes the Dockerfile: https://travis-ci.org/simonduq/contiki-ng/builds/387999221
The PR test issues:
```
Docker image changed, build from Dockerfile
```
* Merge the PR: https://travis-ci.org/simonduq/contiki-ng/builds/388000449
The merge test issues:
```
Docker image changed, build from Dockerfile
This build is for an update of branch develop. Push image to Dockerhub
```
* Do a PR that changes just a readme file: https://travis-ci.org/simonduq/contiki-ng/builds/388000632
The second PR test issues:
```
Docker image unchanged, pull from Dockerhub
```
* Merge the second PR: https://travis-ci.org/simonduq/contiki-ng/builds/388001120
The second merge test issues:
```
Docker image unchanged, pull from Dockerhub
```

All as intended.

PR #570, opened 2018-06-05
Travis: make DOCKER_CHANGED work in both the PR and branch push cases

simonduq:
So... https://github.com/contiki-ng/contiki-ng/pull/566/files, as extensively tested as it was, only worked for branch push tests, not PRs. This PR is an attempt to correctly detect changes to the Dockerfile in both the PR and non-PR cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,12 @@ notifications:
 # Environment setup before test script. Runs for each build
 before_install:
   # Check if anything has changed within the docker directory
-  - DOCKER_CHANGED=`git diff --name-only $TRAVIS_COMMIT_RANGE -- tools/docker | wc -l`
+  - >
+    if [ $TRAVIS_PULL_REQUEST == true ]; then
+      DOCKER_CHANGED=`git diff --name-only $TRAVIS_BRANCH...HEAD -- tools/docker | wc -l`
+    else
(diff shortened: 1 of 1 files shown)


simonduq:
Further clarification: `TRAVIS_COMMIT_RANGE` actually works on PRs, but we suspect it doesn't when a PR gets updated with new commits. The Travis doc is not 100% clear, but looks to me like `TRAVIS_COMMIT_RANGE` should work in all cases: https://docs.travis-ci.com/user/environment-variables/

@g-oikonomou are you 100% sure there was a case where you updated https://github.com/contiki-ng/contiki-ng/pull/547 but the Docker image was not built?

simonduq:
Note that the logs can be confusing at times: when building the image from Dockerfile, one of the first messages is `Pulling from 32bit/ubuntu`, because of the `FROM` command which pulls a generic base image.

g-oikonomou:
> @g-oikonomou are you 100% sure there was a case where you updated #547 but the Docker image was not built?

No, that was not the case. What happened was, I was testing a totally separate branch (not the PR one) on g-oikonomou/contiki-ng. And for that branch, addition of a commit did not trigger a docker image build.

simonduq:
Got it, this PR should not be needed then. Thanks!

PR #571, opened 2018-06-05
added EFR32 port - including I2C hal + its implementation on EFR32

joakimeriksson:
This is a Contiki-NG port for EFR32 and Thunderboard Sense2 which is basically SiLabs version of TI:s SensorTag. 

It has basic support for radio, LEDs (legacy), buttons (legacy), I2C and UART so it can run most of the Contiki-NG software. The radio is not yet complete for running TSCH but that should be fairly easy to fix in a near future.


simonduq:
I didn't follow closely the discussion on the HAL but adding a new platform with legacy button/LED interface doesn't feel right. If the existing HAL just does not work for the EFR32, then we should discuss how to improve it.

The PR needs CI test (check how we do it for jn516x, a platform which also involves proprietary binaries).
Also, we will need a wiki page like for other platforms, can be drafted on the sandbox.

joakimeriksson:
The GPIO hal is not a good match with the EFR32, but I can try to suggest some changes and maybe even try to implement the changed one on the EFR32.

CI is not very easy as the binaries can only be distributed via accepting a license in the Simplicity Studio IDE. There is no other way to get hold of them. It might be possible to develop a fake library with no code in just to be able to compile the code - I guess.

simonduq:
For jn516x we don't even need a fake lib, we just skip linking. Or maybe an alternative is to use Travis secure credentials to download the real libs from your server? Assuming this would comply with the license agreement, not sure... Or a web hook?

joakimeriksson:
Just compilation might work. Will check if we can do a submodule of the header files (and other C-sources).

joakimeriksson:
Anyone that can take a new look at this one - did some significant fixes during the Christmas holidays.

g-oikonomou:
Here are some first thoughts:

* If it works it works and new ports of course desirable
* Great to have an I2C API at last!
* Is this a cortex? We have gone through quite a bit of pain to harmonise the build system, make tagets and usage of dbg-io for all cortex CPUs. Suggest using here.
* For the PORT, MOTES etc, is tools/motelist unsuitable?
* Documentation is unacceptable (basically missing altogether)
* There appears to be a BMP sensor driver that is meant to go under arch/dev. It follows a totally different coding convention
* What is the rgb-leds thing? I don't suppose the board has the same part as the zoul: We have a driver for this under arch/dev/rgb-led

joakimeriksson:
The BMP driver is developed by Bosch and I am not sure how to do with externally produced code. Maybe we should have another ext/dev/... directory for that? (makes no sense to change their code-style).

g-oikonomou:
@joakimeriksson this seems to have stalled. Shall we factor out the I2C HAL and try to get that one in separately?

leoribg:
How is this implementation?

joakimeriksson:
Closing this as a new port is in a PR.

PR #572, opened 2018-06-06
Add guards for service-specific includes in contik-main

lafka:
`Auto-configuration of some services (like shell) require a `#include` in contiki-main.

This has the side effect of unintentionally including various other header files
that cause a conflict, or in worst case break compilation (in the case of enabling
LLSEC for CSMA which triggers and error in TSCH mac header file)

Check constants defined by build before including service specific headers.


lafka:
Needs to include `net/mac/tsch/tsch.h` when `MAC_CONF_WITH_TSCH` is set.

g-oikonomou:
I am not immediately convinced by this approach. It strikes me as though there may be something wrong with the included hears.

I have not followed the discussion on gitter, but I'd love to see a failure example here.

g-oikonomou:
I posted some comments on gitter. For the failure to build with `LLSEC_CONF_ENABLED` under CSMA builds, I suggest we move `#error` to `tsch-security.c`.

For the others, do we have a failure to build use-case?

simonduq:
@g-oikonomou 's proposal works for me

lafka:
Yup that would be fine, i'll change PR

simonduq:
@lafka just checking: this is ready, right?

PR #575, opened 2018-06-08
Dynamic shell command registration

angelos-oikonomopoulos:
Introduce shell command sets, as discussed on gitter.

### DIFF
--- a/os/services/shell/shell-commands.c
+++ b/os/services/shell/shell-commands.c
@@ -345,7 +345,7 @@ PT_THREAD(cmd_log(struct pt *pt, shell_output_func output, char *args))
 static
 PT_THREAD(cmd_help(struct pt *pt, shell_output_func output, char *args))
 {
-  struct shell_command_t *cmd_ptr;
+  const struct shell_command_t *cmd_ptr;
 
   PT_BEGIN(pt);
 
@@ -734,7 +734,7 @@ shell_commands_init(void)
                                     echo_reply_handler);
 }
 /*---------------------------------------------------------------------------*/
-struct shell_command_t shell_commands[] = {
+const struct shell_command_t shell_commands[] = {
   { "help",                 cmd_help,                 "'> help': Shows this help" },
   { "reboot",               cmd_reboot,               "'> reboot': Reboot the board by watchdog_reboot()" },
   { "ip-addr",              cmd_ipaddr,               "'> ip-addr': Shows all IPv6 addresses" },
--- a/os/services/shell/shell-commands.c
+++ b/os/services/shell/shell-commands.c
@@ -45,6 +45,7 @@
 #include "contiki.h"
 #include "shell.h"
 #include "shell-commands.h"
+#include "lib/list.h"
 #include "sys/log.h"
 #include "dev/watchdog.h"
 #include "net/ipv6/uip.h"
@@ -76,7 +77,8 @@ static uint16_t curr_ping_datalen;
 #if TSCH_WITH_SIXTOP
 static shell_command_6top_sub_cmd_t sixtop_sub_cmd = NULL;
 #endif /* TSCH_WITH_SIXTOP */
-
+static struct shell_command_set_t builtin_shell_command_set;
+LIST(shell_command_sets);
 /*---------------------------------------------------------------------------*/
 static const char *
 ds6_nbr_state_to_str(uint8_t state)
@@ -345,15 +347,16 @@ PT_THREAD(cmd_log(struct pt *pt, shell_output_func output, char *args))
 static
 PT_THREAD(cmd_help(struct pt *pt, shell_output_func output, char *args))
 {
-  const struct shell_command_t *cmd_ptr;
-
+  struct shell_command_set_t *set;
+  const struct shell_command_t *cmd;
   PT_BEGIN(pt);
 
   SHELL_OUTPUT(output, "Available commands:\n");
-  cmd_ptr = shell_commands;
-  while(cmd_ptr->name != NULL) {
-    SHELL_OUTPUT(output, "%s\n", cmd_ptr->help);
-    cmd_ptr++;
+  /* Note: we explicitly don't expend any code space to deal with shadowing */
+  for(set = list_head(shell_command_sets); set != NULL; set = list_item_next(set)) {
+    for(cmd = set->commands; cmd->name != NULL; ++cmd) {
+      SHELL_OUTPUT(output, "%s\n", cmd->help);
+    }
   }
 
   PT_END(pt);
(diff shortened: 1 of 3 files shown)


os/services/shell/shell-commands.c
```diff
@@ -734,7 +735,39 @@ shell_commands_init(void)
                                     echo_reply_handler);
 }
 /*---------------------------------------------------------------------------*/
-struct shell_command_t shell_commands[] = {
+void
+shell_command_set_register(struct shell_command_set_t *set)
+{
+  set->next = shell_command_sets_list;
+  shell_command_sets_list = set;
```
g-oikonomou: Please use the list library `os/lib/list.h`, gives you off-the-shelf singly linked list so you don't have to re-implement.

angelos-oikonomopoulos: Ah cool, thanks. Will update the PR.

os/services/shell/shell-commands.c
```diff
@@ -734,7 +735,39 @@ shell_commands_init(void)
...
+  for(curr = shell_command_sets_list->next;
+      curr != NULL;
+      prev = &curr->next, curr = curr->next) {
+    if(curr == set) {
+      *prev = curr->next;
+      set->next = NULL;
+    }
+  }
+  if(curr != set) {
+    printf("Could not deregister command set: not found");
```
simonduq: Here I would just return an error code. Up to the user how to signal the error.

g-oikonomou: and LOG_ERR perhaps

os/services/shell/shell-commands.c
```diff
@@ -734,7 +735,39 @@ shell_commands_init(void)
                                     echo_reply_handler);
 }
 /*---------------------------------------------------------------------------*/
-struct shell_command_t shell_commands[] = {
+void
+shell_command_set_register(struct shell_command_set_t *set)
+{
+  set->next = shell_command_sets_list;
```
simonduq: a sanity check for non-NULL pointer would be nice here

angelos-oikonomopoulos: What should the code be doing on NULL?

g-oikonomou: I think once you switch to the list lib this will become immaterial?

os/services/shell/shell-commands.c
g-oikonomou: Either log, or remove the printf. No strong preference

os/services/shell/shell-commands.c
simonduq: Right

os/services/shell/shell-commands.c
simonduq: IMO just return code is sufficient but feel free to add the log module

os/services/shell/shell-commands.c
angelos-oikonomopoulos: Umm. How does using list_push() make NULL immaterial?

angelos-oikonomopoulos:
Updated. Should address all comments - kindly let me know if I missed something.

simonduq:
It looks like this CI test sends serial input before the shell is initialized (the .js script writes `rpl-set-root` as soon as a node issues `Node ID`). Seems like a bug in the test. @yatch could confirm?

yatch:
@simonduq Yes, I will.

os/services/shell/shell.c
```diff
@@ -89,7 +89,7 @@ output_prompt(shell_output_func output)
 PT_THREAD(shell_input(struct pt *pt, shell_output_func output, const char *cmd))
 {
   static char *args;
-  static struct shell_command_t *cmd_ptr;
+  const struct shell_command_t *cmd_descr = NULL;
```
nfi: The `cmd_descr` variable is used by a protothread - it needs to be `static`.

yatch: It seems this causes the test failures; I've not understood what really happened in the tests of `09-ipv6`, though.
```
$ git log --oneline | head -n1
0e17ee444 Introduce dynamically registered shell command sets

$ git diff --ignore-submodules
diff --git a/os/services/shell/shell.c b/os/services/shell/shell.c
index 359038f3f..c219b1903 100644
--- a/os/services/shell/shell.c
+++ b/os/services/shell/shell.c
@@ -89,7 +89,7 @@ output_prompt(shell_output_func output)
 PT_THREAD(shell_input(struct pt *pt, shell_output_func output, const char *cmd))
 {
   static char *args;
-  const struct shell_command_t *cmd_descr = NULL;
+  static const struct shell_command_t *cmd_descr = NULL;

   PT_BEGIN(pt);


$ contiker make -C tests/09-ipv6
make: Entering directory '/home/user/contiki-ng/tests/09-ipv6'
Running test 01-ping-lla-csma-w-rpl with random Seed 1................................................ OK
Running test 02-ping-ula-csma-w-rpl with random Seed 1........................................ OK
Running test 03-ping-lla-tsch-w-rpl with random Seed 1................................................... OK
Running test 04-ping-ula-tsch-w-rpl with random Seed 1............................................. OK
Running test 05-ping-lla-csma-wo-rpl with random Seed 1.................................. OK
Running test 06-ping-ula-csma-wo-rpl with random Seed 1.................................. OK
Running test 07-ping-lla-tsch-wo-rpl with random Seed 1.................................... OK
Running test 08-ping-ula-tsch-wo-rpl with random Seed 1................................... OK
========== Summary ==========
05-ping-lla-csma-wo-rpl                  TEST OK      1/1
06-ping-ula-csma-wo-rpl                  TEST OK      1/1
07-ping-lla-tsch-wo-rpl                  TEST OK      1/1
08-ping-ula-tsch-wo-rpl                  TEST OK      1/1
make: Leaving directory '/home/user/contiki-ng/tests/09-ipv6'
```

yatch: OK; here is what happened in `01-ping-lla-csma-w-rpl`:

1. `ping` command is issued on node-1
1. `cmd_ping` waits at [L261](https://github.com/angelos-oikonomopoulos/contiki-ng/blob/0e17ee444740420e1d072426612c26cc3cc7a3b6/os/services/shell/shell-commands.c#L261) of of `shell-command.c` 
1. node-1 receives a ping response
1. the event starts being delivered to `cmd_ping` via `shell_input`
1. `shell_input` resumes its job at [L111](https://github.com/angelos-oikonomopoulos/contiki-ng/blob/0e17ee444740420e1d072426612c26cc3cc7a3b6/os/services/shell/shell.c#L111) of `shell.c`

At this moment, `cmd_descr` is `NULL` because of missing `static`. The null pointer reference causes `SIGSEGV`. The same should have happened in the other tests.

nfi: The 09-ipv6 ping test is calling the `ping` shell command which is an asynchronous command. After the initial setup of the ping command, the command protothread will be called again later (either due to ping response or timeout) but at that time `cmd_desc` contains whatever happens to be on the stack instead of the address to the command protothread. Making `cmd_desc` static should solve the problem.

nfi: Right, `cmd_desc` is set to `NULL` before `PT_BEGIN()` so it will always be `NULL` instead of whatever was on the stack when the protothread next executed.

angelos-oikonomopoulos: Thanks for tracking this down!

@yatch re: `4. the event starts being delivered to cmd_ping via shell_input` I don't think I'm following. How does the ping response involve `shell_input`?

AFAIU, the difference here is that `cmd_ping` is asynchronous, so `PT_SPAWN` will end up scheduling it again (i.e. calling it through `cmd_descr`, which will indeed be `NULL`  the second time we enter `shell_input`).

So, IIUC, the reason that I didn't observe that is I only tested with synchronous commands.

The main reason I'm asking is to understand if I should also use static in `cmd_help`. This depends on the semantics of the `shell_output_func`. Is it supposed to be synchronous across all platforms and configurations?

angelos-oikonomopoulos:
Updated with a static in `shell_input`.

os/services/shell/shell.c
yatch: @angelos-oikonomopoulos you don't need to make the variable `static` in `cmd_help`.

A simple rule about this is, if you wait for something in a `PT_THREAD`, you have a high chance to need make variables `static` there (of course, it depends on the context, though).  Otherwise, you don't need to do and you can save a little bit of static area of the memory.

Regarding `shell_input`, it looks like it doesn't have any wait in it; but in fact, `PT_WAIT_THREAD` is called as part of `PT_SPAWN`. 

It all happens under the magic of `pt` and `lc`. These implementation may be interesting to you.

angelos-oikonomopoulos:
Anything more I should take care of? If not, can this go in?

simonduq:
No, let's take this now! Thanks for the contrib :)

PR #577, opened 2018-06-08
CoAP: unified request_state for both blocking and callback APIs and add status codes 

carlosgp143:
This pull request unifies the `coap_request_state` structs into only one for both two APIs with all the common elements. Then, each API has its own extra elements, which are the process and protothread in the case of the blocking-API and the callback function for the callback API. 

Also, status codes are added to have more information about the request. Using these codes one can know if the request is timed out, if it has gotten a response, if there is a block error, etc. 

The LwM2M rd client is also updated to use these codes. Right now it can only check if the response is NULL or not in the callbacks, not being able to distinguish between a timeout, a block error or a finished request. This also allows to remove the `recover_from_delay` mechanism, since the callback will be called with a TIMEOUT code if something is wrong, and then the client can try to re-register.

### DIFF
--- a/os/net/app-layer/coap/coap-blocking-api.c
+++ b/os/net/app-layer/coap/coap-blocking-api.c
@@ -60,37 +60,36 @@
 void
 coap_blocking_request_callback(void *callback_data, coap_message_t *response)
 {
-  coap_request_state_t *state = (coap_request_state_t *)callback_data;
+  coap_blocking_request_state_t *blocking_state = (coap_blocking_request_state_t *)callback_data;
 
-  state->response = response;
-  process_poll(state->process);
+  blocking_state->state.response = response;
+  process_poll(blocking_state->process);
 }
 /*---------------------------------------------------------------------------*/
 PT_THREAD(coap_blocking_request
-          (coap_request_state_t *state, process_event_t ev,
+          (coap_blocking_request_state_t *blocking_state, process_event_t ev,
            coap_endpoint_t *remote_ep,
            coap_message_t *request,
            coap_blocking_response_handler_t request_callback))
 {
-  PT_BEGIN(&state->pt);
+  /* Before PT_BEGIN in order to not be a local variable in the PT_Thread and maintain it */
+  coap_request_state_t *state = &blocking_state->state;
 
-  static uint32_t res_block;
-  static uint8_t more;
-  static uint8_t block_error;
+  PT_BEGIN(&blocking_state->pt);
 
   state->block_num = 0;
   state->response = NULL;
-  state->process = PROCESS_CURRENT();
+  blocking_state->process = PROCESS_CURRENT();
 
-  more = 0;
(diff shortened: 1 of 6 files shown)


joakimeriksson:
Looking a bit more I think I can answer yes on my own question ;-)

carlosgp143:
Exactly! The problem before is that there were no feedback in the callback (like the status that it contains right now), and the rd client could only check if the response was null or not. But response was null in both response successfully finished or timeout. 

Right now, the rd client can just rely on CoAP retransmission system and try to re-register if  the request's time out expires. Also, the 5 seconds selected were a bit to fast to recover, taking into account CoAP retransmission parameters :-)

carlosgp143:
Hi!

I updated the code in the `lwm2m-rd-client` to solve a small problem. Before it was assuming that `coap_send_request(...)` will be successful all the time, and moving to the `..._SENT` state directly. This means that if there is no coap-transaction available, the rd-client would be stuck in that state all the time since no `coap-callback` will be called.

I changed it to check if the `coap_send_request(...)` is successful and then move to the `..._SENT` state,  otherwise remain in the same state to try again if there is a transaction available.

Thanks to @nfi for reporting this issue :-)

nfi:
I think this looks good and it solves multiple issues with the current implementation.

One think I am wondering about is the callbacks in lwm2m-rd-client. These callbacks seem to ignore some request statuses such as `COAP_REQUEST_STATUS_MORE` and `COAP_REQUEST_STATUS_BLOCK_ERROR`. What do you think about handling any unexpected status code as an error and restart the update/registration to ensure that lwm2m-rd-client never gets stuck?

carlosgp143:
You're right, it is just ignoring those two status codes related to blockwise transfers since the response to a registration or update message is supposed to contain just an response code, so it's never going to be a blockwise response. Therefore those two status should not happen for the rd client (if they happen it's probably a bug).

But yes, I agree that to be sure probably is better to include all of them. So your suggestion is to restart the rd client, like in the case of `COAP_REQUEST_STATUS_TIMEOUT` ?. I think it would be good to do that, but not strictly necessary :-)

nfi:
Well, the lwm2m-rd-client use blockwise transfer (block1) when registrering and for me it was not obvious at first that `COAP_REQUEST_STATUS_BLOCK_ERROR` only applied to block2 transfers ;-)

What I am wondering is - is there any risk that the callbacks could get a block-error status instead of timeout if for example lwm2m-rd-client would connect to a buggy CoAP server that includes a block2 option in the response by mistake? If not, then I agree that it might not be strictly necessary to check.

carlosgp143:
Now I'm more convinced with your example about the buggy server :-) And of course it does not heart to add the checks in order to be safer and not get stuck if bugs happen.

So I pushed a fix for this, to have a default case in the callbacks where all the unexpected status values are treated as an error and the rd-client is restarted.

PR #578, opened 2018-06-09
Nightly testbed runs

simonduq:
This adds the code for the nightly testbed runs, as well as the associated parsing script. This has been tried out over the past few weeks, with results shown at: http://simonduq.github.io/testbed/

The idea is to keep track of how performance evolves over time, try out sensitive PRs in a testbed, etc.

The test consists in the root pinging any random node at a one-second interval. Currently, this is run every nights in a private testbed at SICS, 25 nodes, 4h runs, twice a night, using different configurations. The code for the testbed infrastructure is available at: https://github.com/simonduq/iot-testbed


kkrentz:
This sounds and looks great. I hope that it will be possible to do a nightly run of CSL, too.

simonduq:
Yes of course, once we have CSL, we should definitely add it to the nightly!

g-oikonomou:
This is very desirable. I only had a quick look but cannot see anything wrong. @simonduq is this ready for review/merge?

simonduq:
Yes it's ready!

PR #581, opened 2018-06-15
Add partial support for plugtest-server for zoul-platform

twalcari:
The zolertia mote can support the plugtest-server when res_plugtest_large* and res_plugtest_obs are excluded

### DIFF
--- a/examples/coap/coap-plugtest-server/Makefile
+++ b/examples/coap/coap-plugtest-server/Makefile
@@ -2,7 +2,8 @@ CONTIKI_PROJECT = coap-plugtest-server
 all: $(CONTIKI_PROJECT)
 
 # Only intended for native
-PLATFORMS_ONLY = native
+PLATFORMS_ONLY = native, zoul
+
 
 # Include the CoAP implementation
 MODULES += os/net/app-layer/coap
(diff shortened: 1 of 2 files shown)


simonduq:
OK, but then, why just `zoul`?

twalcari:
I was only able to test it with that platform. I don't want to make any claims that I'm unable to verify..

g-oikonomou:
Strongly suggest not using `#if PLATFORM_THIS` in platform-independent code. We've been through plenty of pain to come up with examples that support platform-specific extensions, configuration etc, suggest adopting one of those design patterns.

Here I'd suggest using a platform-dependent header that declares those externs and then also a platform-independent source file with a function called whatever that calls all correct `coap_activate_resource`. We could even have two functions, a generic one that calls all `coap_activate_resource` that are common for all platforms and a second one that calls platform-specific optional ones.

g-oikonomou:
Thinking a bit more about this, the header is probably not necessary, should be poss to do with a single .c file per platform.

g-oikonomou:
@twalcari any updates here? This needs a bit more work before we can merge. Are you planning to work on it further?

twalcari:
I'm unsure on how to proceed. I understand that you want the code to support as many platforms as possible. By getting it working (partially) on the zoul-platform, I discovered that the only blocker was the limited memory available. I suspect that other platforms (excluding sky) all have similar memory limit? So in theory we can support all platforms with similar (or greater) memory limits.

Is there another example in this repo that I can use as a template on how I should structure the code?

g-oikonomou:
@twalcari thanks I made a post earlier in this thread with some recommendations. Other than that, you can also look at the generic MQTT example which has some platform-specific hooks, including for Zoul.

g-oikonomou:
Closing this due to inactivity.

PR #589, opened 2018-06-28
Fix broken rtimer for native

lafka:
`rtimer_arch_schedule/1`  would cast clock_second as `unsigned short` instead of `unsigned int` causing scheduling to set wrong interval.

Example:

```c
rtimer_set(&timer, RTIMER_NOW() + CLOCK_SECOND, 0, callback, null);

/* with the original code the following would be ouputted from `rtimer-arch.c`:
 * > rtimer_arch_schedule time 46487236 46466024 in 46466.24000 seconds
 *
 * With fix the following, corrected version:
 * > rtimer_arch_schedule time 46463604 1000 in 1.0 seconds
 */

### CODE REVIEW

**Comment by lafka** [comment]
The debug needs some work, since `time_t` is a `long int` on x64 and `int` on x86. 

Should i update to use `log` subsystem as well or you want it in a separate PR?

**Comment by simonduq** [comment]
Why do we need a cast there in the first place? Why is it not enough to work directly on `rtimer_clock_t`?
As for the log, my suggestion is to cast to long and print `%ld`.

**Comment by simonduq** [comment]
Why do we need a cast there in the first place? Why is it not enough to work directly on `rtimer_clock_t`?
As for the log, my suggestion is to cast to long and print `%ld`.

**Comment by lafka** [comment]
Removed cast and added correct formating for output. Compiles for launchpad and native platforms

**Comment by simonduq** [comment]
Now that the print format is `%ld` you need to cast the printf argument to `long`. Also, when I asked why we need a cast on line 80, that was a real question, not a request to remove the cast :p

**Comment by lafka** [comment]
The cast is not necessary. Maybe  the code was copy-pasted from a 16bit architecture.. 
 `itimerval` is a struct containing two `timeval`s (a struct with `time_t` and `suseconds_t`) both the fieldss are defined as `unsigned int`.
`rtimer_clock_t` is either a 16, 32, or 64bit unsigned integer depending on platform - for native platform  it will always be `uint32_t` or `uint64_t` (and never a short).

Why CI fails i don't know, i'll look into it as soon as i have time.

**Comment by g-oikonomou** [comment]
This just came to my attention. Seems pretty straightforward, right? The travis failure was random (passes now).

`rtimer_clock_t` is always `uint32_t` for native, so the logging line can be like so (`#include <inttypes.h>` needed):

```C
  PRINTF("rtimer_arch_schedule time %"PRIu32 " %"PRIu32 " in %ld.%ld seconds\n",
         t, c, (long)val.it_value.tv_sec, (long)val.it_value.tv_usec);
```

I'd also get rid of the `#ifndef _WIN32` by the way, but we can do that separately...
`

lafka:
Fixed, compiled fine in my case without including `inttypes.h`. Is there any convention i should follow? Its already included in `contiki-conf.h`, but i can add it in `rtimer-arch.h` as well

PR #603, opened 2018-07-12
Fix log output from uip_icmp6_send function.

debug-ito:
This patch just moves log message output from the beginning to the end of the function.

It prints `UIP_IP_BUF->destipaddr`, but this is updated in the middle of this function. Before this fix, the address was NOT the destination of the ICMPv6 packet, but was whatever destination that previous operation wrote to uip_buf.

### DIFF
--- a/os/net/ipv6/uip-icmp6.c
+++ b/os/net/ipv6/uip-icmp6.c
@@ -253,10 +253,6 @@ uip_icmp6_error_output(uint8_t type, uint8_t code, uint32_t param) {
 void
 uip_icmp6_send(const uip_ipaddr_t *dest, int type, int code, int payload_len)
 {
-  LOG_INFO("Sending ICMPv6 packet to ");
-  LOG_INFO_6ADDR(&UIP_IP_BUF->destipaddr);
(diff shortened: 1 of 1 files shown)


os/net/ipv6/uip-icmp6.c
```diff
@@ -253,10 +253,6 @@ uip_icmp6_error_output(uint8_t type, uint8_t code, uint32_t param) {
 void
 uip_icmp6_send(const uip_ipaddr_t *dest, int type, int code, int payload_len)
 {
-  LOG_INFO("Sending ICMPv6 packet to ");
-  LOG_INFO_6ADDR(&UIP_IP_BUF->destipaddr);
```
yatch: Why don't you replace ` &UIP_IP_BUF->destipaddr` with `dest` here instead of moving the three lines to the bottom?

debug-ito: There is no strong reason for that. Passing `dest` is ok for me, too.

However, I think printing `UIP_IP_BUF->destipaddr` (after being updated) is more true to this function's behavior observed from outside. Correctness of printing `dest` depends on the fact that the function copies it to `UIP_IP_BUF->destipaddr`.

PR #611, opened 2018-07-30
sixtop example: fix incorrect access to uint8_t *cell_list

ohrensessel:
In the cases where `read_cell()` is used to get a single cell from a `uint8_t *cell_list`, the index is incremented byte-wise where it should be incremented by the size of a `cell_list` entry


yatch:
@ohrensessel Thank you! It looks good. Have you tested the change to see if the code worked fine?

ohrensessel:
@yatch yes, I ran a cooja simulation with the fixes from this pr. Printing of the cell_list and adding/removing cells from the schedule now works fine (even if there is more than one cell to schedule).

yatch:
@ohrensessel Great!

PR #620, opened 2018-08-14
Remove gcc-8 warnings

Rehan-MALAK:
Dear all, 

As you may know, gcc-8 introduced a new Warning : stringop-truncation

https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8/

It's intended to detect misused of some string.h functions
This PR is what I currently do with gcc-8 to get ''make -C tests/'' and ''make -C tests/compile-all'' working.

I use :
`strncpy => memcpy` 

and

```c
strncpy(dest,source,sizeof(dest)-1) ;
dest[sizeof(dest)-1]='\0' ;
```

which are two possible ways of dealing with these warnings depending on the situation

for example 

https://github.com/tpetazzoni/libfaketime/commit/0d964363a4352b7ee358f523c3416d07378d4583

and

https://patchwork.kernel.org/patch/10311553/

Can someone have a look at examples/rpl-border-router/webserver/httpd-simple.c since I don't use the webserver feature and I am not sure what is the correct thing to do here. (in my case sizeof(s->filename)=2)

### DIFF
--- a/tools/serial-io/tunslip6.c
+++ b/tools/serial-io/tunslip6.c
@@ -579,8 +579,10 @@ tun_alloc(char *dev, int tap)
    *        IFF_NO_PI - Do not provide packet information
    */
   ifr.ifr_flags = (tap ? IFF_TAP : IFF_TUN) | IFF_NO_PI;
-  if(*dev != 0)
-    strncpy(ifr.ifr_name, dev, IFNAMSIZ);
+  if(*dev != 0) {
+    strncpy(ifr.ifr_name, dev, sizeof(ifr.ifr_name)-1);
+    ifr.ifr_name[sizeof(ifr.ifr_name)-1] = '\0';
+  }
 
   if((err = ioctl(fd, TUNSETIFF, (void *) &ifr)) < 0 ) {
     close(fd);
@@ -806,10 +808,11 @@ main(int argc, char **argv)
 
     case 't':
       if(strncmp("/dev/", optarg, 5) == 0) {
-	strncpy(tundev, optarg + 5, sizeof(tundev));
+        strncpy(tundev, optarg + 5, sizeof(tundev)-1);
       } else {
-	strncpy(tundev, optarg, sizeof(tundev));
+        strncpy(tundev, optarg, sizeof(tundev)-1);
       }
+      tundev[sizeof(tundev)-1] = '\0';
       break;
 
     case 'a':
--- a/tools/serial-io/tunslip6.c
+++ b/tools/serial-io/tunslip6.c
@@ -795,9 +795,9 @@ main(int argc, char **argv)
 
     case 's':
       if(strncmp("/dev/", optarg, 5) == 0) {
-	siodev = optarg + 5;
+        siodev = optarg + 5;
       } else {
-	siodev = optarg;
+        siodev = optarg;
       }
       break;
 
--- a/examples/rpl-border-router/webserver/httpd-simple.c
+++ b/examples/rpl-border-router/webserver/httpd-simple.c
@@ -120,6 +120,7 @@ PT_THREAD(send_headers(struct httpd_state *s, const char *statushdr))
 /*---------------------------------------------------------------------------*/
 const char http_header_200[] = "HTTP/1.0 200 OK\r\nServer: Contiki/2.4 http://www.sics.se/contiki/\r\nConnection: close\r\n";
 const char http_header_404[] = "HTTP/1.0 404 Not found\r\nServer: Contiki/2.4 http://www.sics.se/contiki/\r\nConnection: close\r\n";
+const char http_notfound_html[] = "/notfound.html";
 static
 PT_THREAD(handle_output(struct httpd_state *s))
 {
@@ -128,7 +129,11 @@ PT_THREAD(handle_output(struct httpd_state *s))
   s->script = NULL;
   s->script = httpd_simple_get_script(&s->filename[1]);
   if(s->script == NULL) {
-    strncpy(s->filename, "/notfound.html", sizeof(s->filename));
+    /* Using a local variable in strncpy below avoids a gcc >= 8 stringop    */
+    /* truncation warning. See gcc.gnu.org/bugzilla/show_bug.cgi?id=87028    */
+    const char *notfound_html = http_notfound_html ;
+    strncpy(s->filename, notfound_html, sizeof(s->filename)-1);
+    s->filename[sizeof(s->filename)-1] = '\0';
     PT_WAIT_THREAD(&s->outputpt,
                    send_headers(s, http_header_404));
     PT_WAIT_THREAD(&s->outputpt,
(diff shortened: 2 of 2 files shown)


tools/serial-io/tunslip6.c
```diff
@@ -806,10 +808,11 @@ main(int argc, char **argv)
 
     case 't':
       if(strncmp("/dev/", optarg, 5) == 0) {
-	strncpy(tundev, optarg + 5, sizeof(tundev));
+	strncpy(tundev, optarg + 5, sizeof(tundev)-1);
       } else {
-	strncpy(tundev, optarg, sizeof(tundev));
+	strncpy(tundev, optarg, sizeof(tundev)-1);
       }
+      tundev[sizeof(tundev)-1] = '\0';
```
g-oikonomou: code style: indentation for all lines in this hunk

tools/serial-io/tunslip6.c
```diff
@@ -806,10 +808,11 @@ main(int argc, char **argv)
 
     case 't':
       if(strncmp("/dev/", optarg, 5) == 0) {
-	strncpy(tundev, optarg + 5, sizeof(tundev));
+	strncpy(tundev, optarg + 5, sizeof(tundev)-1);
       } else {
-	strncpy(tundev, optarg, sizeof(tundev));
+	strncpy(tundev, optarg, sizeof(tundev)-1);
```
g-oikonomou: Ditto mempcy

tools/serial-io/tunslip6.c
```diff
@@ -579,8 +579,10 @@ tun_alloc(char *dev, int tap)
    *        IFF_NO_PI - Do not provide packet information
    */
   ifr.ifr_flags = (tap ? IFF_TAP : IFF_TUN) | IFF_NO_PI;
-  if(*dev != 0)
-    strncpy(ifr.ifr_name, dev, IFNAMSIZ);
+  if(*dev != 0) {
+    strncpy(ifr.ifr_name, dev, sizeof(ifr.ifr_name)-1);
```
g-oikonomou: So why is that not a memcpy now? :)

g-oikonomou:
I was just wondering why the patch was replacing some but not all strncpys

Rehan-MALAK:
two different situations : 

___
**tools/serial-io/tunslip6.c** : destination size = copy size so if source size is big enough (for example with ``sudo ./tunslip6 -a 127.0.0.1 -t mybigbigbigtunnelname fd00::1/64``), we won't have our terminal '\0' in  ``ifr.ifr_ifrn.ifrn_name``

with ``strncpy of size n => strncpy or memcmp of size n-1 + '\0' at index n-1`` , gcc doesn't complain anymore and  ``ifr.ifr_ifrn.ifrn_name = "mybigbigbigtunn"`` is a valid string

___

**examples/rpl-border-router/webserver/httpd-simple.c** : 

there seem to be an explicit truncation of ``"/notfound.html"``, but we copy only ``"/n"`` without the terminating '\0'

so I choose strncpy => memcpy 

but I don't know this part of the code so I would feel more confortable if someone could have a look at it

simonduq:
Thanks, looks like we had a bug indeed. But I think your fix is not 100% safe either: you're always copying exactly `sizeof(s->filename)` to the destination buffer. There is risk that you copy too much, i.e. pieces of data that are beyond the source buffer. This could possibly leak internal state.

IMO the better fix would be to stick to `strncpy` and then set the null terminator.

Rehan-MALAK:
Stick to strncpy + '\0'-terminal character was my first suggestion ... but.... 

So let's go from the beginning : 

this is the gcc-8 warnings here for **examples/rpl-border-router/webserver/httpd-simple.c** : 

```
  CC        webserver/httpd-simple.c
webserver/httpd-simple.c: In function ‘handle_output’:
webserver/httpd-simple.c:131:5: error: ‘strncpy’ output truncated copying 2 bytes from a string of length 14 [-Werror=stringop-truncation]
     strncpy(s->filename, "/notfound.html", sizeof(s->filename));
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
webserver/httpd-simple.c: In function ‘handle_input’:
webserver/httpd-simple.c:173:5: error: ‘strncpy’ output truncated copying 2 bytes from a string of length 11 [-Werror=stringop-truncation]
     strncpy(s->filename, http_index_html, sizeof(s->filename));
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
and sizeof(s->filename) = 2 is defined in **examples/rpl-border-router/webserver/httpd-simple.h**

```c
/* The current internal border router webserver ignores the requested file name */
/* and needs no per-connection output buffer, so save some RAM */
#ifndef WEBSERVER_CONF_CFS_PATHLEN
#define HTTPD_PATHLEN 2
#else /* WEBSERVER_CONF_CFS_CONNS */
#define HTTPD_PATHLEN WEBSERVER_CONF_CFS_PATHLEN
#endif /* WEBSERVER_CONF_CFS_CONNS */

...

struct httpd_state {
...
  char filename[HTTPD_PATHLEN];
...
};
```

I tried to do the same transformation as in **tools/serial-io/tunslip6.c** :

strncpy of size n => strncpy of size n-1 AND '\0' at index n-1

but the warning remained : we are copying 1 byte instead of 2 bytes from a string of length 14 or 11...

After investigation, it turned out that with my GCC Debian 8.2.0-4 and GLIBC Debian GLIBC 2.27-5
and compiled with '' gcc -O2 -Wall -Werror -c hello.c -o hello.o'' I get :

```c
#include <stdio.h>
#include <string.h>
struct mystruct {
    char dest[2];
};
const char src[] = "12345";
void function(struct mystruct *s)
{
    strncpy(s->dest,src,sizeof(s->dest)-1);
    s->dest[sizeof(s->dest)-1]='\0';
}
```
=> WARNING

```c
#include <stdio.h>
#include <string.h>
struct mystruct {
    char dest[2];
};
void function(struct mystruct *s)
{
    const char src[] = "12345";
    strncpy(s->dest,src,sizeof(s->dest)-1);
    s->dest[sizeof(s->dest)-1]='\0';
}
```
=> NO WARNING

```c
#include <stdio.h>
#include <string.h>
struct mystruct {
    char dest[2];
};
void function(struct mystruct *s)
{
    strncpy(s->dest,"12345",sizeof(s->dest)-1);
    s->dest[sizeof(s->dest)-1]='\0';
}
```

=> WARNING

So finally, I did what I wanted from the beginning : strncpy everywhere with '\0'-terminal character but only this time, I use a local constant string as the source of the copy :

```c
const char not_found_html[] = "/notfound.html";
```

and

```c
const char http_index_html[] = "/index.html";
```

If this looks strange for you, maybe we can wait further investigation or an explanation from the gcc people.

simonduq:
Alright. I'm fine with your new fix, but could you please add a comment in the code for the rationale of having the string defined like this?

Rehan-MALAK:
We discussed a bit about this warning on https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87028

The gcc people are now working on a patch https://gcc.gnu.org/ml/gcc-patches/2018-08/msg01549.html

Rehan-MALAK:
The fix is not very pretty and the false positive is at gcc side. Newer compilers help to catch bugs for free but, in that case, even if the docker image get upgraded to ubuntu 18.04, the default compiler will be gcc 7 (ubuntu 18.10 may have gcc 8.1). 
What about we split the commit : keep the '\0' terminating character and keep in another branch the gcc 8 specific fix with the local variable so people can find it if they need it.

simonduq:
Yes this sounds great to me!

g-oikonomou:
@simonduq please flag an approval on this if/when you are happy with this so we can merge it

simonduq:
@Rehan-MALAK let us know when the PR is ready, i.e., only contains the `\0` fix but not the gcc 8 workaround

Rehan-MALAK:
new PR #651

Rehan-MALAK:
On https://gcc.gnu.org/ml/gcc-patches/2018-08/msg01691.html, it seems that the false positive could stay for a while... 

> Both come at a cost.  You know I'm usually declaring GCC to be an
> optimizing compiler
> and not a static analysis engine ;)  So I'm not too much convinced when seeing
> disabling/delaying folding here and there to catch some false
> negatives for -Wxyz

simonduq:
Thanks a lot! Feel free to open an issue for the gcc-8 warning

PR #621, opened 2018-08-14
Add option to validate dio before init dag

tdesmet:
This gives the developer the tools to validate a dio before using it to init the dag. If 0 is returned then the dag will not be initiated.

Main use case for this is preventing the device from joining certain networks.

**Some background:** 

We are creating a product that has a local server that the nodes in the network need to communicate with. The server has a whitelist of devices that are allowed to communicate with it. It is possible that more then one network is running (one in company a, the other in company b across the street for example). If the server tells the device it will not communicate with it because it is not whitelisted the device should leave the network and try another one. This validation function will allow us to control if the device should join the network or not, if the dag id is one we have already tried we return 0 and the dio will be skipped.

### DIFF
--- a/os/net/routing/rpl-lite/rpl-conf.h
+++ b/os/net/routing/rpl-lite/rpl-conf.h
@@ -191,6 +191,13 @@
 #define RPL_DEFAULT_LEAF_ONLY 0
 #endif
 
+/*
+ * Function used to validate dio before using it to init dag
+ */
+#ifdef RPL_CONF_VALIDATE_DIO_FUNC
+#define RPL_VALIDATE_DIO_FUNC RPL_CONF_VALIDATE_DIO_FUNC
+#endif
+
 /******************************************************************************/
 /********************************** Timing ************************************/
 /******************************************************************************/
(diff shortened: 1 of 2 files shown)


simonduq:
This is surely a useful feature :)
Was wondering, have you considered filtering by PAN-ID or MAC layer security? Are you with 6TiSCH or not?

tdesmet:
Since all devices leave the factory with the same firmware and in advance we do not know where they end up I do not see an option in using mac layer security. I would need a way to have the user enter the security key into the device. Mac layer security could help to filter out networks that are created by other products/manufacturers.

Pan id filtering could work if the device changes his pan id if he does not find a suitable network, this was my original idea. But with 65535 possible pan ids this could take a long time. If we would have the option to scan for pan id's then this would be a good option. (I think this is what zigbee does, scan first then select one of the found pan id's, or when starting a network scan first and then use a panid that is not in use). If we could implement a scan function that return's the found pan id's then I think this might be a better solution since most radio's filter on pan id I think, which would make it more efficient. I have no idea on where to start to implement this though.

simonduq:
If you're with 6TiSCH you might want to look into the zero-touch join IETF draft: https://datatracker.ietf.org/doc/draft-ietf-6tisch-dtsecurity-zerotouch-join

tdesmet:
Forgot to mention that we are using the cc1310, no tsch. Some people are
working on ports but have not seen anything public.


Op wo 15 aug. 2018 om 11:27 schreef Simon Duquennoy <
notifications@github.com>

PR #622, opened 2018-08-16
Simplify configuration of the CC13xx/CC26xx ROM bootloader

g-oikonomou:
To enable/disable the CC13xx/CC26xx ROM bootloader one needs to fiddle with a number of CCFG-related defines. Currently `board.h` does all the work, but it does not really need to. This commit moves most of the ccfg configuration to `ccxxware-conf.h` and now `board.h` only needs to specify the bootloader enable pin and corresponding level.

This basically reduces copy/pasting and hides from port developers things that they do not need to worry about.

Default behaviour remains unchanged and examples have been adjusted to reflect changes.


atiselsts:
I approve this idea. One thing I'm not sure about is the name change from `ROM_BOOTLOADER_ENABLE` to `CCXXWARE_CONF_ROM_BOOTLOADER_ENABLE`. I'm not happy that every new version of Contiki-NG requires some changes in user configuration files. We need to think more of backwards compatibility.

g-oikonomou:
Point taken. I'll change this PR such that the existing `ROM_BOOTLOADER_ENABLE` still works.

g-oikonomou:
Hi @atiselsts I am looking at your comment now. Are you mainly referring to example configuration files? I was thinking something along the lines of this in `cc13xx-cc26xx-conf.h`:

```
/* Backward compatibility */
#ifdef ROM_BOOTLOADER_ENABLE
#define CCXXWARE_CONF_ROM_BOOTLOADER_ENABLE ROM_BOOTLOADER_ENABLE
#endif

#ifndef CCXXWARE_CONF_ROM_BOOTLOADER_ENABLE
#define CCXXWARE_CONF_ROM_BOOTLOADER_ENABLE              1
#endif
```

Would you consider this satisfactory?

atiselsts:
Yes, I like that. A bit ugly but what can you expect from backwards compatibility...

g-oikonomou:
Pushed

PR #629, opened 2018-08-24
Set correct DNS record expiration time

arurke:
When receiving a DNS response, the TTL field is parsed and used to determine when a record should be flagged as expired. This parsing seems to assume the TTL is stored in an _uint8_t_ array, while it is actually in an _uint16_t_ array - and thus record expirations are incorrect. This PR fixes this.

Some extra scrutiny might be warranted here, since this code is  ~6 years old, and as far as I can tell it has always been like this... 🤔 

Disclaimer: I currently live in contiki - not tested on contiki-ng


arurke:
> you will need two ntohs instead of one ntohl

Indeed. Fixed and rebased.

PR #630, opened 2018-08-26
Fix bug in CoAP retransmission policy

samkumar:
I'm new to Contiki, but I think I found a bug in the CoAP implementation.

When sending a CON request for which no ACK is received, this is what I observe:

1. Packet is sent
2. Wait for ~3 s
3. Packet is retransmitted
4. Wait for ~6 s
5. Packet is retransmitted
6. Wait for ~12 s
7. Packet is retransmitted
8. Wait for ~24 s
9. Packet is retransmitted
10. Transaction times out

Between steps 9 and 10, I would expect the system to wait for ~48 s, but that does not happen. I think this is a bug.

This pull request proposes a simple fix.

### DIFF
--- a/os/net/app-layer/coap/coap-transactions.c
+++ b/os/net/app-layer/coap/coap-transactions.c
@@ -98,12 +98,11 @@ coap_send_transaction(coap_transaction_t *t)
 {
   LOG_DBG("Sending transaction %u\n", t->mid);
 
-  coap_sendto(&t->endpoint, t->message, t->message_len);
-
   if(COAP_TYPE_CON ==
      ((COAP_HEADER_TYPE_MASK & t->message[0]) >> COAP_HEADER_TYPE_POSITION)) {
-    if(t->retrans_counter < COAP_MAX_RETRANSMIT) {
+    if(t->retrans_counter <= COAP_MAX_RETRANSMIT) {
       /* not timed out yet */
+      coap_sendto(&t->endpoint, t->message, t->message_len);
       LOG_DBG("Keeping transaction %u\n", t->mid);
 
       if(t->retrans_counter == 0) {
@@ -138,6 +137,7 @@ coap_send_transaction(coap_transaction_t *t)
       }
     }
   } else {
+    coap_sendto(&t->endpoint, t->message, t->message_len);
     coap_clear_transaction(t);
   }
 }


joakimeriksson:
Quick question: if we clear the transaction immediately after sending the final packet - will we not throw out needed state for the final ACK packet? If it is confirmable we would like to remove the transaction on the ACK (or after a timeout on some max end-to-end ACK timeout)?

samkumar:
@joakimeriksson I agree. After retransmitting the packet for the final time, the system should not clear the transaction immediately. If it does so, it will lose the state needed to properly handle the ACK/response if the final retransmission succeeds.

This PR fixes the issue by adding an RTO backoff after the final retransmission.

samkumar:
@joakimeriksson If you were asking about the call to `coap_sendto` that I added on Line 140, right before a call to `coap_clear_transaction`, note the program will only take that branch for Non-confirmable messages. Confirmable messages will satisfy the predicate of the `if` statement on Line 101, and therefore, will never take that branch.

samkumar:
@nfi Thanks for confirming that this is an issue.

If you really think it's a good idea to use `COAP_RESPONSE_TIMEOUT_TICKS` for the last retransmission, I'm happy to update the pull request with that change.

However, I'd like to point out some reasons why that may not be the right thing to do. The CoAP specification, Section 4.2 (https://tools.ietf.org/html/rfc7252#section-4.2) simply states "If the retransmission counter is less than MAX_RETRANSMIT, the message is retransmitted, the retransmission counter is incremented, and the timeout is doubled." That language seems to indicate that the timeout should be doubled even on the last transmission.

Some protocols with exponential backoff (e.g., TCP) stop doubling the timeout after many retransmissions, but tend to clamp the timeout at a "maximum" value instead of resetting it to the original timeout before backoff (see https://github.com/freebsd/freebsd/blob/master/sys/netinet/tcp_timer.c#L239 for FreeBSD's backoff structure for TCP; note how the last few entries remain at 512. See also https://github.com/freebsd/freebsd/blob/master/sys/netinet/tcp_timer.h#L114, which caps the retransmit timeout to 64 seconds). This sort of strategy is something we could consider for CoAP as well.

But please let me know what you decide. I'm happy to update the pull request with whatever you think is best for Conitki's CoAP implementation. :)

nfi:
@samkumar The CoAP specification text does not separate between response timeout and retransmission timeout which I think might be suboptimal in some cases. But I agree the implementation by default should be as close as possible to the specification so for me it is fine to keep the pull request as is.

PR #636, opened 2018-08-31
CSMA: Allow transmission of packets with broadcast panid

lafka:
All outgoing packets with broadcast panid was discarded. This allows caller to tag any packet to be transmitted on the broadcast pan id (0xffff).

This is usefull for building logic for network discovery or for transmitting data to multiple PAN's.

### DIFF
--- a/os/net/ipv6/sicslowpan.c
+++ b/os/net/ipv6/sicslowpan.c
@@ -1551,6 +1551,7 @@ output(const linkaddr_t *localdest)
   int framer_hdrlen;
   int max_payload;
   int frag_needed;
+  uint16_t uipflags;
 
   /* The MAC address of the destination of the packet */
   linkaddr_t dest;
@@ -1562,6 +1563,8 @@ output(const linkaddr_t *localdest)
   /* reset packetbuf buffer */
   packetbuf_clear();
   packetbuf_ptr = packetbuf_dataptr();
+
+  uipflags = uipbuf_get_attr(UIPBUF_ATTR_FLAGS);
 
   if(callback) {
     /* call the attribution when the callback comes, but set attributes
@@ -1585,6 +1588,10 @@ output(const linkaddr_t *localdest)
   /* copy over the retransmission count from uipbuf attributes */
   packetbuf_set_attr(PACKETBUF_ATTR_MAX_MAC_TRANSMISSIONS,
                      uipbuf_get_attr(UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS));
+
+  if (UIPBUF_ATTR_FLAGS_FRAMER_BROADCAST & uipflags) {
+    packetbuf_set_attr(PACKETBUF_ATTR_BROADCAST_PAN, 1);
+  }
 
 /* Calculate NETSTACK_FRAMER's header length, that will be added in the NETSTACK_MAC */
   packetbuf_set_addr(PACKETBUF_ADDR_RECEIVER, &dest);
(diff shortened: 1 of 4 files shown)


os/net/mac/framer/framer-802154.c
```diff
@@ -92,6 +94,10 @@ create_frame(int do_create)
   framer_802154_setup_params(packetbuf_attr, packetbuf_holds_broadcast(),
                              &params);
 
+  params.dest_pid = 1 == broadcastpan
```
simonduq: shouldn't this be done in `framer_802154_setup_params` instead?

simonduq (commented):
would be good to have @joakimeriksson or @nfi double-check this PR

os/net/mac/framer/framer-802154.c
lafka: yeah makes more sense.

atiselsts:
Hi lafka, we have some questions:
 - what does the standard say about this?
 - (more for an internal discussion between maintainers) should we have the same for TSCH, assuming that standard allows this, for consistency?

lafka:
> * what does the standard say about this?

I'm not to familiar with the spec but from _IEEE Std 802.15.4‐2011_  (the only copy i currently have available) i read this as it being allowed :

_The Destination PAN Identifier field, when present, specifies the unique PAN identifier of the intended
recipient of the frame. A value of 0xffff in this field shall represent the broadcast PAN identifier, which shall be accepted as a valid PAN identifier by all devices currently listening to the channel._

I haven't seen anything that limits the use of broadcast destination panids. 

> * (more for an internal discussion between maintainers) should we have the same for TSCH, assuming that standard allows this, for consistency?

This code should work as it is with TSCH, i'm not familiar with TSCH so i can't say if it makes sense to include it or not.

lafka:
@simonduq do you remember if there's any particular reason for why devices with broadcast drops all outgoing frames? I don't see a reason for denying devices to operate with a broadcast panid.

Otherwise any other points pending from my side?

os/net/ipv6/sicslowpan.c
```diff
@@ -1586,6 +1589,10 @@ output(const linkaddr_t *localdest)
   packetbuf_set_attr(PACKETBUF_ATTR_MAX_MAC_TRANSMISSIONS,
                      uipbuf_get_attr(UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS));
 
+  if (UIPBUF_ATTR_FLAGS_FRAMER_BROADCAST & uipflags) {
```
simonduq: code style: no space between `if` and `(`

simonduq:
Not sure why 0xffff was discarded. Maybe some function used to return this to signal parsing error?
The PR does not compile, so that's one thing pending on your side :)

lafka:
> Not sure why 0xffff was discarded. Maybe some function used to return this to signal parsing error?

Should i remove it? That would allow someone to configure panid 0xFFFF and do all the communication on broadcast panid. Keeping it like it is now a packet would need to be tagged PACKETBUF_ATTR_BROADCAST_PAN.

I'm in favor changing the PR to at least remove the panid check.
Potentially one can also remove the use of  packetbuf/uip attrs but then we need a separate PR for allowing to set an arbitrary PANID for a packet (which has a whole set of problems, for instance how to handle ACKs coming from a different PAN).


> The PR does not compile, so that's one thing pending on your side :)

I'll merge it with develop, remove broadcast panid check and update.

simonduq:
Any update here?

lafka:
Hi,

Unfortunately no, I'll make it up to date as soon as possible.

simonduq:
Hi!

joakimeriksson:
This is an important feature, but we will likely make a slightly different version that would make use of the PHYSICAL / PAN ID in already existing uipbuf attributes and make the 6lowpan code respect that attribute.

PR #638, opened 2018-08-31
CC26xx RF: in polling mode, disable the CPE0 radio interrupt completely

atiselsts:
The polling mode on CC26xx is not behaving as it should. The code that puts the radio in poll mode disables CPE0 on `RX_FRAME_IRQ`. However, we still get loads of CPE0 interrupts on  `IRQ_LAST_FG_COMMAND_DONE` and `IRQ_LAST_COMMAND_DONE` (apparently they are enabled by default).

Now, the ISR looks at the flag status, sees that the `RX_FRAME_IRQ` flag is set, calls     `process_poll(&rf_core_process);` which feeds the packet to the MAC protocol. This is not what we want. At best, it's just more energy consumed, at worst, it could break TSCH which expects to receive packets by itself instead of receiving them directly from the radio driver.

The need for the CPE0 interrupt in the polling mode at all is not clear to me. So I disabled it, tested this change for a while, and it does not seem to break things.

### DIFF
--- a/arch/cpu/cc26xx-cc13xx/rf-core/rf-core.c
+++ b/arch/cpu/cc26xx-cc13xx/rf-core/rf-core.c
@@ -512,8 +512,12 @@ rf_core_setup_interrupts(void)
 void
 rf_core_cmd_done_en(bool fg)
 {
-  uint32_t irq = fg ? IRQ_LAST_FG_COMMAND_DONE : IRQ_LAST_COMMAND_DONE;
+  uint32_t irq = 0;
   const uint32_t enabled_irqs = rf_core_poll_mode ? ENABLED_IRQS_POLL_MODE : ENABLED_IRQS;
+
+  if(!rf_core_poll_mode) {
+    irq = fg ? IRQ_LAST_FG_COMMAND_DONE : IRQ_LAST_COMMAND_DONE;
+  }
 
   HWREG(RFC_DBELL_NONBUF_BASE + RFC_DBELL_O_RFCPEIFG) = enabled_irqs;
   HWREG(RFC_DBELL_NONBUF_BASE + RFC_DBELL_O_RFCPEIEN) = enabled_irqs | irq;


g-oikonomou:
The reason we have enabled `IRQ_LAST_FG_COMMAND_DONE` and `IRQ_LAST_COMMAND_DONE` is so we can put the CM3 to sleep while the RF is TXing (instead of a busy wait). I'll review the pull first chance.

g-oikonomou:
I am looking at this. It should work with what we have now, but I am not fully comfortable. It prevents us from using CPE0 for a different interrupt event in the future. I think what we should be doing is change `rf_core_cmd_done_en()` such that it does not enable the `LAST_COMMAND_DONE` `COMMAND_DONE` interrupt in poll mode.

```
void
rf_core_cmd_done_en(bool fg, bool poll_mode)
{
  uint32_t irq = 0;
  const uint32_t enabled_irqs = poll_mode ? ENABLED_IRQS_POLL_MODE : ENABLED_IRQS;

  if(poll_mode) {
    irq = fg ? IRQ_LAST_FG_COMMAND_DONE : IRQ_LAST_COMMAND_DONE;
  }

  HWREG(RFC_DBELL_NONBUF_BASE + RFC_DBELL_O_RFCPEIFG) = enabled_irqs;
  HWREG(RFC_DBELL_NONBUF_BASE + RFC_DBELL_O_RFCPEIEN) = enabled_irqs | irq;
}
```

What do you think?

atiselsts:
You mean `if(!poll_mode)`, right?

I agree, can make this change. I prefer to wait until #648 is merged because it changes the poll mode from a function parameter to a `rf_core_` variable.

g-oikonomou:
Hi Atis yes, this is what I meant. Happy for this one to get merged once adjusted. Do you still prefer to get #648 first?

atiselsts:
Let's see - the #648 should be ready soon.

atiselsts:
Updated and tested between two nodes.

PR #639, opened 2018-08-31
TSCH: improve the readability of add_link and remove_link messages

atiselsts:
This is just a small change that's pretty self explanatory.

### DIFF
--- a/os/net/mac/tsch/tsch-schedule.c
+++ b/os/net/mac/tsch/tsch-schedule.c
@@ -171,6 +171,45 @@ tsch_schedule_get_link_by_handle(uint16_t handle)
   return NULL;
 }
 /*---------------------------------------------------------------------------*/
+static const char *
+print_link_options(uint16_t link_options)
+{
+  static char buffer[20];
+  unsigned length;
+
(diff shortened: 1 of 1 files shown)


os/net/mac/tsch/tsch-schedule.c
```diff
@@ -260,8 +301,10 @@ tsch_schedule_remove_link(struct tsch_slotframe *slotframe, struct tsch_link *l)
       if(l == current_link) {
         current_link = NULL;
       }
-      LOG_INFO("remove_link %u %u %u %u ",
-             slotframe->handle, l->link_options, l->timeslot, l->channel_offset);
+      LOG_INFO("remove_link sf=%u opt=%s ts=%u ch=%u addr=",
+               slotframe->handle,
+               print_link_options(l->link_options),
```
simonduq: shall we add link type here as well?

atiselsts: Updated. I have no idea why it was not there in the original version...

simonduq: most prob was just a mistake :)

PR #645, opened 2018-09-07
Add qualifier const to ip addresses in uip-ds6-route methods

afonsomota:
While doing work with static routes I noticed that uip-ds6-route methods do not have the const qualifier in IP addresses being passed to or returned from them. Since this project uses const qualifiers in other methods (for instance uip-ds6-nbr), I thought the tendency should be maintained and made these alterations.

This necessity came from creating a route using as the next hop the IP address of a node from the neighbor table, which comes with the const qualifier and hence could not be used in uip_ds6_route_add.

Addresses being returned should also have the const qualifier, as they return pointers from their internal structures. Because of this, I also had to add the const qualifier in a few tcpip.c variables and methods.

Regards,
Afonso Oliveira


afonsomota:
Also had to update a file in test 15. Will run all the tests locally to assure all is ok, then commit the up to date test files.

os/net/ipv6/uip-ds6-route.c
```diff
@@ -133,8 +133,8 @@ assert_nbr_routes_list_sane(void)
 /*---------------------------------------------------------------------------*/
 #if UIP_DS6_NOTIFICATIONS
 static void
-call_route_callback(int event, uip_ipaddr_t *route,
-		    uip_ipaddr_t *nexthop)
+call_route_callback(int event, const uip_ipaddr_t *route,
+		    const uip_ipaddr_t *nexthop)
```
g-oikonomou: Could you fix the code style here? `const` should start under `int event` like so
```
call_route_callback(int event, const uip_ipaddr_t *route,
		    const uip_ipaddr_t *nexthop)
```

os/net/ipv6/uip-ds6-route.c
```diff
@@ -321,8 +321,8 @@ uip_ds6_route_lookup(uip_ipaddr_t *addr)
 }
 /*---------------------------------------------------------------------------*/
 uip_ds6_route_t *
-uip_ds6_route_add(uip_ipaddr_t *ipaddr, uint8_t length,
-		  uip_ipaddr_t *nexthop)
+uip_ds6_route_add(const uip_ipaddr_t *ipaddr, uint8_t length,
+		  const uip_ipaddr_t *nexthop)
```
g-oikonomou: Ditto

afonsomota:
I will correct identation.

As for the "why were there no warnings". It is because there are more functions that should receive the const qualifier from a design perspective. Namely "rpl_parent_get_ipaddr" from rpl-dag.c (classic at least) and "uip_ds6_nbr_ipaddr_from_lladdr" from uip-ds6-nbr, which is the one  "rpl_parent_get_ipaddr" uses to retrieve the IP address of a parent. I was using "uip_ds6_nbr_get_ipaddr". 

Since you did not look at it closely, keep in mind that I had to alter one test, by adding the const qualifier to a function used as a callback. As so, people that have a build using the same function header to use as a callback (call_route_callback) in their personal code would get compilation errors (forced from warnings) until they change it.

PR #649, opened 2018-09-09
Add link stat packet counters

atiselsts:
Another statistic that I'm finding useful in my projects.


simonduq:
Only comment is that it is partly redundant with the existing counters used for ETX. If you have a good idea how both could be somehow merged that's be great, for reduced complexity! Else the PR looks good to me.

simonduq:
ping @atiselsts

atiselsts:
I thought about merging with existing variables, but I don't see how. Fundamentally, their function looks different. The existing ones are more for short term link-performance estimation. The stats I'e added are for meant for long-term, potentially since start of the network, packet counters suitable for off-line analysis.

simonduq:
Fair enough

PR #653, opened 2018-09-11
Add Bluetooth 5 functionalities to CC2640R2

onefreebjy:
This commit adds Bluetooth 5 functionalities on the basis of current structure from BLEach.


g-oikonomou:
There are some doxygen problems here, we can merge this as soon as they have been fixed

PR #655, opened 2018-09-13
update uncrustify configuration

Rehan-MALAK:
uncrustify recently changed the name of one of its option
https://github.com/uncrustify/uncrustify/issues/1476


Rehan-MALAK:
the documentation https://github.com/contiki-ng/contiki-ng/wiki/Code-style says 

> If you wish, you can format all changed resources in your working tree automatically if the tools/code-style/uncrustify-changed.sh script is added as a Git pre-commit hook to your Git configuration.

but when I tried it, the helper script changed substantially lots of part of the code...

g-oikonomou:
I wonder if we should keep both for people using an older version. Will the old option generate an error with new versions?

Rehan-MALAK:
``Unknown symbol 'align_number_left'``

but uncrustify does not emit an error, ($?=0)

it probably just ignores the bad parameter and takes the default in https://github.com/uncrustify/uncrustify/blob/master/documentation/htdocs/default.cfg

```cfg
# Whether to right-align numbers.
align_number_right = false # false/true
```

PR #658, opened 2018-09-15
CC26xx/CC13xx: update the TSCH_CONF_ASSOCIATION_POLL_FREQUENCY parameter to make it associate faster

atiselsts:
The CC26xx/CC13xx nodes have problems joining TSCH networks. The usual way to solve it is by adding busywaiting, which is ugly.

Someone on Gitter suggested that increasing the polling frequency also works - thanks!


simonduq:
Maybe add a comment in the `.h` for the rationale behind this config?

atiselsts:
Updated.

PR #668, opened 2018-09-21
missing header inttypes.h

Rehan-MALAK:
most of the platforms have inttypes.h via the header contiki-conf.h 

but sky and simplelink do not include it by default

this PR adds inttypes.h when needed for these latter platforms...

### DIFF
--- a/examples/multicast/root.c
+++ b/examples/multicast/root.c
@@ -45,6 +45,7 @@
 #include "net/ipv6/multicast/uip-mcast6.h"
 
 #include <string.h>
+#include <inttypes.h>
 
 #define DEBUG DEBUG_PRINT
 #include "net/ipv6/uip-debug.h"
--- a/os/net/ipv6/uip-nd6.c
+++ b/os/net/ipv6/uip-nd6.c
@@ -69,6 +69,7 @@
  */
 
 #include <string.h>
+#include <inttypes.h>
 #include "net/ipv6/uip-icmp6.h"
 #include "net/ipv6/uip-nd6.h"
 #include "net/ipv6/uip-ds6.h"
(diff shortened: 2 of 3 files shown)


atiselsts:
I wondering if there is any reason to avoid including it in `contiki-conf.h` on these platforms?

Rehan-MALAK:
I don't know... I push another branch PR #669 in the case we prefer <inttypes.h> included in contiki-conf.h for platform sky and simplelink

g-oikonomou:
I am in favour of including inttypes in files that need inttypes, instead of using a big sledgehammer called contiki-conf.h to include things that other files forget to include :)

examples/multicast/root.c
```diff
@@ -45,6 +45,7 @@
 #include "net/ipv6/multicast/uip-mcast6.h"
 
 #include <string.h>
+#include <inttypes.h>
```
g-oikonomou: Why do we need this here? Is there a compile failure if we don't include this?

Rehan-MALAK:
example/multicast/root.c is using the the PRIx32 macro : 

```
$ make TARGET=simplelink BOARD=sensortag/cc1350 root.o
  CC        root.c
In file included from ../../os/net/ipv6/uip-debug.h:45:0,
                 from root.c:50:
root.c: In function 'multicast_send':
root.c:86:23: error: expected ')' before 'PRIx32'
   PRINTF(" (msg=0x%08"PRIx32")", uip_ntohl(*((uint32_t *)buf)));
                       ^
../../os/net/net-debug.h:66:28: note: in definition of macro 'PRINTF'
 #define PRINTF(...) printf(__VA_ARGS__)
                            ^~~~~~~~~~~
root.c:86:10: error: conversion lacks type at end of format [-Werror=format=]
   PRINTF(" (msg=0x%08"PRIx32")", uip_ntohl(*((uint32_t *)buf)));
          ^
../../os/net/net-debug.h:66:28: note: in definition of macro 'PRINTF'
 #define PRINTF(...) printf(__VA_ARGS__)
                            ^~~~~~~~~~~
make: *** [../../Makefile.include:355: root.o] Error 1
```

PR #669, opened 2018-09-21
add inttypes.h header to platform sky and simplelink

Rehan-MALAK:
this is an alternative of #668
if we prefer instead including <inttypes.h> in contiki-conf.h for platforms sky and simplelink (as it is the case for all others platforms)


Rehan-MALAK:
#668 seems to be the preferred solution so I close this PR

PR #671, opened 2018-09-23
Create platform-independent SPI select and deselect

g-oikonomou:
The SPI API requires arch-specific select and deselect implementations. This is not required, it is possible to provide platform-independent implementations for both. This PR does exactly that and removes the arch-specific ones.


g-oikonomou:
Shameless ping @fontikos

atiselsts:
Does look god to me. A review from Fontas would be better, ofc, but this is pretty simple change.

PR #672, opened 2018-09-23
Harmonise link rule for all cortex CPUs

g-oikonomou:
This attempts to use a single link rule for all current and future cortex-based ports. Let's see what Travis thinks...


arch/cpu/arm/cortex-m/cm4/Makefile.cm4
```diff
@@ -9,20 +9,10 @@ LDFLAGS += -Wl,-Map=$(CONTIKI_NG_PROJECT_MAP),--cref,--no-warn-mismatch
 
 OBJCOPY_FLAGS += --gap-fill 0xff
 
-### Build syscalls for newlib
-MODULES += os/lib/newlib
```
atiselsts: Why remove this?

g-oikonomou: I was getting link errors that I couldn't really be bothered with debugging :)

g-oikonomou:
```
$ make
using saved target 'nrf52dk'
SDK: /Users/cexgo/nrf-sdk
SoftDevice: /Users/cexgo/nrf-sdk/components/softdevice/s1xx_iot/s1xx-iot-prototype3_nrf52_softdevice.hex
  CC        ../../../../../os/lib/newlib/syscalls.c
  CC        ../resources/res-leds.c
  CC        coap-server.c
  LD        coap-server.elf
obj_nrf52dk/syscalls.o: In function `_sbrk':
/Users/cexgo/contiki-ng/examples/platform-specific/nrf52dk/coap-demo/coap-server/../../../../../os/lib/newlib/syscalls.c:82: undefined reference to `_eheap'
obj_nrf52dk/syscalls.o:(.data.heap_end.4576+0x0): undefined reference to `_heap'
make: *** [coap-server.elf] Error 1
rm coap-server.o
```

g-oikonomou: Any clues? Does not happen with all examples...

atiselsts: Yes, actually. When you look at the original PR https://github.com/contiki-os/contiki/pull/1088, the author also added `_heap` and `_eheap` symbols for cc2538, but not for other CPU. So, these could be added to the linker scripts in `arch/cpu/nrf52832/ld/` (and also maybe for CC26xx?).

g-oikonomou: Thanks Atis. Alternatively, maybe  we change _heap and _eheap to symbols that stock linker scripts actually provide?

g-oikonomou: @atiselsts I've updated this pull, put the MODULE back in for all cortexes and adjusted linker scripts where required.

arch/cpu/cc26xx-cc13xx/cc26xx.ld
```diff
@@ -96,6 +96,8 @@ SECTIONS
     /* These symbols are used by the stack check library. */
     _stack = .;
     _stack_origin = ORIGIN(SRAM) + LENGTH(SRAM);
+    _heap = ORIGIN(SRAM);
+    _eheap = ORIGIN(SRAM) + LENGTH(SRAM);
```
atiselsts: I'm afraid this opened yet another can of worms.

I disagree with this definition of `_heap`. It should be after the static data, not at the start of the memory.

I don't disagree with setting `_eheap` at the end of the RAM.

I suspect that `_heap` is equal to  `_stack` on all platforms (and all platforms that support the stack checker already define the latter, so the former can be added by `LDFLAGS += -Wl,--defsym=_heap=_stack`). Same for `_eheap` and `_stack_origin`.

g-oikonomou: Hm you are of course right. I'll update.

g-oikonomou: @atiselsts what do you think of

```
    /* These symbols are used by the stack check library. */
    _stack = .;
    _stack_origin = ORIGIN(SRAM) + LENGTH(SRAM);
    _heap = _stack;
    _eheap = _stack_origin;
```

Issue #517, opened 2018-05-14
Update Docker image to latest compilers


simonduq:
And make sure to remove `-Wno-unused-const-variable` from `Makefile.include` once this is done. See https://github.com/contiki-ng/contiki-ng/issues/514

simonduq:
Depends on https://github.com/contiki-ng/contiki-ng/issues/678

simonduq:
Postponing to 4.3

joakimeriksson:
Docker and compilers are updated.

PR #676, opened 2018-09-26
Add optimization flags for native builds

atiselsts:
Enables optimizations by default for native builds to be consistent with almost all other platforms (all ARM and msp430 based platforms).

If the performance of an algorithm needs to be compared on different platforms, then it's handy that the compilation flags are the same by default.


atiselsts:
Oh, that's a lot of errors. This appears to be compiler-specific  as I don't see this error locally (gcc 7.3.0).
Perhaps better to drop this PR.

simonduq:
https://github.com/contiki-ng/contiki-ng/issues/517 should help eventually :)
Meanwhile, maybe just don't add `-O2` in the default case, but keep the optional `Os`?

arch/cpu/native/Makefile.native
```diff
@@ -20,6 +20,25 @@ endif
...
+endif
+
+ifeq ($(NATIVE_CAN_OPTIIMIZE),1)
+  ifeq ($(SMALL),1)
+    CFLAGS += -Os
+  else
+    CFLAGS += -O2
+  endif
+endif
+
```
g-oikonomou: Suggest putting the gcc version retrieval inside the linux conditional below (it fails on macos). On macos setting NATIVE_CAN_OPTIIMIZE to 1 unconditionally should be OK

atiselsts: Are you sure about the latter?

I think it would be good to design a GCC_VERSION makefile variable that works on both. Does `gcc -dumpfullversion -dumpversion` work on mac? Is it actually a GCC version format issue or shell script issue?

g-oikonomou:
```
$ gcc -dumpfullversion -dumpversion
4.2.1
```
```
$ gcc --version
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 10.0.0 (clang-1000.10.44.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

Draw your own conclusions... :)

But I don't think we need to try to support mega old macos / XCode command line tool versions

atiselsts: How about this?

g-oikonomou: Yeah that looks good. Tested over here.

g-oikonomou: By the way the "Configured with:" line seems to be going to stderr, you may want to consider

```
GCC_IS_CLANG := $(shell gcc --version 2> /dev/null | grep clang)
```

g-oikonomou:
Looks like this is ready to merge @atiselsts ?

PR #680, opened 2018-09-28
RPL-Lite: fix DAG lifetime, was 60h, now 1h

### LINKED ISSUE #654: RPL_DAG_LIFETIME is set to 60 hours instead of one hour in rpl lite
rpl lite RPL_DAG_LIFETIME is supposed to be set to one hour `#define RPL_DAG_LIFETIME (60 * 60) /* one hour */` but the comments state that it is a multiple of the lifetime unit which is defined as `#define RPL_DEFAULT_LIFETIME_UNIT 60` this makes the RPL_DAG_LIFETIME 60 hours

simonduq:
Fixes #654


simonduq:
Just updated to actually use 8h rather than 1h. RPL-Lite is single-DAG and there is little point expiring DAGs quick.

PR #684, opened 2018-10-02
Improvements to cc2538-rf.c

kkrentz:
This collection of improvements to `cc2538-rf.c` paves the way for #593 .


kkrentz:
Apparently, the return value of `set_channel` should rather be `void` because the range checks therein are either performed prior to calling `set_channel` already or the result of `set_channel` is ignored anyway.

g-oikonomou:
Yeap. I'm with you

PR #685, opened 2018-10-02
Improve usage of CCxxwares and CC2640R2 SDK

g-oikonomou:
This pull makes a few improvements to our usage of CCxxwares and CC2640R2-SDK:

* First and foremost, we entirely remove `ti_lib_rom_XXX` macros and the corresponding `ti-lib-rom.h`, since (somewhat embarassingly) `ti_lib_XXX()` macros automatically select ROM implementations where available anyway.
* We change the build system to always pass the CPU family with `CFLAGS += -DCPU_FAMILY_CCxxxxx` (previously only done for CC13x0). While at it, we change the naming convention to match how TI currently refers to chips (to CC26x0 from CC26xx).
* With the family name now always as a define, we can remove those ugly `ThisLibraryIsFor_CC26x0R2_HaltIfViolated`. Where we need to differentiate among libraries, we do this with the `CPU_FAMILY_xxxx` macros.
* XXwares and CC2640R2-SDK have some very subtle differences (small number of macros rename, small number of functions removed). This pull changes the code such that we can use any of those libraries without `#if` conditionals (small number of such conditionals remains, but may be removed in the future if we upgrade XXwares to a more recent version).


arch/cpu/cc26xx-cc13xx/Makefile.cc13xx
```diff
@@ -2,6 +2,6 @@ TI_XXWARE_PATH = lib/cc13xxware
 
 CONTIKI_CPU_SOURCEFILES += smartrf-settings.c prop-mode.c prop-mode-tx-power.c
 
-CFLAGS += -DCPU_FAMILY_CC13XX=1
```
atiselsts: Hmm, is there a reason for the change (do we expect a new CC13 chip family)? I'm asking because just in the TSCH pull request I did use CPU_FAMILY_CC13XX in one place, this change means that it should be updated again.

g-oikonomou: There is now CC13x0 and CC13x2 / CC26x0 and CC26x2

g-oikonomou: I should add, happy to update this one here to reflect onto the TSCH PR which is kind of likely to get merged first?

atiselsts: The TSCH PR uses the `CPU_FAMILY_CC13XX` flag to determine whether the radio is running with the the slow bitrate (currently assumed to be 50 kbps, which is of course not the most flexible solution).  

Help me to understand - does this Makefile.cc13xx refer only to CC13x0 chips? Also, looks like CC13x2 still has the proprietary Sub-GHz radio.

So, perhaps leave the `-DCPU_FAMILY_CC13XX=1` line as is, and also add `-DCPU_FAMILY_CC13X0=1` for the CC13x0 in particular?

g-oikonomou: It is confusing yes. This CPU dir supports what is now called CC26x0 and CC13x0.

The `simplelink-cc13xx-cc26xx` dir supports all chips of the family, including the various new CC26x2 and CC13x2. The new ones being CM4s.

Happy to undo the diff or to apply your proposal.

PR #686, opened 2018-10-03
TSCH support for CC1200

simonduq:
This PR enables TSCH on CC1200, platform Zoul. A number of generic changes were required to make this happen, mostly w.r.t. configuration of TSCH parameters, on a platform with more than one radio. We now have more flexible configuration of timeslot timings, and PHY parameters that TSCH requires when computing a packet's air time.

A new CC1200 PHY mode is added: 4-GFSK at 1000 kbps. TSCH support is offered for both this new mode (5.8 ms slot) and the existing 50 kbps (31.5 ms slot), but not the 1.2 kbps. In 1.2 kbps, slots longer than one second are needed, and that is currently a problem as our timeslot timing elements currently are stored on 16 bit, i.e., max 65.535 ms.


arch/platform/zoul/contiki-conf.h
```diff
@@ -54,7 +54,72 @@
...
+/* Platform typedefs */
+typedef uint32_t clock_time_t;
+typedef uint32_t uip_stats_t;
+
+/*
+ * rtimer.h typedefs rtimer_clock_t as unsigned short. We need to define
+ * RTIMER_CLOCK_DIFF to override this
+ */
+typedef uint32_t rtimer_clock_t;
+#define RTIMER_CLOCK_DIFF(a, b)     ((int32_t)((a) - (b)))
```
g-oikonomou: I don't think those are needed here (everything from `RTIMER_CLOCK_DIFF` upwards). They all get configured elsewhere.

arch/dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
```diff
@@ -62,7 +63,53 @@
...
+#define CC1200_TSCH_DEFAULT_TS_TIMESLOT_LENGTH  \
+                                                  ( CC1200_TSCH_DEFAULT_TS_TX_OFFSET \
+                                                  + CC1200_TSCH_DEFAULT_TS_MAX_TX \
+                                                  + CC1200_TSCH_DEFAULT_TS_TX_ACK_DELAY \
+                                                  + CC1200_TSCH_DEFAULT_TS_MAX_ACK \
+                                                  + CC1200_TSCH_DEFAULT_SLACK_TIME \
+                                                  )
+
+/* TSCH timeslot timing (mircoseconds) */
+static uint16_t cc1200_50kbps_tsch_timing[tsch_ts_elements_count] = {
```
g-oikonomou: Maybe const?

arch/dev/cc1200/cc1200-868-4gfsk-1000kbps.c
```diff
@@ -0,0 +1,176 @@
...
+#define CC1200_TSCH_DEFAULT_TS_TIMESLOT_LENGTH  \
+                                                  ( CC1200_TSCH_DEFAULT_TS_TX_OFFSET \
+                                                  + CC1200_TSCH_DEFAULT_TS_MAX_TX \
+                                                  + CC1200_TSCH_DEFAULT_TS_TX_ACK_DELAY \
+                                                  + CC1200_TSCH_DEFAULT_TS_MAX_ACK \
+                                                  + CC1200_TSCH_DEFAULT_SLACK_TIME \
+                                                  )
+
+/* TSCH timeslot timing (in rtimer ticks) */
+static uint16_t cc1200_1000kbps_tsch_timing[tsch_ts_elements_count] = {
```
g-oikonomou: Maybe const?

arch/platform/sky/platform.c
```diff
@@ -63,6 +64,22 @@ extern int msp430_dco_required;
 #include "experiment-setup.h"
 #endif
 
+/* TSCH timeslot timing for platforms requiring 15ms slots */
+uint16_t tsch_timeslot_timing_us_15000[tsch_ts_elements_count] = {
```
g-oikonomou: Maybe const?

arch/platform/zoul/contiki-conf.h
simonduq: good catch, fixed!

arch/dev/cc1200/cc1200-802154g-863-870-fsk-50kbps.c
simonduq: yes, fixed

arch/dev/cc1200/cc1200-868-4gfsk-1000kbps.c
simonduq: yes, fixed

arch/platform/sky/platform.c
simonduq: yes, fixed

simonduq:
Great, thanks for the comment. Only thing I didn't address is the remark prototypes on `-conf.h` files. The alternative with conditional includes could result in a series of if/elseif with paltform-specific includes (e.g. the 15ms timing file which is under cc2420). Feels nicer to have the platform configure itself in a standalone manner, i.e., define what to use as timing and declare required symbols if needed (note that sky needs to do this, but cc1200 does not, as the timing template is taken from `struct cc1200_rf_cfg`.

g-oikonomou:
I am not sure I understand why it would lead to #if spaghetti. What I had in mind was something along the lines of:

```
#ifdef TSCH_CONF_ARCH_HDR_PATH
#include TSCH_CONF_ARCH_HDR_PATH
#endif /* TSCH_CONF_ARCH_HDR_PATH */
```

Then in sky's `sky-def.h` we can `#define TSCH_CONF_ARCH_HDR_PATH "path-to-new-header". We move the `extern const uint16_t tsch_timeslot_timing_us_15000[];` there. For zoul we use the same process to move function prototypes there. Would that not work?

My main concern is if we start moving function prototypes and variable declarations to `-conf.h` and `-def.h` files, they stop looking like configuration files and they start looking more like the huge sledgehammer that one can use to include things.

By the way I think some things in `zoul/contiki-conf.h` may conceptually be best placed in `cc2538-def.h` - are they things that we expect users to be allowed to fiddle with?

simonduq:
Definitely agree on keeping `-conf.h` files just focused on configuration, not prototype declaration. But I would find it nice as well to have the file stand-alone, i.e., include anything it happens to use, both symbols and types. After all, it chooses to set a particular configuration flag to a symbol, so isn't it's responsibility to make sure the symbol is defined?

What would you think about moving the symbol to a `.h` as you suggest, but including the `.h` directly in the `-conf.h`?

Also agree on the `-def.h` vs. `-conf.h` comment, will fix.

g-oikonomou:
> After all, it chooses to set a particular configuration flag to a symbol, so isn't it's responsibility to make sure the symbol is defined?

I thought both the configuration and the prototype would move out of the -conf.h with this approach, no?

g-oikonomou:
OK now I see what you mean. I think we have made our points to each other, have a think about it and yell when the design is final and then we can merge this :)

simonduq:
OK, so, working on this, I decided your initial proposal was actually the best trade-off :) This PR is now ready AFAIC.

PR #692, opened 2018-10-12
RPL-Lite: rework DAO sending/resending/refreshing logic

### LINKED ISSUE #498: RPL Lite: do not allow skipping sequence number
With the current `rpl_timers_schedule_dao`, we skip a seqno in case multiple parent switches happen within `RPL_DAO_DELAY` (4s). Would be neater to stick to sequential increase, so the root gets clues about potential DAO losses.

We might also want to rework `handle_dao_timer` entirely, the logic currently is a bit complex.

simonduq:
This PR reworks the DAO sending logic with the following benefits:
* ensures no DAO sequence number is skipped
* clearer with distinct functions for sending a new DAO and resending one
* better handling of refresh: now we schedule a refresh straight after receiving DAO-ACK. Before, we would have waited for the previous resend timer to expire before scheduling the refresh

Fixes #498

I wonder if we want this merged for 4.2 (as per the #498 milestone) or wait after the release to give this 6 months of testing before 4.3. I feel rather confident with the change and Travis is green. But it's easy to get small things wrong with such timer logic.


simonduq:
ping @joakimeriksson

Issue #538, opened 2018-05-23
GPIO API


joakimeriksson:
I am currently porting the GPIO API to the EFR32. But I discovered that the current GPIO hal is a bit hard to implement as there is potentially up to 160 GPIOs and my device happen to have 65 GPIOs.

My suggestion is to change the API slightly - to focus the API on pins and avoid the masks. If we do that we can allow any type of pin type including a struct type or introduce Port and Pin that seems very common to more or less all platforms. I would also like us to consider having slightly shorter names on the functions (if that is an option).

gpio_hal_write(pin, value)
gpio_hal_read(pin, *value)
...

Where pin is either just an int as now or for the EFR32 I would pick a struct like:
struct efr32pin {
  uint8_t port;
  uint8_t pin;
}

Then a typedef on top of that that will be used for the EFR32.

I will try to implement something for the EFR32 that feels easy to configure on the platform, and take a look at CC2538 drivers to see that this will not make things very hard on other platforms.

g-oikonomou:
I cannot see the reason why change the current pin representation from a simple number to a port.pin struct. Common perhaps, but not universal.

The problem is the masking. It applies to functions that manipulate multiple pins in one go (which we can probably live without) but also to interrupt handling logic. This latter we'll need to think about very carefully. This masking is also critical for the operating of the button HAL.

Perhaps a simple solution while you are porting is to allow those masks for some 64 of your pins and leave the rest to the platform code?

We should also think about the possibility of `typedef`ing `gpio_hal_pin_mask_t` to a struct pointer. This can be universal if we need it to be (makes `gpio_hal_event_handler` easier to managge, I think), or we can leave it to the platform but then we'll need more flexibility in the  code under `os/`

g-oikonomou:
Or those masks could simply be a `uint8_t` array and then the platform tells us how many elements...

joakimeriksson:
I have looked at the Zephyr OS API:s for drivers and specifically GPIO - it looks like:

 int gpio_pin_write(struct device *port, u32_t pin, u32_t value)
 int gpio_pin_read(struct device *port, u32_t pin, u32_t *value)
 int gpio_pin_configure(struct device *port, u32_t pin, int flags)

 int gpio_port_write(struct device *port, u32_t value)
 int gpio_port_read(struct device *port, u32_t *value)
 int gpio_port_configure(struct device *port, int flags)

 int gpio_add_callback(struct device *port, struct gpio_callback *callback)
[...]

PR #694, opened 2018-10-12
Extend the GPIO HAL to support a port/pin numbering scheme

g-oikonomou:
This pull changes the GPIO HAL such that it supports a port/pin numbering scheme for pins.

What I am proposing is having two modes, controlled through `GPIO_HAL_PORT_PIN_NUMBERING`:

* 0: Selects what we have currently, whereby pins are numbered sequentially from 0, up to possibly 63. Multiple pins can be manipulated in a single go using a 32-bit or 64-bit mask (mask size dependent on pin count).
* 1 (default): Selects a port/pin numbering. A port is represented by a `uint8_t` and each port can have up to 32 pins, each pin also represented by a `uint8_t`. Multiple pins _within the same port_ can still be manipulated in a single go using a 32-bit mask.

Depending on choice of mode, functions come in two flavours, `port_pin_` and `no_port_pin_`. For example:

```
void gpio_hal_arch_port_pin_set_input(gpio_hal_port_t port, gpio_hal_pin_t pin);
void gpio_hal_arch_no_port_pin_set_input(gpio_hal_pin_t pin);
```

Macros are used to automatically map to the correct one, eg
```
#if GPIO_HAL_PORT_PIN_NUMBERING
#define gpio_hal_arch_pin_set_input(port, pin) \
  gpio_hal_arch_port_pin_set_input(port, pin)
#else
#define gpio_hal_arch_pin_set_input(port, pin) \
  gpio_hal_arch_no_port_pin_set_input(pin)
#endif /* GPIO_HAL_PORT_PIN_NUMBERING */
```

I need to double check a couple of things and also this will need a rebase after #671, #683, #685, so I am flagging this as WIP for now.

Your comments please, @joakimeriksson hopefully this meets your requirements for your new port.


joakimeriksson:
@g-oikonomou  Yes, this would work for me on the EFR32. I will implement it and test it on the EFR - but from what I can see it is good!

joakimeriksson:
I have now tested this fully on EFR32 and this API is fine for me. Maybe in the long run we should go for port, pin style everywhere to avoid so many pre-processor ifs. (and make use of the NULLPORT if it is not used on a specific platform).

g-oikonomou:
OK great!

This needs some conflict fixes and also I need to have a quick look at the simplelink port to make sure it's OK

joakimeriksson:
Yes, but if we pull the EFR32 PR this will get in also. So not sure what is best?

g-oikonomou:
Happy to fix conflicts and merge separately as well as to merge at the same time as the EFR32 port. Depends how quickly the EFR32 can be ready to merge :)

g-oikonomou:
So, this is now all nicely rebased.

I have triple-checked against #571 and there are two differences:

* #571 adds `gpio_hal_arch_init(void)`. I suggest doing this as part of #571 since this is where it's needed
* This pull has an additional commit (768326002), but pulling it into #571 is trivial.

As discussed earlier, these changes here address previous concerns about the API for platforms with a large number of pins and there is no reason to delay this until the EFR32 port is ready.

In other words, please review and merge ;)

Fixes #538

PR #697, opened 2018-10-13
Make sure CONTIKI_BOARD_xyz and CONTIKI_BOARD_STRING are defined

g-oikonomou:
If `BOARD` comes from the command line or from `Makefile.target` then the `CONTIKI_BOARD_xyz` and `CONTIKI_BOARD_STRING` macros get defined correctly:

```
$ cat Makefile.target 
TARGET = srf06-cc26xx
BOARD = srf06/cc26xx

$ make V=1
using saved target 'srf06-cc26xx'
arm-none-eabi-gcc -DCONTIKI=1 -DCONTIKI_TARGET_SRF06_CC26XX=1 -DCONTIKI_TARGET_STRING=\"srf06-cc26xx\" -DCONTIKI_BOARD_SRF06_CC26XX=1 -DCONTIKI_BOARD_STRING=\"srf06/cc26xx\" -Wno-unused-const-variable [...]
```

But if we fall back to a default specified by the platform's Makefile, then those two macros do not get defined:
```
$ rm Makefile.target 
$ make V=1 TARGET=srf06-cc26xx
arm-none-eabi-gcc -DCONTIKI=1 -DCONTIKI_TARGET_SRF06_CC26XX=1 -DCONTIKI_TARGET_STRING=\"srf06-cc26xx\" -Wno-unused-const-variable [...]
```

This currently happens with all platforms that use `BOARD`s.

This pull request fixes the behaviour.


simonduq:
This is ready to merge, right?

PR #698, opened 2018-10-13
Change build output dir structure and filenames

### LINKED ISSUE #395: Platforms overwrite each other's build output
Let's take hello-world as an example, but this is the case for all our examples at the moment.

* `make TARGET=zoul` results in a `hello-world.bin` file.
* `make TARGET=nrf52dk` results in a different `hello-world.bin` file, silently overwriting the zoul one.

This poses a risk of a user programming one device with a firmware built for a different device. This is the case for multiple platforms, not just those two. This is the case for other build outputs too, e.g. the `.elf` and `.hex`.

Options:
* It's fine, ignore it
* Include the platform's name in the filename. Perhaps even the board's name too
* Create `.bin` (and friends) in a subdir

IMHO this is a bug, but I'll just flag it as discussion for now.

Thoughts?

g-oikonomou:
This pull proposes a change of our build system logic in terms of the naming and location of build output artifacts.

We move from the current `obj_$(TARGET)` to something like this:

```
└── build/
    ├── nrf52dk/
    │   └── obj/
    ├── sky/
    │   └── obj/
    ├── srf06-cc26xx/
    │   ├── launchpad/
    │   │   ├── cc1350/
    │   │   │   └── obj/
    │   │   └── cc2650/
    │   │       └── obj/
    │   └── sensortag/
    │       └── cc2650/
    │           └── obj/
    └── zoul/
        ├── firefly/
        │   └── obj/
        └── remote-reva/
            └── obj/
```

Output files are placed as follows:
* `project-goal.$(TARGET)` (e.g. `hello-world.sky`): Still placed in the example's top-level dir.
* Map file, contiki library, other linker output files (e.g. `.hex`, `.bin`) go to `build/$(TARGET)/[ $(BOARD) ]/`. We also change the naming convention of those files, which no longer need to feature the paltform name as part of the filename:
  * All map files: `project-goal.map` (e.g. `hello-world.map`)
  * Ditto all `.hex`, `.elf`, `.bin` files: `project-goal.bin`
* Object and dependency files will go to the corresponding `obj/` dir.

This approach has a number of advantages:
* Platforms won't overwrite each other's outputs - This fixes the problem reported in #395. Basically this pull implements @yatch 's suggestion, but on steroids.
* We no longer need to `make clean` between builds for two different boards.

The top-level `project-goal.$(TARGET)` _will_ in fact get over-written if we build for the same target but a separate board, but at least the original will still be somewhere under `build/`.

Other minor changes:
* We put under verbosity control: cp, mkdir, srec_cat, objcopy, objdump
* A little code duplication removal from the build system for arm devices.

ToDo: Fix the `.upload` target.

Feedback please!

Requires #672, #697 

Fixes #395


simonduq:
OMG this is so much wanted! 💯

g-oikonomou:
So this is pretty-much done except: Cooja is hard-coded to look for a couple of files in the `obj_cooja` directory. I can change this pull to keep putting those specific files in `obj_cooja`, or we can merge this pull and at the same time update the cooja submodule to contiki-ng/cooja#16

simonduq:
I vote for merging the Cooja PR, then adding the submodule update as a commit in this PR

lafka:
This is great!

I compile several different binaries from same source and I like giving them meaningful names so i don't 
 confuse myself to much (ie. leaf.hex or router.hex). It would be great to have an option for overriding `BUILD_DIR*` so i can separate the different types of firmware.


```make
SPECIAL = my-custom-firmware
BUILD_DIR_BOARD = $(BUILD_DIR_TARGET)/$(BOARD)/$(SPECIAL)

# Makefile with `BUILD_DIR_BOARD ?= ....`
include Makefile.include
```

lafka:
Yeah, as you pointed out specific versions for release or development is one reason to split it out. I care about two different scenarios: 

 1. Tagging a build with some information by using filename or path 
 2. Separating binaries for different configuration

The first scenario i use when testing something with versions of contiki. I want two binaries "my-project-4.1.bin" and "my-project-master.bin" which are compiled against different revisions of contiki.
To solve this i normally just copy the binary out and rename it to something useful.

I work with several different "configurations" of contiki all sharing the same source tree and makefiles. The difference between them are sometimes small like a different DIS interval, or large when switching to different routing layer and including several new modules (it only shares same contiki-ng source). Each of the "configurations" can also have different sub-configurations, i think we have around 10 different binaries produced from the same source tree (built using a custom makefile).


As a side note there might be people, like myself, who are interested in keeping the builddir inside a tmpfs to minimize ssd use.

---

The makefile example was to showcase how you could change the build directory by adding a new directory at the end of the path. Say I want to compile two different versions of my project, one with rpl-lite and one with rpl-classic. I can use something like the example below to separate the two builddirs:

```make
# ./Makefile
CONTIKI_PROJECT = my-project
CONTIKI = contiki-ng
TARGET ?= native

ifneq ($(ROUTING),lite)
  ifneq ($(ROUTING),classic)
    $(error "ROUTING was not set to classic or lite"))
  endif
endif

# append rpl-lite or rpl-classic to builddir
BUILD_DIR_BOARD = $(BUILD_DIR_TARGET)/$(BOARD)/rpl-$(ROUTING)

include $(CONTIKI)/Makefile.include
```

Now i should be able to call `make ROUTING=classic` or `make ROUTING=lite` and it will automatically separate the builds.

Hope it's clearer now :)

g-oikonomou:
Thanks for the input @lafka, very clear.

I was thinking that we can introduce a variable called `BUILD_DIR_CONFIG`, with an empty default value. The build system will then put things by default in:

```
build/$(TARGET)/[ $(BOARD) ]/[ $(BUILD_DIR_CONFIG) ]
```

In your Makefile you can then do something along the lines of:

```Makefile
ifneq($(BUILD_DIR_CONFIG),)
  -include $(BUILD_DIR_CONFIG)/Makefile
  ### or
  ### -include $(BUILD_DIR_CONFIG)/Makefile.$(BUILD_DIR_CONFIG)
endif
```

This would allow you to conditionally do all sorts of interesting things, like e.g. configure the main build system differently, include different `MODULES`, create a separate `-conf.h` file per such dir.

The diff is trivial. Do you think that would meet your requirements?

Overriding `BUILD_DIR_BOARD` would also be a possibility, but it would require changes to make sure `make clean` also deletes the overriden path.

Makefile.include
```diff
@@ -319,6 +324,7 @@ distclean:
 		echo Running: $(MAKE) TARGET=$$TARG clean; \
 		$(MAKE) TARGET=$$TARG clean; \
 	done
+	rmdir $(BUILD_DIR)
```
simonduq: prob want to make this conditional or use `rm -f`, else this errors when `$(BUILD_DIR)` is not found

simonduq:
Tried building for `native`,  I expected a copy of `hello-world.native` to exist under `build/native` but couldn't find it there. Not as intended, or?

Makefile.include
g-oikonomou: Yes, good idea. Will have to be -rf

g-oikonomou:
The `.$(TARGET)` files will still be placed in the example's top-level dir.

We can change this so those files can be placed in `build/$(TARGET)` as well as the top-level dir. The background here is that some platforms generate `example.$(TARGET)` directly at the link stage, whereas some others generate a `example.elf` which they they copy to `example.$(TARGET)`.

THoughts? IMHO they definitely have to be at least in the top-level dir

simonduq:
Having the linked binaries under `build/$(TARGET)` would be nice, to keep one binary readily available for each board for instance.

g-oikonomou:
The only embedded platform that does not have a programmable firmware under `build/` is actually sky :P 

Nevertheless, I see your point. Let me knock something together

g-oikonomou:
The platform build system always generates stuff under `build/...`. The top-level Makefile always assumes there will be a `build/.../example.$(TARGET)` and copies it to .

PR #699, opened 2018-10-14
Change platform zoul to use the new motelist

g-oikonomou:
This pull:

* Changes platform zoul to use the Contiki-NG motelist tool instead of the old zolertia-specific one. The latter gets deleted from the repo altogether.
* Removes the `motelist-zoul` target, which was broken. The zoul build system behaviour remains otherwise unchanged.
* Creates a new `motelist-all` make target, available by the build system when building for any of our embedded targets. Tested for CC13x0/CC26x0 and all CC2538s.
* While at it, a new top-level `Makefile.help` is created and we move the `help` and `usage` targets there.


g-oikonomou:
Yes this is ready. Let's merge this when travis passes.

PR #701, opened 2018-10-14
Update docker to use the latest arm-gcc

g-oikonomou:
Just a test here and also to check whether this is what we want? This is not the latest ver.

#517


g-oikonomou:
Ha! I knew there was a reason I had this branch lying around for that long!

simonduq:
So, this is why I'm so keen getting Cooja to work on 64-bit, so we can finally upgrade Docker and compilers inside. And bring Cooja Motes to OSX :)

Issue #704, opened 2018-10-15
It is not possible to compile the TSCH_MAC for Openmote-cc2538


izacarias:
I can't compile the example using TSCH for MAC layer.
- Target platform: openmote-cc2538
- Assuming a complete and functional toolchain running on Linux (Ubuntu 16.04)
- Assuming that you are at the contiki-ng root directory
- Code version: 
```
git show
commit b4a23467fadadcb613bf2dcea790f3939fddc2ff
Merge: c6b795d d85cece
```

1. Navigate to `/examples/hello-world`
2. Include the code `MAKE_MAC = MAKE_MAC_TSCH` to the Project Makefile
3. Run the `make` command: `make TARGET=openmote-cc2538 hello-world`

Expected result: successful compilation

Current result: recipe for target 'obj_openmote-cc2538/tsch-slot-operation.o' failed 

Error log: [contiki-bg-tsch-openmote-cc2538-errorlog.txt](https://github.com/contiki-ng/contiki-ng/files/2480414/contiki-bg-tsch-openmote-cc2538-errorlog.txt)

g-oikonomou:
Thank you, I can confirm this is a bug. Affects openmote-cc2538 as well as - as suspected - cc2538dk. As discussed on gitter, this lot need defined for the two platforms:

```
#define RADIO_PHY_OVERHEAD
#define RADIO_BYTE_AIR_TIME
#define RADIO_DELAY_BEFORE_TX
#define RADIO_DELAY_BEFORE_RX
#define RADIO_DELAY_BEFORE_DETECT
```

I think all can be constants and looks like a straightforward fix.

Also, this is a great reason why we need some more TSCH compile examples on travis :)

PR #706, opened 2018-10-16
Fix TSCH on cc2528dk and openmote-cc2538

simonduq:
Fixes #704

PR #707, opened 2018-10-16
Fix TSCH on cc2528dk and openmote-cc2538

izacarias:
Fixes issue #704


simonduq:
OK, thanks again for the bug report. A fix, together with non-regression tests, was merged in https://github.com/contiki-ng/contiki-ng/pull/706

PR #714, opened 2018-10-18
Safer uIP6 buffer handling

simonduq:
This PR proposes a general cleanup of how uIP6 buffer handling. This started as an attempt to simply make parsing of extension header safe (as pointed out by @nvt , several parsing functions would trigger buffer overflows or infinite loops, on malformed or unexpected input), and ended as a much broader refactoring, with many bugfixes, clearer, and safer code.

In a nutshell:
* `ext_hdr_len` is no longer variable during the parsing of a given packet. It now always contains the total length of all extension headers, and is almost exclusively used to address upper-layer payloads.
* Parsing of extension headers and options is now done from a pointer, rather than the former variable defines `UIP_EXT_BUF` and friends which included an implicit offset of `ext_hdr_len`. This often resulted in confusing code (e.g. backup `ext_hdr_len`, mess with it to access a given ext header, then restore it) and also bugs (e.g., add `ext_hdr_len` one too many times).
* Parsing of extension headers is now defensive, i.e., it should no longer result in buffer overflow or infinite loop when e.g. no upper-layer protocol is found or the chain of headers just points beyond the buffer.
* Lookup and looping on extension headers is not done centrally in the `uipbuf.[ch]` module -- this is where sanity checks are done.
* `UIP_LLH_LEN` was removed, as it is basically always 0 (except for a old sky tool), and no platform uses Ethernet as main interface. If we need the feature again one day, we should handle the link-layer header in the MAC module, just like we already do for 802.15.4 and BLE.

As another positive side effect, looks like we save about 250 bytes of ROM :)

I would definitely not include in 4.2 but would suggest to aim for inclusion right after the 4.2 branch, so that (1) not 1000 conflicts show up and (2) this gets 6 months of testing until next release.

My apologies in advance for any headache caused looking at the diff :p


g-oikonomou:
Definitely agree to exclude this from 4.2, but how very desirable a change this is!

simonduq:
Time to press the big green button!

PR #717, opened 2018-10-19
Add out of tree build tests

### LINKED ISSUE #482: Test out-of-tree builds on travis
Building from outside the tree has been something that we have always kinda wanted and half-supported. I think we should make a more conscious effort to support it. This may (or may not) involve bug fixing (I am thinking #432 here) and it would also be nice to test such out-of-tree builds on travis.

g-oikonomou:
This pull creates `$(HOME)/out-of-tree-tests` in the docker image. It then adds a job on travis that attempts to build a sequence of contiki-ng examples (currently only `hello-world`) from that location.

Locally I have tested two scenarios:
* contiki-ng is a submodule of the `out-of-tree-tests` repo
* contiki-ng is at an entirely different location.

Same behaviour. Marking as work in progress cause we'll have to change the Dockerfile to point to the out of tree repo's final location (presumably `contiki-ng/out-of-tree-tests`)

Closes #482


simonduq:
So, one thing that bothers me is that updates to the external repo might break the Coniki-NG build. What would you think about hardcoding somewhere the `out-of-tree-tests` commit we want to run the test against? A bit submodule-style

g-oikonomou:
I see what you mean, valid point. Perhaps even use a tag?

g-oikonomou:
I've fixed the tests to a specific version and made cloning the repo conditional

simonduq:
There is a conflict here that needs resolving

g-oikonomou:
Conflict fixed here

simonduq:
Base branch should be update to `release-4.2` right?

g-oikonomou:
Yes changed now. I didn't want to do this while there was an ongoing travis build

PR #727, opened 2018-10-22
Fix: Global ip-addr removed if root is rebooted, rejoins network, and then starts acting as root again

### LINKED ISSUE #619: Global ip-addr removed if root is rebooted, rejoins network, and then starts acting as root again
The root seems to have an issue when rebooted in rpl-classic, when other nodes are not rebooted.
The issue can be reproduces using examples/multicast (release/v4.1) in the Cooja simulator:

1. Delete line examples/multicast/root.c:114
2. Add 'MODULES += os/services/shell' to examples/multicast/Makefile
3. Create 3 sink.c nodes with mote ids 1-3.
4. Create 1 root.c node with mote id 4, which is the last mote id (IMPORTANT).
5. Place the motes in a line: root <-> sink <-> sink <-> sink.
6. Start the simulation, send via shell 'rpl-set-root 1' to root and wait for the network to build.
7. Send 'ip-addr' via shell, which should show:
    Node IPv6 addresses:
    -- fd00::204:4:4:4
    -- fe80::204:4:4:4
8. Delete the root mote.
9. Add the root mote again with same position and make sure it is still mote id 4.
10. Start the simulation and keep sending 'ip-addr' via shell until it prints:
    Node IPv6 addresses:
    -- fd00::204:4:4:4
    -- fe80::204:4:4:4
11. Send 'rpl-set-root 1' via shell, and then send 'ip-addr' via shell:
    Node IPv6 addresses:
    -- fe80::204:4:4:4

The address fd00::204:4:4:4 never returns to the ip-addr list, and it is also possible to keep sending 'rpl-set-root 1' via shell.

jeppenodgaard:
Added test to reproduce issue, and proposed fix in rpl-dag, where dag is only freed if dag_id is different.

Fixes #619


jeppenodgaard:
I have no idea why the new test fails in Travis. It does not fail on my PC. I've also cloned this pull request to another PC, and run the test with success.

simonduq:
This is precisely why we run the CI tests in Docker :) Avoids the "works on my machine" problem. Please take a look at the Docker instructions and try to get the test working in there. https://github.com/contiki-ng/contiki-ng/wiki/Docker

jeppenodgaard:
Thank you for the input :)
I run the test manually using Cooja in Docker:
`docker run --privileged --mount type=bind,source=/c/contiki-ng-fork,destination=/home/user/contiki-ng -e DISPLAY=10.0.1.1:0.0 -v /tmp/.X11-unix:/tmp/.X11-unix -v /dev/bus/usb:/dev/bus/usb -ti contiker/contiki-ng cooja`
It runs without errors.
However, I just tried:
`docker run --privileged --mount type=bind,source=/c/contiki-ng-fork,destination=/home/user/contiki-ng -e DISPLAY=10.0.1.1:0.0 -v /tmp/.X11-unix:/tmp/.X11-unix -v /dev/bus/usb:/dev/bus/usb -ti contiker/contiki-ng bash -c "make -C tests/15-rpl-classic 02-rpl-root-reboot-2.testlog"`
That fails exactly like Travis does.
Any idea what the difference between the two is?

simonduq:
No idea what might be going on. But at least now you can reproduce the error :)

jeppenodgaard:
The issue seemed to be that motes joins the rpl instance slower without the graphical interface. Increasing the hardcoded expected join time fixed the issue.

jeppenodgaard:
I don't think so, but I haven't tested it with any other examples. I've changed a bit in rpl-dag.c, so I only think it is tied to the rpl building. Do you have any other examples in mind I can use?

simonduq:
No I mean, I see your Travis test uses multicast. If you could do the same test on a no-multicast RPL example, then we'd be able to run the test not only for Classic but also Lite.

jeppenodgaard:
I ment, no I don't think it is tied to multicast but I'm not certain. And yes, I'm pretty sure it's possible to use another example :)

jeppenodgaard:
I've done the test with the hello-world example, but had to modify the Makefile because the test requires the shell module. Is there any way to add a module as make option? E.g. `make root.cooja TARGET=cooja MODULES += os/services/shell`?

simonduq:
No I don't think so, but you could use the example `libs/shell`, or put new example code directly in the test directory.

jeppenodgaard:
Test is now in tests/14-rpl-lite as well and libs/shell is used. The bug does not exist in rpl-lite it seems :)

simonduq:
Awesome, thanks a lot!

PR #729, opened 2018-10-23
tsch: skip re-synchronization when time source has been already changed

### LINKED ISSUE #712: tsch: re-synchronization can interfere with time source selection by RPL
In `keepalive_packet_sent` in `tsch.c`, the node tries to re-synchronize with `last_eb_nbr_addr` if it fails to receive ACK for the keepalive frame. This can overwrite the time source selected by RPL and set in `tsch_rpl_callback_parent_switch`. As a result, the time source becomes inconsistent with RPL preferred parent.

How can we avoid this problem? I suggest comparing the current time source (obtained by `tsch_queue_get_time_source`) with the KA target (obtained by `packetbuf_addr(PACKETBUF_ADDR_RECEIVER)`). If they are different addresses, it implies someone has already changed the time source, so we don't have to change it anymore.

debug-ito:
Fixes #712

Plus, I split out the re-synchronization as a separate function.


simonduq:
Fixes #712

PR #730, opened 2018-10-24
Unmodulated carrier transmission for cc2538 & cc2420

cperezpenichet:
Modified the cc2538 and cc2420 drivers to enable the radio 
test mode to emit unmodulated carriers (tones).

A carrier can be enabled using the normal radio driver
API:
NETSTACK_CONF_RADIO.set_value(
    RADIO_PARAM_POWER_MODE,
    RADIO_POWER_MODE_CARRIER_ON);

Once enabled, the carrier can be disabled like this:
NETSTACK_CONF_RADIO.set_value(
    RADIO_PARAM_POWER_MODE,
    RADIO_POWER_MODE_CARRIER_OFF);


arch/dev/cc2420/cc2420.c
```diff
@@ -1120,3 +1132,30 @@ set_send_on_cca(uint8_t enable)
...
 /*---------------------------------------------------------------------------*/
+/* Enable or disable radio test mode emmiting modulated or unmodulated 
+ * (carrier) signal. See datasheet page 55.
+*/
+static uint16_t prev_MDMCTRL1, prev_DACTST;
+
+static void
+set_test_mode(uint8_t enable, uint8_t modulated)
+{
+  if (enable) {
```
simonduq: Code style: no space between `if` and `(`

arch/cpu/cc2538/dev/cc2538-rf.c
```diff
@@ -394,6 +396,50 @@ get_sfd_timestamp(void)
...
+    was_on = (mode == RADIO_POWER_MODE_ON);
+    off();
+    prev_FRMCTRL0 = REG(RFCORE_XREG_FRMCTRL0);
+    // This constantly transmits random data
+    printf("FRMCTRL0: %08X\n", (unsigned int)prev_FRMCTRL0);
+    REG(RFCORE_XREG_FRMCTRL0) = 0x00000042; 
+    if (!modulated) {
+      prev_MDMTEST1 = REG(RFCORE_XREG_MDMTEST1);
+      printf("MDMTEST1: %08X\n", (unsigned int)prev_MDMTEST1);
+      // ...adding this we send an unmodulated carrier instead
```
simonduq: Code style: we only use `/* */` comments

arch/cpu/cc2538/dev/cc2538-rf.c
```diff
@@ -394,6 +396,50 @@ get_sfd_timestamp(void)
...
+  
+  if (enable) {
+    if (mode == RADIO_POWER_MODE_CARRIER_ON) {
+      return;
+    }
+    was_on = (mode == RADIO_POWER_MODE_ON);
+    off();
+    prev_FRMCTRL0 = REG(RFCORE_XREG_FRMCTRL0);
+    // This constantly transmits random data
+    printf("FRMCTRL0: %08X\n", (unsigned int)prev_FRMCTRL0);
```
simonduq: why not `PRINTF`?

simonduq:
Great, thanks. Will wait a bit and see if others have comments, and merge if not within a few days.

g-oikonomou:
The `was_on` logic in the cc2538 driver changes seems flawed? As far as I can see we are setting it in one if branch but then using it in the else branch?

cperezpenichet:
The idea was that, once we turn off the carrier, we return the radio to the state it was before the carrier started (either ON or OFF). For this `was_on` is keeping the previous state of the radio between invocations. 

Perhaps a better way is to define that the radio should always go to a predefined state (probably OFF)  when the carrier is turned off. That way there is no need for the `was_on` logic at all. Would that work better?

I will also make the cc2420 and the cc2538 behave in the same way once we agree on the best way to proceed. Right now they don't have a consistent behavior.

g-oikonomou:
Sorry yes, I'm with you now. I think reverting to the previous state is better than forcing a state.

cperezpenichet:
Ok, then I'll go ahead and implement a similar logic for the cc2420 to be consistent.

g-oikonomou:
Some code style fixes here and there and then it looks like we can merge this?

simonduq:
Waiting for your approval here @g-oikonomou :)

g-oikonomou:
Oh! Oops!

PR #732, opened 2018-10-26
Securing CSMA (with session keys)

kkrentz:
This PR adds link layer security to CSMA. In order to use it, it suffices to add `MODULES += os/net/security/akes` to the Makefile and `#include "net/mac/csma/csma-autoconf.h"` to the project-conf. Also, the default key should be replaced by #defining `AKES_SINGLE_CONF_KEY`.

This PR is meant as an alternative to #558. The main difference between the two is that this PR establishes session keys, whereas #558 uses static keys.


joakimeriksson:
So how do we merge the feature-set of both PR:s in the best way?

simonduq:
#558 already received several rounds of revision and the plan was to merge within days. It provides the foundation for link-layer security, with support for dynamic keys (though without a session key management protocol). So I think this PR should build on #558.

Another comment is that, as said before, AKES is not a standard mechanism. As such, *if* we want to take it in, I'd argue this has to be a `service`, not a core OS component in `net/security`.

kkrentz:
We could offer both options, a static network-wide key and session keys. When using a static network-wide key, it will, however, become necessary to persist frame counters and anti-replay data across reboots. At the end of the day, neither of these options is standardized by IEEE 802.15.4 - IEEE 802.15.4 declares an upper layer responsible for providing keys. If we want something standardized, OpenThread and ZigBee IP are two choices. Unfortunately, both seem to require persisting data acrosss reboots, as well.

joakimeriksson:
I would say reboots should be very uncommon in real deployments and as long as upper layers provide the security key that is currently valid I guess a complete "re-join" might be the thing needed at a reboot (unless messing with persistence).

firmwareguru:
It's not clear to me - how are the session keys generated?

kkrentz:
Using a configuration switch, one can choose between group session keys and pairwise session keys. Pairwise session keys are derived from (i) a predistributed shared key and (ii) two cryptographic random numbers, each of which is contributed by one of the communication partners. A node's group session key is randomly generated at start up with the CSPRNG added in #522 and confidentially conveyed to neighboring nodes by generating a temporary pairwise session key like just described.

PR #733, opened 2018-10-27
Fix to queuebuf initialization

kkrentz:
Presently, `queubuf_init` is usually called twice at start up, once by the MAC layer and once by `sicslowpan.c`. This gets problematic if the MAC layer immediately begins to use the queuebuf since the second invocation of `queubuf_init` deletes all queued frames. The solution proposed in this PR is to move the initialization of the queuebuf to `contikimac-main.c`.

### DIFF
--- a/arch/platform/nrf52dk/contiki-conf.h
+++ b/arch/platform/nrf52dk/contiki-conf.h
@@ -74,6 +74,10 @@
 
 /* Packet buffer */
 #define PACKETBUF_CONF_SIZE                     1280  /**< Required IPv6 MTU size */
+
+/* Queuebuf */
+#define QUEUEBUF_CONF_ENABLED                   0
+
 /** @} */
 
 /**
--- a/os/contiki-main.c
+++ b/os/contiki-main.c
@@ -48,6 +48,7 @@
 #include "sys/stack-check.h"
 #include "dev/watchdog.h"
 
+#include "net/queuebuf.h"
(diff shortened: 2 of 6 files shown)


g-oikonomou:
This seems a reasonable change to me. I am pretty certain that the current default for nrf52dk is without queuebuf and this is I think the reason why the test fails.

```
$ make viewconf
using saved target 'nrf52dk'
SDK: /Users/cexgo/nrf5_iot_sdk_3288530
SoftDevice: /Users/cexgo/nrf5_iot_sdk_3288530/components/softdevice/s1xx_iot/s1xx-iot-prototype3_nrf52_softdevice.hex
----------------- Make variables: --------------
##### "TARGET": ________________________________ nrf52dk
##### "BOARD": _________________________________ 
##### "MAKE_MAC": ______________________________ MAKE_MAC_OTHER
##### "MAKE_NET": ______________________________ MAKE_NET_IPV6
##### "MAKE_ROUTING": __________________________ MAKE_ROUTING_NULLROUTING
----------------- C variables: -----------------
[...]
##### "SICSLOWPAN_CONF_FRAG": __________________ == 0
##### "SICSLOWPAN_CONF_COMPRESSION": ___________ == SICSLOWPAN_COMPRESSION_IPHC
[...]
```

I think the second failed test is unrelated.

I'd suggest `#define QUEUEBUF_CONF_ENABLED 0` in the nrf52dk `contiki-conf.h`

PR #743, opened 2018-11-02
General improvements to CSMA

kkrentz:
- Currently, after a frame was sent, CSMA tries to find the pointer to the corresponding `packet_queue` based on the packetbuf attribute `MAC_SEQNO`. This may break when sequence numbers wrap around zero, although very unlikely. Anyway, I suggest adapting CSMA to hand over the pointer to the corresponding `packet_queue` to exclude this issue and to simplify matters.
- If software-based acknowlegement frames are enabled, the incoming frame is parsed twice. I suggest avoiding this by making use of the existing packetbuf attribute `MAC_ACK`.
- Finally, I would suggest discarding frames whose source address equals `linkaddr_node_addr`. Otherwise, a node may end up adding himself as a link layer neighbor (even when using the security mechanisms proposed in #558). The same issue also applies to TSCH.


simonduq:
Have you witnessed packets received with source address of the receiver? What the conditions that trigger this? Filtering sounds good. I wonder if this should be done at MAC or upper. If at MAC we need to make sure this also gets done for TSCH (in this PR, or else opening an issue). Will take a closer look at the rest PR!

os/net/mac/csma/csma.c
```diff
@@ -91,20 +87,11 @@ input_packet(void)
...
-         linkaddr_cmp((linkaddr_t *)&info154.dest_addr,
-                      &linkaddr_node_addr)) {
-        uint8_t ackdata[CSMA_ACK_LEN] = {0, 0, 0};
-
-        ackdata[0] = FRAME802154_ACKFRAME;
-        ackdata[1] = 0;
-        ackdata[2] = info154.seq;
-        NETSTACK_RADIO.send(ackdata, CSMA_ACK_LEN);
-      }
+    if(packetbuf_attr(PACKETBUF_ATTR_MAC_ACK)) {
```
simonduq: in the case where `if(packetbuf_datalen() == CSMA_ACK_LEN)`, isn't there a risk that the frame is not parsed yet, and the packetbuf attribute here is just a leftover from last in or out frame?

os/net/mac/csma/csma-output.c
```diff
@@ -388,30 +391,12 @@ tx_ok(struct packet_queue *q, struct neighbor_queue *n, int num_transmissions)
...
-      break;
-    }
-  }
-
-  if(q == NULL) {
-    LOG_WARN("packet sent: seqno %u not found\n",
-           packetbuf_attr(PACKETBUF_ATTR_MAC_SEQNO));
-    return;
-  } else if(q->ptr == NULL) {
+  if(q->ptr == NULL) {
```
simonduq: maybe re-add the check that `n` and `q` are non-null? (as a defensive programming measure)

os/net/mac/csma/csma.c
```diff
@@ -72,6 +72,8 @@ input_packet(void)
                                          &linkaddr_node_addr) &&
             !packetbuf_holds_broadcast()) {
     LOG_WARN("not for us\n");
+  } else if(linkaddr_cmp(packetbuf_addr(PACKETBUF_ADDR_SENDER), &linkaddr_node_addr)) {
+    LOG_WARN("frame from ourselves\n");
```
simonduq: should this be done here or upper in the stack? If done at MAC we must make sure all MACs also do it

kkrentz:
> Have you witnessed packets received with source address of the receiver? What the conditions that trigger this?

This may happen when somebody launches a replay attack.

> Filtering sounds good. I wonder if this should be done at MAC or upper. If at MAC we need to make sure this also gets done for TSCH (in this PR, or else opening an issue).

The MAC layer seems to be appropriate for me. To avoid duplicate code, I could move this check to `framer_802154`.

os/net/mac/csma/csma.c
kkrentz: No. That if is in the very last else branch, which is only executed if packetbuf_datalen() != CSMA_ACK_LEN and parsing was successful.

os/net/mac/csma/csma-output.c
kkrentz: Since this function is static, I think we can dispense with defensive programming measures here.

simonduq:
> The MAC layer seems to be appropriate for me. To avoid duplicate code, I could move this check to `framer_802154`.

OK let's stick to MAC then. Created an issue for TSCH https://github.com/contiki-ng/contiki-ng/issues/748

os/net/mac/csma/csma-output.c
simonduq: Right, and the current code very clearly ensures the function will not be called with `n` nor `q` set to `NULL`.

I'd personally still add the check for future code evolution, but that's up to you.

os/net/mac/csma/csma.c
simonduq: Got it, I was somehow looking at the wrong indent level. Agree.

simonduq:
So, my remaining questions are:
* Was this tested?
* Any clue what the `original_dataptr` thing was for?

kkrentz:
> * Was this tested?

Yes, but not together with the new security additions, yet.

> * Any clue what the `original_dataptr` thing was for?

That thing was used for parsing a second time prior to sending an acknowledgment frame.

PR #745, opened 2018-11-05
Shell: skip empty lines

simonduq:
With this PR, the shell will skip empty lines without issuing *"Command not found. Type 'help' for a list of commands"* first.


g-oikonomou:
Looks reasonable 👍

PR #752, opened 2018-11-10
Change the way of accessing the default v6 prefix

g-oikonomou:
This pull changes the method of accessing the default v6 prefix. Instead of always using `UIP_DS6_DEFAULT_PREFIX`, we store the default prefix in a variable. We subsequently retrieve the prefix by calling `uip_ds6_default_prefix()`.

`uip_ds6_init()` will set this variable to the value of `UIP_DS6_DEFAULT_PREFIX`, but only if the startup code has not set a different default prefix before `uip_ds6_init()` gets called.

This approach has the following benefits:

* It allows us to change the default prefix at run time.
* It allows the startup code to set a prefix different than the one specified by `UIP_DS6_DEFAULT_PREFIX`, which can be useful if the default prefix comes from a different source (e.g. the command line for native, or a configuration stored on a node's flash)
* In many places the current code assumes that the prefix contains 6 bytes of zeros (e.g. `FDxx::/64`). Changing to a different prefix (e.g. `FD00:ABCD::`) would need extensive code changes. This change here makes it easy to use a prefix of any length.


g-oikonomou:
I've reworked this a bit and edited the OP

os/net/ipv6/uip-ds6.c
```diff
@@ -95,10 +95,29 @@ static uip_ds6_prefix_t *locprefix;
 static const uint8_t iid_prefix[] = { 0x00, 0x00 , 0x00 , 0xff , 0xfe , 0x00 };
 #endif /* (UIP_LLADDR_LEN == 2) */
 
+/* The default prefix */
+static uip_ip6addr_t default_prefix = {
+    .u16 = { 0, 0, 0, 0, 0, 0, 0, 0 }
+};
+/*---------------------------------------------------------------------------*/
+const uip_ip6addr_t *
+uip_ds6_default_prefix()
```
simonduq: code style: missing `void`

simonduq:
Shall we take this one in?

g-oikonomou:
I think this one's fine!

PR #755, opened 2018-11-11
Improve command line parsing for platform native

g-oikonomou:
This pull request tidies-up, improves and extends command line parsing for platform native.

* We provide a way to configure the function used to parse the command line. This allows us to use different command line options under different circumstances.
* We make sure that command line parsing takes place first thing in the boot sequence. Our current border router image parses the command line at the point in time when the `border_router_process` is started, which is too late. By parsing the command line early we can configure things before the boot sequence falls back to defaults, which is an improvement over having to re-configure things a posteriori. 
* We use the existing native border router command line parser for any example that includes this module.
* We introduce a different command line option set for all other IPv6/native examples.

The new command line option set for IPv6/native allow us to:

* Specify the name of the tunnel interface in a fashion identical to what the border router does. This allows us to run multiple native processes, each one bound to a different tunnel interface. Hopefully in the future they may even be able to talk to each other!
* Specify the v6 address of the tunnel interface, again in a fashion identical to what the border router does. While at it, we make sure the default v6 prefix used by the native image matches the prefix used by the tunnel interface.
* Specify the link layer address of the native image.
* Set verbosity control (logging level).

This is what it looks like:

```bash
$ ./hello-world.native -h
usage:  ./hello-world.native [options] [ipaddress]
example: ./hello-world.native -t tun1 -l 01:23:F:7:89:AB:CD:EF -v2 fd00::1/64
If ipaddress is omitted, the default is fd00::1/64
Options are:
 -t tundev      Name of interface (default tun0)
 -l address     Link layer address to assign to the node.
                Must be of a standard MAC address
                format with groups of two hex digits
                separated by colons (':'). The min
                required length depends on the value
                of LINKADDR_SIZE.
 -v level       Set logging level to level.
                Levels correspond to the values of LOG_LEVEL_ macros.
    -v0         Set debug level to LOG_LEVEL_NONE
    -v1         Set debug level to LOG_LEVEL_ERROR
    -v2         Set debug level to LOG_LEVEL_WARN
    -v3         Set debug level to LOG_LEVEL_INFO (default)
    -v4         Set debug level to LOG_LEVEL_DEBUG
```

This pull also adds a helper function that can be used to convert a link layer address from a string representation to a `linkaddr_t`.

Requires #752

### DIFF
--- a/arch/cpu/native/net/tun6-net.c
+++ b/arch/cpu/native/net/tun6-net.c
@@ -35,10 +35,12 @@
 
 #include "net/ipv6/uip.h"
 #include "net/ipv6/uip-ds6.h"
+#include "net/ipv6/uiplib.h"
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdarg.h>
 #include <string.h>
+#include <stdbool.h>
 #include <sys/time.h>
 #include <sys/types.h>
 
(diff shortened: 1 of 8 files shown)


tdesmet:
Something you might want to consider is the option to pass in application specific parameters.
Now when an unknown parameter is found the application refuses to start.

g-oikonomou:
Yes, we need to do that. Do you have a specific use case that works now, but fails with this pull?

Until we can come up with a way for applications to extend the arguments, perhaps we should default to starting instead of failing?

tdesmet:
I had a use case where I needed to pass a key. Because the arguments where parsed in border_router_process I was able to "intercept" them and get the key. Then I would remove remove the key from the argument list so it would not fail during parsing.

I think not failing when finding an unknown argument would be a good start yes.

g-oikonomou:
Yep, noted.

I'll change this to work in progress, clearly not ready yet

simonduq:
Hi!

joakimeriksson:
Parts of this is already available and this is still work in progress. Closing.

PR #763, opened 2018-11-14
Allow configurable linker script

lafka:
Make it possible to overwrite `LDSCRIPT` from other Makefiles

Previously only possible to override linker script using from environment. This allows user to set linker script in their project makefile.


arch/cpu/cc2538/Makefile.cc2538
```diff
@@ -1,7 +1,7 @@
 ifndef SOURCE_LDSCRIPT
 SOURCE_LDSCRIPT = $(CONTIKI_CPU)/cc2538.lds
 endif
-LDSCRIPT = $(OBJECTDIR)/cc2538.ld
+LDSCRIPT ?= $(OBJECTDIR)/cc2538.ld
```
g-oikonomou: This is not required and in fact could potentially cause harm. For this CPU the user can override already using `SOURCE_LDSCRIPT` above.

Issue #608, opened 2018-07-23
RPL clique reconnection problem


ianno:
I'm observing an issue which prevents nodes to reconnect to the network under a particular circumstance. @manoj6932 has reported a similar issue on gitter.
Here is how I think the problem can be replicated:
 
I started from a fresh installation of Contiki-NG v 4.1 (although the problem seems happening also under the develop branch). I'm using the rpl-udp example, modified to use TSCH and Orchestra with default parameters. I'm using 3 ReMOTEs rev b. Then follow the steps:

1. start with
Node A: coordinator, visible by node B
Node B: visible by both A and C
Node C: visible by node B.

2. Network starts, nodes are transmitting and everything is cool. Coordinator shows the routes: B->A; C->B->A.

3. Move B close to C, and make sure that A does not see any other node.

4. After ~5/10 minutes, move node C where B was. This means that:
A sees C
C sees both A and B
B sees C

5. Here is the problem, now the network does not recover, meaning that neither C or B are able to establish a route to A.
Also, moving B closer to A (so that all nodes are visible to each other ~~, as it can be verified by the shell command 'ip-nbr'~~) does not solve the problem.
B and C keep printing: ```dis_input: not in an instance yet, discard```

6. Switching off either B or C, makes the other node disconnect from the network and eventually connect back to A (after a bunch of ```re-synchronizing on B``` in case B is off, printed by C)

On Cooja, same scenario, everything works fine.

It seems that B and C keep synchronizing to each other, ignoring A...

manoj6932:
The Problem seem to happened by RPL_CONF_DEFAULT_LIFETIME_UNIT and RPL_CONF_DEFAULT_LIFETIME in rpl_conf.h . Adjust the value as per your requirement. [Page 51](https://tools.ietf.org/html/rfc6550) **DODAG Configuration**.

ianno:
@manoj6932, unfortunately it seems the problem is not the RPL_DEFAULT_LIFETIME constant. The default value for it is 30  *RPL_DEFAULT_LIFETIME_UNIT = 1800 seconds. In this case, instead, B and C stay disconnected from A way after that limit. Indeed, after that point A does not even list them in its routing table, yet the network does not recover.

ianno:
@yatch I think that what you wrote explains why everything works fine in Cooja. 
What I mean by "A sees C", I mean that the two nodes are close enough to be able to communicate directly. No packets are exchanged though.

[...]

PR #769, opened 2018-11-22
implementation section 6.3 in RFC 8180

tsakoda:
This PR is the title.

Section 6.3 in RFC 8180 delays start sending EB by the RPL node joined network.
When I confirmed the codes, Contiki-NG has not still implemented section 6.3 in RFC 8180.


tsakoda:
This PR relates with https://tools.ietf.org/html/rfc8180#section-6.3

This PR advantage does not made strange TSCH PAN loop.
```
  [A]          [B]             [C]
 coord        router           node
```
For example, 3 nodes set as above.
A starts sending EB, B and C join TSCH PAN.
```
  [A]  <-----  [B]   <-----  [C]
 coord        router         node
```
However, B disconnects with A when B and C don't join RPL DAG, following sitiuation occurs.

```
  [A]          [B]   <-----> [C]
 coord        router         node
```
In this case, if the link between B and C is fine (RSSI is good value), B cannot connect A.
So, B and C create TSCH PAN loop and cannot break it because B and C don't detect loop.

The suppress_sending-EB doesn't send EB until joining RPL DAG.
So, it suppresses above loop problem, and will be stable network.

On the other hand, this PR has disadvantage that some nodes take time to join TSCH PAN.

tsakoda:
This PR implements <https://tools.ietf.org/html/rfc8180#section-6.3>, i.e. it makes a node send EBs ONLY AFTER it has joined to the RPL DODAG.

The advantage of this PR is that it helps prevent strange TSCH PAN loops, which is the cause of #608.

## Example

(1) 3 nodes set as following.

```code
  [A]           [B]           [C]
 coord         router         node
```

A starts sending EB, B and C join TSCH PAN(<->) and RPL DAG(<=>).

```code
  [A]  <-----   [B]   <-----  [C]
 coord <=====  router <=====  node
```

(2) If A moves far away from B and C, B loses the default route to A and searches for the new parent.
However, B has no candidate parent, so B becomes RPL poison status.

(3) B sends DIO including rank infinity, C recieves it, and B and C becomes RPL leave status.

(4) Eventually, B leaves TSCH PAN because it cannot synchronize to A anymore. B starts scanning for EB.

(5) However, C is still sending EBs for a short time until it determines B has left TSCH PAN.

(6) If B catches EB from C, B synchronizes to C. So, B and C create TSCH PAN loop.

```code
  [A]                  [B]   <-----> [C]
 coord                router         node
```


## Effect of this PR

This PR solves the above problem. With this PR enabled, node C stops sending EBs at step (3) because it detects there is no candidate parent and leaves RPL DODAG. So, at step (5), B does not catch EB from C because C has already stopped sending it.

In general, we think this PR makes TSCH PAN more consistent to RPL DODAG.

## Downside of this PR

Possible downside of this PR is that nodes take a little longer time to join TSCH PAN, because EB transmission is delayed until RPL DODAG is properly formed.

os/net/mac/tsch/tsch.c
```diff
@@ -862,6 +862,12 @@ PROCESS_THREAD(tsch_send_eb_process, ev, data)
     unsigned long delay;
 
     if(tsch_is_associated && tsch_current_eb_period > 0) {
+#ifdef TSCH_RPL_CHECK_DODAG_JOINED
```
simonduq: why not just add this check to the if and remove the goto statement?

tsakoda: I agree your suggestion. I will fix it.

tsakoda:
@simonduq 
It's been more a  month since I revised code, but own code was not reviewed.
If you have any problems with my revising, please contact me.

simonduq:
Sorry, I missed the update. Thanks for fixing, now looks good to me!

simonduq:
The `rpl-tsch-z1` test seems to fail though, any clue what might cause this @atiselsts ?

atiselsts:
That looks like a problem with the `rpl-tsch-z1` script, sorry. See #848

simonduq:
Thanks @atiselsts

Issue #574, opened 2018-06-07
The Shell won't build for MAKE_NET_NULLNET


g-oikonomou:
It's clearly documented in the wiki that the shell is meant for "interactive IPv6 host inspection and maintenance.", but perhaps we should do something to prevent users from trying to include this in non-v6 firmwares.

Here is what happens with the shell example. For this particular example we can just change the Makefile to force `MAKE_NET_IPV6`, not sure how to handle other examples that simply include the module.

```
$ make MAKE_NET=MAKE_NET_NULLNET
using saved target 'zoul'
  CC        ../../../arch/cpu/cc2538/./ieee-addr.c
  CC        ../../../arch/cpu/cc2538/cc2538.lds
  CC        ../../../arch/cpu/cc2538/./startup-gcc.c
  CC        example.c
  LD        example.elf
obj_zoul/shell-commands.o: In function `cmd_ipaddr':
shell-commands.c:(.text.cmd_ipaddr+0x70): undefined reference to `uip_ds6_if'
obj_zoul/shell-commands.o: In function `cmd_routes':
shell-commands.c:(.text.cmd_routes+0x1e): undefined reference to `uip_ds6_defrt_choose'
shell-commands.c:(.text.cmd_routes+0x22): undefined reference to `uip_ds6_defrt_lookup'
shell-commands.c:(.text.cmd_routes+0x54): undefined reference to `uip_ds6_route_num_routes'
shell-commands.c:(.text.cmd_routes+0x5c): undefined reference to `uip_ds6_route_num_routes'
shell-commands.c:(.text.cmd_routes+0x70): undefined reference to `uip_ds6_route_head'
shell-commands.c:(.text.cmd_routes+0xc6): undefined reference to `uip_ds6_route_nexthop'
shell-commands.c:(.text.cmd_routes+0xe8): undefined reference to `uip_ds6_route_next'
obj_zoul/shell-commands.o: In function `cmd_ping':
shell-commands.c:(.text.cmd_ping+0x34): undefined reference to `uiplib_ip4addrconv'
shell-commands.c:(.text.cmd_ping+0xaa): undefined reference to `uip_icmp6_send'
obj_zoul/shell-commands.o: In function `cmd_ip_neighbors':
shell-commands.c:(.text.cmd_ip_neighbors+0x10): undefined reference to `uip_ds6_nbr_head'
[...]

PR #772, opened 2018-11-23
IPv6-less shell

simonduq:
This PR makes it possible to use the shell without IPv6, and adds a new command `mac-addr` to print the node's current MAC address.


os/services/shell/shell-commands.c
```diff
@@ -862,7 +888,9 @@ const struct shell_command_t builtin_shell_commands[] = {
   { "rpl-status",           cmd_rpl_status,           "'> rpl-status': Shows a summary of the current RPL state" },
   { "rpl-nbr",              cmd_rpl_nbr,              "'> rpl-nbr': Shows the RPL neighbor table" },
 #endif /* ROUTING_CONF_RPL_LITE */
+#if NETSTACK_CONF_WITH_IPV6
   { "routes",               cmd_routes,               "'> routes': Shows the route entries" },
```
g-oikonomou: Looks good overall. Should we move this to the `#if NETSTACK_CONF_WITH_IPV6`  block above so we can have them all in the same place?

simonduq: agree, will do!

simonduq: done

g-oikonomou:
This should be enough to fix #574 I think

simonduq:
Yes definitely!

PR #773, opened 2018-11-23
Build system: enable compiling with arch directory at arbitrary paths

simonduq:
This is a proposal to enable compiling with arbitrary arch paths, which is useful for platforms, dev, or cpu outside of the Contiki-NG tree. The PR introduces a variable `ARCH_PATH`, which by default points to `$(CONTIKI)/arch` but can be specified by the user so as to point to any location.

For instance:
```bash
make TARGET=<my-custom-target> ARCH_PATH=<path-to-my-custom-arch>
```

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -82,7 +82,10 @@ uniq = $(if $1,$(firstword $1) $(call uniq,$(filter-out $(firstword $1),$1)))
 
 ### Include target makefile (TODO Unsafe?)
 
-target_makefile := $(wildcard $(CONTIKI)/arch/platform/$(TARGET)/Makefile.$(TARGET) ${foreach TDIR, $(TARGETDIRS), $(TDIR)/$(TARGET)/Makefile.$(TARGET)})
+# Configurable arch path
+ARCH_PATH ?= $(CONTIKI)/arch
+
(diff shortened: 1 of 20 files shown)


g-oikonomou:
Will need rebased and updated to reflect more recent stuff. I think we should go ahead and merge this one and then I can look into the other one separately

g-oikonomou:
I'm having a closer look at this change. The way things work now if the user specifies `ARCH_PATH` then this will exclude everything under our current `arch`. For example, `make ARCH_PATH=<somewhere> distclean` will not clean for any of the targets under `arch/`. Likewise, if one is developing a platform under `<somewhere>` they will be unable to include e.g. the CC1200 driver under `arch/dev/cc1200`

Is this intended behaviour?

simonduq:
1. Agree that it would be nice not to exclude `os/arch`. Also useful for an external platform that would want to use of a core cpu/dev.
2. Well, isn't it neater to use `ARCH_PATH` everywhere, so that code can be copied outside of `os` as a basis for new platforms?

g-oikonomou:
Thanks lets continue this discussion under #785 if you get a chance

PR #775, opened 2018-11-25
Scale up the maximum number of CoAP Observers

carlosgp143:
Hi!

This PR tries to scale up the maximum number of simultaneous CoAP (and LwM2M) observers. Right now it is limited to the maximum number of CoAP transactions minus 1, since they all the observers try to get their own transaction from the available list when they notify. The maximum number of transactions is set to 4 by default due to memory usage, so therefore only a very few observers are possible.

In this PR I made a change so that all the observers share one transaction, or in other words, if one observer is notifying, all the rest wait until the notification is done, so at maximum one transaction is taken from the available list for notifying purposes. For doing that, I have separated the list of observers in two different ones, the unactive and the pending observers. When one observer wants to notify, it is placed in the pending list, and as soon as one observer is done and has freed the CoAP transaction, the next one from the list is taken.

Further optimization to allow more observers could be done in a future, since they probably use more memory than needed. But the main limiting aspect was the transaction usage, so that's why I changed that to begin with.


joakimeriksson:
I seems like a few platforms have problems with 20 observers as default. I changed to 5.

carlosgp143:
That is true, thanks a lot @joakimeriksson !

I was developing with the zoul platform and there you could have a lot more observers with no memory problems. But I guess this means that we need to optimize further the memory that each observer allocates for platforms with less available memory :-)

Apart from that, what do you think about this PR as a first step to scale up the maximum number of observers?

joakimeriksson:
I think this is a very good step towards that! We will try to move it in as soon as possible!

carlosgp143:
You're right! It should iterate both lists, the unactive and pending observers, just in case someone removes one observer while it's being notified.

I'll push a new commit repairing that and some other minor stuff soon! :-)

simonduq:
Hi!

PR #784, opened 2018-12-01
Fix the nrf52dk .jlink and .flash make targets

g-oikonomou:
#698 introduced some bugs to the nrf52dk build system. This pull fixes them.

* The `%.jlink` assumes a fixed location for the output hex file. This location is no longer valid, so we fix this.
* The new location of the hex file contains slashes in the path, so we change the sed delimiter to something other than the `/` character
* We include `Makefile.cm4` earlier. This allows us to use some Make variables defined therein.
* We improve rule dependencies: `%.jlink` and `%.flash`

Fixes #782

PR #785, opened 2018-12-02
Various build improvements and bug fixes

g-oikonomou:
Edit: Updated description here: https://github.com/contiki-ng/contiki-ng/pull/785#issuecomment-526945401

This pull proposes various improvements and bugfixes to the build system.

Most importantly, we start using make variables to hold various source tree paths. For example, we introduce something like this:

```
CONTIKI_NG_OS_DIR = os
CONTIKI_NG_NET_DIR = $(CONTIKI_NG_OS_DIR)/net
CONTIKI_NG_MAC_DIR = $(CONTIKI_NG_NET_DIR)/mac
CONTIKI_NG_ROUTING_DIR = $(CONTIKI_NG_NET_DIR)/routing
```

This allows us to use e.g. `MODULES += $(CONTIKI_NG_ROUTING_DIR)/rpl-classic` instead of having to hard-code `os/net/routing` everywhere. This will make source tree restructuring a lot easier in the future, since we will only have to change paths in one place.

This pull also:

* Deletes some obsolete sky/msp430 make targets
* Currently we pass `-I<platform-dir>` twice. This pull fixes this
* Currently we treat `$(CONTIKI)/arch` as a source dir, even though it contains no source files. This pull changes that and removes it from the include search path as well as from `vpath`.
* ~Currently we include `$(CONTIKI)` in in the list of include search dirs (e.g. `-I../..`). This was first added in 837a697dc8, but I cannot see the motivation so I am undoing it unless someone can clarify.~ Edit: I now do understand, so I am reverting this...

Now as per the discussion in #773, I can understand the need for someone to develop platform(s) out-of-tree. However, I must admit I am uncertain about some of the details of the use-case:
* I do not understand why our existing platform/cpu makefiles benefit by using `ARCH_PATH` since - if the user provides a custom `ARCH_PATH` - those Makefiles won't be included in the first place. This pull reverts the logic to ignore `ARCH_PATH` everywhere outside `Makefile.include`
* I do not understand the benefit of using `ARCH_PATH` in the `distclean` target. If we run `make ARCH_PATH=<somewhere> distclean`, we will only clean artifacts corresponding to platforms in `<somewhere>/platform`. Or if we run a simple `make distclean`, we will clean artifacts for in-tree targets, but not for the ones under development externally. Is this intentional? Do we need to be able to do something that the existing `TARGETDIRS` does not allows us to achieve?
* I do not understand the benefit of using `ARCH_PATH` in the `targets` target. If we run `make ARCH_PATH=<somewhere> targets`, we will only get a list of platforms in `<somewhere>/platform`. Ditto about the existing `TARGETDIRS` variable.
* Specifying a custom `ARCH_PATH` removes the ability to use existing drivers under `arch/dev`. I think this is undesirable.
* I do understand the benefit of using it as part of the target makefile detection `target_makefile`.

I'm therefore marking this work-in-progress until I've understood what we need to be able to achieve. I'll update the pull once I know what I'm doing :)

Requires #784


simonduq:
Whats' the status on this?

g-oikonomou:
Still working on this

g-oikonomou:
So @simonduq I am looking at this again.

In #773 we did `ARCH_PATH ?= $(CONTIKI)/arch`. We then use `$(ARCH_PATH)` to make references to cpu and platform dirs, but we left `arch/dev/` untouched.

For example we still do `MODULES += arch/dev/cc1200` at places.

It comes across as though we want to be able to move the cpu and platform dirs out of tree while retaining the ability to use stuff under `arch/dev` as `MODULES`. In other words, is the intention for `ARCH_PATH` to only influence `arch/cpu` and `arch/platform`, but not `arch/dev`?

Am I on the right track here?

g-oikonomou:
Furthermore, I see that $773 allows relocation of `arch/cpu/arm`. Is this actually required by the use-case, or can it be assumed to always be under `$(CONTIKI)`?

g-oikonomou:
I have updated this pull and this is ready for review. Below updated description.

This pull proposes various improvements and bugfixes to the build system.

Most importantly, we start using make variables to hold various source tree paths. For example, we introduce something like this:

```
CONTIKI_NG_OS_DIR = os
CONTIKI_NG_NET_DIR = $(CONTIKI_NG_OS_DIR)/net
CONTIKI_NG_MAC_DIR = $(CONTIKI_NG_NET_DIR)/mac
CONTIKI_NG_ROUTING_DIR = $(CONTIKI_NG_NET_DIR)/routing
```

This allows us to use e.g. `MODULES += $(CONTIKI_NG_ROUTING_DIR)/rpl-classic` instead of having to hard-code `os/net/routing` everywhere. This will make source tree restructuring a lot easier in the future, since we will only have to change paths in one place.

These variables are hosted in a new top-level Makefile called `Makefile.dir-variables`. This allows us to include this from example Makefiles and then use path variables to specify example-specific `MODULES`.

This pull retains compatibility with the existing `ARCH_PATH`, with the following logic:

* A user can specify `ARCH_PATH` to relocate `arch/cpu` and `arch/platform` to an out-of tree location.
* `arch/dev` and `arch/cpu/arm` are _not_ relocated and are always assumed to be under `$(CONTIKI)`. In doing so, a user can do the following without need to copy the respective dir over:
   i) use the arm build infrastructure when developing an arm-based platform out-of-tree
  ii) Include drivers under `arch/dev` when developing a platform out-of-tree

This pull also applies some minor changes:

* Deletes some obsolete sky/msp430 make targets
* Currently we pass `-I<platform-dir>` twice. This pull fixes this
* Currently we treat `$(CONTIKI)/arch` as a source dir, even though it contains no source files. This pull changes that and removes it from the include search path as well as from `vpath`.

Makefile.dir-variables
```diff
@@ -0,0 +1,67 @@
...
+### in this separate Makefile here, which is included by Makefile.include
+### This allows examples that wish to specify additional MODULES to include
+### this Makefile and append to MODULES using the path variables defined here
+
+# Variables for the various directories under os/
+# Those are sometimes used relative to the top-level dir in which case we don't
+# need to prepend $(CONTIKI), as well as relative to the example dir, in which
+# case we do. So we define the variables to suit the former case and we will
+# prepend $(CONTIKI) explicitly where needed.
+CONTIKI_NG_OS_DIR = os
```
simonduq: shouldn't all the variables in this file be defined with `?=`? (e.g. for example makefile to override)

g-oikonomou:
> I wonder if it would make sense to also allow relocating arch/dev?

I am struggling to come up with a use-case for that one. `arch/dev` contains platform-independent code. If a user wanted to develop new code in there as well as use one of the existing modules then they would have to copy the latter to the out-of-tree location as well. Would it not be easier for the user to simply put their new code in some other location? Maybe there is a use-case I cannot think of?

> I'm not sure about those under os/net (mac, routing etc.).

The `os/net/mac` and `os/net/routing` patterns were being repeated in various places hence the thought to create a variable for those as well

Makefile.dir-variables
g-oikonomou: So with the exception of `ARCH_PATH` we are not allowing the user to override the location of any other parts of the tree. Why would we want to start doing so now?

simonduq:
OK right, the intention is more cleanup and de-duplication than configurability. I'm happy with how the PR is now then.

PR #792, opened 2018-12-11
sys/process: thread-safe event queue

### LINKED ISSUE #764: It is not safe to call process_post in an interrupt handler
I'm not sure if this is expected or not, but it is not safe to call `process_post` in an interrupt handler. This is because it may cause inconsistency in the event queue (in `sys/process.c`).

The consumer of the event queue is `do_event` function (run in normal context), and the producer is `process_post` function. Because operation on `fevent` and `nevents` are not atomic in `do_event`, in an interrupt handler `process_post` may manipulate the event queue while it's in an inconsistent state. This may cause losing a posted event or dispatching an event that was already dispatched before.

I don't think so many modules call `process_post` in an interrupt context, but at least the tsch module does (in `tsch_disassociate`, which may be called in `tsch_slot_operation` thread) (although that is relatively a rare case).

I think we should either

1. make the event queue thread-safe, probably by using `ringbufindex`.
2. or, add warning to the document of `process_post`, and fix existing calls to `process_post` in interrupt contexts. For the tsch case above, we can use `process_poll` instead.

debug-ito:
This p-r fixes #764, i.e. it makes it safe to call `process_post` in an interrupt handler.

To protect access to the event queue (the shared resource), I used sys/mutex. I could use sys/critical instead. They have different pros and cons.

- Use sys/mutex

    - (pro) Duration where interrupts are disabled is relatively short, or on some platforms (such as arm) there is even no need to disable interrupts.
    - (con) We need to add a new return code to `process_post`, `PROCESS_ERR_LOCKED`, which means the mutex is locked when it's called.

- Use sys/critical

    - (pro) We don't need to change any external API of sys/process.
    - (con) We would need to disable interrupts to manipulate the event queue. This could affect accuracy of tsch-slot-operation.

I think adding `PROCESS_ERR_LOCKED` has little impact because `process_post` could return `PROCESS_ERR_FULL` anyway. So, I chose sys/mutex.

I ran an experiment with real devices in which it called `process_post` rapidly in a sys/rtimer callback. With this p-r, no events were lost or duplicated.

### DIFF
--- a/os/sys/process.c
+++ b/os/sys/process.c
@@ -47,6 +47,7 @@
 
 #include "contiki.h"
 #include "sys/process.h"
+#include "sys/mutex.h"
 
 /*
  * Pointer to the currently running process structure.
@@ -65,6 +66,7 @@ struct event_data {
   struct process *p;
 };
 
+static mutex_t events_mutex;
 static process_num_events_t nevents, fevent;
 static struct event_data events[PROCESS_CONF_NUMEVENTS];
 
@@ -209,6 +211,7 @@ process_init(void)
 {
   lastevent = PROCESS_EVENT_MAX;
 
+  events_mutex = MUTEX_STATUS_UNLOCKED;
   nevents = fevent = 0;
 #if PROCESS_CONF_STATS
   process_maxevents = 0;
@@ -250,6 +253,14 @@ do_event(void)
   struct process *receiver;
   struct process *p;
 
+  if(!mutex_try_lock(&events_mutex)) {
+    return;
+  }
+  if(nevents == 0) {
+    mutex_unlock(&events_mutex);
+    return;
+  }
+  
   /*
    * If there are any events in the queue, take the first one and walk
    * through the list of processes to see if the event should be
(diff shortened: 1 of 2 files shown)


atiselsts:
I'm almost convinced by the reasoning, but not quite... I'm with you regarding the overhead of critical sections vs. mutexes.

My problem is that this new approach with the new failure mode with `PROCESS_ERR_LOCKED` leaves the system broken in many places. Almost no code in Contiki-NG checks for the return value of `process_post()` (one notable exception being the etimer code). Even macros that use `process_post()`, for example the macro `PROCESS_PAUSE()` do not check for this!! Previously as a system programmer I knew that I had to make sure the event buffer never overflows; if it did, I felt it was my fault and that undefined behavior is expected afterwards. With the new approach, the failure is essentially random. (As an anecdote, we deployed ~1000 Contiki devices for up to 1 year; the monitoring data I've seen showed that the event buffer never got full after an initial design flaw in the code was fixed. On the other hand, random race conditions that only happen with very low probabilty will definitely happen if you keep the system running for long enough.)

Anyway, perhaps this situation calls for some like spinlock (https://unix.stackexchange.com/questions/5088/what-is-a-spinlock-in-linux) I suppose the mutex API in Contiki-NG could be redesigned / extended (@g-oikonomou ?)

debug-ito:
Thanks for comment!

Yes, I've also noticed most calls of `process_post` don't check the return code. In my opinion, it's a bug and should be fixed by another p-r, whether or not we have `PROCESS_ERR_LOCKED`.

Note that `PROCESS_ERR_LOCKED` response is a very rare event. It can be returned only if `process_post` is called in an interrupt handler. So far, I've found only one such call in Contiki-NG code base, i.e., in `tsch_disassociate` called from `tsch_slot_operation` thread. We can substitute `process_poll` for that call.

atiselsts:
There may be only one instance in the core code (what is now called "os"), but there are quite a few more in platform code: cc26xx BLE, cc26xx RNG, msp430 button, probably more. Plus, people are developing their own drivers / extending existing ones and so far they were able to use `process_post()` freely in interrupt handlers (after extending the cc26xx BLE driver with a couple more events in a private branch I don't fancy migrating to poll will be nontrivial). Anyway, OS API changes should not break existing software, and especially not in such an obscure way that no one will notice.

So, fundamentally, I think there are two options:
1) do not allow to call `process_post` in an ISR (and be very loud and clear about that).
2) fix the behaviour in `process_post` in an ISR so that it never fails (as long as the event queue is not full).

In the first case, no API changes are needed. I'm all for the second case, but this PR is not there yet.

debug-ito:
I agree with you @atiselsts. Your option 2 is better than my p-r. I just don't know how to achieve that.

g-oikonomou:
> Anyway, perhaps this situation calls for some like spinlock...

I thought about spinlocks when I was doing the mutex work. I cannot immediately think how an interrupt would ever exit a spinlock, hence I never did anything about it. Open to suggestions, it would be a nice thing to have. Unfortunately what we have now is that an ISR will either acquire a resource or has to be able to do something meaningful if the resource is locked.

> do not allow to call process_post in an ISR (and be very loud and clear about that).

I think this is what we should be doing and I think we are not loud enough. Ditto `ctimer_set` in interrupts, something I have been discussing with @nfi offline in the past :)

debug-ito:
I thought what we needed here was a lock-free multi-producer queue like in https://www.linuxjournal.com/content/lock-free-multi-producer-multi-consumer-queue-ring-buffer

I'm not sure we can implement the one in the above blog post in Contiki-NG. I chose to use sys/mutex to lock the queue because the lock-free implementation seemed too complex for me.

debug-ito:
I just examined the code of cc26xx BLE (ble-hal-cc26xx) and found that it heavily relied on the event queue mechanism. Now I think it's worth the effort to try to implement the lock-free queue. I'll try that...

debug-ito:
The event queue is basically a multi-producer single-consumer queue. Because the consumer (`do_event` function) always runs in non-ISR context, we have three cases to consider.

1. Producer (in ISR) preempts Consumer
2. Producer (in ISR) preempts Producer (in non-ISR)
3. Producer (in ISR) preempts Producer (in ISR)

I think case 3 is rare, but possible.

Good news is that the consumer does not preempt anyone. So, if we switch to using `ringbufindex` we can solve case 1.

Cases 2 and 3 are a little tricky. To resolve the race condition between two producers, we need to manipulate `put_ptr` (in `ringbufindex`) atomically. 

Suppose we had an abstract method of atomic compare-and-swap (CAS),

```c
/**
 * If value of `target` is equal to `old`, set `target` to `new` and return true.
 * Otherwise, it just returns false.
 * The whole compare-and-swap operation is done atomically.
 */
int atomic_cas_uint8(uint8_t *target, uint8_t old, uint8_t new);
```

then I think we could implement atomic put like

```c
uint8_t now_ptr;
uint8_t next_ptr;

while(1) {
  now_ptr = put_ptr;
  next_ptr = (put_ptr + 1) & mask;
  if(((now_ptr - get_ptr) & mask) == mask) {
     return ERROR_QUEUE_IS_FULL;
  }
  if(atomic_cas_uint8(&put_ptr, now_ptr, next_ptr)) {
    break;
  }
}

/* Put event object at now_ptr */

```

Maybe we need to place some `memory_barrier`s...

Anyway, to make the above implementation, we need platform-independent `atomic_cas_uint8` (perhaps in sys/atomic or something). We can give it a default implementation using sys/critical, and more efficient implementation for ARM.

debug-ito:
I'm gonna make three pull-requests that implement an alternative solution to this pull-request.

1. Add sys/atomic, which provides atomic compare-and-swap (CAS) API.
2. Extend lib/ringbufindex with atomic put/get API using sys/atomic.
3. Re-implement the event queue in sys/process using ringbufindex and its atomic API to fix #764

This approach solves the problem discussed in this thread. With the new atomic API, ringbufindex can perform lock-free put. So, there is gonna be no error mode of `PROCESS_ERR_LOCKED`.

simonduq:
Hi!

joakimeriksson:
Closing this since the documented and intended behaviour is to not call post from interrupt.

PR #803, opened 2018-12-18
make compilation with SICSLOWPAN_COMPRESSION_IPV6 possible

hudson-ayers:
Without this fix, setting SICSLOWPAN_CONF_COMPRESSION = SICSLOWPAN_COMPRESSION_IPV6 in a project-conf file leads to compilation failures because uncompress_hdr_iphc() is not defined.

### DIFF
--- a/os/net/ipv6/sicslowpan.c
+++ b/os/net/ipv6/sicslowpan.c
@@ -1397,7 +1397,7 @@ digest_6lorh_hdr(void)
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  * \endverbatim
  */
- #if SICSLOWPAN_COMPRESSION == SICSLOWPAN_COMPRESSION_IPV6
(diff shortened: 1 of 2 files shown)


simonduq:
What would you think about moving `SICSLOWPAN_COMPRESSION > SICSLOWPAN_COMPRESSION_IPV6` to the `if` instead of having it as pre-processor macro?

hudson-ayers:
I think that is probably fine given how little code is nested within the if statement, though I do think as a general rule that people are likely to select SICSLOWPAN_COMPRESSION_IPV6 because of code size concerns, so I prefer to exclude code entirely via pre-processor macros when possible.

simonduq:
But having it in the `if` will also result in the code being excluded. Basically same end result, but more maintainable source code, and the guarantee that the branch is always parsed by the compiler.

simonduq:
ping @hudson-ayers

joakimeriksson:
I think we need to be able to configure away IPHC as it takes significant size, but also we need to document the assumption that we order the defines of compression mechanisms in the header file. I can take a stab on this and try to push/PR to this PR.

PR #814, opened 2019-01-09
fix temperature sensor reading in nrf52dk platform, which was stuck a…

zhitaoh:
The blink-hello example for the nrf52dk platform display temperature samples stuck at zero:
...
temp: 0.00
temp: 0.00
temp: 0.00
...

After fixing the temperature sensor driver's value() function by adding a proper sequence of HAL statements from the NRF SDK, the blink-hello example shows correct temperature values in degree Celsius:
...
temp: 22.50
temp: 22.25
temp: 22.75
...

### DIFF
--- a/arch/platform/nrf52dk/dev/temperature-sensor.c
+++ b/arch/platform/nrf52dk/dev/temperature-sensor.c
@@ -64,7 +64,15 @@ static int
 value(int type)
 {
 #ifndef SOFTDEVICE_PRESENT
-  return nrf_temp_read();
+  int32_t volatile temp;
+
+  NRF_TEMP->TASKS_START = 1;
+  while(NRF_TEMP->EVENTS_DATARDY == 0);
+  NRF_TEMP->EVENTS_DATARDY = 0;
+  temp = nrf_temp_read();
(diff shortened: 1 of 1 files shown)


atiselsts:
Maybe it's better to use the new `RTIMER_BUSYWAIT_UNTIL` macro (https://github.com/contiki-ng/contiki-ng/blob/develop/os/sys/rtimer.h) with a time limit instead of the unconditional `while` loop here:

    while(NRF_TEMP->EVENTS_DATARDY == 0);

g-oikonomou:
Agree with Atis. What time durations are we looking at? If too long, best re-written as asynchronous, with events?

zhitaoh:
Agree! I was half lazy to check the timing requirement and half curious to see if you guys would slack off with the reviewing. The datasheet says it takes typically 36 us to take a sample - I think it is short enough to just busy wait for the result. Have updated using `RTIMER_BUSYWAIT_UNTIL()` as Atis suggested.

g-oikonomou:
@atiselsts are you happy here?

atiselsts:
:+1:

PR #815, opened 2019-01-09
Add AES-128 driver for CC26x0/CC13x0

atiselsts:


g-oikonomou:
Should work for CC13x0 with trivial changes, right? Merging anyway, as this is super desirable.

g-oikonomou:
> Should work for CC13x0 with trivial changes, right?

Or I could have read the PR title... 😃

PR #819, opened 2019-01-14
Add back Zolertia Z1 platform

atiselsts:
Add the code, tools and some regression tests. Tried to avoid  changes if they were not necessary.


g-oikonomou:
Is the current `tools/motelist` not suitable for Z1s?

atiselsts:
> Is the current `tools/motelist` not suitable for Z1s?

I suppose it is - don't have a Mac to test it! Do you know that it should work with Z1 on Mac?

g-oikonomou:
Happy to test on macos, linux too unless someone else does so first. We still need the bsl tools and I am not 100% certain whether the build system or those tools expect specific output from motelist

atiselsts:
I don't immediately see how to adapt the new motelist for Z1's build system. A working example for some other platform would be nice :P

g-oikonomou:
> I don't immediately see how to adapt the new motelist for Z1's build system. A working example for some other platform would be nice :P

Shame on you now!

g-oikonomou:
```
$ python tools/motelist/motelist.py 
Port                     Serial    VID     PID     Product      Vendor      
-----------------------  --------  ------  ------  -----------  ------------
/dev/tty.SLAB_USBtoUART  Z1RC5336  0x10C4  0xEA60  Zolertia Z1  Silicon Labs
```

PR #820, opened 2019-01-14
correct time intervals of example CoAP periodic resources

zhitaoh:
The PERIODIC_RESOURCE macro expects a time interval represented by No. coap timer ticks, which are one millisecond in length. Therefore one second = 1000 x 1 ms ticks, not CLOCK_SECOND system ticks.

These wrong intervals were however not noticeable on the native platform, where each CLOCK_SECOND happens to include also 1000 ticks.


joakimeriksson:
Ok, good catch!

PR #825, opened 2019-01-17
Add sys/atomic

debug-ito:
First of three-step alternative for #792.

This pull-request adds a new core library sys/atomic.{c,h}, which defines platform-independent API for atomic operations. Currently, it has `atomic_cas_uint8` only, which does atomic compare-and-swap (CAS). Generic implementation and ARM-specific implementation are provided.

This kind of atomic operation is necessary to implement lock-free data structures.


debug-ito:
Any updates?

atiselsts:
I have some small comments:
- In `mutex.h` there's this line: `typedef uint_fast8_t mutex_t;`. So, in general, a mutex is not guaranteed to be 8-bits on all platforms (it might be 8 bits or longer). The `mutex-cortex.h` does typedef it to `uint8_t`, but there is still a potential for confusion when modifying the code or porting it to other platforms. As a minimum, I suggest adding comment explaining why it's ok to cast `mutex_t *` to `uin8t_t *`.
- the  `volatile` qualifier - is it really needed in this case? If no, should keep it just because the existing API has it?
- `cas` is not an immediately obvious acronym, perhaps just use `compare_and_swap`?

debug-ito:
Thanks for comments!

- I added comment to mutex-cortex.h. Is it ok?
- Do you mean the `volatile` in `atomic_generic_cas_uint8` and `atomic_cortex_cas_uint8`? To be honest, I'm not sure if it's really necessary. I just added `volatile` because mutex.h uses it. Could you give me advice?
- In my opinion, CAS is a [relatively common term in computer science especially in the context of atomicity](https://en.wikipedia.org/wiki/Compare-and-swap), the acronym is expanded in the doxygen entry and `atomic_compare_and_swap_uint8` would be too long for me. However, I'm ok with using `compare_and_swap` in function names if you insist. Do you insist?

atiselsts:
Yes, that is ok and I don't insist on the other changes - perhaps @g-oikonomou can comment on the `volatile`. Not sure why the regression test fails.

debug-ito:
Hmm, when I ran the test locally on my machine, the test `rpl-classic` was successful.

atiselsts:
That is probably a random glitch. Could you try to push again (make some trivial change, commit, and push)?

g-oikonomou:
`target` can change after the `__LDREXB` and before the `__STREXB` and we don't want the compiler to make any assumptions as to the contrary. Remember these can be used without disabling the global interrupt.

g-oikonomou:
The only thing I am not 100% certain about is the idea of mapping the current `mutex_cortex_try_lock()` to a more generic function that takes 3 arguments. I wonder whether we should be keeping both irrespective of the code duplication in order to keep `mutex_cortex_try_lock` a bit simpler.

debug-ito:
Thanks, @g-oikonomou!

So, I think setting `volatile` is important, because the reason why we use `__LDREXB` and `__STREXB` is that they work as expected even if the content of `target` changes.

About `mutex_cortex_try_lock`, I'm ok with keeping the current implementation untouched. This should be safer in terms of backward compatibility.
Shall I make a commit to revert the change about `mutex_cortex_try_lock`, or should I make a new, updated pull-request without that change?

atiselsts:
> `target` can change after the `__LDREXB` and before the `__STREXB` and we don't want the compiler to make any assumptions as to the contrary. Remember these can be used without disabling the global interrupt.

Not sure that this matters. Here's how I see it. It's the `uint8_t` that is declared in the function signature to be volatile, not the `*` (pointer) - `volatile uint8_t *`, not `uint8_t *volatile`. So, the `volatile` only starts to matter when the pointer is dereferenced. And the pointer is not dereferenced by the Contiki-NG function code itself, so there is no functional differences whether the `uint8_t` is declared volatile or not.

The intrinsic functions `__LDREXB` and `__STREXB` presumably do dereference the pointer, so I agree that they should declare their parameter as volatile, and this is what the apparently already do: https://www.keil.com/pack/doc/CMSIS/Core/html/group__intrinsic__CPU__gr.html

g-oikonomou:
> Not sure that this matters.

Hm yea I think you may be right

@debug-ito please don't open a new pull. Update your branch with any changes and force push it to your repo

debug-ito:
@g-oikonomou Thanks. I reverted the changes to mutex-cortex.h and pushed that.

@atiselsts I agree with you. `atomic_cortex_cas_uint8` does not have to set `volatile` for `target`. However, I think we should keep the `volatile` qualifier, because other implementations (such as `atomic_generic_cas_uint8`) may need it. We should use the same function signature for different implementations.

g-oikonomou:
@debug-ito Thanks. Let's also remove the volatile from `atomic_generic_cas_uint8()` and friends and we can merge this. We can deal with volatile in the mutex code in a separate pull.

debug-ito:
Thanks. I just removed the `volatile`s and pushed it.

debug-ito:
@g-oikonomou Could you merge this? Should I still revise something?

debug-ito:
Thanks!

Issue #826, opened 2019-01-17
BLE on Simplelink platform not working


firmwareguru:
I've tried to get the BLE beacon functionality going with the new simplelink platform and it's not going well.  The example app locks up in the RF_pendCmd() in sched.c on the first beacon broadcast attempt.

I have "ported" the `platform-specific/cc26x0-cc13x0/cc26x0-demo.c` example to build with the simplelink platform - this just required renaming a couple of defines.

I have tried this on the launchpad/cc1350 as well as a cc1352-based board - same result.  I can build the cc26x0-demo example for the cc26x0-cc13x0 target for the same launchpad/cc1350 board and it works fine - beacons are captured by my phone.

@tiepettersen I am wondering if you can help out here?  

First step is to get BLE beacon daemon going with the simplelink target.  Next I'd like to get the ble-hal ported over to simplelink as well, but first things first.

tiepettersen:
@firmwareguru, sounds strange. I'll take a look.

tiepettersen:
Ok, I found out what the issue is. The CFE RF Patch is wrong, as a special CPE patch is required when running multi-protocol. Somehow the correct patches did not propagate when pushing for a PR. I'm developing a fix now. I'll also test that BLE beaconing works as well.

firmwareguru:
Super, thanks.

tiepettersen:
So I'm not able to push out a PR before the weekend, but I have developed a fix for CC13x2/CC26x2 devices. However, there seems to be an underlying problem of running multi-protocol on CC13x0/CC26x0 devices that I will need to work out. Aiming for a PR next monday.

firmwareguru:
I look forward to trying it out.

tiepettersen:
Fixed by #829

tiepettersen:
This can now be closed.

PR #829, opened 2019-01-21
Fixed BLE advertisement on SimpleLink

tiepettersen:
Reported by issue #826.

A couple of issues was found:
* Wrong CPE patches in RF settings for CC13x2/CC26x2 devices
* Wrong RF mode in RF settings for CC13x0/CC26x0 devices
* Wrong whitening init value for the BLE advertisement commands for BLE/BLE5
* BLE MAC address was not copied to RAM before set in the bleAdvPar
  construct
* BLE advertisements were not transmitted in intervals as specified by
  BLE spec
* Very subtle logic error when ieee-mode/prop-mode scheduled RX for CCA request/RSSI measurement


arch/cpu/simplelink-cc13xx-cc26xx/rf/sched.c
```diff
@@ -479,45 +491,171 @@ ble_open(RF_Params *params)
...
-      &rf_ble,
-      (RF_Op *)&ble_cmd_beacon,
-      &sched_params,
-      cb,
-      bm_event);
+  /*
+   * The most efficient way to schedule the command is as follows:
+   *   1. Schedule the BLE advertisement chain
+   *   2. Reschedule the RX command IF it was running.
+   *   3. Pend on the BLE avertisement chain
```
firmwareguru: a(d)vertisement

arch/cpu/simplelink-cc13xx-cc26xx/rf/sched.c
```diff
@@ -479,45 +491,171 @@ ble_open(RF_Params *params)
...
+
+  /* If no channels are mapped, then early return OK */
+  if((bm_channel & BLE_ADV_CHANNEL_ALL) == 0) {
+    return RF_RESULT_OK;
+  }
 
+  initial_adv = init_ble_adv_array(ble_cmd_adv_nc_array, bm_channel);
+
+  if(initial_adv == NULL) {
+    LOG_ERR("Initializing BLE Advertismenet chain failed\n");
```
firmwareguru: sp: Advertisement

tiepettersen:
I will check out the `rf_ble_get_tx_power()` issue. However, during development I observed BLE advertisements with a very high dBm (-35, -40 dBm). I'll see if I'm able to reproduce what you saw.

tiepettersen:
@firmwareguru, I was able to reproduce the issue you reported on CC13xx devices. It was a combination of wrong RF front-end mode configurations as well outdated RF switching logic for CC13xx devices. This only affected CC13xx devices, not CC26xx devices. I've pushed an update which should fix the sensitivity.

firmwareguru:
Thanks for getting this straightened out @tiepettersen.  

The launchpad/cc1350 platform is working now; rf power API reports a value of 5, and the beacon power levels are nominal at > -50 dBm.

I have a question about the `RF_BIAS_MODE` setting.  More specifically to the CC1352R1 launchpad, where RX_TX is connected to just the sub-GHz path and not the 2.4 path.  Does it make sense to assign the same bias mode to both ble radio setup and prop-mode setup commands?  You've changed this to "EXTERNAL" bias mode that is applied to both setup commands.

tiepettersen:
@firmwareguru, you are correct that the front-end mode configurations should indeed match the hardware. The setup I implemented expects both the 2.4 and Sub-1 paths to have the same front-end mode configurations, which obviously won't work for some boards. I'll extend the current implementation to support specifying the front-end mode configurations individually for each path if needed.

arch/platform/simplelink/cc13xx-cc26xx/launchpad/cc1312r1/rf-conf.h
```diff
@@ -0,0 +1,60 @@
...
+/*---------------------------------------------------------------------------*/
+#include "rf/rf.h"
+/*---------------------------------------------------------------------------*/
+/**
+ * \name  Board-specific front-end configurations for the radio.
+ *
+ * @{
+ */
+#define RF_CONF_FRONT_END_MODE  RF_FRONT_END_MODE_DIFFERENTIAL
+#define RF_CONF_BIAS_MODE       RF_BIAS_MODE_EXTERNAL
```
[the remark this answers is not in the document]
tiepettersen: Seems to be a mistake. I'll remove the ones defined in the Board.h.

tiepettersen:
@firmwareguru, the latest RF front-end mode configurations for the LaunchPad CC1352R1 I've pushed configured both paths with differential+external bias. This is because the LaunchPad with the release silicon is designed as such, while the LaunchPad with the pre-release silicon is not. Simply open up launchpad/cc1352r1/rf-conf.h and change the configuration to match your version.

g-oikonomou:
Doesn't seem to be @tiepettersen , I have restarted the failed job. Let's see

Issue #774, opened 2018-11-24
Won't compile assembly files .s or .S


firmwareguru:
Compilation of assembly files with .s or .S is not working, at least when added like this, for example

`CONTIKI_TARGET_SOURCEFILES += pdm2pcm_cc26x2_gnu.s`

Results in

`process_begin: CreateProcess(NULL, as -o build/keystone/r1//obj/pdm2pcm_cc26x2_gnu.o ../../../../arch/platform/keystone/dev/pdm2pcm_cc26x2_gnu.S, ...) failed.`

(Note: using simplelink platform)

Looking through the makefiles it's pretty easy to see why.  Neither `AS` nor `ASFLAGS` are defined and the command invocation line is missing an option.

Here is what works in my case.  Note the `-c` added before `$<`.

**Makefile.include**

```
AS := $(CC)
ASFLAGS := $(CFLAGS)

....

ifndef CUSTOM_RULE_S_TO_OBJECTDIR_O
$(OBJECTDIR)/%.o: %.S | $(OBJECTDIR)
	$(TRACE_AS)
	$(Q)$(AS) $(ASFLAGS) -o $@ -c $<
$(OBJECTDIR)/%.o: %.s | $(OBJECTDIR)
	$(TRACE_AS)
	$(Q)$(AS) $(ASFLAGS) -o $@ -c $<
endif
```

https://github.com/contiki-ng/contiki-ng/issues/302
https://github.com/contiki-ng/contiki-ng/pull/321

@g-oikonomou

firmwareguru:
Sure I can do that.  What I proposed up top is a bit of a hack so I'll need you to weigh on how you think it should be resolved.



Sent from my Samsung Galaxy smartphone.


-------- Original message --------

g-oikonomou:
@firmware Evan is there any progress here?

firmwareguru:
I'm working on a patch, but I can't figure out where AS is getting set in the build system.  But it is getting set to:
`AS=as`. 

I'd like to use `AS ?= $(CC)` or `ifndef AS ...` but that isn't working because of the aforementioned issue.  @g-oikonomou do you know where it is getting set?  I've searched all makefiles including arch/cpu/arm...

g-oikonomou:
I think what you are observing is GNU make setting `AS` to `as` by default since we are not setting it elsewhere. https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

[...]

PR #843, opened 2019-02-09
Fixes contiki-ng/contiki-ng#774 Won't compile asm files

firmwareguru:
Reference issue #774.

This fix takes care of compiling both .s and .S for me.

Using `gcc` rather than `as` in `arch/cpu/arm/Makefile.arm`  to compile asm files offers all of the available preprocessor goodies, so that's what I've chosen to implement.

### DIFF
--- a/arch/cpu/arm/Makefile.arm
+++ b/arch/cpu/arm/Makefile.arm
@@ -10,6 +10,7 @@ CC       = arm-none-eabi-gcc
 CPP      = arm-none-eabi-cpp
 LD       = arm-none-eabi-gcc
 AR       = arm-none-eabi-ar
+AS       = arm-none-eabi-gcc
 OBJCOPY  = arm-none-eabi-objcopy
 OBJDUMP  = arm-none-eabi-objdump
 NM       = arm-none-eabi-nm
@@ -24,6 +25,11 @@ CFLAGS += -fshort-enums -fomit-frame-pointer -fno-builtin
 ifeq ($(WERROR),1)
   CFLAGS += -Werror
 endif
+
+### Pass CFLAGS along to assembly files in the build
+ASFLAGS += $(CFLAGS)
+### Specify '-c' option to assemble only and not link
+ASFLAGS += -c 
 
 LDFLAGS += -mthumb -mlittle-endian
 


g-oikonomou:
Hi @firmwareguru thanks for this. We want this.

Using `arm-none-eabi-gcc` instead of `-as` sounds fine to me, in fact I have done the same in a local branch too.

What I am not 100% comfortable about is blindly using `CFLAGS` as `ASFLAGS`. Could you please add some motivation? Is there a reason why not simply use `ASFLAGS`?

These are the changes that I have had to make locally:

```
arch/cpu/arm/Makefile.arm:ASFLAGS += -mthumb -mabi=aapcs -mlittle-endian
arch/cpu/arm/cortex-m/cm4/Makefile.cm4:ASFLAGS += -mcpu=cortex-m4
arch/cpu/.../Makefile.<my-new-cpu>:ASFLAGS += -mfloat-abi=hard -mfpu=fpv4-sp-d16 -specs=nano.specs
arch/cpu/.../Makefile.<my-new-cpu>:ASFLAGS += -c -x assembler-with-cpp
```

Obviously:

* Some of those are specific to the hardware I am working on and
* We are missing out on all the `-D` and `-I`,

but I just wanted to share how I am approaching the same problem.

The main line of interest here is the last one in the above snippet. This allows us to pass `-c` without touching the top-level `Makefile.include`. `-x assembler-with-cpp` allows us to put pre-processor stuff in the assembly.

Would such approach meet your requirements?

g-oikonomou:
Also, if we are going to set `ASFLAGS` to `CFLAGS`, I'd rather do it with `+=`: Allows the user to append more.

firmwareguru:
It seems like offering the asm files the same environment as the .c files in which to build would be the desired outcome, so that's why I used gcc rather than as, and then set ASFLAGS to CFLAGS.

But,  I realize there are possibly some more "traditional" assembly use cases that might want the file extension to determine what kind of pre-processing to apply (i.e. .s vs .S, for which assembler behavior is effectively overridden with the -x assembler-with-cpp option)

For me the "hassle" of maintaining a separate, but identical, ASFLAGS isn't warranted.

I defer to you to specify if you want to offer that extra level of control w.r.t ASFLAGS.

Makefile.include
```diff
@@ -347,10 +349,10 @@ endif
 ifndef CUSTOM_RULE_S_TO_OBJECTDIR_O
 $(OBJECTDIR)/%.o: %.S | $(OBJECTDIR)
 	$(TRACE_AS)
-	$(Q)$(AS) $(ASFLAGS) -o $@ $<
+	$(Q)$(AS) $(ASFLAGS) -o $@ -c $<
```
g-oikonomou: Please remove this change

Makefile.include
```diff
@@ -337,6 +337,8 @@ distclean:
 
 -include $(ARCH_PATH)/platform/$(TARGET)/Makefile.customrules-$(TARGET)
 
+ASFLAGS ?= $(CFLAGS)
```
g-oikonomou: This needs to change to `+=`

Makefile.include
```diff
@@ -347,10 +349,10 @@ endif
 ifndef CUSTOM_RULE_S_TO_OBJECTDIR_O
 $(OBJECTDIR)/%.o: %.S | $(OBJECTDIR)
 	$(TRACE_AS)
-	$(Q)$(AS) $(ASFLAGS) -o $@ $<
+	$(Q)$(AS) $(ASFLAGS) -o $@ -c $<
 $(OBJECTDIR)/%.o: %.s | $(OBJECTDIR)
 	$(TRACE_AS)
-	$(Q)$(AS) $(ASFLAGS) -o $@ $<
+	$(Q)$(AS) $(ASFLAGS) -o $@ -c $<
```
g-oikonomou: Please remove this change

Makefile.include
g-oikonomou: And also moved to `Makefile.arm`

g-oikonomou:
@firmwareguru Hi Evan, do you think you will have time to make changes here? It'd be great to include this in the upcoming release. Please let me know if you don't have time.

firmwareguru:
Yes, working on it!

firmwareguru:
I'm proposing this for makefile.arm.  I don't see any use cases for not having `-c` a part of the ASFLAGS for all ARM CPUs... but you have better visibility on the big picture than I.  So tell me exactly what you want to see here and I'll do it.

```
CFLAGS += -mthumb -mabi=aapcs -mlittle-endian
CFLAGS += -Wall
CFLAGS += -std=c99
CFLAGS += -ffunction-sections -fdata-sections -fno-strict-aliasing
CFLAGS += -fshort-enums -fomit-frame-pointer -fno-builtin
ifeq ($(WERROR),1)
  CFLAGS += -Werror
endif

### Pass our CFLAGS along to assembly files in the build
ASFLAGS += $(CFLAGS)
### Specify '-c' option to assemble only and not link
ASFLAGS += -c 

LDFLAGS += -mthumb -mlittle-endian
```

firmwareguru:
Hmm, looks like I rebased then committed another change then pushed this up, and made a mess of things.  Probably should close this and start over.

g-oikonomou:
Just fix the branch locally and force push

firmwareguru:
Ok, changes made and branch fixed.  Don't rebase on public branches!  

The end result is that only `makefile.arm` is modified to enable compilation of `.s` and `.S` files.

Issue #835, opened 2019-01-29
We are refreshing watchdogs inside interrupt contexts


g-oikonomou:
As per the title, I spotted this using a TSCH-enabled image but it's not a flaw in TSCH per se.

Basically `RTIMER_BUSYWAIT_UNTIL_ABS()` calls `watchdog_periodic()`. This seems innocent, except when `RTIMER_BUSYWAIT_UNTIL_ABS()` gets used inside an interrupt context. This is the case with TSCH-enabled images: `TSCH_SCHEDULE_AND_YIELD()` happens inside an interrupt context and ends up refreshing watchdogs.

The result of all this is that a firmware otherwise crashed may never recover because interrupts will be firing frequently enough to reset the WDT before it triggers a reset. The externally observable behaviour of this is a WDT-enabled device that appears totally dead but which is not getting reset by its WDT.

This, IMHO, is a critical bug and I am flagging this as such. We can document `RTIMER_BUSYWAIT_UNTIL_ABS()` as not safe to run within an interrupt context, or we can remove the call to `watchdog_periodic()` from within the macro. This would likely mean we'd need to add calls to `watchdog_periodic()` in other places. We must also document that it's a bad idea to call `watchdog_periodic()` from inside interrupt contexts.

This was introduced by #562

g-oikonomou:
Fixed in #846 and #946

PR #846, opened 2019-02-18
Do not refresh watchdog in rtimer busy-wait macros

simonduq:
This removes the watchdog refresh from the rtimer busy-wait macros. Partially addresses https://github.com/contiki-ng/contiki-ng/issues/835

I think removing the refresh should not break anything here. Before these macros, the only busy-wait done with watchdog refresh was in `cc1200.c` and this was meant to handle busy waiting during `transmit` at very low data rates. No data rate low enough to result in 1s airtime is supported in NG though, so this shouldn't be an issue.

### DIFF
--- a/os/sys/rtimer.h
+++ b/os/sys/rtimer.h
@@ -57,6 +57,11 @@
 #include "dev/watchdog.h"
 #include <stdbool.h>
 
+#if CONTIKI_TARGET_COOJA
+#include "lib/simEnvChange.h"
+#include "sys/cooja_mt.h"
+#endif
+
 /** \brief The rtimer size (in bytes) */
 #ifdef RTIMER_CONF_CLOCK_SIZE
 #define RTIMER_CLOCK_SIZE RTIMER_CONF_CLOCK_SIZE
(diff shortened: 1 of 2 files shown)


os/sys/rtimer.h
```diff
@@ -189,14 +194,24 @@ void rtimer_arch_schedule(rtimer_clock_t t);
 #endif /* RTIMER_CONF_GUARD_TIME */
 
 /** \brief Busy-wait until a condition. Start time is t0, max wait time is max_time */
+#if CONTIKI_TARGET_COOJA
```
g-oikonomou: I would suggest `#ifndef RTIMER_BUSYWAIT_UNTIL_ABS` here. And then the cooja-specific stuff can go in a cooja-specific header somewhere.

os/sys/rtimer.h
```diff
@@ -57,6 +57,11 @@
 #include "dev/watchdog.h"
 #include <stdbool.h>
 
+#if CONTIKI_TARGET_COOJA
+#include "lib/simEnvChange.h"
+#include "sys/cooja_mt.h"
+#endif
+
```
g-oikonomou: Meh platform-specific code under `os/`:)

simonduq:
Platform-specific def makes a lot more sense, will do

simonduq:
Done

MartinNPN:
Hello,

wasn't the watchdog triggering not introduced to ensure that larger ACK waiting times work proper at cc1310/cc1350, cause this is function is used at the rf driver ?

Issue #833, opened 2019-01-28
Cooja mote: Transmission with radio turned off puts it back on.


pciolkosz:
https://github.com/contiki-ng/contiki-ng/blob/17f4c688880df3b242036dc121ce215795a11911/arch/platform/cooja/dev/cooja-radio.c#L214

Turning radio off in Cooja mote is not blocking outgoing packets. Even if it is a design choice, it does it wrong, because sometimes it allows to receive packet if there is incoming transmission around that time.

yatch:
Indeed, seems it should return `RADIO_TX_ERR` there. I checked the history of the file but I cannot find why this code was introduced: https://github.com/contiki-ng/contiki-ng/commit/bee0ea4834259b18d57ad5edb479c008cb3197c9#diff-9366018bcde46e7af8a0e01013daca30

> because sometimes it allows to receive packet if there is incoming transmission around that time.

By the way, just out of curiosity, did you see this actually happened in Cooja?

pciolkosz:
Yes, I saw confirmed transmission to a node with radio turned off and believe this was the cause.

g-oikonomou:
@yatch are we looking at a bug here?

yatch:
@g-oikonomou Yes; I cannot see any sensible explanation for that. I will send a PR later.

yatch:
As per the discussion at PR #855, the reported behavior is common in H/W radios. 

@pciolkosz 

> because sometimes it allows to receive packet if there is incoming transmission around that time.

I checked the code in cooja as well, but I cannot see how this can happen. It seems the radio won't receive anything during transmission. If you can give more information, I would see further. I'm closing this issue for now.

simonduq:
So I'm not so sure about the internals of cooja motes on the java side, but, maybe, what happens is:
1. we set `simRadioHWOn`
2. we call `cooja_mt_yield`
3. Cooja will run one tick and call us back after that. Perhaps it will accept any incoming packet during that tick?
4. we unset `simRadioHWOn`

In that case what we need is probably a cooja fix :)
Or maybe we don't even need to set `simRadioHWOn` in order to transmit?

yatch:
To me, It seems the radio doesn't accept an incoming frame during transmission.

https://github.com/contiki-ng/cooja/blob/63538bbb882ba06a7b8cf97c11ce2fe4d22e4f88/java/org/contikios/cooja/contikimote/interfaces/ContikiRadio.java#L183

[...]

PR #855, opened 2019-02-22
Fix a possible bug in TSCH communication

yatch:
If CCA is disabled, a node having `TSCH_RADIO_ON_DURING_TIMESLOT` set with `0` cannot transmit a frame during a TSCH TX slot because the radio is off when transmission is triggered. In addition, such a node cannot send back an ACK frame in a TSCH RX slot.

The fix is a by-product of Issue #833, that is closed by this PR.

### DIFF
--- a/arch/platform/cooja/dev/cooja-radio.c
+++ b/arch/platform/cooja/dev/cooja-radio.c
@@ -197,8 +197,6 @@ channel_clear(void)
 static int
 radio_send(const void *payload, unsigned short payload_len)
 {
-  int radiostate = simRadioHWOn;
-
   /* Simulate turnaround time of 2ms for packets, 1ms for acks*/
 #if COOJA_SIMULATE_TURNAROUND
   simProcessRunValue = 1;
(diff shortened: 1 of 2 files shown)


simonduq:
OK, does that make any actual difference? Looks like in both cases, `simRadioHWOn` will be set, then the packet transmitted, then unset.

Also, unless we really need a change to fix the cooja bug, I think it's a good idea to keep the same `transmit` behavior for cooja motes and H/W, that is, automatically turn radio on, send, then restore previous on/off state.

yatch:
Oh, thanks! I didn't know that this is kind of a common behavior of a radio H/W. I was wondering why `#define TSCH_CONF_RADIO_ON_DURING_TIMESLOT 1` is found only in `arch/cpu/cc26x0-cc13x0` and `arch/cpu/simplelink-cc13xx-cc26xx/`, although other platforms support TSCH. Now, that makes sense.

I'm fine to keep the current `cooja-radio.c`. This conversation here could help developers understand why the cooja radio can be turned on temporarily during transmission.

PR #867, opened 2019-03-06
Enhanced safety in the IPv6 extension header processing

nvt:
This PR fixes a variety of vulnerabilities in the processing of IPv6 extension headers.

* All extension header length variables are now 16 bits instead of 8 bits to avoid overflows observed in the processing of certain packets.

* Check that uipbuf calls in rpl-classic and rpl-lite succeed before continuing with packet processing.

* Reset uip_ext_len to 0 when creating an ICMP error message. In some cases, the processing of extension headers would involve a bad uip_ext_len, leading to the creation of a new ICMP error message. But the invalid uip_ext_len was later used again when checksumming the outgoing ICMP packet, and thus accessing memory out of bounds.

* Extension header removal functions now return true or false to indicate whether the operation was successful.

* Explicitly check whether the source IP address is a multicast address in uip_icmp6_error_output(), rather than having a comment that states that we assume that the caller has checked it.

* Immediately check the return value of uipbuf_get_next_header(), so we do not risk dereferencing a null pointer.


kostis:
Is it possible to review and merge this PR soon (or at least make it up-to-date with the latest `develop`) ?

kostis:
If it's not too much work, can I suggest that appropriate test(s) also gets added so that vulnerabilities and crashes identical/similar to those that this PR fixes are avoided in the future?

simonduq:
Would be great with new CI tests for these fixes I agree :)
Simply feeding a native node with a faulty input should do, what do you think @nvt

nvt:
Yes, I can arrange a crash test with a bunch of problematic input packets.

simonduq:
@nvt any chance we get the test within the next week or two for 4.3?

nvt:
@simonduq Done. :-)

g-oikonomou:
So we're good to merge this as well as #871, and the tests are in separate pull. Right?

PR #883, opened 2019-03-18
Add "atomic" API to ringbufindex

debug-ito:
Second of three-step alternative for #792. The 1st one was #825.

This pull-request adds `ringbufindex_atomic_get` and `ringbufindex_atomic_put` functions to lib/ringbufindex. Under the hood, these functions use `atomic_cas_uint8` introduced in #825. As a result, the atomic get (put) is interrupt-safe with another atomic get (put), respectively.

Now that ringbufindex has two types of APIs, the conventional "peek" API and the new "atomic" API, this pull-request adds a guideline on which API the user should use to the document of ringbufindex. Thread-safety of these APIs are also discussed in the document.

Finally, this pull-request adds an example in examples/libs/ringbufindex-interrupt. This example demonstrates how the "peek" and "atomic" APIs can be used to resolve asynchronous access to a single message queue. Using this example, I confirmed that the new "atomic" API could arbitrate asynchronous get or put to a single queue.

### DIFF
--- a/examples/libs/ringbufindex-interrupt/Makefile
+++ b/examples/libs/ringbufindex-interrupt/Makefile
@@ -1,0 +1,5 @@
+CONTIKI_PROJECT = ringbufindex-interrupt
+all: $(CONTIKI_PROJECT)
+
+CONTIKI = ../../..
+include $(CONTIKI)/Makefile.include
--- a/examples/libs/ringbufindex-interrupt/README.md
+++ b/examples/libs/ringbufindex-interrupt/README.md
@@ -1,0 +1,79 @@
+# ringbufindex-interrupt example
+
+This example demonstrates and tests interrupt-safety of
+lib/ringbufindex.
+
+## What it does
+
+It prepares a message queue managed by ringbufindex. Two producers
+put some messages into the queue. One producer runs in normal
+context, and the other in interrupt context (using rtimer). At the
+same time, two consumers (in normal and interrupt contexts,
+respectively) get messages from the queue and stores them in
+dedicated memory regions.
+
+                          put             get
+       NORMAL (Producer) --+               +--> (Consumer) --> [store]
+                           |               |
+                           +-> --------- >-+
+                               | queue |
+                           +-> --------- >-+
+                           |               |
+    INTERRUPT (Producer) --+               +--> (Consumer) --> [store]
+
+After all activities are finished, it checks the content of the two
+message stores. If the stores do not contain the exact messages put
+by the producers, it prints error messages.
+
+After the check, it resets itself.
+
+## Configuration
+
+In project-conf, you can configure number of messages two produers
+put to the queue and two consumers get from the queue. You can also
+configure whether the program uses "peek" or "atomic" APIs to access
+the queue. See ringbufindex's doc for detail about those APIs.
+
+## Expectations
+
+This example should run without error in the following four cases.
+
+- The "put interrupts get" case
+
+        NORMAL_PUT_NUM    == 0
+        INTERRUPT_PUT_NUM  > 0
+        NORMAL_GET_NUM     > 0
+        INTERRUPT_GET_NUM == 0
+        PUT_USE_PEEK      == 1
+        GET_USE_PEEK      == 1
+
+- The "get interrupts put" case
+
+        NORMAL_PUT_NUM     > 0
+        INTERRUPT_PUT_NUM == 0
+        NORMAL_GET_NUM    == 0
+        INTERRUPT_GET_NUM  > 0
+        PUT_USE_PEEK      == 1
+        GET_USE_PEEK      == 1
+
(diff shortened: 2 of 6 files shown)


debug-ito:
Could anyone review this?

simonduq:
So wait, are you saying ringbuf and ringbufindex are not interrupt-safe? Interrupt-safety is the reason we have them in the first place, so I hope you mean something else :)

debug-ito:
I'm not sure about ringbuf, but unfortunately ringbufindex is not interrupt-safe in some cases. For example, if one is putting an item A (with the "peek" API) in a normal context and another interrupts it and puts item B, the item A is overwritten by B. I confirmed this problem with the example  examples/libs/ringbufindex-interrupt.

ringbufindex is still interrupt-safe in "put interrupts get" and "get interrupts put" cases. See [my updated documentation](https://github.com/debug-ito/contiki-ng/blob/atomic-ringbufindex/os/lib/ringbufindex.c#L41).

simonduq:
Thanks for the clarification, now I understand what this is about :)

debug-ito:
Could anyone review this p-r?

debug-ito:
Can we include this in v4.4?

simonduq:
So, I'm worried about how confusing the API is.

Ringbuf(index) is currently meant to be atomic under the assumption of a single reader and writer. I feel we should either leave it as it is, or change it to support multiple readers and writers (and maybe arbitrary lengths as opposed to power-of-two). But the proposed API here mixes both. Why would we even keep the existing `put` and `get` now with the `atomic_` versions?

I'd like to hear more opinions from the merge team.

debug-ito:
The reason why there are "peek" and "atomic" versions for "get" and "put" is that we need to support both existing and additional scenarios of interruption.

The existing scenario is, as you pointed out, the single-reader single-writer case. The reader can interrupt the writer and the writer can interrupt the reader, but the reader (writer) does not interrupt another reader (writer).

The additional scenario I want to support is that the writer can interrupt another writer, which can happen in theory if you call `process_post` both in normal and interrupt contexts. This scenario is possible if you use the "atomic" version for the writer. Unfortunately, the current "atomic" version does not support cross interruption, that is, the ("peek" or "atomic") reader must not interrupt the "atomic" writer. This limination is not a problem for the event queue, because there is only one reader running in normal context.

See also [the updated document on this topic](https://github.com/contiki-ng/contiki-ng/pull/883/files#diff-1d8462914f98b27fd5f22f62143fd33eR67).

I admit that this design is confusing and far from complete. It is a result of compromise. It was too difficult for me to design and implement a lock-free queue that support all interrupt cases with reasonable overhead. [This article](https://www.linuxjournal.com/content/lock-free-multi-producer-multi-consumer-queue-ring-buffer), which I used as a reference, discusses the complete implementation, but I'm not even sure if we can do it on Contiki-NG and ringbufindex. Maybe we should try it.

Another possible way is to make a new ringbuf module for the "atomic" API and keep ringbufindex intact. However, I think that the new module is gonna have something like the "peek" API, because the "atomic" API alone does not support cross interruption.

simonduq:
So, the `atomic` version can have multiple readers or multiple readers, but does not allow reader<->writer interruption?

debug-ito:
> So, the `atomic` version can have multiple readers or multiple readers, but does not allow reader<->writer interruption?

Correct. That's why we need the conventional "peek" API anyway.

Maybe the word "atomic" is misleading. Or maybe we can just use sys/critical to protect access to a queue.

simonduq:
So, don't get me wrong, I understand the limitations of the current API. But I'm concerned about complexity, both code-wise and API-wise. The semantics are really complex. And we don't have a use for the new API anywhere in NG, do we? Or did you have another contrib pending that would build on the new API?

My (weakly held) opinion is that we either want a new ringbuf, fully atomic with no caveat, that replaces the one we have today, or simply stick to the current one.

But again it's not just my call, please others chime in!

debug-ito:
> Or did you have another contrib pending that would build on the new API?

Exactly. If this p-r is merged, I'll submit another p-r that re-implements the event queue (sys/process.c) with the new API. That will fix #764.

joakimeriksson:
@debug-ito - could we target a fully atomic API so that it really is atomic - or possibly have a separate module and not changed the old "semantics" in the same module? I would love a fully atomic version - since that would not need the user of the API to understand specific limitations in detail.

debug-ito:
Thanks for comment. Maybe it's possible to implement a fully atomic version. I'll think about it.

simonduq:
@debug-ito so, this is now addressed in https://github.com/contiki-ng/contiki-ng/pull/1151 ?
Do we close this one then?

debug-ito:
Thanks for comment. Yes, this will be addressed by #1151, so you can close this.

g-oikonomou:
Thank you. Closing

Issue #887, opened 2019-03-30
Trickle Timer Library Behaviour


edrose:
I've noticed some undesirable behaviour in how the trickle timer library handles timers that aren't currently running.

If I call `trickle_timer_inconsistency()` or `trickle_timer_reset_event()` on a timer that has been configured but is not currently running, a new interval will be calculated and a `ctimer` callback scheduled. In the case of timers that have never been set using `trickle_timer_set()` the timer expiry will call a function pointed to by an uninitialised function pointer, causing undefined behaviour if that pointer happens to not be `NULL`. If the timer has been set before, then it'll simply restart the timer with the callback pointer last provided in `trickle_timer_set`. In both cases subsequent calls to `trickle_timer_is_running()` will indicate that the timer is running.

From a user perspective, I had timers running with `NULL` callback function pointers running which did nothing because calls to `trickle_timer_inconsistency()` were made before the timer started. Calling `trickle_timer_set()` would have solved my issue, but this was never called in the application since `trickle_timer_is_running()` indicated that it was already running. This resulted in trickle timer callbacks simply never occurring.

@g-oikonomou

g-oikonomou:
I'll confirm this as a bug.

When I wrote the library I was assuming that trickle timers would only be manipulated (reset, inconsistency etc) while running. I am not sure what would be the preferred way to handle this, but clearly something we should fix.

g-oikonomou:
So I think we have multiple problems here:

* The library should never set the `ctimer` for a suspended trickle. `ctimer_set()` should only ever happen when the protocol explicitly requests a `trickle_timer_set()`. Therefore I think that  `trickle_timer_inconsistency()` should change such that it does absolutely nothing for suspended trickles.
* The library should behave the same for trickles that have never been started before as it does for trickles that have in the past been running but that are now stopped.
* Lastly and importantly, `trickle_timer_is_running()` should always return a consistent result.

> If I call trickle_timer_inconsistency() or trickle_timer_reset_event() on a timer that has been configured but is not currently running, a new interval will be calculated and a ctimer callback scheduled.

This should not happening, we don't want trickles starting themselves! Do you think a simple check inside `trickle_timer_inconsistency()` would fix this?

```C
void
trickle_timer_inconsistency(struct trickle_timer *tt)
{
  if(tt->i_cur == TRICKLE_TIMER_IS_STOPPED) {
    return;
  }
[...]
}
```

[...]

PR #896, opened 2019-04-09
Implemented bug fix to trickle timer

edrose:
Fix for issue #887.

This modifies the trickle timer library to guard some functions to prevent undesirable behaviour with the library. It also adds a couple of lines to timer configurations to ensure a consistent state of a timer is maintained throughout it's lifetime.

There are four main changes:

- Guard the `trickle_timer_consistency()` function so it returns and does nothing if a timer is stopped

- Guard the `trickle_timer_inconsistency()` function so it returns and does nothing if a timer is stopped

- Explicitly set `tt->i_cur` to `TRICKLE_TIMER_IS_STOPPED` in `trickle_timer_config()` to ensure that a timer created by that function is in the 'stopped' state before it's started.

- Explicitly set `tt->cb` to `NULL` in `trickle_timer_config()` so an undefined function can never be used as a callback.

### DIFF
--- a/os/lib/trickle-timer.c
+++ b/os/lib/trickle-timer.c
@@ -81,7 +81,7 @@ static void double_interval(void *ptr);
 static uint32_t
 wide_rand(void)
 {
-  return ((uint32_t)random_rand() << 16 | random_rand());
+  return (uint32_t)random_rand() << 16 | random_rand();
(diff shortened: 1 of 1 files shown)


g-oikonomou:
Fixes #887

Issue #800, opened 2018-12-17
Maximal packet size unification


atiselsts:
At the moment there are a large number of configuration settings that try to guess the maximal payload size the radio driver can transmit (which is 125 bytes for IEEE 802.15.4 2.4 GHz radios):

* NETSTACK_RADIO_MAX_PAYLOAD_LEN
* SICSLOWPAN_MAC_MAX_PAYLOAD
* TSCH_PACKET_MAX_LEN
* CSMA_MAC_LEN

Some of them are used at places, some (`SICSLOWPAN_MAC_MAX_PAYLOAD`) are never used. These should be unified; ideally, the radio driver should tell the max packet size it can handle. However, this must be a `#define` since buffer sizes depend on this.
Some radio drivers use `127` or other constants directly; some radio drivers define their own constants for max packet sizes (which may be different for Rx and Tx, for example if Rx includes the FCS but Tx does not).

Also, the radio drivers should have a unified semantics when the upper layers try to send a packet that is too big. My proposal is to validate the size at the start of `transmit()`; do not accept 0-size or oversized packets (return `RADIO_TX_ERR`).

I'm not 100% sure how the packetbuf size is related to this. It's set to 128 at the moment.

atiselsts:
Are underscore #define names acceptable for constants? Then we could do this in a clean way without changing much code - allow each radio driver to define its own limits and make the other `#defines` dependent on that:

```
/* New code in  cc2538 radio driver's header */
#define cc2538_rf_driver_max_payload_len 125

/* Existing code in cc2538 platform's header */
#define NETSTACK_RADIO cc2538_rf_driver

/* In the netstack.h header */
#define NETSTACK_RADIO_MAX_PAYLOAD_LEN_XX(radio) radio##_max_payload_len
#define NETSTACK_RADIO_MAX_PAYLOAD_LEN_X(radio) NETSTACK_RADIO_MAX_PAYLOAD_LEN_XX(radio)
#define NETSTACK_RADIO_MAX_PAYLOAD_LEN NETSTACK_RADIO_MAX_PAYLOAD_LEN_X(NETSTACK_RADIO)

/* In a TSCH header */
#define TSCH_PACKET_MAX_LEN MIN(NETSTACK_RADIO_MAX_PAYLOAD_LEN, PACKETBUF_SIZE)

```

debug-ito:
[...]

Issue #799, opened 2018-12-17
TSCH packet size issue


atiselsts:
TSCH can add a few bytes to a packet when adding the security header. This can cause a packet to become oversized. At the moment the radio drivers do not always handle oversized packets well, for example the CC26xx driver accepts the packet and returns the `RADIO_TX_OK` code from the `transmit()` function.

There is a security problem in here. Consider this scenario:
   - node X joins the network
   - node X sends slightly oversized packets destined to some other node through node A
   - A tries but fails to forward the packet, but does not detect that, as the radio driver returns TX_OK
   - A does not get an ACK for the packet it never sent.
   - node X repeats sending the oversized packets with high rate
   - A thinks it's link to the next hop is bad. A selects alternative routing path. The same problem repeats with the alternative nexthop.
   - A drops from the RPL network as it thinks all nexthops are bad.

Node X can be malicious or honest but misconfigured node.

There is a Cooja script and config files here to repeat the problem.
https://github.com/atiselsts/contiki-ng/blob/bugfix/packet-sizes/examples/rpl-udp/rpl-udp-cooja.csc

The various values of `UDP_PAYLOAD_SIZE` in the `project-conf.h` file correspond to different behaviors.

simonduq:
So, I think what we want is to fix the MAC header len estimate for TSCH. The faulty function is:
https://github.com/contiki-ng/contiki-ng/blame/develop/os/net/mac/tsch/tsch.c#L1160

as it calls `NETSTACK_FRAMER.length()` without setting the security parameters first.
So what we'd want in principle is to set the required packetbuf attributes exactly as in:
https://github.com/contiki-ng/contiki-ng/blame/develop/os/net/mac/tsch/tsch.c#L1046

*but*

I'm confused now because it looks like the framer function used to compute the header length:
https://github.com/contiki-ng/contiki-ng/blob/develop/os/net/mac/framer/frame802154.c#L358

includes the security auxiliary header len but not the MIC len!
And if that's correct then we also have a bug in CSMA.
Am I missing something there?

atiselsts:
The security problem is now fixed with #931.

The functionality problem remains in this corner case: silent failure in forwarding when fragmentation is disabled. But it's probably just a drawback of the 6LoWPAN specification, rather than something Contiki-NG specific. Closing the issue.

PR #908, opened 2019-04-18
Maximal packet size unification

atiselsts:
This PR solves network stability problem reported in #799 by implementing the strategy proposed in #800.

The idea is that every radio driver should define a publicly visible constant that describes the maximal packet size (payload) supported by that driver. For 802.15.4 drivers it is 125 (the max MAC layer frame is 127 bytes, but that includes 2 bytes FCS). For example, along with `cc2538_rf_driver` this should be defined:

  #define cc2538_rf_driver_max_payload_len       125

This constant is then mapped to the name `NETSTACK_RADIO_MAX_PAYLOAD_LEN`, selected through the `NETSTACK_RADIO` define.

In `prepare` and `transmit` functions of the radio drivers now always return `RADIO_TX_ERR` code if the packet is too big to be sent.

This PR also removes the `#define SICSLOWPAN_CONF_MAC_MAX_PAYLOAD`, that was not used in the code.

This PR does not aim to improve the behaviour at the sicslowpan layer (in the scenario described in #799, the sicslowpan layer still thinks the packet size is small enough to be send without fragmentation). It just sanity checks the size at radio driver's level and makes sure to return error code to upper layers instead of failing silently.


arch/platform/jn516x/dev/micromac-radio.c
```diff
@@ -463,11 +467,8 @@ prepare(const void *payload, unsigned short payload_len)
 static int
 send(const void *payload, unsigned short payload_len)
 {
-  if(prepare(payload, payload_len) == 0) {
-    return transmit(payload_len);
-  } else {
-    return RADIO_TX_ERR;
-  }
+  prepare(payload, payload_len);
```
simonduq: why no longer check the return value of `prepare`?

arch/dev/cc1200/cc1200-conf.h
```diff
@@ -72,6 +72,8 @@
 #else
 #define CC1200_MAX_PAYLOAD_LEN           127
 #endif
+
+#define cc1200_driver_max_payload_len    CC1200_MAX_PAYLOAD_LEN
```
simonduq: Shouldn't this be 125? Won't this result in 127+2 bytes CRC frames, even without 802.15.4g?

arch/platform/jn516x/dev/micromac-radio.c
atiselsts: not strictly needed. but I can change it back

arch/dev/cc1200/cc1200-conf.h
atiselsts: Not sure - you're the expert on this radio! I got the impression  that 127 is ok after reading the comment above

simonduq:
Looks good to me. Is this read?

Issue #897, opened 2019-04-12
Multicast CI Tests not testing all Drivers


edrose:
I've discovered a bug in the Travis CI tests for the multicast drivers. The test isn't being run for all of the intended. The issue is with how the build system is dealing with the parameters passed to it. The tests I'm referring to are numbers 15-18 in the `tests/07-simulation-base` folder, however it's possible other tests are also affected by this.

The problem stems from how the multicast engines are swapped out between the tests. Setting the multicast engine in contiki-ng is done using the preprocessor macro `UIP_MCAST6_CONF_ENGINE`; this can either be done in `project-conf.h` or using a command line macro definition when building. When cooja builds motes for the tests, the engine is changed using the command line method. When using a clean working directory, this works as expected and the specified engine is used. Cooja creates the mote, and removes the object file for the top-level program (`root.c`/`forwarder.c`/`sink.c`). When the next test is run, make only recompiles the top level program, since that object file was deleted and no other files have changed. This is then linked to the rest of the OS which was compiled for the previous test with the previous test's multicast driver. As a result, the only test being performed is for `roll-tm`, which just happens to be the first test the automated script gets to.

[I've created a working example for this in the `trash/travis-issue` branch in my fork](https://github.com/contiki-ng/contiki-ng/compare/develop...edrose:trash/travis-issue). In this branch, I've deliberately sabotaged the smrf driver so that it never accepts any packets, and this behaviour can be confirmed by running the cooja test at `tests/07-simulation-base/16-cooja-multicast-11-hops-smrf.csc` which will fail when built with a clean build tree. However, when `distclean`ing the `examples/multicast`folder and then running `make` inside of the `tests/07-simulation-base` folder, all tests will pass. [This is reflected in the travis tests that have passed here.](https://travis-ci.org/edrose/contiki-ng/builds/519201020).

The solution to this is to add the `-B` option to the command for creating cooja motes in affected tests, [as I've done in this branch fixing the issue for my MPL implementation.](https://github.com/edrose/contiki-ng/compare/feat/mpl...edrose:bug/travis-tests) I haven't merged this into my feature branch since it's an issue that needs to be fixed separately. There also may be more places where this needs to be fixed that I'm unaware of so I thought I'd create this issue explaining the problem so it can be properly fixed everywhere that it occurs.

edrose:
[...]

PR #918, opened 2019-04-21
Fixed issues Running Tests Locally

edrose:
This fixes a further issue raised in #897 whereby tests run on a local machine by entering the `tests` directory and running `make` would cause test fails due to unclean build directories. After these changes, local testing works as expected with the exception of out of tree tests, which fail of the out of tree directory is not properly configured.


simonduq:
Thanks for the fix!

Issue #296, opened 2018-02-16
Add support for MPL (RFC7731)


g-oikonomou:
We currently lack support for MPL ([RFC7731](https://tools.ietf.org/html/rfc7731)).

As I have discussed in the past, our current trickle-multicast engine is based on version 3 of the respective I.D., but there have been significant changes since then.

I'm adding this to our long-term roadmap.

g-oikonomou:
#919

PR #919, opened 2019-04-22
Implemented MPL

edrose:
This implements the MPL multicast engine, as defined in [RFC7731  Multicast Protocol for Low-Power and Lossy Networks (MPL)](https://tools.ietf.org/html/rfc7731).

There are four parts to the PR:
- Implementation of the multicast engine itself
- Small modifications to the uIP stack to support processing MPL Hop-by-Hop headers and MPL ICMP Messages
- Updating the multicast example to support MPL, and making MPL the default multicast engine since it supersedes roll-tm
- Creating travis tests for MPL to run on each push

---
**A Note on MPL Addresses**

Multicast devices using MPL form a ["domain"](https://tools.ietf.org/html/rfc7731#section-4.1) rather than a "group". A domain is a scope zone - ["a connected region of topology of a given scope"](https://tools.ietf.org/html/rfc4007#section-5). Therefore all intermediate nodes using MPL forwarding must both support MPL and be members of the correct MPL domain. By default all MPL forwarders subscribe to the well-known address `ALL_MPL_FORWARDERS` which is `FF03::FC`. It is for this reason that the multicast examples have been modified with MPL specific code, so the intermediate nodes are in the same domain as the root and sink nodes.

MPL nodes also subscribe to two addresses per domain: an address of scope 3, 4, or 5 which deals with data messages, and an address of scope 2 (link local) which deals with control messages. If only the higher scoped address is subscribed to by a program, the MPL implementation will notice and subscribe to the link local version, however it may take up to a minute for that to happen. 

I think those two things might catch people out and may need documentation. 

---
**Future Work** 

Off the top of my head, one future improvement could be to support disabling of Reactive Forwarding. MPL defines two forwarding methods: Proactive Forwarding schedules re-transmissions of messages as soon as they are received, regardless of whether any neighbours are known to be missing them; Reactive Forwarding occurs when a neighbour signals that it's missing a message using MPL ICMP Control messages, and the re-transmissions are scheduled based on that. Currently Proactive Forwarding can be disabled, however Reactive Forwarding is always enabled. Being able to disable reactive forwarding is supported by the spec, and could also drastically reduce the compiled size of the engine at the expense of message loss correction.


edrose:
This also would resolve #296.

g-oikonomou:
As discussed folks, I've been following this work for quite a while myself and been offering some advice. 👍 from me, but it'd be nice if someone else were to review this PR.

os/net/ipv6/multicast/mpl.c
```diff
@@ -0,0 +1,1807 @@
...
+ * \brief Clear a seed id value to zero
+ * a: Value to clear
+ */
+#define seed_id_clr(a) (memset((a), 0, sizeof(seed_id_t)))
+/*---------------------------------------------------------------------------*/
+/* Buffered message set
+ *  This is implemented as a linked list since the majority of operations
+ *  involve finding the minimum sequence number and iterating up the list.
+ */
+struct mpl_msg {
```
simonduq: consider re-arranging the fields to play nicely with 32-bit alignment?

os/net/ipv6/multicast/mpl.c
```diff
@@ -0,0 +1,1807 @@
...
+  uip_ip6addr_t srcipaddr;
+  uint8_t seq;
+  uint8_t data[UIP_BUFSIZE];
+  uint16_t size; /* Side of the data stored above */
+  uint8_t e; /* Expiration count for trickle timer */
+  struct mpl_msg *next; /* Next message in the set, or NULL if this is largest */
```
simonduq: any reason not to use the list module?

os/net/ipv6/multicast/mpl.c
```diff
@@ -0,0 +1,1807 @@
...
+  /* Iterate over seed set to create payload */
+  for(locssptr = &seed_set[MPL_SEED_SET_SIZE - 1]; locssptr >= seed_set; locssptr--) {
+    if(SEED_SET_IS_USED(locssptr) && locssptr->domain == dom) {
+      locsiptr->min_seqno = locssptr->min_seqno;
+      SEED_INFO_CLR_LEN(locsiptr);
+      SEED_INFO_CLR_S(locsiptr);
+
+      /* Try setting our source address to global */
+      addr = uip_ds6_get_global(ADDR_PREFERRED);
+      if (addr) {
```
simonduq: code style: no space between ìf` and `(`. If seen other instanced of this in this PR, not only on `if` statements, please fix

os/net/ipv6/multicast/mpl.c
```diff
@@ -0,0 +1,1807 @@
...
+        SEED_INFO_SET_S(locsiptr, 1);
+        break;
+        case 2:
+        seed_id_host_to_net(&((struct seed_info_s2*)locsiptr)->seed_id, &locssptr->seed_id);
+        SEED_INFO_SET_S(locsiptr, 2);
+        break;
+      }
+
+      /* Populate the seed info message vector */
+      memset(vector, 0, sizeof(uint8_t) * 32);
```
simonduq: why not sizeof(vector) ?

os/net/ipv6/multicast/mpl.c
```diff
@@ -0,0 +1,1807 @@
...
+        locsiptr = ((void *)locsiptr) + sizeof(struct seed_info_s2) + (sizeof(uint8_t) * vec_size);
+        payload_len += sizeof(struct seed_info_s2) + (sizeof(uint8_t) * vec_size);
+        break;
+      case 3:
+        memcpy(((void *)locsiptr) + sizeof(struct seed_info_s3), vector, sizeof(uint8_t) * vec_size);
+        locsiptr = ((void *)locsiptr) + sizeof(struct seed_info_s3) + (sizeof(uint8_t) * vec_size);
+        payload_len += sizeof(struct seed_info_s3) + (sizeof(uint8_t) * vec_size);
+        break;
+      }
+    }
```
simonduq: looks like a lot of redundancy in the code above, both within `case` clauses (e.g. `sizeof(uint8_t) * vec_size` etc.) and among clauses (all closes do more or less the same thing, only with different offsets and sizes)

os/net/ipv6/multicast/mpl.c
```diff
@@ -0,0 +1,1807 @@
...
+    if(trickle_timer_is_running(&locdsptr->tt)) {
+      mpl_trickle_timer_inconsistency(locdsptr);
+    }
+  } else {
+    LOG_INFO("Domain is consistent \n");
+    trickle_timer_consistency(&locdsptr->tt);
+  }
+
+discard:
+  uip_len = 0;
```
simonduq: You might need to use `uipbuf_clear` instead

edrose:
Hi @simonduq,

Thanks for the feedback, I'll go through it now and respond to your comments. I'll try and implement any changes this afternoon as well.

Ed

os/net/ipv6/multicast/mpl.c
edrose: I think the reason for this was that I initially didn't design it to be a linked list. It evolved that way once I realised how many loops were being used to access the elements in order. I agree that it should be modified to use the linked list library though, so I'll fix this.

os/net/ipv6/multicast/mpl.c
edrose: > looks like a lot of redundancy in the code above, within case clauses

Ah yes I see. I can sort that.

>and among clauses (all closes do more or less the same thing, only with different offsets and sizes)

How would you suggest tackling this? I can move most of the `+ vec_size` parts out of the cases, however I feel that affects the readability as the pointer arithmetic is then divided in two and it's less obvious as to it's intention.

simonduq: You're probably right, do what you think is best and I'll take another look!

edrose: I've left it like this.

```c
switch(SEED_INFO_GET_S(locsiptr)) {
      case 0:
        memcpy(((void *)locsiptr) + sizeof(struct seed_info), vector, vec_size);
        locsiptr = ((void *)locsiptr) + sizeof(struct seed_info) + vec_size;
        payload_len += sizeof(struct seed_info) + vec_size;
        break;
      case 1:
        memcpy(((void *)locsiptr) + sizeof(struct seed_info_s1), vector, vec_size);
        locsiptr = ((void *)locsiptr) + sizeof(struct seed_info_s1) + vec_size;
        payload_len += sizeof(struct seed_info_s1) + vec_size;
        break;
      case 2:
        memcpy(((void *)locsiptr) + sizeof(struct seed_info_s2), vector, vec_size);
        locsiptr = ((void *)locsiptr) + sizeof(struct seed_info_s2) + vec_size;
        payload_len += sizeof(struct seed_info_s2) + vec_size;
        break;
      case 3:
        memcpy(((void *)locsiptr) + sizeof(struct seed_info_s3), vector, vec_size);
        locsiptr = ((void *)locsiptr) + sizeof(struct seed_info_s3) + vec_size;
        payload_len += sizeof(struct seed_info_s3) + vec_size;
        break;
}
```
---
Do you think it would be better to move the `+ vec_size` parts from the 2nd and 3rd lines in each case into their own lines at the bottom (like below), or do you think I should keep it as it is?

```c
 locsiptr = ((void*)locsiptr) + vec_size;
 payload_len += vec_size;
```

g-oikonomou: I am perhaps missing something, but isn't the only difference the size of the seed info structure? Why not store this in a variable, say `seed_info_len` in the four cases and then

```C
    memcpy(((void *)locsiptr) + seed_info_len, vector, vec_size);
    locsiptr = ((void *)locsiptr) + seed_info_len + vec_size;
    payload_len += seed_info_len + vec_size;
```

Could that work?

edrose: No that works and I like that solution. Thanks!

edrose:
I believe all of those points have been fixed.

I've rebased them into my clean branch, but if you want to see [a diff of the changes you can see it on my messy branch here.](https://github.com/edrose/contiki-ng/compare/f9f98bad521a740e1a383b0cea8fad8a3263ef6a...edrose:feat/mpl) There are quite a few changes due to the move to the list library, however, so I'm not sure how useful that is.

simonduq:
Thanks a lot, looks great!

PR #922, opened 2019-04-22
Fix CC13x0 prop mode under CSMA

g-oikonomou:
CC13x0 prop mode is broken under CSMA.

The culprit is the `receiving_packet()` function: The current version of this function relies on the value of the `MDMSOFT` interrupt flag. The radio sets this flag when it detects a sync word, but it does _not_ clear it automatically.

The current version of the function checks the flag to determine whether frame RX has started. It then assumes that the caller will keep calling the function repetitively until frame RX has finished, at which point we can clear the flag. This assumption is correct for TSCH and the logic works fine.

This assumption however is _not_ correct for CSMA, which will sometimes make a one-off call to this function here. In these cases, the current logic of how/when to clear `MDMSOFT` will fail and the function will very often erroneously signal ongoing frame reception. This will cause aborted frame transmission attempts.

This pull fixes the bug by providing two flavours of the function. One flavour is for TSCH, whereby we retain the current logic. The other flavour is for CSMA, in which case we perform a clear channel assessment and we signal ongoing frame reception if the channel is found busy.

While at it, the pull disables the `MDMSOFT` interrupt: We don't need the radio to raise an interrupt since we are not acknowledging it with a handler in the main CPU. The radio will set the flag irrespectively.


atiselsts:
In the future, we should better handle noisy environments, but for now this should be good enough

g-oikonomou:
> In the future, we should better handle noisy environments, but for now this should be good enough

Agree.

The only way I can think of that would achieve this would be to rely on `MDMSOFT`. As I have commented before as well as in the code, for CSMA the problem is choosing a logic to clear this flag reliably. Perhaps we can enable all relevant `RX_xyz` interrupts and clear it in there:
 * `RX_OK`
 * `RX_NOK` (CRC error)
 * `RX_IGNORED` (CRC OK, but ignored due to frame filtering)
 * `RX_ABORTED`
 * Perhaps also `RX_BUF_FULL` (although I suspect this one will not happen without also `RX_OK`, so perhaps redundant.

Then `receiving_packet()` simply inspects this value. Possible problem:  Next sync word gets detected before the interrupt fires, and we end up clearing the flag erroneously.

PR #926, opened 2019-04-23
build system: warn if using compiler versions known to be problematic

atiselsts:
This hopefully will make Contiki-NG more user friendly. The idea is to nudge users to upgrade / downgrade their toolchains when their currently installed version is one of those that don't work well with Contiki-NG.

### DIFF
--- a/Makefile.gcc
+++ b/Makefile.gcc
@@ -1,0 +1,27 @@
+# -*- makefile -*-
+#
+# GCC specific definitions and actions
+#
+
+GCC_MAJOR_VERSION := $(shell $(CC) -v 2>&1 | grep "gcc version" | cut -b 13)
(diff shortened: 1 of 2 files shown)


debug-ito:
Hi @atiselsts 

I wonder what kind of breakage arm-none-eabi-gcc 6.3.x causes. Is there any issue report on it? Could you show me some examples?

Thanks in advance.

atiselsts:
@debug-ito The issue I've seen is this: an executable is produced successfully by the compiler, but after flashing it into a device it fails to start. E.g. on CC2650 launchpad.

Don't think there was ever an issue on this github repo, but there was a complain about the same recently (before this PR) on contiki-ng gitter about the same thing.

debug-ito:
Thanks!

PR #927, opened 2019-04-23
fix automatically setting TSCH node as a coordinator when it is the RPL root

atiselsts:
The intention is that `NETSTACK_ROUTING.root_start();` should be sufficient to also set TSCH in the coordinator mode. This is not currently happening.

The new code uses the magic number `1`. However, from my reading of the existing code that is ok, as the normalized root rank is always going to be equal to 1. Also, one is a special number, exception from the "no magic numbers" rule.


atiselsts:
My bad, the problem was somewhere else and the existing code is ok

PR #931, opened 2019-04-25
TSCH: account for security header in max_payload

simonduq:
This is a partial fix for https://github.com/contiki-ng/contiki-ng/issues/799

What I believe is still missing is the accounting of the security MIC in the framer.
Affects both TSCH and CSMA.


atiselsts:
Looking at the code, yes I don't see where MIC is taken into account, since that is not part of the header. So the MIC length probably is still a problem that remains to be fixed.

`LLSEC802154_MIC_LEN(packetbuf_attr(PACKETBUF_ATTR_SECURITY_LEVEL))` should give MIC size in bytes, no?

atiselsts:
@simonduq could you add accounting for the MIC length so we can merge this?

atiselsts:
Alternatively, I suppose I can add a commit myself to this PR, using the `LLSEC802154_MIC_LEN` macro.

simonduq:
Please feel free to go ahead and add a commit :) Thanks!

simonduq:
OK I did add the MIC!

Please review; and note this is untested.

Issue #882, opened 2019-03-15
Unsure about the "Sanity-check the timestamp" logic in tsch.c


atiselsts:
Why do we need this check?

It seems to slow down joining a lot in Cooja (using coojamotes). After disabling the check, they join just fine. The same may be happening with real hardware (CC1310), at least we see this check failing.

atiselsts:
#944 fixes this.

PR #942, opened 2019-05-02
Reduce noise on TSCH joining

atiselsts:
I've seen a number of people complaining and being worried about the "failed to parse EB" messages.
Almost every time that means the node has received something that is not an EB, but part of the normal network traffic (data or ACK packet).
This PR changes the logging severity of these messages from ERR to INFO.
If also silently discards "packets" that have zero size without attemping to parse them.

Finally, it increases the allowed time discrepancy for EB packets (see issue #882) from single TSCH slot to two seconds, which is empirically observed to be a max value that is seen in normal operation.


g-oikonomou:
Accidental commit on the submodule?

PR #943, opened 2019-05-02
SimpleLink tickless mode

tiepettersen:
Tickless mode was implemented by maintaining a wakeup clock instance in the clock module. Every time the device enters idle and wants to drop to some low-power mode, the wakeup clock is armed with the following logic:
* If etimer is pending, the etimer wakeup time is the the next etimer expiration time. If etimer is not pending, then no timeout.
* If Watchdog is enabled, the watchdog wakeup time is the Watchdog expiration time with a 1 ms margin. If watchdog is not enabled, then no timeout.
    * The margin is added to ensure we are able to possibly enter some low-power mode, wakeup, and clear the watchdog timer before it times out within the current expiration time.
* The wakeup time is the minimum timeout of etimer and watchdog timeout.
* If both timeouts are no timeout, then the wakeup clock is not armed.
* If the etimer is already expired or the watchdog is too close to timeout, then the wakeup clock is not armed, AND the device does not enter some low-power mode. Instead, it continuous back to the main loop.


g-oikonomou:
Hi Edvard, this is of course a very desirable feature. Other than the discussion we had offline, I don't see any show-stoppers here. Just yell when this is ready and we can merge.

PR #946, opened 2019-05-05
Remove watchdog_periodic() calls that may happen inside an interrupt context

### LINKED ISSUE #835: We are refreshing watchdogs inside interrupt contexts

g-oikonomou:
This pull removes calls to `watchdog_periodic()` from parts of the code that may get called inside an interrupt context.

There are two more calls that I identified as suspicious:

* `tsch-slot-operation::tsch_get_lock()`

https://github.com/contiki-ng/contiki-ng/blob/develop/os/net/mac/tsch/tsch-slot-operation.c#L207

I am not so sure about this one, but I think this also happens inside an interrupt so I've removed it. Happy to put it back if it's safe.

* `sicslowpan.c::send_packet()`

https://github.com/contiki-ng/contiki-ng/blob/develop/os/net/ipv6/sicslowpan.c#L1472

I think that one should be OK and I've left it there.

Fixes #835

### DIFF
--- a/arch/cpu/cc2538/usb/usb-arch.c
+++ b/arch/cpu/cc2538/usb/usb-arch.c
@@ -1114,13 +1114,8 @@ ep_tx(uint8_t ep_hw)
     len -= copy;
     ep->buffer->left -= copy;
 
-    /*
-     * Delay somewhat if the previous packet has not yet left the IN FIFO,
-     * making sure the dog doesn't bark while we're waiting
-     */
-    while(REG(USB_CSIL) & USB_CSIL_INPKT_RDY) {
-      watchdog_periodic();
-    }
+    /* Delay somewhat if the previous packet has not yet left the IN FIFO */
+    while(REG(USB_CSIL) & USB_CSIL_INPKT_RDY);
 
     write_hw_buffer(EP_INDEX(ep_hw), ep->buffer->data, copy);
     ep->buffer->data += copy;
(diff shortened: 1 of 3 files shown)


g-oikonomou:
> I am not so sure about this one, but I think this also happens inside an interrupt so I've removed it. Happy to put it back if it's safe.

Actually I'm a little bit confused by this. Under what circumstances will `tsch_get_lock()` get called while `tsch_slot_operation` is 1? And if that does indeed happen, how will this lock get released? Are we thinking nested interrupts here?

simonduq:
`tsch_get_lock` is only called outside of interrupt for processes to get hold of the TSCH lock, which will prevent any further slot operation until lock release. The busy wait will always be outside of interrupt, and will never exceed one timeslot length in duration. So, `watchdog_periodic` call is safe even though unnecessary.

The sicslowpan thing is definitely safe (only called outside of interrupt context)

g-oikonomou:
OK thanks, I've removed this hunk. Ready for review here then

PR #953, opened 2019-05-08
coap-engine,coap-transactions: Fix handling of separate response

akindestam:
A response from a server might come as a "separate response" where first an
empty message ACK is sent to tell the client there is no need to retransmit, and
that it should expect a response message soon. (RFC 7252 §5.2.2)

To stop the client from retransmitting when our message has already been ACKed,
a field was added to coap_transaction_t that blocks the coap_sendto call in
coap_send_transaction, with the retrans_timer continuing to fire like normal so
that we can still time out if the response from the server never shows up.

The response message is sent with a new Message ID chosen by the server, and if
its type is CON it should be ACK:ed by the client to inform the server that the
response has been successfully received.

To be able to handle all of the above we also need to fix how transactions are
managed and matched; the response from the server won't be using the same
Message ID as in the request. According to RFC, request/response should be
matched using the token and endpoint information only (RFC 7252 §5.3), so
transactions were refactored to match on this information instead of matching on
Message ID (except in the case of an empty message ACK).

### DIFF
--- a/examples/coap/coap-example-server/resources/res-b1-sep-b2.c
+++ b/examples/coap/coap-example-server/resources/res-b1-sep-b2.c
@@ -76,7 +76,9 @@ res_post_handler(coap_message_t *request, coap_message_t *response, uint8_t *buf
 
     /* Send first block */
     coap_transaction_t *transaction = NULL;
-    if((transaction = coap_new_transaction(request_metadata.mid, &request_metadata.endpoint))) {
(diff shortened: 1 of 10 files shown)


akindestam:
Rebased on latest contiki-ng/develop. Also added some missing includes

simonduq:
Hi!

arurke:
Hi @akindestam, if you are still present and have the motivation, could you rebase this PR? I am interested in testing and reviewing it.

PR #954, opened 2019-05-08
coap-blocking-api: Signal timeout to the callback function

akindestam:
Currently there is no convenient way for the caller of COAP_BLOCKING_REQUEST
to tell if the request timed out.

This change calls the request_callback with a NULL pointer as the argument
on timeout. Thus the caller will be able to handle a timeout by checking
for NULL in the callback function.

### DIFF
--- a/examples/lwm2m-ipso-objects/example-server.c
+++ b/examples/lwm2m-ipso-objects/example-server.c
@@ -172,6 +172,12 @@ client_chunk_handler(coap_message_t *response)
 {
   const uint8_t *chunk;
   unsigned int format;
+
+  if(response == NULL) {
+    PRINTF("\nRequest timed out\n");
+    return;
+  }
+
   int len = coap_get_payload(response, &chunk);
   coap_get_header_content_format(response, &format);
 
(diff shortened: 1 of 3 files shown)


akindestam:
Rebased on latest develop

joakimeriksson:
I think there are more example code that use this API which might also need some NULL checks. The coap-client example? Otherwise I think it makes sense to indicate timeout!

examples/lwm2m-ipso-objects/example-server.c
```diff
@@ -172,6 +172,12 @@ client_chunk_handler(coap_message_t *response)
 {
   const uint8_t *chunk;
   unsigned int format;
+
+  if (response == NULL) {
```
joakimeriksson: should be without space after if:
if(response == NULL) {

akindestam:
Rebased on latest develop and updated with fixed if statement formatting, added explicit NULL check to coap-example-client (there was an implicit NULL check in coap_get_payload, but having an explicit example showing that response may be NULL is probably wise)

PR #957, opened 2019-05-10
remove the special TSCH EB frequency config from Cooja

atiselsts:
Apparently the Cooja motes in the past needed a specific config for TSCH. This reduces the usefulness of the Cooja mote as a test platform for TSCH enhancement development / benchmarking. After merging in #944 this issue might be fixed.

Let's see what Travis thinks!

### DIFF
--- a/arch/platform/cooja/contiki-conf.h
+++ b/arch/platform/cooja/contiki-conf.h
@@ -130,15 +130,6 @@ typedef unsigned long clock_time_t;
 
 #define UIP_ARCH_IPCHKSUM        1
 
-#if MAC_CONF_WITH_TSCH
-/* A bug in cooja causes many EBs to be missed at scan. Increase EB
-   frequency to shorten the join process */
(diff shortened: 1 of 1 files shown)


atiselsts:
For the start, this one needs #958 , otherwise the reduced EB frequency causes those examples to stop working.

simonduq:
Ah so you're saying the Cooja TSCH beacon misses might have been due to the timestamp sanity check fixed in https://github.com/contiki-ng/contiki-ng/pull/944 ?

Real good news if this is fixed now!

simonduq:
Did you check if the misses are actually gone? If I remember right, without the 4s EB config, tests would also work before. Only, the scan was taking way too long due to misses EBs.

simonduq:
Where are we on this?

atiselsts:
I will look at this later, there are many other PRs / potential PRs right now that may be higher priority. Will close for now.

PR #960, opened 2019-05-12
Provide non-returning assert() implementation

### LINKED ISSUE #959: lib/assert.h gets in the way and behaves in an unexpected fashion
Our `lib/assert.h` will always get included instead of an `assert.h` that may be shipping with the toolchain. This is because we pass `-I$(CONTIKI)/os/lib` in the command line. So our `assert.h` will get included even if a developer tries to indicate an intention to use the system header by using the angle-bracket form `#inculde <assert.h>`.

Now most toolchains we use ship `assert.h` since it is part of the C standard library. However, msp430-gcc (version 4.7.0 at least) does not, so we cannot just remove our own `lib/assert.[ch]`.

The situation is problematic, especially if we try to include third party libraries, where the developer anticipates standard behaviour from the `assert()` macro.

Specifically, when building with arm-gcc and our configuration, `assert()` expands to `__assert_func()` which is `__attribute__ ((__noreturn__))`. This allows the toolchain to make certain assumptions that cannot be made when using our own `assert.h`. In some cases, this causes the compiler to emit warnings for code that would otherwise build fine if the system `assert.h` were to be included.

The first thing that comes to mind as a fix is to only include our own `assert.h` when building with the msp430-gcc toolchain. However then the rest of the code would use the system-provided `assert()` which behaves as above. I am a little bit concerned by this, since we use `#include "lib/assert.h"` in various places, including under `os/` (e.g. in `csma-output.c`). Are we expecting any specific non-standard behaviour from `assert()` in those places?

Would we be breaking anything if we were to default to the system-provided `assert.h` where possible?

g-oikonomou:
As reported in #959, our `os/lib/assert.h` will always get included instead of an `assert.h` that may be shipping with the toolchain. In some cases this is very undesirable and breaks builds, while I cannot immediately think of any use-cases where we may actually want to be doing this.

Currently, `assert.h` is provided by all relevant toolchains except our recommended msp430-gcc v4.7.0.

This pull recommends moving our own `assert.[ch]` to their own `os/lib/assert/` directory. This allows us to explicitly select them by adding them as a `MODULE` if required. In doing so:

* We can enable the respective functionality where the system does not provide it. We do so for all msp430 builds
* We can overwrite the system-provided header if there is reason to do so

@yatch I see you are using `assert()` in various places in the 6tisch/6top code and tests. Do you need non-default behaviour anywhere? Any other comments here or in #959 ?

Fixes #959

### DIFF
--- a/os/lib/assert.c
+++ b/os/lib/assert.c
@@ -27,6 +27,7 @@
  * SUCH DAMAGE. 
  *
  */
+#include "lib/assert.h"
 
 #include <stdio.h>
 
@@ -34,8 +35,8 @@ void
 _xassert(const char *file, int lineno)
 {
   printf("Assertion failed: file %s, line %d.\n", file, lineno);
-  /*
-   * loop for a while;
-   * call _reset_vector__();
-   */
+
+#if !ASSERT_RETURNS
+  while(1);
+#endif
 }
--- a/os/lib/assert.c
+++ b/os/lib/assert.c
@@ -37,6 +37,8 @@ _xassert(const char *file, int lineno)
   printf("Assertion failed: file %s, line %d.\n", file, lineno);
 
 #if !ASSERT_RETURNS
+  printf("The firmware will stop running\n");
+  printf("A watchdog timer may restart this device\n");
   while(1);
 #endif
 }
--- a/arch/platform/cooja/contiki-conf.h
+++ b/arch/platform/cooja/contiki-conf.h
@@ -48,6 +48,8 @@
 
 #define COOJA 1
 
+#define ASSERT_CONF_RETURNS  1
+
 #define LEDS_CONF_LEGACY_API 1
 
 #ifndef EEPROM_CONF_SIZE
--- a/os/lib/assert.h
+++ b/os/lib/assert.h
@@ -31,13 +31,25 @@
 #ifndef ASSERT_H_
 #define ASSERT_H_
 
+#include "sys/cc.h"
+
+#ifdef ASSERT_CONF_RETURNS
+#define ASSERT_RETURNS ASSERT_CONF_RETURNS
+#else
+#define ASSERT_RETURNS 0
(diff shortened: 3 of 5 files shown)


g-oikonomou:
OK so now I see the reason why overwrite the system-provided `assert.h`...

That didn't quite work now, did it?!

g-oikonomou:
So it looks like for arm platforms at least we are nowhere near being able to build with the system-provided `assert()`.

I have changed this pull to propose a different approach. We conditionally provide a non-returning `assert()` implementation. I am proposing making this the default to match the system-provided behaviour. This at least allows us to build code that implicitly (intentionally or accidentally) assumes a non-returning `assert()` implementation.

I am not sure why this test fails, I cannot even run the test locally inside docker.

yatch:
@g-oikonomou 

> I see you are using assert() in various places in the 6tisch/6top code and tests. Do you need non-default behaviour anywhere?

Thanks for pinging me. My answer is "no". I don't see any problem in your proposal.

os/lib/assert.c
```diff
@@ -27,15 +27,16 @@
...
 _xassert(const char *file, int lineno)
 {
   printf("Assertion failed: file %s, line %d.\n", file, lineno);
-  /*
-   * loop for a while;
-   * call _reset_vector__();
-   */
+
+#if !ASSERT_RETURNS
+  while(1);
```
yatch: It would be nicer to users to print something before entering the while loop, like "The program (firmware) stops running."

g-oikonomou: Something like that?

yatch: Looks good. Thanks!

g-oikonomou:
@yatch don't hesitate to click the big green merge button if you are satisfied with this :P

yatch:
@g-oikonomou Thanks! Done 🎉

PR #964, opened 2019-05-15
Update Readme.md

simonduq:


yatch:
Hmm. Why does `continuous-integration/travis-ci/push` fail...? (Not sure differences between `travis-ci/push` and `tranvis-ci/pr`, though...)

PR #970, opened 2019-05-23
Merge master back into develop

g-oikonomou:
Merge master back into develop


simonduq:
Warning here: do not press "update branch", but rather do a forced merge.

Updating the branch would result in a commit to master (the source branch)

g-oikonomou:
I wonder whether we should be using an intermediate branch here in the future, just to make sure any accidental merges don't mess around with master

simonduq:
I think we should yes!

simonduq:
But we should still not do "upgrade branch" even on that one, so what we get is still a clean master-to-develop merge

g-oikonomou:
Yeah. But if someone does so accidentally, we won't mess master and a simple force-push to the branch in question will fix

PR #900, opened 2019-04-16
Packet processing test for the network stack.

nvt:
This PR adds a parsing test with binary input files that have been known to cause crashes or hangs in the network stack. Current targets are the sicslowpan and uip input functions, but more can be added later.

At the time of posting this, the test is expected to fail on the current develop branch, but PRs are being prepared to fix the issues.

Some statistics from different combinations of branches in my contiki-ng repo:

* Branch: develop

packet-injector-uip              TEST FAIL
Succeeded:  13
Timed out:  7
Failed   :  15

packet-injector-sicslowpan       TEST FAIL
Succeeded:  28
Timed out:  7
Failed   :  31

* Branch: develop + fix-rpl-ext-header-removal

packet-injector-uip              TEST OK
Succeeded:  35
Timed out:  0
Failed   :  0

packet-injector-sicslowpan       TEST FAIL
Succeeded:  41
Timed out:  0
Failed   :  25

* Branch: develop + fix-rpl-ext-header-removal + check-uip-len

packet-injector-uip              TEST OK
Succeeded:  35
Timed out:  0
Failed   :  0

packet-injector-sicslowpan       TEST FAIL
Succeeded:  41
Timed out:  0
Failed   :  25

* Branch: develop + fix-rpl-ext-header-removal + check-uip-len + sicslowpan-reassembly-fixes

packet-injector-uip              TEST OK
Succeeded:  35
Timed out:  0
Failed   :  0

packet-injector-sicslowpan       TEST OK
Succeeded:  66
Timed out:  0
Failed   :  0

PR #972, opened 2019-05-24
sicslowpan vulnerability fixes

nvt:
This PR fixes a number of problems with packet processing in the sicslowpan implementation. PR #900 should pass with the help of these changes.

For future work, it could be good to redesign the sicslowpan module, as it has become a bit convoluted.

Thanks to Tobias Scharnowski for reporting the issues with the add_fragment and store_fragment functions, and the possible use of previous packet data when reassembling a packet.

### DIFF
--- a/os/net/ipv6/sicslowpan.c
+++ b/os/net/ipv6/sicslowpan.c
@@ -415,6 +415,14 @@ copy_frags2uip(int context)
 {
   int i;
 
+  /* Check length fields before proceeding. */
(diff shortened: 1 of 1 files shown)


simonduq:
This seems reasonable to me. Would like to hear what @nfi @joakimeriksson think

joakimeriksson:
I am hoping to take a pass at that tomorrow and hopefully merge!

PR #974, opened 2019-05-25
Rework the logic around the maximum payload size supported by radios

### LINKED ISSUE #971: change zoul RF from 2.4G to 868MHz
Hi
I'm running a test with 11 zoul (1 Root, 10 client).
I use spectrum analyzer to see real tx signal,
I only saw signal on 2.4GHz.
What should I do to set zoul RF from 2.4GHz to 868MHz.

g-oikonomou:
As discussed internally, there are some issues with the current `_driver_max_payload_len` macros and the way they propagate. The situation is particularly problematic with platform-independent / generic radio drivers, namely cc1200, cc2420 and nullradio.

This pull reworks the logic around how we retrieve and consider a radio driver's maximum payload length. Basically, we make this information unavailable at compile-time and instead we retrieve it at run time by asking the radio driver whenever we need it.

* We introduce a new `RADIO_CONST_MAX_PAYLOAD_LEN`. It is mandatory to support this constant in all radio driver `get_value()` implementations.
* We change all radio drivers to implement the above.
* We change CSMA logic: `CSMA_MAC_LEN` goes away. `NETSTACK_MAC.max_payload()` now queries the radio.
* We change TSCH logic. `TSCH_PACKET_MAX_LEN` goes away, and:
  * The length of some TSCH packet buffers was based on `TSCH_PACKET_MAX_LEN` and basically relying on knowing the radio's max payload length at compile time . Since this is no longer the case, we make all those buffers to have a length dependent on `PACKETBUF_SIZE` instead.
  * TSCH driver's `NETSTACK_MAC.max_payload()` now queries the radio.
* We obsolete all `_driver_max_payload_len` macros and the compile-time constant `NETSTACK_RADIO_MAX_PAYLOAD_LEN`

On the transmit path, TSCH does this:

```
/* prepare packet to send: copy to radio buffer */
if(packet_ready && NETSTACK_RADIO.prepare(packet, packet_len) == 0) { /* 0 means success */
```
`packet_len` comes from the queuebuf, so it's basically `PACKETBUF_SIZE` or lower. I think this is the case with as well as without LLSEC.

I think it's OK to leave as-is and let the radio reject the packet if it ends up being too long. Alternatively, we can proactively make sure that `packet_len` is <= the max payload supported by the radio. But if it is not, then there is no way to recover anyway that I can think of. Happy for any feedback on this.

Fixes #971


debug-ito:
Thanks for the work. I like the idea! I have some comments, though.

- For tsch MAC, we could just `#define TSCH_PACKET_MAX_LEN PACKETBUF_SIZE` instead of removing the macro. I think that would make the diff simpler.
- (Probably this should be in another p-r) I would like the buffer size of cooja's radio to be configurable, instead of hard-coded as 125. That is because sometimes we want to send huge frames in simulated experiments.
- In `max_payload` method of csma and tsch MAC, `max_radio_payload_len` may be used uninitialized if `get_value` returns error.

g-oikonomou:
Thanks for the review @debug-ito 

> * For tsch MAC, we could just #define TSCH_PACKET_MAX_LEN PACKETBUF_SIZE instead of removing the macro. I think that would make the diff simpler.
>
> * I would like the buffer size of cooja's radio to be configurable, instead of hard-coded as 125

Done

> In max_payload method of csma and tsch MAC, max_radio_payload_len may be used uninitialized if get_value returns error.

I have changed those functions to return 0 if the radio does not support `RADIO_CONST_MAX_PAYLOAD_LEN`. I have added checks in the two MAC `init()` functions to make sure the radio can report this correctly - as per the OP this will have to be mandatory.

debug-ito:
Thanks!

joakimeriksson:
I think this looks good - the failed check was bad luck on the TSCH example (that we need to figure out I guess).

PR #985, opened 2019-06-14
Allow MPL Reactive Forwarding to be Disabled

edrose:
This PR adds a config switch that allows for the reactive forwarding strategy
in MPL to be disabled, something that was overlooked in my original implementation. With reactive forwarding disabled *most* of the caching of messages is disabled, and no MPL Control messages are exchanged. Instead only proactive forwarding is used where received messages are immediately scheduled for transmission by a trickle timer, so are forwarded after a small random delay. Some inconsistencies in the domain can also be detected by the 'M' flag within the data message which will trigger a retransmission. Messages are removed from the cache after the trickle timer has expired `DATA_MESSAGE_TIMER_EXPIRATIONS` times. Disabling reactive forwarding allows for smaller caches to be used, and also for the code relating to MPL Control messages to also be omitted at compile time.

### DIFF
--- a/os/net/ipv6/multicast/mpl.c
+++ b/os/net/ipv6/multicast/mpl.c
@@ -77,6 +77,10 @@
 #if MPL_SEED_ID_TYPE == 2 && MPL_SEED_ID_H > 0x00
 #warning MPL Seed ID upper 64 bits set yet not used due to Seed ID type setting
 #endif
+/* Check a forwarding strategy is enabled */
+#if !MPL_PROACTIVE_FORWARDING && !MPL_REACTIVE_FORWARDING
+#error "No forwarding strategy is enabled."
+#endif
 /*---------------------------------------------------------------------------*/
 /* Data Representation */
 /*---------------------------------------------------------------------------*/
@@ -330,6 +334,7 @@ struct mpl_hbho_s3 {
  * Seed ID length being sent, so all the different representations it can take
  * are shown here.
  */
+#if MPL_REACTIVE_FORWARDING
 struct seed_info {
   uint8_t min_seqno;
   uint8_t bm_len_S; /* First 6 bits bm-len, last 2 S */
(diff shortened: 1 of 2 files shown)


edrose:
Having reread the spec, it's a little ambiguous as to whether disabling reactive forwarding is actually part of the spec or not. For proactive forwarding, the spec lists the ability to disable it as a protocol configuration switch, yet there isn't a switch for reactive forwarding. On the other hand, all through the spec it says things like "if reactive forwarding is enabled" which suggests it can be disabled. I'm not 100% sure how to properly interpret that.

os/net/ipv6/multicast/mpl.c
```diff
@@ -899,6 +912,10 @@ data_message_expiration(void *ptr, uint8_t suppress)
   if(locmmptr->e > MPL_DATA_MESSAGE_TIMER_EXPIRATIONS) {
     /* Terminate the trickle timer here if we've already expired enough times */
     trickle_timer_stop(&locmmptr->tt);
+    #if !MPL_REACTIVE_FORWARDING
```
g-oikonomou: please un-indent the cpp lines here

os/net/ipv6/multicast/mpl.c
```diff
@@ -483,6 +493,7 @@ buffer_free(struct mpl_msg *msg)
   if(trickle_timer_is_running(&msg->tt)) {
     trickle_timer_stop(&msg->tt);
   }
+  list_remove(msg->seed->min_seq, msg);
```
g-oikonomou: This looks suspicious. Needed for both modes?

edrose: That was a preexisting bug within the `buffer_free` function that caused no problems with reactive forwarding always enabled, however became an issue when it was disabled. Should that have been done in a separate pull request?

g-oikonomou: > Should that have been done in a separate pull request?

No. Just making sure it does not break anything

edrose:
Having looked into this some more, I don't believe this PR is the correct approach. The spec says that reactive forwarding can be disabled by setting CONTROL_MESSAGE_TIMER_EXPIRATIONS to zero. I'll look at using this flag to remove unused bits of code from the driver, rather than defining a new one.

PR #993, opened 2019-06-28
Fixed bug in CC1310 prop RF settings

tiepettersen:
A bug was introduced in the CC1310 prop RF settings when fixing a bug
with BLE beaconing. In short, prop RF settings are shared between CC1310
and CC1350, and rfMode was set to RF_MODE_MULTIPLE which is only
supported on CC1350. This is now split between CC1310 and CC1350.

### DIFF
--- a/arch/cpu/simplelink-cc13xx-cc26xx/rf-settings/cc13x0/prop-settings.c
+++ b/arch/cpu/simplelink-cc13xx-cc26xx/rf-settings/cc13x0/prop-settings.c
@@ -66,6 +66,19 @@
 #include "prop-settings.h"
 /*---------------------------------------------------------------------------*/
 /* TI-RTOS RF Mode Object */
+#if defined(DEVICE_CC1310)
+RF_Mode rf_prop_mode =
+{
+  /*
+   * CC1310 only supports RF_MODE_PROPRIETARY_SUB_1, and hence we need to
+   * separate the rfMode setting between CC1310 and CC1350*.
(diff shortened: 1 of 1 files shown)


arch/cpu/simplelink-cc13xx-cc26xx/rf-settings/cc13x0/prop-settings.c
```diff
@@ -66,13 +66,27 @@
 #include "prop-settings.h"
 /*---------------------------------------------------------------------------*/
 /* TI-RTOS RF Mode Object */
+#if defined(DEVICE_CC1310)
```
g-oikonomou:
```
#if defined(DEVICE_CC1310)
#define RF_MODE RF_MODE_PROPRIETARY_SUB_1
#else
#define RF_MODE RF_MODE_MULTIPLE
#endif
```

and then we can have a single definition for the `rf_prop_mode` struct?

tiepettersen: I will call it RF_PROP_MODE, as RF_MODE is already used by the generic RF configuration.

g-oikonomou: Any name for the macro, really. I just pulled this one out of my head without checking elsewhere. I don't feel strongly, just a thought

Issue #994, opened 2019-06-28
Scheduling of ASF bursty links possibly breaks hard cells


mbaddeley:
So in tsch-slot-operation.c, burst_link_requested and burst_link_scheduled allow nodes to opportunistically schedule another TX if they have more packets in the TX queue (as per ASF, I believe).

```
/* We requested an extra slot and got an ack. This means the extra slot will be scheduled at the received */
    if(burst_link_requested) {
        burst_link_scheduled = 1;
    }
```

I was playing around with statically defined schedules, and I noticed that when I explicitly set cells as a hard RX, a node still might try to schedule a bursty TX, causing collisions with other TXs from other nodes in the schedule. e.g. the schedule should be...

```
1 - Rx | Rx | Rx
2 - Tx | Rx | Rx
3 - Rx | Tx | Rx
4 - Rx | Rx | Tx
```

... but node 2 has additional packets in its TX queue, it schedules a burst slot, and this happens...

```
1 - Rx | Rx  | Rx
2 - Tx | Tx* | Rx
3 - Rx | Tx* | Rx
4 - Rx | Rx  | Tx

*collision
```

Is this is a bug, or am I just not getting something?

atiselsts:
My understanding is that this is the expected behavior as per the IEEE 802.15.4 standard. 

With the MSF and other scheduling functions that adapt to the traffic, enabling the burst bit reduces performance. With purely autonomous scheduling the discussion is more open - for some applications the burst bit helps.

yatch:
I don't think it's a bug. It seems kind of a policy, which prioritizes transmission over reception. IEEE 802.15.4-2015 doesn't say anything about how the transmitter should behave in this particular case (see Section 7.2.1.3), although the pending bit itself is underspecified....

However, as you mentioned, bursty TX may break a communication schedule. Busty TX may be undesirable in some cases. In my working project, I've introduced a configuration macro with which a user (developer) can decide to use the bursty TX or not. I will propose it soon.

mbaddeley:
Sounds good! I think it would be good to have this in NG at some point so it's on by default but if a user wants to break from IEEE 802.15.4 then they can, as it would be useful when making your own scheduling functions. In the meantime, like @yatch, I'll define my own macro :)

yatch:
[...]

PR #995, opened 2019-06-28
tsch: introduce TSCH_CONF_PENDING_BIT_ENABLED

yatch:
This PR introduces `TSCH_CONF_PENDING_BIT_ENABLED` which gives a choice, whether to enable the pending bit, defined in Section 7.2.1.3 of IEEE 802.15.4-2015, or not. Having the pending bit enabled, a transmitter can perform the bursty TX.

See discussion in Issue #994.

### DIFF
--- a/os/net/mac/tsch/tsch-conf.h
+++ b/os/net/mac/tsch/tsch-conf.h
@@ -362,6 +362,15 @@
 #define TSCH_WITH_LINK_SELECTOR (BUILD_WITH_ORCHESTRA)
 #endif /* TSCH_CONF_WITH_LINK_SELECTOR */
 
+/* Having the pending bit on in frame, following frames to the same
+ * destination can be transmitted consecutively over unscheduled
+ * links. This enables bursty TX. (default enabled) */
+#ifdef TSCH_CONF_PENDING_BIT_ENABLED
+#define TSCH_PENDING_BIT_ENABLED TSCH_CONF_PENDING_BIT_ENABLED
+#else
+#define TSCH_PENDING_BIT_ENABLED 1
+#endif
+
 /******** Configuration: CSMA *******/
 
 /* TSCH CSMA-CA parameters, see IEEE 802.15.4e-2012 */
--- a/os/net/mac/tsch/tsch-slot-operation.c
+++ b/os/net/mac/tsch/tsch-slot-operation.c
@@ -497,12 +497,14 @@ PT_THREAD(tsch_tx_slot(struct pt *pt, struct rtimer *t))
       is_broadcast = current_neighbor->is_broadcast;
       /* Unicast. More packets in queue for the neighbor? */
       burst_link_requested = 0;
+#if TSCH_PENDING_BIT_ENABLED
       if(!is_broadcast
              && tsch_current_burst_count + 1 < TSCH_BURST_MAX_LEN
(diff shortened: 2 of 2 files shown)


yatch:
Test code comes later...

atiselsts:
I'm actually not so sure we need yet another configuration macro. On the sender side, the same effect can be achieved with setting `TSCH_BURST_MAX_LEN` to 0 or 1. Is this not enough for 99% of use cases?

mbaddeley:
> TSCH_BURST_MAX_LEN

Didn't see this previously. I think your'e right. What's the 1% of use-cases though?

yatch:
@atiselsts Thanks for the comment. Maybe it's enough :-) 

`TSCH_BURST_MAX_LEN` affects only the transmitter side. The receiver still processes the frame pending bit even with `TSCH_BURST_MAX_LEN` 0 or whatever. For interoperability, it'd better to have a way to disable the feature completely.

mbaddeley:
```
/* Set an upper bound on burst length. Set to 0 to never set the frame pending
 * bit, i.e., never trigger a burst. Note that receiver-side support for burst
 * is always enabled, as it is part of IEEE 802.1.5.4-2015 (Section 7.2.1.3)*/
#ifdef TSCH_CONF_BURST_MAX_LEN
#define TSCH_BURST_MAX_LEN TSCH_CONF_BURST_MAX_LEN
#else
#define TSCH_BURST_MAX_LEN 32
#endif
```

So it seems that the standard dictates that it should always be enabled at the RX?

yatch:
> or interoperability, it'd be better to have a way to disable the feature completely.

"Interoperability" may be not a right word... :-/ 

> So it seems that the standard dictates that it should always be enabled at the RX?

Anyway, good point. I checked IEEE 802.15.4-2015, again. It sounds like so :-D 

As I said, Section 7.2.1.3 of IEEE 802.15.4-2015 needs more text, though... I believe we will have more text in the next revision.

mbaddeley:
Yeah, even though it says that it should always be enabled at the RX, you're right in that it would be nice to be able to turn the RX check off, rather than relying on the TX not to set it in the first place, but it's not that big an issue I suppose.

simonduq:
Do you have a particular scenario in mind @yatch where the sender sets the frame pending but does not expect the receiver to keep on listening at next slot?

yatch:
To me, the pending bit is kind of an optional feature, which personally I want to disable completely when it's not necessary, in order to prevent any unexpected behaviors.

But, honestly, `TSCH_BURST_MAX_LEN` would be enough for my current project, implementing MSF :-)

One possible convincing scenario is a DoS attack. An attacker can send an unprotected unicast frame having the pending bit on so as to keep the victim listening, which causes energy consumption for nothing. If the victim doesn't need the pending bit for a good reason, it'd be better to disable the feature in the first place.

simonduq:
With standards I like being liberal on the receive end and conservative on the sender; in this case always be able to interpret the bit pending flag, whether it's currently in use in a given network or not. Helps with interop among stacks as well as among upgrades of a given firmware.

But no strong opinion; if OP or anybody feels like they would like this PR merged, let's merge it! (otherwise let's just close)

@yatch note that there's a conflict

yatch:
I'm going to close this for now. Thanks!

simonduq:
Thanks/sorry!

PR #997, opened 2019-07-03
Re-organize structure and documentation of rtimer.h

debug-ito:
This pull-request re-organizes the structure and documentation of rtimer.h:

- API doc for `rtimer_set` is updated to match the current implementation. It was out of date, I think.
- Configuration macros are put together at the top of the header.
- Functions that must be defined in architecture-specific source files are put together at the bottom.

    - Some platforms (e.g. `native`) define those symbols as macros, so the function declaration is now conditinal (just like gpio-hal.h).

- Add documentation on macros that must be defined in rtimer-arch.h

I don't have any idea how I should describe `RTIMER_ERR_` constants, so I added a TODO comment to them.

BTW, [the usage example of rtimer in Wiki](https://github.com/contiki-ng/contiki-ng/wiki/Documentation:-Timers#the-rtimer-library) is out of date.

### DIFF
--- a/os/sys/rtimer.h
+++ b/os/sys/rtimer.h
@@ -57,6 +57,8 @@
 #include "dev/watchdog.h"
 #include <stdbool.h>
 
+/*---------------------------------------------------------------------------*/
+
 /** \brief The rtimer size (in bytes) */
 #ifdef RTIMER_CONF_CLOCK_SIZE
 #define RTIMER_CLOCK_SIZE RTIMER_CONF_CLOCK_SIZE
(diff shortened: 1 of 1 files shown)


debug-ito:
Thanks for review!

I'm a little confused. Which of the following is your opinion?

1. Revert https://github.com/contiki-ng/contiki-ng/commit/d40b6f3739a7ac542f1e41c204e4724a4fd12ca2 and add a comment that rtimer-arch.h needs to define `rtimer_arch_now`.
2. Remove the conditional macros introduced in https://github.com/contiki-ng/contiki-ng/commit/d40b6f3739a7ac542f1e41c204e4724a4fd12ca2 and revise arch/cpu/native/rtimer-arch so that it defines `rtimer_arch_now` as a function instead of a macro.

I'm ok with either of the two options.

yatch:
Sorry for the confusion; option 1.

debug-ito:
OK, I just committed option 1.

debug-ito:
I'm on it. Should I push the squashed commits to another branch and make a new p-r? Or, can I force-push to the same branch?

yatch:
Please force-push it to the same branch here :+1:

debug-ito:
Thanks for advice. I force-pushed the new branch.

PR #1002, opened 2019-07-09
Move resolv to a separate module / service

joakimeriksson:
Also fixed logging of resolve and uip-nameserver to make use of the Logging system and added two shell commands - nslookup <host> and wget <URL> for testing connectivity to the Internet (and between devices).

I am currently working with another PR that enable NAT64 on native platform so that these two shell commands can be used more easily (+ a NAT64/DNS64 script in python - to avoid deploying Jool or Tayga to enable NAT64). But that will come as a separate PR (and will likely include NAT64/DNS64 support for native-border-router also).


os/net/ipv6/ip64-addr.c
```diff
@@ -34,9 +34,12 @@
...
 #include <string.h>
 
-#define printf(...)
+/* Log configuration */
+#include "sys/log.h"
+#define LOG_MODULE "IPv6"
+#define LOG_LEVEL LOG_LEVEL_NONE
 
-static uip_ip6addr_t ip64_prefix = {{ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff, 0, 0, 0, 0}};
+static uip_ip6addr_t ip64_prefix = {{ 0, 0x64, 0xff, 0x9b, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}};
```
yatch: I think, we may need this change in a separate commit, explaining in its commit log why the prefix is modified. It seems this change correspond to https://tools.ietf.org/html/rfc6052#section-2.1?

joakimeriksson: Yes, this is changed since the normally used NAT64 prefix seems to be 64:ff9b::/96. I can split it into a separate commit if that makes sense - or document it in  the source code.

yatch: Thanks! :+1:

PR #1003, opened 2019-07-10
lib: memb: simplify handling of allocation status

evdenis:
During formal verification of memb* functions it was discovered that
reference counting in the functions could be significantly simplified.

Here is the citation from the work:
"This verification case study also allowed to detect a potentially
harmful situation. The memb_free function used to decrement the count
associated to the given block, instead of setting it to 0. An awkward
consequence of this is that calling memb_free on a block with an unusual
count (e.g., greater than 2) would not actually free it. While this
should not happen under normal circumstances, we have decided to replace
that decrement operation by a set to 0. This choice makes memb_free both
simpler and more robust, easing the verification process..."

"Formal Verification of a Memory Allocation Module of Contiki with
Frama-C: a Case Study" Frédéric Mangano, Simon Duquennoy, Nikolai Kosmatov

This commit adopts the verification results from the work.

Signed-off-by: Denis Efremov <efremov@linux.com>
Signed-off-by: Nikolay Kosmatov <nikolaikosmatov@gmail.com>
Signed-off-by: Allan Blanchard <allan.blanchard@inria.fr>
Signed-off-by: Frederic Loulergue <frederic.loulergue@nau.edu>

Cc: @simonduq, @AllanBlanchard, @fredericloulergue

### DIFF
--- a/os/lib/memb.c
+++ b/os/lib/memb.c
@@ -51,7 +51,7 @@
 void
 memb_init(struct memb *m)
 {
-  memset(m->count, 0, m->num);
+  memset(m->used, 0, m->num);
   memset(m->mem, 0, m->size * m->num);
 }
 /*---------------------------------------------------------------------------*/
@@ -61,11 +61,10 @@ memb_alloc(struct memb *m)
   int i;
 
   for(i = 0; i < m->num; ++i) {
-    if(m->count[i] == 0) {
-      /* If this block was unused, we increase the reference count to
-	 indicate that it now is used and return a pointer to the
-	 memory block. */
-      ++(m->count[i]);
+    if(m->used[i] == false) {
+      /* If this block was unused, we set the used flag on
+	 and return a pointer to the memory block. */
+      m->used[i] = true;
       return (void *)((char *)m->mem + (i * m->size));
     }
   }
@@ -75,7 +74,7 @@ memb_alloc(struct memb *m)
   return NULL;
 }
 /*---------------------------------------------------------------------------*/
-char
+int
 memb_free(struct memb *m, void *ptr)
 {
   int i;
@@ -85,15 +84,13 @@ memb_free(struct memb *m, void *ptr)
      which the pointer "ptr" points to. */
   ptr2 = (char *)m->mem;
   for(i = 0; i < m->num; ++i) {
-    
     if(ptr2 == (char *)ptr) {
-      /* We've found to block to which "ptr" points so we decrease the
-	 reference count and return the new value of it. */
-      if(m->count[i] > 0) {
-	/* Make sure that we don't deallocate free memory. */
-	--(m->count[i]);
-      }
-      return m->count[i];
+      /* We've found the block to which "ptr" points, so we check the allocation
+         status to detect the double-free error and free the block. */
+      if (m->used[i] == false)
+        return -1;
+      m->used[i] = false;
+      return 0;
     }
     ptr2 += m->size;
   }
@@ -114,7 +111,7 @@ memb_numfree(struct memb *m)
   int num_free = 0;
 
   for(i = 0; i < m->num; ++i) {
-    if(m->count[i] == 0) {
+    if(m->used[i] == false) {
       ++num_free;
     }
   }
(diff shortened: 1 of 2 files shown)


os/lib/memb.c
```diff
@@ -62,10 +62,10 @@ memb_alloc(struct memb *m)
 
   for(i = 0; i < m->num; ++i) {
     if(m->count[i] == 0) {
-      /* If this block was unused, we increase the reference count to
+      /* If this block was unused, we set the count flag to
 	 indicate that it now is used and return a pointer to the
 	 memory block. */
-      ++(m->count[i]);
+      m->count[i] = 1;
```
yatch: Then, I think it'd be better to rename the member, `count` to `used`.

Now, I'm not sure why the type of the member is `char`... This one may be changed to `bool`? With these changes, this line would be:
```
m->used[i] = true;
```

evdenis: Done. Thanks for the review

yatch:
As you noticed, I wrote some tests for `memb` (https://github.com/contiki-ng/contiki-ng/pull/1007). I didn't know `memb_free()` accepts double-free...

Does the formal verification consider double-free as an acceptable operation? Seems, returning -1 from `memb_free()` to  a free memory block is a right thing to do... I'm not sure what was done in the formal verification, though...

evdenis:
I'm not the original author of the verification work. However, as far as I could see from the formal specs https://github.com/evdenis/Contiki-NG/blob/develop/os/lib/memb.h#L170-L174 the double-free is considered acceptable. I know from the original authors of the research that they were not trying to alter the existing behavior of the functions and were trying to write the specifications as close to the existing behavior as it possible. It looks like this case was missed.

I agree with you, that the double free should be treated as an erroneous situation. Since at almost all calls the result of memb_free is ignored, what do you think about inserting a bug macro in addition to -1? Something similar to the Linux kernel detection https://lore.kernel.org/kernel-hardening/1502468246-1262-1-git-send-email-alex.popov@linux.com/

os/lib/memb.c
```diff
@@ -85,15 +84,13 @@ memb_free(struct memb *m, void *ptr)
...
-	 reference count and return the new value of it. */
-      if(m->count[i] > 0) {
-	/* Make sure that we don't deallocate free memory. */
-	--(m->count[i]);
-      }
-      return m->count[i];
+      /* We've found the block to which "ptr" points, so we check the allocation
+         status to detect the double-free error and free the block. */
+      if (m->used[i] == false)
+        return -1;
```
evdenis: The check inserted.

os/lib/memb.c
```diff
@@ -85,15 +84,13 @@ memb_free(struct memb *m, void *ptr)
      which the pointer "ptr" points to. */
   ptr2 = (char *)m->mem;
   for(i = 0; i < m->num; ++i) {
-    
     if(ptr2 == (char *)ptr) {
-      /* We've found to block to which "ptr" points so we decrease the
-	 reference count and return the new value of it. */
-      if(m->count[i] > 0) {
-	/* Make sure that we don't deallocate free memory. */
-	--(m->count[i]);
```
evdenis: It looks like in the original code there was a clear intention to omit the double-free bug.

joakimeriksson: I do think that the original code might even have had the idea to allow calling alloc multiple times. But I do think your fix seems to clean up the code and make it more easy to read in general.

yatch:
@evdenis Thank you for your comment (and update)! I'm sorry, the following was just out of my curiosity. But, I'm happy to have the check for double-free :-)

> Does the formal verification consider double-free as an acceptable operation? Seems, returning -1 from memb_free() to a free memory block is a right thing to do... I'm not sure what was done in the formal verification, though...

And, thank you for your contribution! Thanks to you and this PR, I know more about `memb` :+1:

os/lib/memb.c
nkosmatov: I think you are right. I even had the impression that the original code had the intention to count references for some reason (the name count, ++/-- operations support this idea), which was unfinished and not really used. Such unfinished intentions can be dangerous, thank you for your corrections. As soon as time allows, we will try to verify the new version.

PR #1015, opened 2019-07-23
RPL-Lite: don't reset DIO timer if the current interval is equal to Imin

### LINKED ISSUE #910: RPL-lite: DIO may not be sent if there is continuous DIS.
I was testing a large number of device simulation and many devices did not join after long period of time. The issue was the root device was not sending DIO.

rpl_timers_dio_reset in os/net/routing/rpl-lite/rpl_timers.c is getting called every time a DIS is received.  This sets a new timer to send DIO. When there are many devices sending DIS, it is enough that the DIO sending timer never expires, so the DIO is not sent.

Is it necessary to create the timer every time is it reset? Wouldn't it be better to leave already running timer if it was started with the minimum DIO interval?

yatch:
Otherwise, heavy traffic of DIS prevents routers from sending DIOs. This issue is found only in `rpl-lite`. 

Fixes #910


atiselsts:
@simonduq Any objections to merging this? I'm not familiar with the RPL lite code, but this PR appears to fix a real problem, cites RFC 6206 for backing, includes new tests, and the new changes seem to make sense. Overall very well done.

PR #1017, opened 2019-07-29
Implemented Snmp v1 and v2c

Yagoor:
This is an light weight implementation of SNMP. It implements a simple API for custom mib implementation but it also has a default mib.
I'm currently implementing some functionalities, meaning that this is still a work in progress. But I would like to get some feedback while I'm still working on it. 
Currently only the SNMP v1 and v2c are implemented. The SNMP v3 cannot be fully implemented hence why I didn't implement it partially.

Currently I'm still working on the unit tests and documentation. But the core code is already in place and working.


yatch:
It's really nice that example and test code are provided!! Personally, I'm not sure we need SNMP, though...

One comment from me is on license. I saw your changes contain **GPL** code, which cannot be part of a software distributed under BSD 3-Clause License, from my understanding. Contiki-NG uses BSD 3-Clause License.

Yagoor:
Oh alright. Thanks for letting me know the license issue.
It is mainly in the ASN1.1 encoding and decoding sections.
Shouldn't be too hard to fix it. 
I'll work on that as well.

Yagoor:
Due to license issue. I'll close this pull request.

PR #1020, opened 2019-08-06
Implemented Snmp v1 and v2c

Yagoor:
Dear all,

This is a lightweight implementation of SNMP.
Currently, only the SNMP v1 and v2c are implemented. The SNMP v3 cannot be fully implemented hence why I didn't implement it partially.

Currently, I'm still working on unit tests and documentation. But the core code is already in place and working.


Yagoor:
It would be interesting to know from the community which data types are more relevant. There are several data types on the ASN1.1 that can be used in the SNMP protocol. But I do not see a point in implementing all of them and increasing the code footprint.

I would also like to know if the best approach is to provide by default some MIB implementations, like the IF-MIB or the SYSTEM-MIB.

I'm currently working on a MIB for the Contiki. I already got a Private Enterprise Number from IANA. The Contiki-ng number is 54352. The complete list is here: https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers

joakimeriksson:
Cool stuff! Some things I would like and that I saw in the diffs:

1. A short description of the use cases you would see for this in a 6LoWPAN context (e.g. resource constrained IoT devices setup).

2. There are a bunch of formatting changes which I do not think should be in this PR. We would like to have the fewest possible diffs possible to make it easy to do the code review and would like to avoid any formatting changes that are not strongly motivated.

Yagoor:
Hi @joakimeriksson 

1. This is what my master thesis is entirely about. There are many scenarios in the WSN context that would be interesting to use the SNMP. It is good to check the sysuptime of a sensor that is deployed in a very remote area, for example, is order to check if it is reaching its lifetime end. Or check the power usage of a sensor that again is in a very remote area. If this sensor should be working on a very unexcpeted scenario, such as a dam breaking alert. You don't want to wait for it to break to realize that the sensor is dead. That is why I only implemented the Get functionalities. The Set would be irrelevant since WSN networks should be able to reconfigure itself. And to Set we would need to implement the SNMP v3 with its authentication securities.

2. I'll rollback those files that I changed outside the snmp folder. I ran the uncrustify script in all files that I changed therefore those files are not following the coding style.

Yagoor:
I added one example of a SNMP Server using the API and I reverted the formatting changes.
What would be the next natural step?

examples/snmp/server/project-conf.h
```diff
@@ -0,0 +1,34 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+/*---------------------------------------------------------------------------*/
+
+#define LOG_CONF_LEVEL_SNMP     LOG_LEVEL_NONE
+#define SNMP_CONF_DEBUG         0
```
g-oikonomou: I would propose getting rid of `SNMP_CONF_DEBUG` since it feels redundant.

In the places where it is being used, I recommend replacing the CPP conditional with a check against the value of `LOG_LEVEL`

Something like:
```c
#if LOG_LEVEL == LOG_LEVEL_DBG
```

This is for example done in https://github.com/contiki-ng/contiki-ng/blob/66f542fc14273b14e89d1e2a9eb1d72edcca345a/os/services/resolv/resolv.c#L1338

os/net/app-layer/snmp/snmp-ber.c
```diff
@@ -0,0 +1,274 @@
...
+ */
+/*---------------------------------------------------------------------------*/
+
+/**
+ * \file
+ *      An implementation of the Simple Network Management Protocol (RFC 3411-3418)
+ * \author
+ *      Yago Fontoura do Rosario <yago.rosario@hotmail.com.br
+ */
+
```
g-oikonomou: Please always include "contiki.h" at the top of every file, this will could prevent the appearance of subtle bugs in the future.

os/net/app-layer/snmp/snmp-ber.c
```diff
@@ -0,0 +1,274 @@
...
+  if(type != BER_DATA_TYPE_OCTET_STRING) {
+    /*
+     * Sanity check
+     * Invalid type in buffer
+     */
+    return NULL;
+  }
+
+  uint8_t i, length_bytes;
```
g-oikonomou: Please move to the start of the function. We typically don't intermingle variable definitions with code (for historic reasons)

os/net/app-layer/snmp/snmp-engine.c
```diff
@@ -0,0 +1,252 @@
...
+ */
+/*---------------------------------------------------------------------------*/
+
+/**
+ * \file
+ *      An implementation of the Simple Network Management Protocol (RFC 3411-3418)
+ * \author
+ *      Yago Fontoura do Rosario <yago.rosario@hotmail.com.br
+ */
+
```
g-oikonomou: Ditto inclusion of "contiki.h"

os/net/app-layer/snmp/snmp.h
```diff
@@ -0,0 +1,220 @@
...
+ */
+
+#ifndef SNMP_H_
+#define SNMP_H_
+
+#include "contiki.h"
+#include "contiki-net.h"
+
+#include "sys/log.h"
+#include "sys/energest.h"
```
g-oikonomou: Why energest?

os/net/app-layer/snmp/snmp.h
```diff
@@ -0,0 +1,220 @@
...
+       * @remarks Do not use strlen on the string since it might not be null terminated
+       */
+      uint32_t length;
+    } string;
+    /**
+     * @brief A pointer to the beggining of a oid array
+     */
+    uint32_t *oid;
+  } value;
+} snmp_varbind_t;
```
g-oikonomou: Both `snmp_header_t` and `snmp_varbind_t` are very likely to get padded by the compiler. Is this OK? I did not check how they are being used, but if they are used to create/parse packets such padding could create trouble.

os/net/app-layer/snmp/snmp.h
Yagoor: That is a good question. I forgot to remove it.

os/net/app-layer/snmp/snmp.h
Yagoor: Padding isn't a problem. SNMP uses the BER encoding, so the structs are only holding the information before encoding or after decoding. I tried to avoid padding, but since I used unions, I have to admit that I forgot to reorder the struct.

os/net/app-layer/snmp/snmp-engine.c
Yagoor: Done

os/net/app-layer/snmp/snmp-ber.c
Yagoor: This was a mistake. I also don't like this.

os/net/app-layer/snmp/snmp-ber.c
Yagoor: Done

Yagoor:
Thank you for the review.

First of all, my history was messy and after implementing the changes you requested I rebased the branch, it should be easier to see the changes as well.

The answer for your questions:

1 - I tested this on the native, cooja and I'll test it in the cc26x0-cc13x0, which is the only physical platform I have.

2 - There is a easy way to test if it is SNMP-compliant. I'll implement the Test and commit it here. I know that it is SNMP compliant now, but I do see the point to have a regression test.

os/net/app-layer/snmp/snmp.h
g-oikonomou: OK so this is just a question about total size then

Yagoor:
I created the SNMP compliant test. 

Is there any other request I should work on?

g-oikonomou:
As far as I can tell we have a single test for SNMP execution and it uses platform native. Right? In that case, I'd suggest moving the test under `08-native-runs`. It's OK to add smitools and other  packages to tho docker image.

Yagoor:
I rebased the branch into develop. Added the SNMP compliance test to the native-runs test set and added the packages into the Docker image.

g-oikonomou:
@Yagoor Hi Yago I see in the snmp-server example you have not added any `PLATFORMS_ONLY` or `_EXCLUDE`. Is this expected to work on all our supported h/w?

Other than that, this looks really good!

Yagoor:
@g-oikonomou. I believe it is supposed to work in all h/w supported by Contiki-NG. I don't see why it wouldn't work. =D

g-oikonomou:
```
$ make TARGET=sky
  MKDIR     build/sky/obj
  CC        ../../os/contiki-main.c
  CC        ../../arch/platform/sky/./contiki-sky-platform.c
[...]
  CC        ../../os/net/routing/rpl-lite/rpl.c
  AR        build/sky/contiki-ng-sky.a
  CC        snmp-server.c
snmp-server.c:36:21: fatal error: strings.h: No such file or directory
compilation terminated.
make: *** [snmp-server.o] Error 1
```

Yagoor:
Oh. Alright.
I'll test all the platforms and updated the Makefile accordingly. 
My apologies.

g-oikonomou:
Perhaps we also need a couple more compile tests for known working platforms

Yagoor:
I took a good look and I couldn't understand why I had the string and strings headers there. This was probably something that I used for debugging while developing and forgot to remove. 
I adapted the compile-all script to check if this example runs in every platform and it worked fine.
Do you have any platforms that you'd like me to add the test for? Like some cc26x0-cc13x0, for example.

This is the output from the compile-all script. I made it compile only this example, for simplicity. 
image

g-oikonomou:
I'd say sky and any one cortex platform. With the native execution test we are basically testing 3 toolchians and that's enough I think

g-oikonomou:
I should add, I'd say it's OK to remove the native compile-test since we have the execution test

Yagoor:
Alright. I added the sky and the cc2538dk compile and removed the native compile. Everything should be green. :D

PR #1025, opened 2019-08-14
Typo in tools/viewconf/viewconf.c

CrySpy:
I found a typo in viewconf.c which makes it so the value of TSCH_CONF_DEFAULT_HOPPING_SEQUENCE appears incorrectly as the text TTSCH_CONF_DEFAULT_HOPPING_SEQUENCE instead of whatever value(s) it is set to.


yatch:
Thank you for the fix.

PR #1044, opened 2019-08-31
Add new GPIO, button and LED HAL support to platform Cooja

### LINKED ISSUE #1031: Cooja buttons
In Cooja I create a Cooja Mote and compile a project that uses the button API. i.e. with button_hal_button_count defined and function button_hal_get_by_index. It compiles in Cooja ok when I try and Create it, it doesn't link. There are a slew of error messages the most relevent is probably "undefined symbol: button_hal_button_count"

g-oikonomou:
This pull implements the new GPIO, button and LED HALs for platform Cooja. Respective examples will now work.

On the java-side of Cooja we have the following assumptions:

* Cooja motes have 3 LEDs (Green, Red, Yellow)
* Cooja motes have 1 button

For that reason, thus pull implements four "virtual" pins for platform Cooja (On the C-side of things). Pins 0, 1 and 2 are used for the three LEDs respectively. Pin 3 for the single button.

Cooja also makes a hard-coded expectation for a symbol called `button_sensor`. So while this pull deletes the old `button_sensor` driver, we still expose a stub symbol to satisfy the dependency. We can thus avoid changing Cooja sources as part of this pull.

Closes #1031


atiselsts:
I tried to run the `examples/dev/button-hal/` example in Cooja, by creating a new simulation and adding a Cooja mote to it. It detects the button all right:

```
00:00.382	ID:1	Button HAL example.
00:00.382	ID:1	Device button count: 1.
00:00.382	ID:1	User button on pin 3 with ID=0, Logic=Negative, Pull=Pull Up
```

But when I select "Click button on Contiki 1", nothing is printed. Shouldn't there be?

g-oikonomou:
> But when I select "Click button on Contiki 1", nothing is printed. Shouldn't there be?

I tried pressing the button on the mote interfaces -> button window and it was working as expected. I will have to check what you described

g-oikonomou:
@atiselsts good catch, unlike the button page on the mote interface viewer, those "button presses" are instantaneous and they were getting neutralised by the s/w debounce logic. Fixed now and tested.

asereq:
Maybe it is needed to add PLATFORM_HAS_LEDS in some of the platform configuration file; this could add the leds functionality to the coap-example-server (and maybe other examples).

PR #1047, opened 2019-09-02
fix initialization of link stats

atiselsts:


os/net/link-stats.c
```diff
@@ -147,6 +147,7 @@ link_stats_packet_sent(const linkaddr_t *lladdr, int status, int numtx)
     /* Add the neighbor */
     stats = nbr_table_add_lladdr(link_stats, lladdr, NBR_TABLE_REASON_LINK_STATS, NULL);
     if(stats != NULL) {
+      memset(stats, 0, sizeof(*stats));
```
simonduq: the two `memset` should not be necessary, as this is already done by `nbr_table`. Or am I missing something?

atiselsts: True, the nbr table code does memset already.

atiselsts:
Perhaps don't merge this yet, I have more link stats changes in the Amber project that I'll ask to be released.

simonduq:
Is this ready?

atiselsts:
Yes, I believe so.

PR #1048, opened 2019-09-02
fix the removal of extension header on fallback interface

atiselsts:
Someone more familiar with this code, please check if this is the right approach!


g-oikonomou:
Changed base to branch `release-4.4`

PR #1051, opened 2019-09-02
minor improvements and cleanups in TSCH

atiselsts:
* Add `tsch_adaptive_timesync_reset()` function
* Rename `is_broadcast` to `do_wait_for_ack` to better reflect how it is used in code (and there can be non-ACK'able unicast messages, after all)
* Return `NULL` instead of `0` from `tsch_queue_add_packet`
* Allow to compile the code when `TSCH_EB_PERIOD` is set to 0

### DIFF
--- a/os/net/mac/tsch/tsch.c
+++ b/os/net/mac/tsch/tsch.c
@@ -526,6 +526,7 @@ tsch_disassociate(void)
 {
   if(tsch_is_associated == 1) {
     tsch_is_associated = 0;
+    tsch_adaptive_timesync_reset();
     process_poll(&tsch_process);
   }
 }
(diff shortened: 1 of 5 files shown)


os/net/mac/tsch/tsch.c
```diff
@@ -847,7 +848,7 @@ PROCESS_THREAD(tsch_send_eb_process, ev, data)
 
   /* Set an initial delay except for coordinator, which should send an EB asap */
   if(!tsch_is_coordinator) {
-    etimer_set(&eb_timer, random_rand() % TSCH_EB_PERIOD);
+    etimer_set(&eb_timer, TSCH_EB_PERIOD ? random_rand() % TSCH_EB_PERIOD : 0);
```
simonduq: Out of pure curiosity, what is the purpose of `TSCH_EB_PERIOD == 0`?
Looks like we might end up in an infinite busy-wait with this setting here, no?

PR #1053, opened 2019-09-02
add multiple channel option to Orchestra unicast slotframe

atiselsts:
This increases the performance of Orchestra by reducing the number of collisions, especially helpful in dense networks.

### DIFF
--- a/os/net/mac/tsch/tsch-schedule.c
+++ b/os/net/mac/tsch/tsch-schedule.c
@@ -224,9 +224,6 @@ tsch_schedule_add_link(struct tsch_slotframe *slotframe,
     if(timeslot > (slotframe->size.val - 1)) {
       LOG_ERR("! add_link invalid timeslot: %u\n", timeslot);
       return NULL;
-    } else if(channel_offset > 15) {
-      LOG_ERR("! add_link invalid channel_offset: %u\n", channel_offset);
-      return NULL;
     }
 
     /* Start with removing the link currently installed at this timeslot (needed
--- a/os/net/mac/tsch/tsch-slot-operation.c
+++ b/os/net/mac/tsch/tsch-slot-operation.c
@@ -238,10 +238,20 @@ tsch_release_lock(void)
 
 /* Return channel from ASN and channel offset */
 uint8_t
-tsch_calculate_channel(struct tsch_asn_t *asn, uint8_t channel_offset)
+tsch_calculate_channel(struct tsch_asn_t *asn, uint16_t channel_offset, struct tsch_packet *p)
 {
-  uint16_t index_of_0 = TSCH_ASN_MOD(*asn, tsch_hopping_sequence_length);
-  uint16_t index_of_offset = (index_of_0 + channel_offset) % tsch_hopping_sequence_length.val;
+  uint16_t index_of_0, index_of_offset;
+#if TSCH_WITH_LINK_SELECTOR
+  if(p != NULL) {
+    uint16_t packet_channel_offset = queuebuf_attr(p->qb, PACKETBUF_ATTR_TSCH_CHANNEL_OFFSET);
+    if(packet_channel_offset != 0xffff) {
+      /* The schedule specifies a channel offset for this one; use it */
+      channel_offset = packet_channel_offset;
+    }
+  }
+#endif
+  index_of_0 = TSCH_ASN_MOD(*asn, tsch_hopping_sequence_length);
+  index_of_offset = (index_of_0 + channel_offset) % tsch_hopping_sequence_length.val;
   return tsch_hopping_sequence[index_of_offset];
 }
 
@@ -1014,7 +1024,8 @@ PT_THREAD(tsch_slot_operation(struct rtimer *t, void *ptr))
           burst_link_scheduled = 0;
         } else {
           /* Hop channel */
-          tsch_current_channel = tsch_calculate_channel(&tsch_current_asn, current_link->channel_offset);
+          tsch_current_channel = tsch_calculate_channel(&tsch_current_asn,
+              current_link->channel_offset, current_packet);
         }
         NETSTACK_RADIO.set_value(RADIO_PARAM_CHANNEL, tsch_current_channel);
         /* Turn the radio on already here if configured so; necessary for radios with slow startup */
--- a/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-ns.c
+++ b/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-ns.c
@@ -44,7 +44,6 @@
 #include "net/packetbuf.h"
 
 static uint16_t slotframe_handle = 0;
-static uint16_t channel_offset = 0;
 static struct tsch_slotframe *sf_unicast;
 
 /*---------------------------------------------------------------------------*/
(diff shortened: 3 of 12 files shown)


atiselsts:
Using multiple channel offsets for the EB rule also feels like a good idea. What do you think? I will send in out for testing it in one of my projects.

atiselsts:
I agree the hash function should be improved. I'd prefer using `sax()` from MSF or another more complex hash function. I would also prefer to make it time variant, if we go this way- for example, take into account the ASN besides the MAC address.

New attribute for the channel could work, I suppose.
Do you think the logic in tsch-slot-operation should be like this (where `packet_attr_channel_offset` is the attribute value set by the scheduler):
```
if(current_packet == NULL || packet_attr_channel_offset == 0xffff) {
  channel_offset = current_link->channel_offset;
} else {
  channel_offset = packet_attr_channel_offset;
}
```

simonduq:
Yes exactly, right before calling `tsch_calculate_channel` (and only when `burst_link_scheduled` is 0).

I have nothing against adding ASN to the hash function nor providing SAX as a readily-available option!

atiselsts:
I've changed the code as requested, moving the selection functionality to the Orchestra module.

For now the hashing remains simple - I realized that as the channel offset selection code is moved to Orchestra, time-variant hash function is not easy to do anymore. I'll leave that for future improvements.

g-oikonomou:
Changed base to branch `release-4.4`. Rebase possibly required.

os/net/mac/tsch/tsch-slot-operation.c
```diff
@@ -238,10 +238,20 @@ tsch_release_lock(void)
 
 /* Return channel from ASN and channel offset */
 uint8_t
-tsch_calculate_channel(struct tsch_asn_t *asn, uint8_t channel_offset)
+tsch_calculate_channel(struct tsch_asn_t *asn, uint16_t channel_offset, struct tsch_packet *p)
```
simonduq: could we keep `channel_offset` on 8 bits and keep the check that it's no greater than 15? The code below can still read the packetbuf attribute, and assign it to `channel_offset` only if != 0xffff

os/net/mac/tsch/tsch-schedule.c
```diff
@@ -224,9 +224,6 @@ tsch_schedule_add_link(struct tsch_slotframe *slotframe,
     if(timeslot > (slotframe->size.val - 1)) {
       LOG_ERR("! add_link invalid timeslot: %u\n", timeslot);
       return NULL;
-    } else if(channel_offset > 15) {
-      LOG_ERR("! add_link invalid channel_offset: %u\n", channel_offset);
-      return NULL;
```
simonduq: I would like it if we could keep this check

os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-ns.c
```diff
@@ -81,6 +80,12 @@ select_packet(uint16_t *slotframe, uint16_t *timeslot)
     if(timeslot != NULL) {
       *timeslot = get_node_timeslot(dest);
     }
+#if ORCHESTRA_MULTIPLE_CHANNEL_OFFSETS
+    /* set per-packet channel offset */
+    if(channel_offset != NULL) {
+      *channel_offset = ORCHESTRA_LINKADDR_HASH(dest) % (ORCHESTRA_MAX_HASH + 1);
```
simonduq: likewise, this could be % 16

simonduq:
One more thought: I think the choffset selection should not be between 0 and 15 but rather within a given limited range, assigned to the orchestra rule. This would preserve the property that orchestra slotframes are all isolated on their own (set of) channel offset(s).

os/net/mac/tsch/tsch-slot-operation.c
atiselsts: Channel offset is already 16 bits in the `struct tsch_link` which is passed as a parameter when this function is called (`tsch_calculate_channel(&tsch_current_asn, current_link->channel_offset, current_packet);`) so restricting it to 8 bits in the parameter seems like an oversight. I would like to keep the change to `uint16_t` (or if you're absolutely sure, change the value in `struct tsch_link` and other places to 8 bits)

os/net/mac/tsch/tsch-schedule.c
atiselsts: I wonder why? There can be hopping sequences with more than 16 elements. I don't think standard says anything about this. Plus, many of the current hopping sequences have less than 16 elements, including the default one! As you know, with hopping sequence 4 elements long, the channel offsets 1, 5, 9 and 13 are all going to map to the same physical channel - so why not allow channel offsets 17, 21 etc. as well?

atiselsts:
> One more thought: I think the choffset selection should not be between 0 and 15 but rather within a given limited range, assigned to the orchestra rule. This would preserve the property that orchestra slotframes are all isolated on their own (set of) channel offset(s).

I like the idea of reserving "own" channel offsets for Orchestra slotframes. One simple way would be to use the range starting from `slotframe_handle` value for the unicast slotframe, in this case no extra config would be needed.

simonduq:
> I like the idea of reserving "own" channel offsets for Orchestra slotframes. One simple way would be to use the range starting from slotframe_handle value for the unicast slotframe, in this case no extra config would be needed.

Yes I'm not sure what the best way to configure. One thing is the unicast slotframe doesn't come last as we it to take precedence over default_common

os/net/mac/tsch/tsch-schedule.c
simonduq: well yes you're right; I had been somehow assuming that the standard mandated this but doesn't seem to be any other constraint than it being 16bit unsigned :)

simonduq:
Any update on the choffset per-rule allocation @atiselsts ? Any hope to have this soon enough for the release?

atiselsts:
I'm not sure about the API yet. Do you have a good idea how the channel offset range should be configured / decided?

It would be nice to have this feature in 4.4. Testing showed quite an improvement compared with the single channel offset approach.

simonduq:
Maybe the current thing that goes through `ORCHESTRA_RULES` and assigne offsets 0, 1, 2... needs to be replaced with something else. Maybe each Orchestra rule needs its own define? And for the ones with multiple channel offsets, both a MIN and MAX defs?

os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-storing.c
```diff
@@ -74,6 +74,17 @@ get_node_timeslot(const linkaddr_t *addr)
   }
 }
 /*---------------------------------------------------------------------------*/
+static uint16_t
+get_node_channel_offset(const linkaddr_t *addr)
+{
+  if(addr != NULL && ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET > ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET) {
```
simonduq: shouldn't we allow `ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET == ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET` ?

atiselsts: True that. Updated now...

simonduq:
Awesome, thanks!

PR #1067, opened 2019-09-20
Cherry-pick from https://github.com/contiki-ng/contiki-ng/pull/497

simonduq:
This adds one of the commits of https://github.com/contiki-ng/contiki-ng/pull/497 , which adds support for both CR and LF in the shell.


g-oikonomou:
Should we also keep 3cdfa07de5f7dbe5d24a12ad8ffffb3c4ec9b1d3 ?

simonduq:
My understanding is that this is identical in behavior to what we have in develop

g-oikonomou:
Right. I hadn't realised!

PR #1074, opened 2019-09-25
Fix cc26x0/cc13x0 watchdog and LPM sleep timers

arurke:
In the cc26x0/cc13x0 platform, the watchdog timer interval is set to [0xFFFFF](https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/cc26x0-cc13x0/dev/contiki-watchdog.c#L55), which yields a ~0.7 second interval (watchdog ticks at 48 MHz / 32). However, the maximum allowed sleep, MAX_SLEEP_TIME, is one second. Thus the watchdog will bark if sleeping the maximum amount.

This PR increases the watchdog timer to 1.1 sec, to avoid this.
——
Slightly related to #1052, as that may bring forward this problem (since MCU would sleep more).


g-oikonomou:
I've always considered the watchdog reload value a magic number and never quite spent any time to figure out the conversion to SI time units. Do you have any info? Perhaps what we want to do is configure using msec and use macro to convert to the reload value assuming 48MHz? Are you saying the WDT ticks at 1.5MHz in run mode? Do you have a reference?

arurke:
Yes, see e.g. [old TI doc.](http://software-dl.ti.com/dsps/dsps_public_sw/sdo_sb/targetcontent/tirtos/2_20_00_06/exports/tirtos_full_2_20_00_06/products/tidrivers_cc13xx_cc26xx_2_20_00_08/docs/doxygen/html/_watchdog_c_c26_x_x_8h.html): "The Watchdog counts down at a rate of the device clock SCLK_HF (48 MHz) divided by a fixed-division ratio of 32, which equals to 1.5 MHz.".

> Perhaps what we want to do is configure using msec and use macro to convert to the reload value assuming 48MHz?

Seems CONTIKI_WATCHDOG_TIMER_TOP is not used anywhere else so I guess we get the power to define it. Something ala this perhaps
```
/* Watchdog timer interval, in milliseconds */
#ifdef CONTIKI_WATCHDOG_CONF_TIMER_TOP
#define CONTIKI_WATCHDOG_TIMER_TOP CONTIKI_WATCHDOG_CONF_TIMER_TOP
#else
#define CONTIKI_WATCHDOG_TIMER_TOP 1100
#endif

/* Convert timer interval from ms to ticks. Clock runs at 48 MHz / 32 */
#define WATCHDOG_TIMER_TOP_TICK ((48000 / 32) * CONTIKI_WATCHDOG_TIMER_TOP)
```
---
One additional comment:
The watchdog is patted just before we enter into sleep, but as far as I can tell, not afterwards. If so, and we spend 1 second sleeping, there would be only 100 ms left to execute `process_run()` before it would bark. One could therefore argue for e.g. 2 sec interval, to guarantee a full second of execution afterwards. This did not seem a concern for the previous implementation so maybe I am missing something?

g-oikonomou:
Thanks!

> Something ala this perhaps
Yeah looks nice I think

> The watchdog is patted just before we enter into sleep, but as far as I can tell, not afterwards. If so, and we spend 1 second sleeping

I am not quite sure how we could end up sleeping for an entire second? Surely for 1 sec we'd end up in deep sleep in which case the dog stops ticking. Unless there is a use-case whereby deep sleep is disabled, is this what you have in mind?

arurke:
> Unless there is a use-case whereby deep sleep is disabled, is this what
you have in mind?

Exactly, the LPM allows modules to block deep sleep.

arurke:
Alternatively we could pat it when waking up, as is done in ‘wake_up()’ after deep sleep

g-oikonomou:
Or in https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/cc26x0-cc13x0/lpm.c#L384

arurke:
:+1: I added these two changes now

PR #1077, opened 2019-09-28
Misc. RPL-Lite fixes

### LINKED ISSUE #1026: Probing stops after global repair.
The problem is this line and the next one:
https://github.com/contiki-ng/contiki-ng/blob/7b4de0585224d877de53d1402df28d4d31f3eef9/os/net/routing/rpl-lite/rpl-dag.c#L231

While other timers gets rescheduled, the probing timer is not scheduled in init_dag_from_dio function, but in process_dio_init_dag, which is not called in case of global repair. This causes the node to stop probing. This sometimes leads to a group of nodes joining the new version, but not being able to find any preferred parent in it, because of high etx to the DIO sender.

### LINKED ISSUE #1033: Global repair does not remove default route.
Hi,
I believe there is a bug in the global_repair_non_root function. It calls init_dag_from_dio to reinitialize the DAG, which calls init_dag, which does:
https://github.com/contiki-ng/contiki-ng/blob/7b4de0585224d877de53d1402df28d4d31f3eef9/os/net/routing/rpl-lite/rpl-dag.c#L483
This sets the preferred parent to NULL without calling rpl_neighbor_set_preferred_parent, which removes and adds default routes to preferred parent. This may lead to a node with no preferred parent sending packets to it's previous preferred parent. If a node has a preferred parent and performs global repair, this preferred parent is left as a default route. If after the repair, this node is left with no parent, it will use this leftover route as the default one.

simonduq:
Fixes #1033
Fixes #1026 

And also re-works how the preferred parent is cleared upon removing all neighbors.


g-oikonomou:
Changed base to branch `release-4.4`. Rebase possibly required.

Issue #1072, opened 2019-09-25
TSCH calls unsafe functions from interrupt context


atiselsts:
According to #841, ctimer operations are not safe in interrupt context. However, TSCH manipulates the `keepalive_timer` from `tsch_tx_slot` and `tsch_rx_slot` protothread context. It happens via the `tsch_timesync_update()` and `tsch_schedule_keepalive()` functions.

g-oikonomou:
#1078 fixes this

PR #1078, opened 2019-09-29
TSCH: move all ctimer manipulations outside from the interrupt context

atiselsts:
The idea of this PR is to do the scheduling of keepalives only from process context: see issue #1072. 

The code has been tested in a CC1310 testbed.


debug-ito:
Great work! I think this significantly improves memory integrity. Below are my comments.

- Instead of adding `tsch_schedule_keepalive_async`, I think we should make `tsch_schedule_keepalive` and `tsch_schedule_keepalive_immediately` safe. That way, we can eliminate the possibility of dangerous operation on `keepalive_timer`.

    - In this pull-request, you forgot to revise the call to `tsch_schedule_keepalive` in `tsch_tx_slot`. That is exactly why revising `tsch_schedule_keepalive` is much safer than adding a new safe function.
    - To make `tsch_schedule_keepalive_immediately` safe, I suggest to extend the semantics of `do_update_keepalive`: 0 = not necessary to update the timer, 1 = set the timer with `tsch_current_ka_timeout` delay, 2 = set the timer with zero delay.

- In `tsch_schedule_keepalive_async` (or any routine that sets `do_update_keepalive` to non-zero), I think we should call `process_poll(&tsch_pending_events_process)`.

atiselsts:
Good points, @debug-ito
Originally, this was a quick fix to get Contiki-NG TSCH working in IoT lab testbed.
Looking more carefully at the keepalive scheduling code,  I decided to rewrite some parts, get rid of the `tsch_schedule_keepalive_immediately()` function and to instead, add an argument to `tsch_schedule_keepalive()`, as well as handle your general points. The PR is updated now.

g-oikonomou:
Changed base to branch `release-4.4`

debug-ito:
Thanks! Looks good to me.

os/net/mac/tsch/tsch.c
```diff
@@ -317,25 +325,58 @@ keepalive_send(void *ptr)
...
+
+      case KEEPALIVE_SCHEDULE_OR_STOP:
+        if(tsch_current_ka_timeout > 0) {
+          /* Pick a delay in the range [tsch_current_ka_timeout*0.9, tsch_current_ka_timeout[ */
+          unsigned long delay;
+          if(tsch_current_ka_timeout >= 10) {
+            delay = (tsch_current_ka_timeout - tsch_current_ka_timeout / 10)
+                + random_rand() % (tsch_current_ka_timeout / 10);
+          } else {
+            delay = 1;
```
simonduq: should this be some delay between 1 and `tsch_current_ka_timeout`?

os/net/mac/tsch/tsch.c
```diff
@@ -317,25 +325,58 @@ keepalive_send(void *ptr)
...
-/* Set ctimer to send a keepalive message immediately */
-void
-tsch_schedule_keepalive_immediately(void)
+static void
+tsch_keepalive_process_pending(void)
 {
-  /* Pick a delay in the range [tsch_current_ka_timeout*0.9, tsch_current_ka_timeout[ */
-  if(!tsch_is_coordinator && tsch_is_associated) {
-    ctimer_set(&keepalive_timer, 0, keepalive_send, NULL);
+  if(keepalive_status != KEEPALIVE_SCHEDULING_UNCHANGED) {
```
simonduq: Initialize `keepalive_status` to `KEEPALIVE_SCHEDULING_UNCHANGED` in `tsch_reset`?
Doesn't seem strictly necessary at the moment, but more as a defensive programming

os/net/mac/tsch/tsch.c
```diff
@@ -317,25 +325,58 @@ keepalive_send(void *ptr)
...
+          if(tsch_current_ka_timeout >= 10) {
+            delay = (tsch_current_ka_timeout - tsch_current_ka_timeout / 10)
+                + random_rand() % (tsch_current_ka_timeout / 10);
+          } else {
+            delay = 1;
+          }
+          ctimer_set(&keepalive_timer, delay, keepalive_send, NULL);
+        } else {
+          /* zero timeout set, stop sending keepalives */
+          ctimer_stop(&keepalive_timer);
```
simonduq: don't we need to stop the ctimer if timeout is 0 regardless of `keepalive_status`? (risk of never stopping if an immediate KA was requested?)

os/net/mac/tsch/tsch.c
simonduq: or just maybe between `tsch_current_ka_timeout - 1` and `tsch_current_ka_timeout`, for more consistent behavior?

os/net/mac/tsch/tsch.c
atiselsts: For immediate KA, keepalive_send() stops the timer in the first line.

os/net/mac/tsch/tsch.c
atiselsts: ok, will do

os/net/mac/tsch/tsch.c
atiselsts: I think that `tsch_current_ka_timeout - 1` is the only value to use, to be pedantic about it. Can change `1` to that!

os/net/mac/tsch/tsch.c
simonduq: got it!

Issue #1054, opened 2019-09-03
Neighbor table consistency


atiselsts:
There are a number of modules in Contiki-NG that use neighbor tables:
- RPL candidate parents
- link stats
- IPv6 neighbors (called `ds6_neighbors`)
- TSCH neighbors (not an actual `NBR_TABLE`, but used in a similar role)
- other tables (to be reviewed!)

**The problem:** There is no attempt to ensure that the tables contain the same entries.

This can (and does) lead to issues. One is that the TSCH neighbors and link stats are removed much more aggressively than RPL candidate parents.
- If there is an entry for RPL parent, but a TSCH neighbor cannot be added for this parent, the parent cannot be probed.
- If the entry for link stats is deleted for a RPL parent, it can be re-added back later with a better metric. This causes the routing tree convergence time to become very long.

**Proposed solution**: introduce a global consistency mechanism between the tables.

Details:
- Could use an approach with bitmaps to implement that: only remove a neighbor when all of TSCH, RPL and link stats have freed that neighbor, i.e. the bitmap is zero.
- Common neighbor replacement policy can now be implemented, e.g. the "least recently used" (LRU) policy.
- Additionally, there should be a locking / prioritization option (the existing one may be sufficient?). For example, a TSCH timesource or RPL preferred parent should never be removed to free space for a new neighbor that is neither one of these.
- Allow to disable the IPv6 neighbor table completely, to save RAM. As discussed with George, in the normal Contiki-NG use case there is no need to keep that table at all, because the MAC address / link address mapping can be done on the fly.

simonduq:
Not denying that there are issues, but "There is no attempt [..]" is a false statement :p
The very purpose of `nbr-table` was to attempt to sync all tables; for context see https://github.com/contiki-os/contiki/pull/288

Now I agree that TSCH being not a `nbr-table` bring back inter-table inconsistencies.

Also, note that there is already in place both a bitmap to keep track of all tables in the system, a lock and a replacement policy, as well as `UIP_CONF_ND6_AUTOFILL_NBR_CACHE` to operate without IPv6-MAC mapping... but given the frustrations expressed here I assume NG and new expectations have outgrown this good old `nbr-table` :p 😢

simonduq:
[...]

PR #1079, opened 2019-09-29
Resetting of the link stats

atiselsts:
This will improve the consistency between the RPL and link-stats tables, partially solving the issue #1054 .
This is based on PR #1047 instead of develop t o avoid conflicts once 1047 is merged.

### DIFF
--- a/os/net/link-stats.c
+++ b/os/net/link-stats.c
@@ -127,6 +127,17 @@ guess_etx_from_rssi(const struct link_stats *stats)
 }
 #endif /* LINK_STATS_INIT_ETX_FROM_RSSI */
 /*---------------------------------------------------------------------------*/
+/* Both for initialization and reset */
+static void
+link_stats_entry_init(struct link_stats *stats)
+{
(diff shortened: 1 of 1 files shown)


simonduq:
Maybe clearing only ETX is the right thing to do when losing TSCH sync, but I'd say not when leaving the RPL DAG

atiselsts:
Perhaps one problem I have is that the semantics of the link stats are not clear to me.
Are they temporary cache for RPL?
Or are they history of how the node's links were performing?
If it's the second option, then link stats remain valid when the node has left the network. If it's the first, it could be clarified in the docs / wiki.

Still, the main issue here is that some parts of state are removed while others are not. Which parts should be removed and under what conditions? At the moment the code in different sitations does both of these:
1) remove RPL parent without removing link stats entry
2) remove link stats entry without removing RPL parent
This is quite confusing to understand and debug.

atiselsts:
I feel that this PR is not going anywhere, a more fundamental change would be preferable.
I'm closing this for now.

Issue #1089, opened 2019-10-11
cc1310 radio stops listening in busy network


tdesmet:
I have a network with 60 cc1310. Yesterday there was a moment with a lot of network traffic and after this moment I see a lot of nodes that are not reachable. Sniffing with sensniff I see the nodes sending out DIS packets and other nodes responding. But the affected nodes act like they do not hear the responses.

I have been debugging and noticed that the affected nodes don't seem to receive any packet anymore.

This seems related to this PR https://github.com/contiki-ng/contiki-ng/pull/567

The code base we are running on has this pr merged into it so there seems to be another issue where the radio stops listening for some reason.

g-oikonomou:
CSMA or TSCH?

tdesmet:
CSMA, cc26x0-cc13x0

g-oikonomou:
So under CSMA, there is a condition whereby the driver turns the radio off after TX. This is irrelevant for TSCH. Hence my question. If we have a different problem in the RF driver we need of course to look at it.

Also, #197 :)

tdesmet:
Can you point me to this condition to see if it is related? Thanks in advance

g-oikonomou:
#778, contiki-os/contiki#2161 and possibly others

tdesmet:
Yes I read those but cannot see how they are related to only csma :-)

Anyway I though that #567 was a fix for those.

g-oikonomou:
> Yes I read those but cannot see how they are related to only csma :-)

I think the fact that TSCH powercycles the radio anyway is what makes it disappear. I could be wrong of course, I do not understand exactly what the root cause is. Or whether they are related in the first place, just that they have similar manifestation

tdesmet:
Ok thank you, I am going to investigate further

I wonder if the simplelink platform has the same problems. Any clue?

g-oikonomou:
I do not know. But if we can create an example that breaks cc13x0-cc26x0 then we can try on the same h/w using simplelink. If the latter doesn't crash then added confidence that we need to be looking at cc26x0-cc13x0 prop mode driver.

tdesmet:
Ok I enabled debug in the radio driver and now see these messages

```
RXQ was full, re-enabling radio!
rf_cmd_prop_rx: send_cmd ret=0, CMDSTA=0x00000c23, status=0x0000
```

[...]

Issue #1084, opened 2019-10-07
Travis fails when Xenial is selected


yatch:
# Issue
Tests requiring `BUILD_COOJA` fail when Xenial is selected for the test session.

This, https://travis-ci.org/yatch/contiki-ng/builds/594673095, is an example of the issue. The following error stopped the session:

```
$ if [ ${BUILD_COOJA:-false} = true ] ; then ant -q -f $CNG_HOST_PATH/tools/cooja/build.xml jar ; fi
The program 'ant' is currently not installed. To run 'ant' please ask your administrator to install the package 'ant'
The command "if [ ${BUILD_COOJA:-false} = true ] ; then ant -q -f $CNG_HOST_PATH/tools/cooja/build.xml jar ; fi" failed and exited with 127 during .
```

This test was performed against `develop` branch in my personal repository, that points to the head commit of the official `develop` branch. Looking into the build system information, it is found `xenial` was selected.

```
Build system information
Build language: generic
Build group: stable
Build dist: xenial
```

Tests against the same commit in the official repository, https://travis-ci.org/contiki-ng/contiki-ng/builds/591679211, are green. The build system is `trusty`.

```
Build system information
Build language: generic
Build group: stable
Build dist: trusty
```

# Why this happens?
* Xenial has been the default image, but it seems this default switch has not been done completely: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment
* We don't specify `dist` in our `.travis-ci`; there is a chance Xenial is selected
* The Xenial image doesn't have `ant` installed: https://docs.travis-ci.com/user/languages/java/#using-ant-on-ubuntu-xenial-1604
* Building Cooja fails with the Xenial image because `ant` is missing.

# (short-term) Workaround
Add `dist: trusty` to `.travis-ci`

# Solution
Follow https://docs.travis-ci.com/user/languages/java/#using-ant-on-ubuntu-xenial-1604
```
dist: xenial
language: java
addons:
  apt:
    packages:
      - ant
```

yatch:
Related #1088

g-oikonomou:
Yeap. We are on skype now. We are merging your fix once travis passes.

Thanks!

yatch:
@g-oikonomou :+1:

g-oikonomou:
@yatch why don't you join us on skype if you can? (Timezone permitting!)

g-oikonomou:
Fixed in #1085

g-oikonomou:
[...]

PR #1090, opened 2019-10-11
Fix rx problem on cc1310 radio

tdesmet:
Fixes https://github.com/contiki-ng/contiki-ng/issues/1089

When the radio receives a lot of packes the rx buffer becomes full and the radio stops listening. 

before when handling the rf_core_rx_is_full flag this was logged:
```
RXQ was full, re-enabling radio!
rf_cmd_prop_rx: send_cmd ret=0, CMDSTA=0x00000c23, status=0x0000
```

The current implementation of rf_is_on tested if the status of the rx command is ACTIVE but if an error occurs in the rx command then the radio is still on and just not listening. The different usages of this function indicate that sometimes we want to know if the rx command is in the ACTIVE state and other times we just want to know if the radio is on.

So I decided to hold a flag that is set if the radio is turned on and is cleared when the radio is turned off.

Because of this the radio is no longer turned on or off when the rx command is in an error state like when the rx buffer is full. This solves a lot of issues.

After this work the rf_core_rx_is_full flag is no longer needed, we can just look at the status of the rx command. If the status is not active the rx command is send again.

It might be needed to add a check if the command is not in the PENDING state but rf_cmd_prop_rx waits for the command to reach the ACTIVE state or times out so it does not seem to be required in the current code.

### DIFF
--- a/arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
+++ b/arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
@@ -45,7 +45,6 @@
 #include "net/netstack.h"
 #include "sys/energest.h"
 #include "sys/clock.h"
-#include "sys/critical.h"
 #include "sys/rtimer.h"
 #include "sys/cc.h"
 #include "lpm.h"
@@ -129,6 +128,8 @@ static volatile rfc_dataEntry_t *packet_being_received;
 /*---------------------------------------------------------------------------*/
 static int on(void);
 static int off(void);
+
+static bool rf_setup_is_completed = false;
 
 static rfc_propRxOutput_t rx_stats;
 /*---------------------------------------------------------------------------*/
@@ -263,13 +264,23 @@ volatile static uint8_t *rx_read_entry;
 static uint8_t tx_buf[TX_BUF_HDR_LEN + TX_BUF_PAYLOAD_LEN] CC_ALIGN(4);
 /*---------------------------------------------------------------------------*/
 static uint8_t
-rf_is_on(void)
+rx_is_on(void)
 {
   if(!rf_core_is_accessible()) {
     return 0;
   }
 
   return smartrf_settings_cmd_prop_rx_adv.status == RF_CORE_RADIO_OP_STATUS_ACTIVE;
+}
+/*---------------------------------------------------------------------------*/
+static uint8_t
+rf_is_on(void)
+{
+  if (!rf_core_is_accessible()) {
+    return 0;
+  }
+
+  return rf_setup_is_completed;
 }
 /*---------------------------------------------------------------------------*/
 static uint8_t
@@ -502,7 +513,7 @@ rx_on_prop(void)
 {
   int ret;
 
-  if(rf_is_on()) {
+  if(rx_is_on()) {
     PRINTF("rx_on_prop: We were on. PD=%u, RX=0x%04x\n",
            rf_core_is_accessible(), smartrf_settings_cmd_prop_rx_adv.status);
     return RF_CORE_CMD_OK;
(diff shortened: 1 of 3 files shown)


tdesmet:
travis is complaining about ant not being installed ...

g-oikonomou:
#1084

tdesmet:
@g-oikonomou I have new fix and updated the description, testing shows that the radio keeps working now and turning rx back on does not return the weird 0x00000c23 CMDSTA

atiselsts:
There are some conflicts now, can you resolve them? We could include this in version 4.8.

atiselsts:
thanks, will test this

arch/cpu/cc26x0-cc13x0/rf-core/rf-core.c
```diff
@@ -746,8 +747,6 @@ cc26xx_rf_cpe1_isr(void)
 
   if(HWREG(RFC_DBELL_NONBUF_BASE + RFC_DBELL_O_RFCPEIFG) & IRQ_RX_BUF_FULL) {
     PRINTF("\nRF: BUF_FULL\n\n");
-    /* set a flag that the buffer is full*/
-    rf_core_rx_is_full = true;
```
arurke, line 750: The definition and declaration of `rf_core_rx_is_full` in this file and the header can also be removed now?

arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
```diff
@@ -130,6 +129,8 @@ static volatile rfc_dataEntry_t *packet_being_received;
 static int on(void);
 static int off(void);
 
+static uint8_t rf_status = 0;
```
arurke, line 132: Should we be clearer about the semantics of this new flag? For example using an enum with sensible names, or a bool like `rf_is_on`?

arurke:
@tdesmet We would like to include this in the upcoming release. Would you be able to address the comments before next week?

tdesmet:
Sorry, I forgot about this. I made the requested changes.

arurke:
Apologies @tdesmet , could you squash the commits?

tdesmet:
I think I finally got it right...

arurke:
I did a _sanity check_ with two custom boards with cc1310, no issues.

arurke:
@joakimeriksson @atiselsts I think we may have slipped on this one. iirc we were happy about it, and our plan was to merge it after the v4.8 release to have it "in the wild" for some time before v4.9.

arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
```diff
@@ -816,7 +827,6 @@ release_data_entry(void)
 {
   rfc_dataEntryGeneral_t *entry = (rfc_dataEntryGeneral_t *)rx_read_entry;
   uint8_t *data_ptr = &entry->data;
-  int_master_status_t interrupt_status;
```
atiselsts, line 819: I suppose this critical section was there to make the read-and-set operation of `rf_core_rx_is_full` atomic. As the variable is now removed, it could be safe to remove the critical section as well.

atiselsts (approved):
This is rather low confidence, as the state-machine logic of the driver is complex, and it has been a while since I looked into it, but generally looks ok. If we want to accept it in `develop` so more people test it, now could be a good time.

arurke:
@tdesmet Would you be able to address the comment from Atis, and we can get this merged?

arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
tdesmet: What critical section? I might be missing something but I don't see it...

atiselsts: It was more of a comment to myself, really, to remember that I checked this code and it seemed safe.

Previously there was a `critical_enter()` / `critical_exit()` block, your PR removes these guards.

arurke:
@tdesmet Sorry for the pings, would mind rebasing as well

Issue #1069, opened 2019-09-21
Newer CoAP clients fail on docker


g-oikonomou:
In our current docker image we use version 0.5.1 of the `coap-cli` package:

```
$ coap --version
0.5.1
$ coap get coap://[fd00::302:304:506:708]/.well-known/core
(2.05)	</.well-known/core>;ct=40,</test/hello>;title="Hello world: ?len=0..";rt="Text",</debug/mirror>;title="Returns your decoded message";rt="Debug",</test/chunks>;title="Blockwise demo";rt="Data",</test/separate>;title="Separate demo",</test/push>;title="Periodic demo";obs,</test/sub>;title="Sub-resource demo",</test/b1sepb2>;title="Block1 + Separate + Block2 demo"
```
Newer versions (0.6.0 or newer) will however fail

```
$ coap --version
0.7.0
$ coap get coap://[fd00::302:304:506:708]/.well-known/core
Invalid URL. Protocol is not given or URL is malformed.
```

(When trying natively on macOS I am getting a different error - a DNS resolve failure. Our theory is that something is messed up with URL parsing in newer versions)

This means that we are currently stuck: If a pull touches `Dockerfile` then travis will re-generate the container and the "bad" version of coap-cli will get installed. This is exactly what causes #1060 to fail (the pull touches `Dockerfile` to add ~vagrant~ edit: valgrind tests).

@alexandruioanp

g-oikonomou:
In #1060 873c0ef2d24d40bcfa8e2cc2b366ab6d751ae630 therefore the proposal is to pin the coap client package to the known working version until this has been resolved.

Yagoor:
I took a look into this issue and I believe it shouldn't be classified as a bug in Contiki-NG.

The issue happens in this section of the coap-cli code:
![image](https://user-images.githubusercontent.com/6549332/65886673-0d0f7c00-e39d-11e9-94bd-d9d04f262504.png)

By removing the try catch to see the real error, I got this:
![image](https://user-images.githubusercontent.com/6549332/65886776-3b8d5700-e39d-11e9-95ce-0fc6d1f19172.png)

Therefore this is outside the Contiki-NG scope. My guess is that out node.js is outdated and the new version of the coap-cli requires the latest version or a newer one at least.

Yagoor:
My suggestion would be to use a more stable version of the client. By stable I mean not using the node. This is one example.

[...]

PR #1093, opened 2019-10-12
Substitute the npm coap client for the one in libcoap

g-oikonomou:
libcoap is a more mature coap implementation so we remove the old npm client (`coap-cli`) in favour of the one provided with libcoap (`coap-client`).

Unfortunately a package with binaries is not provided for Ubuntu 16.04, so we have to compile from sources. We only need to do this when rebuilding the docker image.

This PR also opportunistically moves `native-ping` and `native-coap` tests from the `tun-rpl-br` travis task to the `native-runs` one.

Suggest including in 4.4 unless objections


g-oikonomou:
Draft for now: I have removed some stuff to make travis builds quicker for testing

g-oikonomou:
Also changed our vagrant bootstrapping script and ready for review

alexandruioanp:
If #1115 is merged, this can be installed as a package.

g-oikonomou:
I have reworked this PR.

We now install libcoap as a package instead of compiling from sources. We apply this change to both our docker container as well as our vagrant image.

#1069, also applies to the vagrant image. This PR fixes the situation.

Requires #1140 but otherwise ready for review, so I am removing the wip label

PR #1106, opened 2019-10-25
RPL CLassic: only allow DAGs with a parent in rpl_has_joined

### LINKED ISSUE #933: TSCH is sending EB when the RPL prefered parent is NULL
In RPL classic, the `joined` flag of a dag might be set even if the node does not have a preferred parent. As a result, TSCH may continue to send EB and create infinite timesource loops. I did not investigate whether this problem also applies to RPL lite.

According to [RFC 8180](https://tools.ietf.org/html/rfc8180#section-6.3):

>  When a RPL node joins the network, it MUST NOT send EBs before having
>   acquired a RPL Rank to avoid inconsistencies in the time
>   synchronization structure.  This applies to other routing protocols
>   with their corresponding routing metrics.  As soon as a node acquires
>   routing information (e.g., a RPL Rank, see Section 5.1.1), it SHOULD
>   start sending EBs.

Since the spirit / desired outcome of this paragraph is "to avoid inconsistencies in the time synchronization structure" I think that the RPL "has joined" functionality should be improved to also take into account parent state.

### LINKED ISSUE #1105: RPL_LOOP_ERROR_DROP is not configurable in rpl-lite
In rpl-lite, `RPL_LOOP_ERROR_DROP` macro is used in rpl-dag.c, but this macro is never defined or configurable. Instead, rpl-classic defines its configuration in rpl-conf.h, but it does not use that macro.

I think we should move the configuration of `RPL_LOOP_ERROR_DROP` in rpl-classic/rpl-conf.h to rpl-lite/rpl-conf.h.

simonduq:
Fixes #933
Fixes #1105

### DIFF
--- a/os/net/routing/rpl-classic/rpl-conf.h
+++ b/os/net/routing/rpl-classic/rpl-conf.h
@@ -91,17 +91,6 @@
 #define RPL_CONF_STATS 0
 #endif /* RPL_CONF_STATS */
 
-/* Set to 1 to drop packets when a forwarding loop is detected
- * on a packet that already had an error signaled, as per RFC6550 - 11.2.2.2.
- * Disabled by default for more reliability: even in the event of a loop,
- * packets get a chance to eventually find their way to the destination. */
-#ifdef RPL_CONF_LOOP_ERROR_DROP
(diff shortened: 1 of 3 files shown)


g-oikonomou:
@atiselsts @joakimeriksson  how do you feel about this? Looks reasonable from here

PR #1107, opened 2019-10-26
Fix and test TSCH-enabled border router for platform Zoul

g-oikonomou:
The border-router example is currently broken on `master` as well as on `develop` for platform zoul if we try to enable TSCH:

```
$ make MAKE_MAC=MAKE_MAC_TSCH
using saved target 'zoul'
  MKDIR     build/zoul/remote-reva/obj
  CC        ../../arch/platform/zoul/./platform.c
[...]
  CC        border-router.c
  LD        build/zoul/remote-reva/border-router.elf
/usr/local/gcc-arm-none-eabi-7-2017-q4-major/bin/../lib/gcc/arm-none-eabi/7.2.1/../../../../arm-none-eabi/bin/ld: build/zoul/remote-reva/border-router.elf section `.stack' will not fit in region `FRSRAM'
/usr/local/gcc-arm-none-eabi-7-2017-q4-major/bin/../lib/gcc/arm-none-eabi/7.2.1/../../../../arm-none-eabi/bin/ld: region `FRSRAM' overflowed by 1216 bytes
make: *** [build/zoul/remote-reva/border-router.elf] Error 1
rm build/zoul/remote-reva/obj/startup-gcc.o border-router.o
```

The default zoul build configuration allows the chip to enter PM2. When PM2 is allowed we only use 16 out of 32KB of the CC2538's SRAM. This is because the low 16KB do not have retention in PM2.

However, since the border router's radio is always on, the zoul will never actually drop to PM2. This commit therefore disables PM2 altogether in the firmware, allowing us to use the entire 32KB of the chip's SRAM.

This pull also adds a test for the border router example with TSCH enabled for platform zoul.


simonduq:
Shouldn't we do this under `os/services/rpl-border-router` instead?
(one possible limitation might be our max level of nested module inclusion though, but this is something that we'll need fixing eventually)

g-oikonomou:
I think doing this as per the PR is fine. This is consistent with how we already provide conditional configuration for the sky platform

simonduq:
I mean the Zoul fix is something needed in principle for any RPL BR right, so in principle I'd still lean towards having this where the BR is implemented. For sky I see it slightly differently, it's more like "let's get squeeze this particular example into a sky".

g-oikonomou:
There we go

PR #1108, opened 2019-10-26
Add support for multiple docker containers

g-oikonomou:
This pull allows us to create and use a separate docker container image each time `tools/docker` gets changed. This means that `tools/docker` changes in one branch will not influence travis tests running against a different branch (potentially causing them to fail, as was the case for #1084 & #1085). Each branch will from now on simply pull and use (or build) the correct docker image.

Each image is tagged with the hash of the most recent commit that changed `tools/docker`. Let's call this `CONTIKER_TAG`. This is retrieved by running:

```
$ git log -1 --oneline -- tools/docker/ | cut -d" " -f1
5310e7ae5
```

Travis tests will attempt to pull `contiker/contiki-ng:$CONTIKER_TAG`.

Then:
* If the pull succeeds then the PR / branch does not change `tools/docker`
* If `tools/docker` changed then the `docker pull` will fail. In this case, or if any other error occurs, we build a new docker image. This new docker image will be tagged `contiker/contiki-ng:$CONTIKER_TAG`

Lastly, if we are running on `contiki-ng/contiki-ng` and if the build was triggered by a branch, push `contiker/contiki-ng:$CONTIKER_TAG` to dockerhub.

To determine whether we are running on `contiki-ng/contiki-ng` (and not on a fork), we change the test to:

```
if [ $TRAVIS_REPO_SLUG == contiki-ng/contiki-ng ]
```

This is because `$TRAVIS_SECURE_ENV_VARS` can be true even outside `contiki-ng/contiki-ng`. It is on `g-oikonomou/contiki-ng` at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,30 +7,39 @@ notifications:
 
 # Environment setup before test script. Runs for each build
 before_install:
-  # Check if anything has changed within the docker directory
-  - DOCKER_CHANGED=`git diff --name-only $TRAVIS_COMMIT_RANGE -- tools/docker | wc -l`
-  # If Docker directory has not changed, pull image from Dockerhub. Else, build
-  # image from Dockerifle. This needs to be done for each job. Any build error
-  # will count as Travis test failure. In case this updates develop, push new
-  # image to Dockerhub (secure credentials only readable on bulids to
-  # contiki-ng/contiki-ng branches, not forks or PRs)
+  # Construct the correct container image tag corresponding to this branch/pull
+  - CONTIKER_TAG=`git log -1 --oneline -- tools/docker/ | cut -d" " -f1`
+  - DOCKER_IMG=$DOCKER_BASE_IMG:$CONTIKER_TAG
+  # Try to download the image from dockerhub. If it works, use it.
+  #
+  # If however it fails then we are most likely looking at a branch or pull
+  # that touched tools/docker. In this case, build the image.
+  #
+  # Any build error will count as Travis test failure.
+  #
+  # If the test was triggered by a branch update (e.g. a PR merge) then push
+  # the new image to dockerhub. This will only happen for builds against
+  # contiki-ng/contiki-ng, not for builds on forks.
+  - echo $DOCKER_IMG
   - >
-    if [ $DOCKER_CHANGED == 0 ]; then
-      echo "Docker image unchanged, pull from Dockerhub"
-      docker pull $DOCKER_IMG;
-    else
-      echo "Docker image changed, build from Dockerfile"
+    echo "Pulling image $DOCKER_IMG from dockerhub";
+    docker pull $DOCKER_IMG;
+    if [ $? != 0 ]; then
+      echo $CONTIKER_TAG "does not exist or pull failed";
+      echo "This is normal for PR builds and for builds on forks";
+      echo "Building from dockerfile";
       docker build tools/docker -t $DOCKER_IMG;
       if [ $? != 0 ]; then
         echo "Failed to build Docker image"
         exit 1
       fi
-      if [ $TRAVIS_SECURE_ENV_VARS == true ] && [ $TRAVIS_PULL_REQUEST == false ] && [ $TRAVIS_BRANCH == 'develop' ]; then
-        echo "This build is for an update of branch develop. Push image to Dockerhub"
+      if [ $TRAVIS_REPO_SLUG == contiki-ng/contiki-ng ] && [ $TRAVIS_PULL_REQUEST == false ]; then
+        echo "This is a branch build that updates the container. Push image to Dockerhub"
         echo $DOCKERHUB_PASSWD | docker login --username contiker --password-stdin
         docker push $DOCKER_IMG;
       fi
     fi
+
   # Build Cooja conditionally
   - if [ ${BUILD_COOJA:-false} = true ] ; then
       ant -q -f $CNG_HOST_PATH/tools/cooja/build.xml jar ;
@@ -58,7 +67,7 @@ script:
 env:
   # Global environment variables, i.e., set for all builds
   global:
-    - DOCKER_IMG='contiker/contiki-ng'
+    - DOCKER_BASE_IMG='contiker/contiki-ng'
     - CNG_HOST_PATH=`pwd`
     - OUT_OF_TREE_TEST_PATH=$HOME/out-of-tree-tests
     - OUT_OF_TREE_TEST_VER=2869ae7
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,12 @@ before_install:
         echo "This is a branch build that updates the container. Push image to Dockerhub"
         echo $DOCKERHUB_PASSWD | docker login --username contiker --password-stdin
         docker push $DOCKER_IMG;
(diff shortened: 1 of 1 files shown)


g-oikonomou:
A pleasant side-effect is that when a pull gets merged, then the build triggered by the merge will be a lot faster: The first travis job that builds the image will push it; jobs starting later will then succeed in pulling it and will not have to build it.

simonduq:
Awesome :)
One question is how do we handle the `latest` tag, for non-CI uses of the docker image? Maybe push to it whenever the Travis run is a `develop` merge?

.travis.yml
```diff
@@ -7,30 +7,39 @@ notifications:
...
+      echo "This is normal for PR builds and for builds on forks";
+      echo "Building from dockerfile";
       docker build tools/docker -t $DOCKER_IMG;
       if [ $? != 0 ]; then
         echo "Failed to build Docker image"
         exit 1
       fi
-      if [ $TRAVIS_SECURE_ENV_VARS == true ] && [ $TRAVIS_PULL_REQUEST == false ] && [ $TRAVIS_BRANCH == 'develop' ]; then
-        echo "This build is for an update of branch develop. Push image to Dockerhub"
+      if [ $TRAVIS_REPO_SLUG == contiki-ng/contiki-ng ] && [ $TRAVIS_PULL_REQUEST == false ]; then
```
simonduq: why the change from `TRAVIS_SECURE_ENV_VARS`?

.travis.yml
```diff
@@ -7,30 +7,39 @@ notifications:
...
+  #
+  # If however it fails then we are most likely looking at a branch or pull
+  # that touched tools/docker. In this case, build the image.
+  #
+  # Any build error will count as Travis test failure.
+  #
+  # If the test was triggered by a branch update (e.g. a PR merge) then push
+  # the new image to dockerhub. This will only happen for builds against
+  # contiki-ng/contiki-ng, not for builds on forks.
+  - echo $DOCKER_IMG
```
simonduq: Maybe a more verbose log here?

.travis.yml
g-oikonomou: As per the OP: `$TRAVIS_SECURE_ENV_VARS` can be true outside `contiki-ng/contiki-ng` even though the actual credentials are not available

g-oikonomou:
>  One question is how do we handle the latest tag, for non-CI uses of the docker image? Maybe push to it whenever the Travis run is a develop merge?

Good question, I think you are right. Do we want to do something about master too?

We can have `:master`, `:develop` == `:latest` maybe?

.travis.yml
g-oikonomou: Done that

.travis.yml
simonduq: Sorry I was lazy and read the code not the text :p

simonduq: Maybe what we want is check both vars?

simonduq: I hope the secure creds are indeed only available where we want them to...

simonduq:
> We can have :master, :develop == :latest maybe?

Yes that sounds right.
Maybe (for future work not this PR) we'd also want a per-release tag, so that we have a ready, testable environment for any release (even though in practice Travis changes things every now and then and renders old .travis.yml non-runnable :/)

.travis.yml
g-oikonomou: So this is what I saw on a [branch build](https://travis-ci.org/g-oikonomou/contiki-ng/jobs/603332312) under `g-oikonomou/contiki-ng`

```
$ echo "Pulling image $DOCKER_IMG from dockerhub"; docker pull $DOCKER_IMG; if [ $? != 0 ]; then
    echo $CONTIKER_TAG " does not exist or pull failed.";
    echo "Building from dockerfile";
    docker build tools/docker -t $DOCKER_IMG;
    if [ $? != 0 ]; then
      echo "Failed to build Docker image"
      exit 1
    fi
    echo "Would push $DOCKER_IMG";
    if [ $TRAVIS_SECURE_ENV_VARS == true ] && [ $TRAVIS_PULL_REQUEST == false ]; then
      echo "This is a branch build that updates the container. Push image to Dockerhub"
      echo $DOCKERHUB_PASSWD | docker login --username contiker --password-stdin
      docker push $DOCKER_IMG;
    fi
  fi

Pulling image contiker/contiki-ng:afdedad from dockerhub
Error response from daemon: manifest for contiker/contiki-ng:afdedad not found
afdedad  does not exist or pull failed.
Building from dockerfile
Sending build context to Docker daemon  6.144kB


Step 1/23 : FROM 32bit/ubuntu:16.04
[...]
Step 23/23 : CMD bash --login
 ---> Running in c3c9614ed927
 ---> d93b180c1c99
Removing intermediate container c3c9614ed927
Successfully built d93b180c1c99
Successfully tagged contiker/contiki-ng:afdedad
Would push contiker/contiki-ng:afdedad
This is a branch build that updates the container. Push image to Dockerhub
Error: Cannot perform an interactive login from a non TTY device
The push refers to a repository [docker.io/contiker/contiki-ng]
84d210f3adb0: Preparing
307f5dc1c117: Preparing
0074e8cb5ff8: Preparing
eb10466b0f03: Preparing
e91e41d09bfd: Preparing
9126e1fda065: Preparing
7d0b734d3160: Preparing
b008139cb6b0: Preparing
0dc69be94b41: Preparing
856263a8407c: Preparing
45ddea7619dd: Preparing
339fd2f0e879: Preparing
9126e1fda065: Waiting
7d0b734d3160: Waiting
b008139cb6b0: Waiting
0dc69be94b41: Waiting
856263a8407c: Waiting
45ddea7619dd: Waiting
339fd2f0e879: Waiting
denied: requested access to the resource is denied

**Comment by g-oikonomou** [comment]
> Maybe (for future work not this PR) we'd also want a per-release tag

Yes. This is what I was experimenting with a few days ago. Unfortunately we will end up with messy regex if statements to do the build/push from inside travis. Perhaps simpler to just tag the release containers each time we release. We will have the image built already, so we can simply add a second tag to it, without a need to rebuild

**Comment by g-oikonomou** [comment]
So, something like that:

```
        if [ $TRAVIS_BRANCH == develop ] || [ $TRAVIS_BRANCH == master ]; then
          echo "This is a merge in branch $TRAVIS_BRANCH";
          echo "Push also $DOCKER_BASE_IMG:$TRAVIS_BRANCH";
          docker tag $DOCKER_IMG $DOCKER_BASE_IMG:$TRAVIS_BRANCH;
          docker push $DOCKER_BASE_IMG:$TRAVIS_BRANCH;
        fi
```

g-oikonomou:
The latest commit will also tag and push `contiker/contiki-ng:master` or `:develop` when this is a build triggered by a merge into one of those two branches

.travis.yml
simonduq: Right, looks all good. Looking at the docs on `TRAVIS_SECURE_ENV_VARS` it appears that this is set whenever secure vars are configured (regardless of whether available on a given job).

simonduq:
Looks good to me. Feel free to merge if/when you feel confident enough this works as intended

g-oikonomou:
I'd say after the release

simonduq:
Sounds wise

PR #1112, opened 2019-10-31
Merge master back into develop

simonduq:


simonduq:
This is only 23 commits as `release-4.4` was merged into `develop` already not long ago

PR #1122, opened 2019-11-07
Update rpl-dag.c

vconsales:
I noticed that the function `int rpl_has_joined(void)` in rpl-classic returns 0 if it is called by the root of the dodag. Because of that issue the root dodag will never send a tsch beacon since the condition
`TSCH_RPL_CHECK_DODAG_JOINED()`  inside the process `tsch_send_eb_process` always fails.
The fix is straightforward: if I am the root of rpl dag I am joined to the dag by definition.
I tried this code with with Zolertia zoul RE-Mote revision B.


g-oikonomou:
Possible bug / bugfix. Need someone to look at this.

PR #1124, opened 2019-11-11
fix PROCESS_EVENT_EXIT not called

Luka234:
# Bug description

```
process_init();
process_start(&TestProcess, 0);
process_poll(&TestProcess);
process_run();
process_exit(&TestProcess);   // this line will NOT call PROCESS_EVENT_EXIT event in TestProcess
```

- process_current state remains 'dirty'
   after process poll or event
 - if process_exit is called for last polled
   process, that process will NOT receive
   PROCESS_EVENT_EXIT event

`process_exit` will call `exit_process(p, PROCESS_CURRENT());`.
https://github.com/contiki-ng/contiki-ng/blob/86934a16dc44fe0f90af5f151d1478d6658f4a27/os/sys/process.c#L204

exit_process function will call PROCESS_EVENT_EXIT event only if current process is not exiting process.
https://github.com/contiki-ng/contiki-ng/blob/86934a16dc44fe0f90af5f151d1478d6658f4a27/os/sys/process.c#L153

process_current status will (wrongly) point to last polled (do_poll) or event sent (do_event) process.

# After this fix is applied

`process_current` state is saved and restored inside function call_process.

# How to reproduce

```
PROCESS(TestProcess, "");
int num_of_received_exit_events = 0;
int num_of_exit_commands = 0;
PROCESS_THREAD(TestProcess, ev, data)
{
    if (ev == PROCESS_EVENT_EXIT)
    {
        num_of_received_exit_events ++;
    }

    PROCESS_BEGIN();
    while(1)
    {
        PROCESS_YIELD();
    }
    PROCESS_END();
}

void main()
{
    process_init();
    while (1)
    {
        process_start(&TestProcess, 0);
        process_poll(&TestProcess);  // after this process_current will point to TestProcess
        process_run();
        num_of_exit_commands ++;
        process_exit(&TestProcess);  // this will not fire PROCESS_EVENT_EXIT because of 'dirty' process_current
     }
}
```

### DIFF
--- a/os/sys/process.c
+++ b/os/sys/process.c
@@ -175,6 +175,7 @@ static void
 call_process(struct process *p, process_event_t ev, process_data_t data)
 {
   int ret;
+  struct process *old_current = process_current;
 
 #if DEBUG
   if(p->state == PROCESS_STATE_CALLED) {
@@ -196,6 +197,8 @@ call_process(struct process *p, process_event_t ev, process_data_t data)
       p->state = PROCESS_STATE_RUNNING;
     }
   }
+
+  process_current = old_current;
 }
 /*---------------------------------------------------------------------------*/
 void
@@ -361,10 +364,7 @@ process_post(struct process *p, process_event_t ev, process_data_t data)
 void
 process_post_synch(struct process *p, process_event_t ev, process_data_t data)
 {
-  struct process *caller = process_current;
-
   call_process(p, ev, data);
-  process_current = caller;
 }
 /*---------------------------------------------------------------------------*/
 void


g-oikonomou:
Hi @Luka234 thanks for this, but I am afraid I am not so sure about it.

In the example you are providing to reproduce this you have replaced the main function. What you are observing happens because you are then calling `process_exit()` from outside a process context, and I am pretty certain you are not meant to do that. 

This is [discussed in a bit more detail here](https://github.com/contiki-ng/contiki-ng/wiki/Documentation:-Processes-and-events#stopping-processes):

> Another process kills the process by calling process_exit().

Using our existing `main()` function, `process_exit()` will always get called from within the context of some other process (I am willing to be corrected here). The last polled process will therefore only be stored in `process_current` for as long as it is executing. Once it yields, then it will be replaced in `process_current` by the next process to be called. Therefore the problem you are seeing goes away.

Can you perhaps provide a way to reproduce this without changing `main()` (and without calling `process_exit()` inside an interrupt, which should also not be done)?

Luka234:
I'm using just part of contiki in our embedded devices - it is bare metal with contiki procesess used for some system parts. In that context, calling process_start and process_exit can be called outside of contiki process (but of course, from interrupt we are calling only process_poll). I don't see any inherent reason why contiki processes shouldn't be used as described. We are using it in that way for many years now. Only in our latest project we used process_exit for some resource cleaning and encountered problem. Localy we have applied fix. I thought this fix would increase usability value of contiki processes without any sacrifice. On the other hand I've never used full contiki OS so I can't be sure are there some fine grained consequences of applying this fix.

g-oikonomou:
Interesting use-case. My main concern is that `process.[ch]` have been stable for like 20 years and I am disinclined to start messing around with them in order to support non-mainstream cases.

Have you perhaps looked at `PROCESS_CONTEXT_BEGIN()` & `PROCESS_CONTEXT_END()`?

Luka234:
I understand your concern. I could avoid using proposed fix but it would be unnatural in our codebase (and error prone).
We will use changed process.c and you can return to this in the future if you find more reasons for it.
Best regards

g-oikonomou:
Hi @Luka234. We just had a discussion internally about this. We are disinclined to change the process API in order to support a use-case that is not how we envision Contiki-NG being used. I am thus closing this. Thanks for your interest in the project!

Issue #1136, opened 2019-11-23
The mqtt-client example causes cooja to die a terrible death


g-oikonomou:
To reproduce, create a new sim inside docker. Create two cooja motes: 1x border router and 1x mqtt-client. Follow the necessary steps in order for the BR to obtain a prefix and start the network.

If the MQTT client and the BR are outside each other's radio range, or while the BR is still searching for a prefix things work fine. As soon as they get within range though, cooja dies with this output:

```
[...]
     [java]  INFO [Thread-6] (Simulation.java:253) - Simulation main loop started, system time: 1574522544845
     [java]  INFO [Thread-7] (SerialSocketServer.java:478) - Forwarder: socket -> serial port
     [java]  INFO [Thread-5] (SerialSocketServer.java:425) - Client connected: /127.0.0.1
     [java] #
     [java] # A fatal error has been detected by the Java Runtime Environment:
     [java] #
     [java] #  SIGSEGV (0xb) at pc=0xc12f51ff, pid=7796, tid=0xc2073b40
     [java] #
     [java] # JRE version: OpenJDK Runtime Environment (8.0_222-b10) (build 1.8.0_222-8u222-b10-1ubuntu1~16.04.1-b10)
     [java] # Java VM: OpenJDK Server VM (25.222-b10 mixed mode linux-x86 )
     [java] # Problematic frame:
     [java] # C  [mtype679.cooja+0x1b1ff]  uip_icmp6_send+0x5d
     [java] #
     [java] # Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
     [java] #
     [java] # An error report file with more information is saved as:
     [java] # /home/user/contiki-ng/tools/cooja/build/hs_err_pid7796.log
     [java] Java Result: 139
```

So it appears as though the problem will only occur while the mqtt client is trying to join the RPL network, or after it has done so.

The problem only appears with RPL lite, not with classic. The problem only appears if there is an mqtt-client in the simulation, other examples work fine.

Affects 4.2, 4.3 and latest `develop`

yatch:
I reproduced the issue. `SIGSEGV` occurred here, L239, where `dest` is `NULL`:
```
 229 void
 230 uip_icmp6_send(const uip_ipaddr_t *dest, int type, int code, int payload_len)
 231 {
 232   UIP_IP_BUF->vtc = 0x60;                                                                                                               233   UIP_IP_BUF->tcflow = 0;
[...]

PR #1137, opened 2019-11-23
Fix Issue #1136 (issue of mqtt-client example + rpl-lite)

yatch:
See Issue #1136 for details. 

@g-oikonomou Can you confirm this fix works on your side? I did a test using the `.csc` files attached to Issue #1136 .

### DIFF
--- a/examples/mqtt-client/mqtt-client.c
+++ b/examples/mqtt-client/mqtt-client.c
@@ -250,6 +250,16 @@ static const uint8_t mqtt_client_extension_count = 0;
 /*---------------------------------------------------------------------------*/
 PROCESS(mqtt_client_process, "MQTT Client");
 /*---------------------------------------------------------------------------*/
+static bool
+have_connectivity(void)
+{
+  if(uip_ds6_get_global(ADDR_PREFERRED) == NULL ||
+     uip_ds6_defrt_choose() == NULL) {
+    return false;
+  }
+  return true;
+}
+/*---------------------------------------------------------------------------*/
 static int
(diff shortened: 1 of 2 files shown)


g-oikonomou:
The fix works yes.

examples/mqtt-client/mqtt-client.c
```diff
@@ -591,12 +601,12 @@ connect_to_broker(void)
 static void
 ping_parent(void)
 {
-  if(uip_ds6_get_global(ADDR_PREFERRED) == NULL) {
-    return;
+  if(have_connectivity()) {
+    uip_icmp6_send(uip_ds6_defrt_choose(), ICMP6_ECHO_REQUEST, 0,
+                   ECHO_REQ_PAYLOAD_LEN);
+  } else {
+    /* we don't have connectivity; cannot send a ping */
```
g-oikonomou: Do we need the else block here? (Unless we add some logging)

yatch: Let me add `LOG_ERR` here :-)

g-oikonomou: I wouldn't call this quite an error. More of a "not ready yet". I'd either warn or dbg.

yatch: oh, alright!

yatch:
@g-oikonomou Thank you for the review. I put `LOG_WARN` and performed force-push.

g-oikonomou:
OK 👍 will merge when back near the puter, unless someone else gets there first. Thanks!

PR #1142, opened 2019-11-25
[WIP] Proposed docker container refactor

pipex:
The current contiker docker image although is very useful, it is quite
large in size (1.2GB compressed, 3.5 uncompressed). I propose a refactor
of the Dockerfile that allows to produce smaller intermediate images for
different targets, making it faster to get started with Contiki-NG

I divided the dockerfile into 5 targets, each one extending the one
above with additional functionality

- minimal, containing build packages and tools to build contiki examples
  in native (~450MB uncompressed)
- cooja, including the packages in minimal, plus the msp430 toochain and java packages to test
  simulations in cooja (~1GB uncompressed)
- arm, to build arm-based projects (~1.5GB uncompressed)
- nrf52, which adds nrf52 sdk (~1.6GB uncompressed)
- nxp, adding nxp build tools plus the renode tool and its dependencies
  (~3GB uncompressed)
- docs, which adds additional packages to build the documentation
  (~3.4GB uncompressed)

Each target can be built using the following command
```
docker build contiker/contiki-ng:$TARGET --target $TARGET .
```

To compensate a little for the space added by the additional layers, I
also removed some packages that I think should be unnecessary (e.g. g++,
dpkg-dev in the build-essential package, or java-11 in the default-jdk
package, amonst others).

The full docker image *should* be fully backwards compatible with the
old image, although I would appreciate if someone could review it.

What is left to do in the PR

- [x] Refactor dockerfile
- [ ] Review new containers and run tests
- [ ] Define new target image names in refactor
- [ ] Update Travis to create new targets
- [ ] Update documentation

Let me know if this appears interesting to you and I can keep working on it

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -5,7 +5,6 @@ FROM i386/ubuntu:18.04 as minimal
 ENV DEBIAN_FRONTEND noninteractive
 ARG USER=user
 ARG UID=1000
-ARG GID=100
 
 # Update path with toolchains
 ENV PATH=${PATH}:/opt/bin
@@ -44,7 +43,7 @@ RUN apt-get -qq update && \
 
 # Create user, enable X forwarding, add to group dialout
 # -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix
-RUN useradd -m -s /bin/bash -N -u $UID $USER && \
+RUN useradd -m -s /bin/bash -u $UID $USER && \
     echo "${USER} ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers && \
     chmod 0440 /etc/sudoers && \
     usermod -aG dialout $USER
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -1,53 +1,131 @@
-FROM i386/ubuntu:18.04
+# The minimal image should be enough to build contiki for native
+# targets
+FROM i386/ubuntu:18.04 as minimal
 
 ENV DEBIAN_FRONTEND noninteractive
+ARG USER=user
+ARG UID=1000
+ARG GID=100
+
+# Update path with toolchains
+ENV PATH=${PATH}:/opt/bin
+
+# Base Tools
+RUN apt-get -qq update && \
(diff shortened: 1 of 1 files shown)


g-oikonomou:
I'm 100% sold on this

To be honest with you, I'd take remove the `docs` container and simply include docs in `minimal`

As you are most likely well aware, travis automatically pushes docker files to dockerhub every time `tools/docker` gets touched. How do you envision this changing as part of this pull?

pipex:
I created `docs` just to reduce the size of the `minimal` image, but it's only around 80MB, which probably doesn't justify the additional layer.

Regarding travis, my guess is that the easiest would be to build the intermediate targets with their respective tags right [before pushing](https://github.com/contiki-ng/contiki-ng/blob/f87846cfa082e4b4cc43f8827ec9eaebcbbaa3ba/.travis.yml#L41). Since the full image is already built, the intermediate images should build very quickly.

A more involved solution would be to build the image as it is being tested, i.e. for the native tests build the minimal image, for the msp430 tests build the cooja image, etc. This would have the advantage of speeding the testing process, although it would add complexity to the travis script.

By the way, speaking of Travis. I see that [one of the tests failed](https://travis-ci.org/contiki-ng/contiki-ng/jobs/617349104?utm_medium=notification&utm_source=github_status), however I don't understand why, and cannot seem to replicate the failure in my local setup.

What do you think the target names should be? I would propose
- native
- cooja
- arm
- nRF52
- nxp (or riscv?)

What do you think?

pipex:
ping

LeoVerto:
I'm pretty sure there is even more potential to slim the image down by not including the entire `renode` repository in the built image and instead using a builder image to build the .deb and then installing it in the following image.

simonduq:
One suggestion would be to have one for simulation, with both Cooja and Re-node.
The NXP one should not include Re-node (note that the supported NXP platforms are not RISC-V ones).

W.r.t. pushing to dockerhub from travis, I guess we want to do the same as today, but for each target, with an image tag such as `$TARGET-$COMMIT_HASH` and `$TARGET-$BRANCH`

simonduq:
Hi!

Issue #1145, opened 2019-11-27
TSCH base drift ppm for CC1310


hseokman:
arch/cpu/cc26x0-cc13x0/cc13xx-cc26xx-def.h defines TSCH_CONF_BASE_DRIFT_PPM as -977.
It is calculated for 10ms time slot.
Since cc1310 uses 40ms time slot, should this be changed accordingly?
Slot length 40000 usec which is 1311 ticks
1311 ticks is 40008.544922 usec
-8.544921 / 40008.544921 = -0.0002135774.
Therefore I think it needs to be -214 for cc1310.

I don't quite follow the calculation for the 10 ms timeslot. Should this be
-9.765625 / 10009.765625 = -0.00097560, so -976 instead of -977?

Also this base drift is not compensated for the first sleep.
tsch_slot_operation_start() from os/net/mac/tsch/tsch-slot-operation.c
does not call tsch_timesync_adaptive_compensate. There is no drift calculated for the association, but base drift rate still needs to accounted.

atiselsts:
I think you are right on all accounts. Can you make a PR?

hseokman:
I just created PR. I think I'm mistaken in PPM calculation. It's supposed to be delta/target, so -977 for the 2.4 Ghz is correct.

atiselsts:
Let's take the 10 millisecond slots.

My reasoning is this:
10009.765625 microseconds is the time duration of one uncompensated TSCH slot,

Now assume compensation with base drift 977 ppm. The actual slot length becomes:
 10009.765625 * (1 - 0.000977) = 9999.986083984375 microseconds - not bad.

Now assume compensation with base drift 976 ppm:
10009.765625 * (1 - 0.000976) = 9999.99609375 - better!

hseokman:
I guess the actual calculation depends on how the compensation math is done. At first I thought your way because that seems just natural about compensation, but the ppm is really 'Variance/Target'. 
I looked at the how the compensation work and it basically tries to add 'time_delta_usec * ppm' to the calculation every time its called. It also records remainder and tick conversion error every time to calculate the next compensation.
When the compensate method is called for the 10 ms, the device actually delays for 328 ticks which is 10009.765625 usec 
for drift -977 ppm
It adds 10000 * (-0.000977) = -9.77 usec for compensation
Error: 10009.765625 - 9.77 = 9999.995625
for drift -976 ppm
It adds 10000 * (-0.000976) = -9.76 usec for compensation
[...]

PR #1148, opened 2019-12-04
Correct base drift for CC13XX SubGhz, Compensate base drift at association

hseokman:
This addresses issue #1145 .

CC1310 uses slot length of 40000 usec instead of 10000 usec.
The base drift rate is different from 2.4 Ghz.

When the first EB is received, the next time to wake up calculation needs to include the base drift compensation.


atiselsts:
Thanks for this.

I have two comments:
1) Also change the 2.4 GHz base drift to -976 - I think your calculation in the issue 1145 is correct. Not sure about the comment in the header file.
2) Please add /* */ comments to the preprocessor conditional `#else` and `#endif` branches, especially the last two `#endif` in row.

hseokman:
1. I think my calculation was an error, and the current comment is right. I should have divided by the target not measured to calculated the ppm.
2. I added comment to the preprocessor. I don't know if this is what you meant.

PR #1151, opened 2019-12-11
Add mpmc-ring, a multi-producer multi-consumer lock-free ring buffer

debug-ito:
An alternative solution to #883. Instead of modifying ringbufindex, I made a new data structure.

This pull-request adds a new data structure, os/lib/mpmc-ring, and its test program, examples/libs/mpmc-ring-interrupt. mpmc-ring is like ringbufindex, but it supports multi-producer multi-consumer scenarios. To do that, mpmc-ring requires extra memory to keep track of parallel transactions. I also added comments to ringbufindex's document that it's for the single-producer single-consumer scenario only.

## Test

I tested mpmc-ring with the example mpmc-ring-interrupt. The example program sets up two producers and two consumers in normal and interrupt contexts, streams messages through the queue and checks the output. See [README of the example](https://github.com/debug-ito/contiki-ng/blob/pr-mpmc-ring/examples/libs/mpmc-ring-interrupt/README.md) for detail.

Test environment:

- code: https://github.com/contiki-ng/contiki-ng/commit/c51a1998dc503ee6983a2180f587f42ccb191644
- hardware: CC1310 LaunchPad
- TARGET: simplelink
- BOARD: launchpad/cc1310
- compiler: the one in the Contiker

I ran the example with the following six configurations.

| Case | `START_GET_NUM` | `PUT_INTERVAL` | `GET_INTERVAL` | `QUEUE_LEN` | Note  |
| ---- | --------------- | -------------- | -------------- | ----------- | ----- |
| 1.   | 0               | 1              | 1              | 64          | The queue is always nearly empty. |
| 2.   | 32              | 1              | 1              | 64          | The queue is almost always half-occupied. |
| 3.   | 63              | 1              | 1              | 64          | The queue is almost always nearly full. |
| 4.   | 0               | 1              | 2              | 64          | The producer is faster.   |
| 5.   | 63              | 2              | 1              | 64          | The consumer is faster.   |
| 6.   | 0               | 1              | 2              | 256         | The producer is faster and the queue size is the maximum.  |

Using six different LaunchPads, I ran the above six cases in parallel for about 115 hours. Each LaunchPad ran about 70,000 trials with no errors.

### DIFF
--- a/examples/libs/mpmc-ring-interrupt/Makefile
+++ b/examples/libs/mpmc-ring-interrupt/Makefile
@@ -1,0 +1,8 @@
+CONTIKI_PROJECT = mpmc-ring-interrupt
+all: $(CONTIKI_PROJECT)
+
+MAKE_MAC=MAKE_MAC_NULLMAC
+MAKE_NET=MAKE_NET_NULLNET
+
+CONTIKI = ../../..
+include $(CONTIKI)/Makefile.include
--- a/examples/libs/mpmc-ring-interrupt/README.md
+++ b/examples/libs/mpmc-ring-interrupt/README.md
@@ -1,0 +1,45 @@
+# mpmc-ring-interrupt example
+
+This example demonstrates and tests interrupt-safety of
+lib/mpmc-ring.
+
+## What it does
+
+It prepares a message queue managed by mpmc-ring. Two producers put
+some messages into the queue. One producer runs in normal context,
+and the other in interrupt context (using rtimer). At the same time,
+two consumers (in normal and interrupt contexts, respectively) get
+messages from the queue and stores them in dedicated memory regions.
+
+                          put             get
+       NORMAL (Producer) --+               +--> (Consumer) --> [store]
+                           |               |
+                           +-> --------- >-+
+                               | queue |
+                           +-> --------- >-+
+                           |               |
+    INTERRUPT (Producer) --+               +--> (Consumer) --> [store]
+
+After all activities are finished, it checks the content of the two
+message stores. If the stores do not contain the exact messages put
+by the producers, it prints error messages.
+
+If the check succeeds, it re-initializes the producers, consumers
+and queue, and run the same test again. If the check fails, it
+resets.
+
+## Configuration
+
+In project-conf, you can configure number of messages two produers
+put to the queue and two consumers get from the queue. You can also
+configure speed of the producers and consumers relative to each
+other.
+
+## Expectations
+
+This example should run without any error logs even though there are
+multiple producers and consumers.
+
+## Author
+
+Toshio Ito <toshio9.ito@toshiba.co.jp>
--- a/examples/libs/mpmc-ring-interrupt/mpmc-ring-interrupt.c
+++ b/examples/libs/mpmc-ring-interrupt/mpmc-ring-interrupt.c
@@ -1,0 +1,593 @@
+/*
+ * Copyright (C) 2020, Toshiba Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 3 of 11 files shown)


yatch:
I'm just impressed with what you have done for this PR..., code, documentation, and experiments... 

Related Issue #764

debug-ito:
Thanks, but unfortunately I just came up with a scenario that probably breaks `mpmc_ring`, which is not covered by mpmc-ring-interrupt test.

1. A process in normal context calls `put_begin`. Inside the function, `ring->put_ptr` is saved to `tmp_put`.
2. An interrupt happens. It calls `put_begin`, `put_commit`, `get_begin`, `get_commit` several times. Those calls move `ring->get_ptr` and `ring->put_ptr` to arbitrary positions.
3. Return to the normal process from the interrupt. Now `tmp_put` can be in an invalid region.

So, obviously the current `mpmc-ring` requires some conditions to work properly. I'm not sure I can fix it to support ALL multi-producer multi-consumer scenarios.

debug-ito:
Hmm, maybe the ring should keep min and max of get/put ptrs? and we can remove GETTING and PUTTING states?

simonduq:
I added the WIP label just judging from the PR title -- but let me double-check: is this still work in progress?

debug-ito:
Thanks for comment. Yes, this is still work-in-progress.

debug-ito:
I re-implemented `mpmc-ring` and force-pushed it. The new implementation is borrowed from Rust implementation, [heapless::mpmc](https://docs.rs/heapless/0.5.5/heapless/mpmc/index.html). I think this is more robust to many interrupt scenarios and simpler in terms of memory access pattern.

## Updated API

Due to change in implementation, the API of mpmc-ring is now a little different from ringbufindex. mpmc-ring defines a new struct `mpmc_ring_index_t` for indices, while ringbufindex just uses plain integer for it.

```c
MPMC_RING(ring, 32);           // Declare and define a mpmc-ring as static variable.
struct your_data elements[32]; // Define memory region to keep the data in the queue.

void some_func() {
  mpmc_ring_index_t index;
  if(mpmc_ring_put_begin(&ring, &index)) { // Obtain index
    // Queue has room for new data
    write_to(&elements[index.i]);          // Access queue element at index.i
    mpmc_ring_put_commit(&ring, &index);   // Finish putting
  }
}
```

For detailed usage, see the code in the simulation test.

## Simulation test

I added Cooja-based tests to tests/07-simulation-base/08-mpmc-ring.csc. It tests basic functionality of mpmc-ring and some corner cases, for different queue lengths.

## Real device test

Just like I did in [the original post](https://github.com/contiki-ng/contiki-ng/pull/1151#issue-351695702), I ran examples/lib/mpmc-ring-interrupt program on real devices.

- code: https://github.com/debug-ito/contiki-ng/commit/7c9b95418ce2df6c8d8dbe8da75c48e67832ac81
- hardware: CC1310 LaunchPad
- TARGET: simplelink
- BOARD: launchpad/cc1310
- compiler: the one in the Contiker

I ran cases 1. to 5. for 94 hours using five different LaunchPads. (Case 6. is removed because now the maximum length of mpmc-ring queue is 64). Each LaunchPad ran about 60,000 trials with no error.

simonduq:
Hi!

arurke:
@debug-ito Any chance you are still around and want to rebase this? Even though this is part of a larger effort in #764, and I understand that effort will not be continued, the mpmc-ring is useful in its own right. :+1: for the examples and tests as well. I can run some tests on my own to further the confidence.

debug-ito:
I'm sorry but I don't have time or energy to work on this pull-request any more. If anyone took over it and merged it to the main branch, that would be great.

PR #1162, opened 2019-12-17
TSCH scheduling: prioritize TSCH links by the number of packets in the queue

atiselsts:
This PR adds two changes:
1) Key TSCH links by channel offset as well as by timeslot; 
2) When there are multiple links at the same slotframe in the same timeslot, prioritize the one with the most packets to send (as determined by looking at the destination address of the link).

Test cases and code that actually uses this functionality remain to be added.


debug-ito:
Looks cool! Thanks. I have two comments.

1: `target_neighbor` field is added to `tsch_link`. Is it necessary? I think the `addr` field already plays the role, and no one seems to initialize the `target_neighbor`.

2: I wonder what we should do when two links WITHOUT TX flag are scheduled at the same timeslot but different channel offsets (i.e. implicit `else` clause at https://github.com/contiki-ng/contiki-ng/pull/1162/files#diff-2bf15cd0c720d583e9643da96e9b5e5cR410 ). Current implementation (i.e. just don't update the `curr_best`) is ok, but we could make it customizable instead.

For example, define a link comparator

```c
typedef const struct tsch_link * (* tsch_link_comparator)(const struct tsch_link *a, const struct tsch_link *b);
```

and add a configuration macro.

```c
#ifdef TSCH_CONF_LINK_COMPARATOR
#define TSCH_LINK_COMPARATOR TSCH_CONF_LINK_COMPARATOR
#else
#define TSCH_LINK_COMPARATOR(a,b) (a)
#endif
```

We can use `TSCH_LINK_COMPARATOR` to get the best link. Users can define their own link comparator and set it to `TSCH_CONF_LINK_COMPARATOR`.

Maybe we can do that even for the case where the two links have TX flag, and provide a default link comparator that selects the link with more packets towards the target.

atiselsts:
Thanks. Agree on both points. The `target_neighbor` is simply a leftover of the previous version. I will make changes in the PR.

I would go for your last idea regarding the second point.

On a similar note, I believe that there also should be a more flexible way to prioritize different slotframes over each another: currently it is not possible to ever prioritize Rx links over Tx links, which is bad, as in the current Orchestra schedule this means that sometimes schedule mismatch will occur. If you have ideas

debug-ito:
> currently it is not possible to ever prioritize Rx links over Tx links

This is basically because the standard says a device prioritizes Tx links over Rx links (sec. 6.2.4.6, IEEE 802.15.4-2015). I'm ok with giving users more flexibility about it, as long as the default behavior complies with the standard.

atiselsts:
To continue this digression,  the use case I have in mind is prioritizing the EB slotframe over other Orchestra slotframes. In my mind, the important thing to be standard compatible is not follow the standard letter by letter, but to be fully interoperable with nodes that _do_ follow the standard letter by letter. A node that prioritizes Rx'ing in the EB slotframe over sending packet in some other slotframe remains interoperable with a standard reference implementation node.

Also, on a different note, my reading of the standard always was that it's slightly ambiguous - it says to prioritize Tx over Rx, and lower slotframe handle over higher slotframe handle. The order of these prioritization rules is only implied in the text, not fully spelled out.

debug-ito:
> A node that prioritizes Rx'ing in the EB slotframe over sending packet in some other slotframe remains interoperable with a standard reference implementation node.

I think so, and I'm ok with that behavior.

> The order of these prioritization rules is only implied in the text, not fully spelled out.

I totally agree with it!

atiselsts:
@yatch could you comment on this? The PR touches some sixtop code.

yatch:
OK.

examples/6tisch/etsi-plugtest-2017/sf-plugtest.c
```diff
@@ -402,9 +409,10 @@ delete_req_handler(const linkaddr_t *peer_addr,
   assert(cell_list_len == sizeof(sf_plugtest_cell_t));
   memcpy(&pending_cell, cell, sizeof(pending_cell));
   timeslot = pending_cell.slot_offset[0] + (pending_cell.slot_offset[1] << 8);
+  channel_offset = pending_cell.channel_offset[0] + (pending_cell.channel_offset[1] << 8);
 
   if((slotframe = tsch_schedule_get_slotframe_by_handle(0)) == NULL ||
-     (link = tsch_schedule_get_link_by_timeslot(slotframe, timeslot)) == NULL ||
+    (link = tsch_schedule_get_link_by_timeslot(slotframe, timeslot, channel_offset)) == NULL ||
```
yatch: Don't we need a heading white space here for the correct indentation? Similar things are seen in other places.

atiselsts:
I added an example - now ready to merge from my perspective!

yatch:
@atiselsts Great!
@debug-ito Thank you for your review, and triggering the good discussion.

PR #1164, opened 2019-12-17
TSCH EB timer and documentation fix

atiselsts:
Two things:
1. Do not start TSCH EB sending process when the TSCH_EB_PERIOD is set to zero. Otherwise, the timer runs every tick and wastes CPU resources.
2. Document that `tsch_set_eb_period(0)` is not sufficient to turn off EB sending if RPL is used.

### DIFF
--- a/os/net/mac/tsch/tsch-conf.h
+++ b/os/net/mac/tsch/tsch-conf.h
@@ -71,14 +71,20 @@
 #define TSCH_DESYNC_THRESHOLD (2 * TSCH_MAX_KEEPALIVE_TIMEOUT)
 #endif
 
-/* Period between two consecutive EBs */
+/* The default period between two consecutive EBs (not taking into account any randomization).
+ * When TSCH_CONF_EB_PERIOD is set to 0, sending EBs is disabled completely; the EB process is not started.
+ * Otherwise, if RPL is used, TSCH_CONF_EB_PERIOD used only before joining the RPL network;
+ * afterwards, the EB period is set dynamically based on RPL DIO period, updated whenever
+ * the DIO period changes, and is upper bounded by TSCH_MAX_EB_PERIOD.
+ */
 #ifdef TSCH_CONF_EB_PERIOD
 #define TSCH_EB_PERIOD TSCH_CONF_EB_PERIOD
 #else
 #define TSCH_EB_PERIOD (16 * CLOCK_SECOND)
 #endif
 
-/* Max Period between two consecutive EBs */
+/* Max Period between two consecutive EBs.
+ * Has no effect when TSCH_EB_PERIOD is zero. */
 #ifdef TSCH_CONF_MAX_EB_PERIOD
 #define TSCH_MAX_EB_PERIOD TSCH_CONF_MAX_EB_PERIOD
 #else
--- a/tests/17-tun-rpl-br/02-border-router-cooja-tsch.csc
+++ b/tests/17-tun-rpl-br/02-border-router-cooja-tsch.csc
@@ -26,7 +26,7 @@
       <description>Cooja Mote Type #1</description>
       <source>[CONTIKI_DIR]/examples/rpl-border-router/border-router.c</source>
       <commands>make TARGET=cooja clean
-make -j border-router.cooja TARGET=cooja MAKE_MAC=MAKE_MAC_TSCH</commands>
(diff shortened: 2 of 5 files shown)


debug-ito:
`turn_on` function in tsch.c is the only place that starts the `tsch_send_eb_process`. If we disable it here, I think the node cannot start sending EBs even if it calls `tsch_set_eb_period` with non-zero period later.

If you intend to extend the meaning of `TSCH_CONF_EB_PERIOD` so that setting it to zero disables EB transmissions COMPLETELY, we have to explain that in the document in tsch-conf.h. However, I would rather add a new configuration macro to enable/disable EB transmissions. That would be simpler and backward-compatible.

atiselsts:
> I think the node cannot start sending EBs even if it calls tsch_set_eb_period with non-zero period later.

Yes, this is what I had in mind - if the `TSCH_CONF_EB_PERIOD` is zero, the node cannot send EBs. This may be worth thinking about, however, I saw that setting this configuration define to zero is what Contiki-NG users find intuitive.

Regarding backwards compatibility, there is none to worry about.  #1051 is only recently merged, and before that with `TSCH_CONF_EB_PERIOD` value of 0 it was impossible to compile the code.

debug-ito:
Thanks for reply.

> before that with `TSCH_CONF_EB_PERIOD` value of 0 it was impossible to compile the code.

I see.

In that case, I think we can treat `TSCH_CONF_EB_PERIOD == 0` as disabling EBs altogether. We should still explain that in the document of `TSCH_CONF_EB_PERIOD`, though.

atiselsts:
Updated, please check this now.

debug-ito:
Thanks! This is a cool feature addition.

tests/17-tun-rpl-br/02-border-router-cooja-tsch.csc
```diff
@@ -52,7 +52,7 @@ make -j border-router.cooja TARGET=cooja MAKE_MAC=MAKE_MAC_TSCH</commands>
       <description>Cooja Mote Type #2</description>
       <source>[CONTIKI_DIR]/examples/hello-world/hello-world.c</source>
       <commands>make TARGET=cooja clean
-make -j hello-world.cooja TARGET=cooja MAKE_MAC=MAKE_MAC_TSCH</commands>
+make -j hello-world.cooja TARGET=cooja MAKE_MAC=MAKE_MAC_TSCH DEFINES=TSCH_CONF_DEFAULT_HOPPING_SEQUENCE=TSCH_HOPPING_SEQUENCE_1_1</commands>
```
simonduq: so, no more coverage of channel hopping in cooja?

os/net/mac/tsch/tsch-conf.h
```diff
@@ -71,14 +71,20 @@
 #define TSCH_DESYNC_THRESHOLD (2 * TSCH_MAX_KEEPALIVE_TIMEOUT)
 #endif
 
-/* Period between two consecutive EBs */
+/* The default period between two consecutive EBs (not taking into account any randomization).
+ * When TSCH_CONF_EB_PERIOD set to 0, sending EBs is disabled completely; the EB process is not started.
```
simonduq: typo: "When TSCH_CONF_EB_PERIOD *is* set"

tests/17-tun-rpl-br/02-border-router-cooja-tsch.csc
atiselsts: As you know if there are multiple channels, the joining time becomes potentially unbounded. In practice, this means that the test timeout should be increased, especially if the 4 channels default is used. IMO it's better to avoid doing it for this test as its a real-time test (speedlimit=100%) and instead leave channel hopping testing for other tests, which can be executed much faster. However, if you want, I can increase the test timeout (from the current 120 seconds) and retain the channel hopping.

simonduq: Do we have tests with channel hopping on speedlimit > 100%?

atiselsts: Sure, `tests/07-simulation-base` has a number of TSCH tests with the default channel hopping sequence and unlimited speed, if that's what you mean.

simonduq: Then it's good with me.

os/net/mac/tsch/tsch-conf.h
atiselsts: fixed

PR #1165, opened 2019-12-19
avoid adding link stats back to neighbors that are dead

### LINKED ISSUE #938: Link stats: unsuccessful probing can improve link stats etx metric
Let's say we have topology:
 
    RPL root <-> A <-> B, C, D, E

Sequence of events:
- the RPL root disappears
- node A continuously increases etx to root and advertises it to downstream nodes
- node A leaves the TSCH network and deletes all link stats (let's say it has `etx_1` to root)
- node A rejoins TSCH and RPL networks through some of the downstream nodes
- node A sends a probing message to root
- node A fails to get reply to the probing message
- node A adds a new link stats with some default etx + noack penalty (let's call it `etx_2`)

Now even though the probing to root did not get any reply, `etx_2` is likely to be lower than `etx_1`. This will cause the downstream nodes to decrease their etx to root as well.

Expected behavior: the RPL etx metrics reach higher and higher numbers due to counting to infinity; all nodes eventually leave the RPL and TSCH networks.

Actual behavior: counting to infinity is slow or does not happen, as the etx keeps resetting.

atiselsts:
This should greatly improve network rebuilding time when the root node is rebooted or replaced.

Fixes #938

### DIFF
--- a/os/net/link-stats.c
+++ b/os/net/link-stats.c
@@ -144,6 +144,11 @@ link_stats_packet_sent(const linkaddr_t *lladdr, int status, int numtx)
 
   stats = nbr_table_get_from_lladdr(link_stats, lladdr);
   if(stats == NULL) {
+    /* If transmission failed, do not add the neighbor, as the neighbor might not exist anymore */
+    if(status != MAC_TX_OK) {
+      return;
+    }
+
     /* Add the neighbor */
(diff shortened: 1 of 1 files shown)


debug-ito:
Although I have some questions in #938, the change in this pull-request makes sense to me.

However, we could have a better comment to explain why we "do not add the neighbor if transmission failed". In my opinion, that's because we are not sure if that neighbor really exists when the transmission failed.

Issue #877, opened 2019-03-14
TSCH timesource loops


AdamShier1:
This problem is related to rpl preferred parent loops but for tsch timesources. 

Say you have a configuration such as:

A<----B<-----D, where node A is the border router, node B sends to node A (ie node A is his preferred parent and his tsch timesource), and node D sends to node A through B (ie node B is his preferred parent and his tsch timesource). 

If A is restarted, B and D will no longer be able to send messages to him, and in the normal scenario would disassociate and rejoin from node A's new eb. If B is restarted soon after A restarts, B **can** join D's tsch network if he receives an eb from D faster than one from A. Now B and D are eachothers timesources, and will keep eachother looping in a tsch network with no border router, forever waiting for a DIO message which will never come. 

So A is left sending eb's but with noone in the network, and B and D are in the 'old' tsch network, with eachother as their timesources, so the keepalives will be sent and received with no perceived problems.

atiselsts:
Would be good to see a Cooja test for this. I'm afraid you're right and the loop will never timeout, since
1) there will be KA packets being sent and ACKed even when B and D do not have any RPL parent;
2) `tsch_rpl_callback_parent_switch` will fail to reset the on D timesource when D resets the RPL parent.

atiselsts:
The TSCH timesource loop can form even if the root is not reset. This was discussed recently on Gitter. When a node gets no ACK to a keepalive, it may select another node as it's timesource (the `resynchronize` function). In this situation the RPL routing parent and the TSCH timesource similarly become two different nodes.

AdamShier1:
This problem is essentially the same as #608, which in many scenarios was fixed by #769 supressing eb's when you leave a network, the problem is that everyone doesn't necessarily leave the network at the same time, allowing people who left first to sync off others still sending eb's because their still thinking their in a RPL network.

One possibility as mentioned is you dont set a timesource until you've picked a preferred parent, so there would need to be some timer that says if you dont have a preferred parent/rpl network in T time (possibly a function of max DIO timer), then disassociate. Ideally, you would do some kind of 'poisoning' before disassociating to reduce the probability that you rejoin that same loop again.

atiselsts:
Thanks for the reference. At least that PR makes sure that the timesource
loops don't stay unresolved forever! That's a very good start.

atiselsts:
Fixed in #1166

PR #1166, opened 2019-12-19
keep TSCH timesource state consistent with RPL parent state

### LINKED ISSUE #608: RPL clique reconnection problem

### LINKED ISSUE #890: TSCH timesource selection issues
This is relevant to the #877. Infinite loops may show up in the TSCH timesource tree. Instead of completely hijacking Adam's issue with multiple related problems, here is a new one.

First, a minor comment: the condition in the https://github.com/contiki-ng/contiki-ng/blob/develop/os/net/mac/tsch/tsch.c#L367 seems wrong? It will update the `last_eb` with the current Eb's source address if and only if there is no timesource, or the current `last_eb` address is not the same as the timesource address. As you can see, this will very soon converge to the `last_eb` address being equal to the timesource's address. I'm either misunderstanding something or guessing the if statement is meant to check the frame's source address instead of the `last_eb` address.

Fortunately, the current behavior minimizes the chance that an alternative timesource is selected and so reduces the probability the loop shows up.

Setup to replicate the timesource loop:
Network topology: 1 is the coordinator, 2 and 3 are peer nodes are in direct reachability of 1.
Routing is RPL lite, default settings (`examples/6tisch/simple-node/project-conf.h`).

With the `develop` branch:
* 0:00 - start all nodes
* around 1:00: remove root
* 1:15-1:30 - the node 2 switches to 3 as its timesource and 3 switches to 2 as its timesource
* 1:50 - RPL switches parent to NULL

With the `develop` branch, and after changing the line 367 in `tsch.c`:

    if(ts == NULL || !linkaddr_cmp(&last_eb_nbr_addr, &ts->addr)) {

to this:

    if(ts == NULL || !linkaddr_cmp((linkaddr_t *)&frame.src_addr, &ts->addr)) {

* 0:00 - start all nodes
* around 1:00: remove root
* 1:06 - the node 2 switches to 3 as its timesource and 3 switches to 2 as its timesource
* 12:04 - RPL switches parent from 1 to the peer node (2 or 3)
* 12:04-12:33 - RPL continues switching between no parent (NULL) to the peer node as a parent
* 12:33 - RPL switches parent to NULL

In both cases the TSCH timesource never becomes NULL. This will cause "islands" to form in the network and drift away from the coordinator's clock. Since they never leave TSCH network, they will never get to resync again (unless accidentally drifting back in the TSCH timing and hearing some packets).

My recommendation is to remove the alternative TSCH timesource selection, and in general keep the tsch timesources consistent with RPL parent (e.g. leave the TSCH network when RPL parent becomes NULL).

atiselsts:
Fixes #890 and #877

Potentially fixes #608

### DIFF
--- a/os/net/mac/tsch/tsch-rpl.c
+++ b/os/net/mac/tsch/tsch-rpl.c
@@ -128,9 +128,8 @@ tsch_rpl_callback_new_dio_interval(clock_time_t dio_interval)
 void
 tsch_rpl_callback_parent_switch(rpl_parent_t *old, rpl_parent_t *new)
 {
-  /* Map the TSCH time source on the RPL preferred parent (but stick to the
-   * current time source if there is no preferred aarent) */
-  if(tsch_is_associated == 1 && new != NULL) {
+  /* Map the TSCH time source on the RPL preferred parent */
+  if(tsch_is_associated == 1) {
     tsch_queue_update_time_source(
       (const linkaddr_t *)uip_ds6_nbr_lladdr_from_ipaddr(
         rpl_parent_get_ipaddr(new)));


debug-ito:
Hmm, I thought about the implication of this change, but in the end, I think it's ok.

- When a node joins a TSCH/RPL network, first it associates with the TSCH PAN (sets the time source) and then it determines the RPL preferred parent. This means there has to be a time period when the node has a TSCH time source but no RPL preferred parent. This change allows that time period, I think, because `RPL_CALLBACK_PARENT_SWITCH` is called only when `old != new`. In the joining phase, `old == new == NULL`, I think.
- What will happen after calling `tsch_queue_update_time_source(NULL)` in `RPL_CALLBACK_PARENT_SWITCH`? I think the node cannot update the `last_sync_asn` in `tsch-slot-operation.c` anymore, and call `tsch_disassociate()` eventually. We don't need to call `tsch_disassociate()` in `RPL_CALLBACK_PARENT_SWITCH` (but maybe we should??)

Well, maybe we should run some simulation experiments...

debug-ito:
Although it's ok for me, the change in the behavior of TSCH layer is significant. Maybe we should have a way to roll it back.

For example, we can add a new configuration macro that switches between the old behavior (i.e. change time source to non-NULL RPL parent only) and the new behavior. The default behavior can be the new one, I think.

atiselsts:
I'm against introducing any more configuration constants for features like this. In the long run, they make the OS more difficult to use and understand.
So, I'm afraid I disagree with you here.

debug-ito:
OK. Please make sure you advertise this change in the ChangeLog of the next version.

PR #1169, opened 2020-01-03
Revert TSCH config changes for nightly tests

simonduq:
The original commit was introduced to only enable TSCH multi-choffset explicitly once switching to `TSCH_CONFIG_OPTIMS3`. But the commit did enable multi-choffset for `CONFIG_TSCH`, by mistake.

This PR reverts the commit, so that we're back to the following:
* `CONFIG_TSCH`: all defaults
* `CONFIG_TSCH_OPTIMS`: all defaults + basic optims
* `CONFIG_TSCH_OPTIMS2`: all defaults + more optims (not run in testbed so far, but will enable this soon)

All will now include multi-choffset.

Note: it does look like multi-choffset enhanced the performance for `CONFIG_TSCH`. See results in https://contiki-ng.github.io/setups/test-tsch.html, and compare runs prior to the merge on October 30, 2019, to the runs after this date. I only scrolled, did not run stats, but would say PDR is higher.


atiselsts:
Hi Simon,
The results have indeed improved slightly. I checked the results, old ones going back to the start of August. After excluding one outlier data point on 7th October:
PDR: old=99.83 new=99.89

I ran ANOVA test and the difference between the PDR restuls is (barely) statistically significant, with p value 0.04

But I still don't how did the PR enable multiple channel offsets; when the condition `#if CONFIG_OPTIMS < 3` is true, just one channel offset is used. Are you absolutely sure multiple channel offsets were used? The difference in PDR could be explained by other factors, you know.

simonduq:
Great :)

I do think we have multi-choffset enabled, because the test for `#if CONFIG_OPTIMS < 3` is nested inside `#if CONFIG_OPTIMS >= 1`, and in the `CONFIG_TSCH` makefile flag leaves `CONFIG_OPTIMS` unset.

simonduq:
Also note that PDR diffs always less important than they actually are when closing 100%.
If you look at the end-to-end loss rate this is 0.17% vs 0.11%.

simonduq:
BTW new "optim2" setup is doing pretty good so far!
https://contiki-ng.github.io/setups/test-tsch-optims2.html

This is without multi-choffset yet (will be after this PR is merged) but with fine-tuned RPL and TSCH traffic, as well as more frequent link probing.

PR #1170, opened 2020-01-04
Fix AES-CCM* to work with a_len up to 0xfeff and m_len up to 0xffff

simonduq:
This PR fixes the AES-CCM* library for payloads/headers beyond 240 bytes (current implem is for up to 255 but ends in infinite loop for `m_len` greater than 240).

The new limit for `a_len` is 0xfeff and for `m_len` 0xffff.
This is useful for e.g. use of AES-CCM in upper layers, or for 802.15.4g with larger frames.
Also adds relevant unit tests.

This is contributed by Yanzi Networks AB (www.yanzi.se).


os/lib/ccm-star.c
```diff
@@ -42,64 +42,68 @@
...
     uint8_t *result,
     uint8_t mic_len)
 {
   uint8_t x[AES_128_BLOCK_SIZE];
-  uint8_t pos;
+  uint32_t pos; /* 32-bits as can need to exceed m_len to reach end of loop */
   uint8_t i;
-  
-  set_iv(x, CCM_STAR_AUTH_FLAGS(a_len, mic_len), nonce, m_len);
+
```
yatch: Would it be nicer to check if the given `mic_len` is valid?
```
[Section 2, RFC 3610]

   For the generic CCM mode there are two parameter choices.  The first
   choice is M, the size of the authentication field.  .... (snip) ...
   Valid values are 4, 6, 8, 10, 12, 14, and 16 octets.  The second
```

os/lib/hexconv.h
```diff
@@ -0,0 +1,50 @@
...
+ *         Niclas Finne <nfi@sics.se>
+ */
+
+#ifndef HEXCONV_H_
+#define HEXCONV_H_
+
+#include <stdint.h>
+
+int hexconv_hexlify(const uint8_t *data, int data_len,
+		    char *text, int text_size);
```
yatch: Indentation

tests/08-native-runs/07-aes-ccm/Readme.md
```diff
@@ -0,0 +1,2 @@
+To generate test vectors, run:
+./generate-test-vectors.py > test-vectors.c
```
yatch: Add
```
pip install pycryptodome
```
or something to resolve the dependency.

tests/08-native-runs/07-aes-ccm/test-aesccm.c
```diff
@@ -0,0 +1,184 @@
...
+#include "contiki.h"
+#include "lib/random.h"
+#include "unit-test.h"
+#include "lib/ccm-star.h"
+#include "lib/hexconv.h"
+#include <string.h>
+#include <stdio.h>
+#include <stdbool.h>
+
+#define MICLEN 8
```
yatch: Can we have tests for `MICLEN == 4` and `MICLEN == `16`, which can be used in IEEE 802.15.4 as well (Table 9-4, IEEE 802.15.4-2015)

tests/utils.sh
```diff
@@ -21,3 +76,99 @@ function kill_bg( )
...
+  printf "* %-60s        " "$NAME ..." | tee -a $BASENAME.testlog;
+}
+
+function report_end_failure( )
+{
+  TIMEOUT=$1
+  if [ -z $TIMEOUT ]; then
+    printf "FAIL (%u seconds)\n" $SECONDS | tee -a $BASENAME.testlog;
+  else
+    printf "FAIL (%u seconds timeout)\n" $TIMEOUT | tee -a $BASENAME.testlog;
```
yatch: This works perfectly... I confirmed with the commit of 84e07a9. This is great!
```
-- Starting test 07-aes-ccm
* clean ...                                                           OK   (6 seconds)
* compile ...                                                         OK   (20 seconds)
* start ./test-aesccm.native ...                                      OK   (0/30 seconds)
* run ./test-aesccm.native ...                                        FAIL (120 seconds timeout)
* run ./test-aesccm.native ...                                        SKIP
-- Killing background jobs
kill -9 1827
../utils.sh: line 41:  1827 Killed                  $TEST &> $RUNLOG
```

os/lib/ccm-star.c
```diff
@@ -42,64 +42,68 @@
 #include "lib/aes-128.h"
 #include <string.h>
 
-/* see RFC 3610 */
-#define CCM_STAR_AUTH_FLAGS(Adata, M) ((Adata ? (1u << 6) : 0) | (((M - 2u) >> 1) << 3) | 1u)
+/* As per RFC 3610. M == 2 (m_len is two bytes long) */
```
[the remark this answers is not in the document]
simonduq: > Here, `M == 2` should be `L == 2`??
Good point, fixing the comment

> And, `M` is always `2` in this implementation, which I'd mention here as well.
Hmm I don't think so: M is the MIC len, right?

os/lib/ccm-star.c
simonduq: agree

tests/08-native-runs/07-aes-ccm/Readme.md
simonduq: Did that. Also, switched to python3

os/lib/hexconv.h
simonduq: fixed

os/lib/ccm-star.c
simonduq: done

tests/utils.sh
simonduq: Oh yes we've been using that one intensively :D

simonduq: Though just spotted a buggy log output: the SKIP line should have been for "check" not for "run". Committed a fix for that

os/lib/ccm-star.c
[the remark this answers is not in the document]
yatch: I guess I was confused. Ignore the second point. Sorry for that (>_<)

tests/08-native-runs/07-aes-ccm/Readme.md
yatch: 👍 I confirmed the code actually was able to run with Python 2 and Python 3 :-)

tests/utils.sh
yatch: :+1

yatch: 👍

simonduq:
Thanks. Just pushed another minor thing.

PR #1171, opened 2020-01-07
add option to disable router solicitation

tdesmet:
right now when rpl is used router advertisements are disabled, but router solicitation is enabled. Giving the weird behavior that leaf nodes send out router solicitations but never receive a response (router advertisement). So when using rpl I suggest that the sending of router solicitations is disabled.

### DIFF
--- a/os/contiki-default-conf.h
+++ b/os/contiki-default-conf.h
@@ -130,6 +130,16 @@
 #endif /* UIP_CONF_TCP */
 #endif /* UIP_CONF_TCP_CONNS */
 
+/* UIP_CONF_ND6_SEND_RS enables standard IPv6 Router Solicitating.
+ * We enable it by default when IPv6 is used without RPL. */
+#ifndef UIP_CONF_ND6_SEND_RS
+#if (NETSTACK_CONF_WITH_IPV6 && !UIP_CONF_IPV6_RPL)
+#define UIP_CONF_ND6_SEND_RS 1
+#else /* NETSTACK_CONF_WITH_IPV6 && !UIP_CONF_IPV6_RPL */
+#define UIP_CONF_ND6_SEND_RS 0
+#endif /* NETSTACK_CONF_WITH_IPV6 && !UIP_CONF_IPV6_RPL */
+#endif /* UIP_CONF_ND6_SEND_RS */
+
 /* UIP_CONF_ND6_SEND_RA enables standard IPv6 Router Advertisement.
  * We enable it by default when IPv6 is used without RPL. */
 #ifndef UIP_CONF_ND6_SEND_RA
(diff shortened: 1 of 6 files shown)


yatch:
> right now when rpl is used router advertisements are disabled, but router solicitation is enabled.

I think this doesn't reflect well what is implemented. `UIP_CONF_ROUTER` and `UIP_CONF_IPV6_RPL` are separate settings, although there is a dependency found in `contiki-default-conf.h`. Here is how I read the code:

* by default, you're a router https://github.com/contiki-ng/contiki-ng/blob/346d4dec7eca625b062c5a4dbc49dfe469535f6f/os/contiki-default-conf.h#L91
  * if RPL is enabled, don't send RAs to advertise yourself (instead, send DIOs) https://github.com/contiki-ng/contiki-ng/blob/346d4dec7eca625b062c5a4dbc49dfe469535f6f/os/contiki-default-conf.h#L139
  * otherwise, send RAs https://github.com/contiki-ng/contiki-ng/blob/346d4dec7eca625b062c5a4dbc49dfe469535f6f/os/contiki-default-conf.h#L137
* if you're an IPv6 host (`UIP_CONF_ROUTER`==0), send RSes to discover routers always https://github.com/contiki-ng/contiki-ng/blob/346d4dec7eca625b062c5a4dbc49dfe469535f6f/os/net/ipv6/uip-ds6.c#L730

What we need would be, to suppress RSes when you're an IPv6 host without having RPL enabled (that is, `RPL_LEAF_ONLY==1`)?

> Giving the weird behavior that leaf nodes send out router solicitations but never receive a response (router advertisement).

If the node is a normal IPv6 router,  this seems a sensible behavior. RPL routers around the node may need to respond to RSes by the node as IPv6 routers in this particular case.

> So when using rpl I suggest that the sending of router solicitations is disabled.

Then, why don't we use `UIP_CONF_IPV6_RPL` directly to decide whether to send RSes or not, by the way? The code pointed at https://github.com/contiki-ng/contiki-ng/pull/1171/files#diff-63fb182faff93d943763f6defdbc7355R404 could be like this:

```
#if !UIP_CONF_ROUTER && !UIP_CONF_IPV6_RPL
    if(data == &uip_ds6_timer_rs &&
        etimer_expired(&uip_ds6_timer_rs)) {
      uip_ds6_send_rs();
```

simonduq:
Any update @tdesmet ?

tdesmet:
Sorry no

Have been working on other stuff and the changes in this pr where enough
for our project.

I also have no idea on what is the best solution, my knowledge of all this
is limited so I am not the best person to give input on a better solution.

Op wo 15 apr. 2020 om 13:54 schreef Simon Duquennoy <
notifications@github.com>

yatch:
@tdesmet Thanks for your response. I'm going to close this issue. If you come up with another idea or solution, open a new one ;-)

PR #1173, opened 2020-01-09
Fix out-of-bounds uIP data pointer on some TCP URG segments

TheJokr:
This is a bug/vulnerability I discovered while working on a [paper on TCP specification conformance](https://arxiv.org/abs/2002.05400). One of the implementations we tested was uIP (standalone v1.0, so quite old), which generated segmentation faults when we ran a test regarding TCP's urgent pointer. We determined the cause within uIP (see below), but didn't submit a fix for it initially because we thought the project was inactive (judging from its [GitHub repo](https://github.com/adamdunkels/uip)).

If the UIP_URGDATA macro is zero (the default), the amount of urgent data is deducted from the uip_appdata pointer and from uip_len without any size check. This can cause out-of-bounds reads (for the former) and unsigned integer underflow (for the latter) when the urgent pointer points past the current segment.

Now that we discovered its use within contiki(-ng), the (quite simple) fix in this PR is to cap the effective urgent data length at urg_len independent of UIP_URGDATA's value. The already-present tmp16 variable is used to hold the length in the absence of uip_urglen on builds with UIP_URGDAT == 0.

### DIFF
--- a/os/net/ipv6/uip6.c
+++ b/os/net/ipv6/uip6.c
@@ -2039,12 +2039,15 @@ uip_process(uint8_t flag)
     /* Check the URG flag. If this is set, the segment carries urgent
          data that we must pass to the application. */
     if((UIP_TCP_BUF->flags & TCP_URG) != 0) {
-#if UIP_URGDATA > 0
-      uip_urglen = (UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1];
-      if(uip_urglen > uip_len) {
-        /* There is more urgent data in the next segment to come. */
-        uip_urglen = uip_len;
+      tmp16 = (UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1];
+      if(tmp16 > uip_len) {
+        /* There is more urgent data in the next segment to come. 
+	     Cap the urgent data length at the segment length for
+	     further processing. */
+        tmp16 = uip_len;
       }
+#if UIP_URGDATA > 0
+      uip_urglen = tmp16;
       uip_add_rcv_nxt(uip_urglen);
       uip_len -= uip_urglen;
       uip_urgdata = uip_appdata;
@@ -2052,8 +2055,9 @@ uip_process(uint8_t flag)
     } else {
       uip_urglen = 0;
 #else /* UIP_URGDATA > 0 */
-      uip_appdata = ((char *)uip_appdata) + ((UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1]);
-      uip_len -= (UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1];
+      /* Ignore and discard any urgent data in this segment. */
+      uip_appdata = ((char *)uip_appdata) + tmp16;
+      uip_len -= tmp16;
 #endif /* UIP_URGDATA > 0 */
     }
 


yatch:
@TheJokr Great 👍 I'm going to review the current TCP urgent mechanism implementation as well.

> This is a bug/vulnerability I discovered while working on a paper on TCP specification conformance. 

Is the paper publicly available? I'm interested.

TheJokr:
> Is the paper publicly available? I'm interested.

Awesome! It's not quite finished yet, but it got accepted at [PAM 2020](https://pam2020.cs.uoregon.edu/Accepted_papers.html). I'll leave a comment here once the preprint is available on arXiv.

TheJokr:
Hey, any progress on this @yatch? The preprint is available now: https://arxiv.org/abs/2002.05400

yatch:
Thanks for the kind reminder 😅😅😅

TheJokr:
The tool we developed for testing the specification conformance is open source now: [TheJokr/tcpreq](https://github.com/TheJokr/tcpreq). If you need instructions for our uIP test setup, please notify me.

yatch:
Thank you... I know I have to have some time on this. Hopefully soon.

os/net/ipv6/uip6.c
```diff
@@ -2039,21 +2039,22 @@ uip_process(uint8_t flag)
     /* Check the URG flag. If this is set, the segment carries urgent
          data that we must pass to the application. */
     if((UIP_TCP_BUF->flags & TCP_URG) != 0) {
-#if UIP_URGDATA > 0
-      uip_urglen = (UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1];
-      if(uip_urglen > uip_len) {
+      tmp16 = (UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1];
+      if(tmp16 > uip_len) {
         /* There is more urgent data in the next segment to come. */
```
yatch, line 2044: Could you put a comment saying that we "cap the effective urgent data length at `urg_len`" or something ?  What I read from the code including the original one is that, urgent data continued in following segments is just discarded, although this is the case in the original code only with `UIP_URGDATA > 0` as you found.

os/net/ipv6/uip6.c
```diff
@@ -2039,21 +2039,22 @@ uip_process(uint8_t flag)
     /* Check the URG flag. If this is set, the segment carries urgent
          data that we must pass to the application. */
     if((UIP_TCP_BUF->flags & TCP_URG) != 0) {
-#if UIP_URGDATA > 0
-      uip_urglen = (UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1];
-      if(uip_urglen > uip_len) {
+      tmp16 = (UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1];
```
yatch, line 2042: Using `tmp16` here should be fine. 👍

os/net/ipv6/uip6.c
```diff
@@ -2039,21 +2039,22 @@ uip_process(uint8_t flag)
...
       uip_add_rcv_nxt(uip_urglen);
       uip_len -= uip_urglen;
       uip_urgdata = uip_appdata;
       uip_appdata += uip_urglen;
     } else {
       uip_urglen = 0;
 #else /* UIP_URGDATA > 0 */
-      uip_appdata = ((char *)uip_appdata) + ((UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1]);
-      uip_len -= (UIP_TCP_BUF->urgp[0] << 8) | UIP_TCP_BUF->urgp[1];
+      uip_appdata = ((char *)uip_appdata) + tmp16;
```
yatch, line 2056: It would be nice to say in a comment that we ignore urgent data in this segment.

os/net/ipv6/uip6.c
TheJokr: Done. You are right, the problem was that the old code discarded more data than there was in the segment, so the pointer pointed "before" the received segment in memory. I only noticed because uIP then segfaulted on Linux.

os/net/ipv6/uip6.c
TheJokr: Done.

yatch:
@TheJokr  Thank you, again, for the great contribution! Thanks to this PR, I got to know RFC 6093, and now I understand TCP URG better than before ;-)

PR #1194, opened 2020-02-19
coap: fix initialization of random MID

LDVG:
Depending on the arch/platform, the standard library `rand()` may or may not
have been seeded (for instance, `arch/cpu/simplelink-cc13xx-cc26xx`
contains its own PRNG implementation and therefore never calls `srand()`
during initialization[1]). The proper way to generate a random number is
therefore to call the contiki-ng library function `random_rand()` and
not `rand()`.

[1] See `random_init()` and `random_rand()` in [arch/cpu/simplelink-cc13xx-cc26xx/dev/random.c](https://github.com/contiki-ng/contiki-ng/blob/5f1088dd1fd5417960be50f8d792df9d93ed6ba1/arch/cpu/simplelink-cc13xx-cc26xx/dev/random.c) compared to [os/lib/random.c]( https://github.com/contiki-ng/contiki-ng/blob/5f1088dd1fd5417960be50f8d792df9d93ed6ba1/os/lib/random.c).


LDVG:
It appears `lib/random.{c,h}` needs to be added to `CORE_FILES` in `Makefile.contiki` in the `tests/18-coap-lwm2m/example-lwm2m-standalone` submodule -- and the (unused) inclusion of `sys/clock.h` in `os/lib/random.c` must be removed before the tests will build. I'm not sure how to synchronize that with this PR considering that it's a submodule.

atiselsts:
Can you raise a PR in https://github.com/contiki-ng/example-lwm2m-standalone ?

LDVG:
> Can you raise a PR in https://github.com/contiki-ng/example-lwm2m-standalone ?

I have pushed the necessary changes to my PR (see linked PR) and the `example-lwm2m-standalone` repository. Not sure if you'd like me to update the submodule in the tests folder, since the SHA might change when the PR is merged?

joakimeriksson:
This will need to be merged before getting it green I assume? Even when merging lwm2m-standalone now - it will not be "green" until this is "in" so we need to disregard the "redness" in that specific test-case.

LDVG:
I updated the `example-lwm2m-standalone` submodule after it was merged, tests are running now. :slightly_smiling_face:

joakimeriksson:
> I updated the `example-lwm2m-standalone` submodule after my PR in that repository was merged, tests are running now. 🙂

Many thanks - then we should be able to merge when this is "green".

LDVG:
> Many thanks - then we should be able to merge when this is "green".

Indeed, it looks like the tests passed just fine now. :slightly_smiling_face:

PR #1197, opened 2020-02-24
MQTT: increase size of variable i to avoid potential infinite loop

simonduq:
If compiled without any optimization, this debug loop can result in an infinite loop.

Special thanks to H2020 Vessedia (https://vessedia.eu/) the vulnerability report.


alexandruioanp:
Could you share a link to the full report?

simonduq:
The report will be made public in the first week of March.

simonduq:
Good idea @joakimeriksson , added conditional compilation

PR #1200, opened 2020-03-05
DIO after the rank changed

matinpf:
I think maybe better to a node after changed own rank that awareness other nodes by sending multicast DIO message. If you are considering this modification arises unnecessary DIO traffic, we can add a threshold to prevent this. For example, if change been more than (minhop-1) thus, the DIO will send.

*in the current implementation, some time node rank change but other nodes don't get awareness about this change for several minutes


matinpf:
How can I request a recheck?!

simonduq:
The feature per se is welcome and note it's already implemented in RPL-Lite with `RPL_SIGNIFICANT_CHANGE_THRESHOLD`. If we want it in Classic I'll encourage using the same define and an implem as similar as possible as Lite.

The code style and overall quality (global var should be static, what's `-1*` etc.) is way off. See 
https://github.com/contiki-ng/contiki-ng/wiki/Code-style for code style.

simonduq:
Closing.

PR #1205, opened 2020-03-09
Link based Orchestra rule

atiselsts:
The Orchestra link-based rule has been designed based on the insights from:
1) An Empirical Survey of Autonomous Scheduling Methods for TSCH, Atis Elsts, Seohyang Kim, Hyung-Sin Kim, Chongkwon Kim, IEEE Access, 2020.
2) ALICE: autonomous link-based cell scheduling for TSCH, Seohyang Kim, Hyung-Sin Kim, Chongkwon Kim, IEEE Access, ACM/IEEE IPSN, 2019.

It potentially offers higher performance (better energy/reliability tradeoff) compared to the other Orchestra rules, especially for applications where both upstream and downstream traffic is frequent.

### DIFF
--- a/examples/6tisch/simple-node/Makefile
+++ b/examples/6tisch/simple-node/Makefile
@@ -22,24 +22,26 @@ include $(CONTIKI)/Makefile.dir-variables
 MODULES += $(CONTIKI_NG_SERVICES_DIR)/shell
 
 ifeq ($(MAKE_WITH_ORCHESTRA),1)
-MODULES += $(CONTIKI_NG_SERVICES_DIR)/orchestra
+  MODULES += $(CONTIKI_NG_SERVICES_DIR)/orchestra
 
-ifeq ($(MAKE_WITH_LINK_BASED_ORCHESTRA),1)
-ifeq ($(MAKE_WITH_STORING_ROUTING),1)
-CFLAGS += -DORCHESTRA_CONF_RULES="{&eb_per_time_source,&unicast_per_neighbor_link_based,&default_common}"
-else
-$(error "Inconsistent configuration")
-endif # MAKE_WITH_LINK_BASED_ORCHESTRA
-endif # MAKE_WITH_STORING_ROUTING
-
-endif # MAKE_WITH_ORCHESTRA
-
-ifeq ($(MAKE_WITH_STORING_ROUTING),1)
-MAKE_ROUTING = MAKE_ROUTING_RPL_CLASSIC
-CFLAGS += -DRPL_CONF_WITH_STORING=1
-ifneq ($(MAKE_WITH_LINK_BASED_ORCHESTRA),1)
-CFLAGS += -DORCHESTRA_CONF_RULES="{&eb_per_time_source,&unicast_per_neighbor_rpl_storing,&default_common}"
+  ifeq ($(MAKE_WITH_STORING_ROUTING),1)
+    ifeq ($(MAKE_WITH_LINK_BASED_ORCHESTRA),1)
+      # enable the `link_based` rule
+      CFLAGS += -DORCHESTRA_CONF_RULES="{&eb_per_time_source,&unicast_per_neighbor_link_based,&default_common}"
+    else
+      # enable the `rpl_storing` rule
+      CFLAGS += -DORCHESTRA_CONF_RULES="{&eb_per_time_source,&unicast_per_neighbor_rpl_storing,&default_common}"
+    endif
+  else
+    ifeq ($(MAKE_WITH_LINK_BASED_ORCHESTRA),1)
+      $(error "Inconsistent configuration")
+    endif
+  endif
 endif
+
+ifeq ($(MAKE_WITH_STORING_ROUTING),1)
+  MAKE_ROUTING = MAKE_ROUTING_RPL_CLASSIC
+  CFLAGS += -DRPL_CONF_MOP=RPL_MOP_STORING_NO_MULTICAST
 endif
 
 ifeq ($(MAKE_WITH_SECURITY),1)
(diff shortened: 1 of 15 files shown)


examples/6tisch/simple-node/Makefile
```diff
@@ -19,6 +23,23 @@ MODULES += $(CONTIKI_NG_SERVICES_DIR)/shell
...
+else
+$(error "Inconsistent configuration")
+endif # MAKE_WITH_LINK_BASED_ORCHESTRA
+endif # MAKE_WITH_STORING_ROUTING
+
+endif # MAKE_WITH_ORCHESTRA
+
+ifeq ($(MAKE_WITH_STORING_ROUTING),1)
+MAKE_ROUTING = MAKE_ROUTING_RPL_CLASSIC
+CFLAGS += -DRPL_CONF_WITH_STORING=1
```
simonduq, line 39: This might be a mis-understanding: this flag only *includes* storing, does not have the root *enable* it.
That being said, it is enabled by default in Classic.
`RPL_CONF_MOP` is what tells to root whether to build a storing or non-storing network.

examples/6tisch/simple-node/Makefile
```diff
@@ -19,6 +23,23 @@ MODULES += $(CONTIKI_NG_SERVICES_DIR)/shell
...
+ifneq ($(MAKE_WITH_LINK_BASED_ORCHESTRA),1)
+CFLAGS += -DORCHESTRA_CONF_RULES="{&eb_per_time_source,&unicast_per_neighbor_rpl_storing,&default_common}"
+endif
```
simonduq, line 42: Can't the conditions above be re-factored for readability? The first checks link-based then storing, the second is outside of block `ifeq ($(MAKE_WITH_ORCHESTRA),1)` yet it configures Orchestra, and checks storing then link-based. Looks like as bug even, or am I missing something?

os/net/mac/tsch/tsch-schedule.c
```diff
@@ -226,9 +226,11 @@ tsch_schedule_add_link(struct tsch_slotframe *slotframe,
...
     }
 
-    /* Start with removing the link currently installed at this timeslot (needed
-     * to keep neighbor state in sync with link options etc.) */
-    tsch_schedule_remove_link_by_timeslot(slotframe, timeslot, channel_offset);
+    if(do_remove) {
+      /* Start with removing the link currently installed at this timeslot (needed
+       * to keep neighbor state in sync with link options etc.) */
+      tsch_schedule_remove_link_by_timeslot(slotframe, timeslot, channel_offset);
+    }
```
simonduq, line 233: Does having multiple slots at the same coordinates have any implications in the scheduler? I feel like not, as it's not much different to handle as having multiple choffsets at the same timeslot. Just something to think about.

os/net/mac/tsch/tsch.c
```diff
@@ -686,7 +686,8 @@ tsch_associate(const struct input_packet *input_eb, rtimer_clock_t timestamp)
         tsch_schedule_add_link(sf,
             ies.ie_tsch_slotframe_and_link.links[i].link_options,
             LINK_TYPE_ADVERTISING, &tsch_broadcast_address,
-            ies.ie_tsch_slotframe_and_link.links[i].timeslot, ies.ie_tsch_slotframe_and_link.links[i].channel_offset);
+            ies.ie_tsch_slotframe_and_link.links[i].timeslot,
+            ies.ie_tsch_slotframe_and_link.links[i].channel_offset, 0);
```
simonduq, line 690: Is this change required? (If not, I think it should be reverted. But I can live with it if link-based requires it).

examples/6tisch/simple-node/Makefile
atiselsts: Not sure if there is an action required?

examples/6tisch/simple-node/Makefile
atiselsts: Refactored

os/net/mac/tsch/tsch.c
atiselsts: No, definitely not required. Was not sure what's the intention here, but if removing conflicting links is what we need, I'll revert

examples/6tisch/simple-node/Makefile
simonduq: I would say configure `RPL_CONF_MOP` instead

os/net/mac/tsch/tsch-schedule.c
atiselsts: Ok, there is actually an issue. 

The active cell selection will be working fine, at least not worse than previously. The PR #1162 solves the case when there are multiple Tx links at the same timeslot. In general, we still don't have a good solution for multiple Rx links - just the first one always is used. Round robin would be better, but adds some complexity.

But the real issue is in the unlikely but possible case when the Orchestra code has added Tx and Rx links at the same (timeslot, choffset) and wants to remove one of them. The wrong link may be removed, since we don't look at link options when removing links. Not sure what's the best way to fix this: adding a function that searches for links including their options would be one way.

atiselsts: I modified `tsch_schedule_remove_link_by_timeslot` to remove all matching links.

This will have no direct effect on the new rule, since it does no call the function `tsch_schedule_remove_link_by_timeslot` after the changes.

Instead, there is a new static function `remove_unicast_link` in the file `orchestra-rule-unicast-link-based.c` that also looks at link's options to decide whether that link should be removed. If it is, the code calls `tsch_schedule_remove_link` with the link to remove as an argument.

PR #1222, opened 2020-04-09
Cooja mote improvements

atiselsts:
* Add radio energest level tracking to Cooja motes
* Fix random init for cooja motes so that each mote gets a different random seed

### DIFF
--- a/arch/platform/cooja/dev/cooja-radio.c
+++ b/arch/platform/cooja/dev/cooja-radio.c
@@ -38,6 +38,7 @@
 
 #include "net/packetbuf.h"
 #include "net/netstack.h"
+#include "sys/energest.h"
 
 #include "dev/radio.h"
 #include "dev/cooja-radio.h"
@@ -139,6 +140,7 @@ radio_LQI(void)
 static int
 radio_on(void)
 {
+  ENERGEST_ON(ENERGEST_TYPE_LISTEN);
   simRadioHWOn = 1;
   return 1;
 }
@@ -146,6 +148,7 @@ radio_on(void)
 static int
 radio_off(void)
 {
+  ENERGEST_OFF(ENERGEST_TYPE_LISTEN);
   simRadioHWOn = 0;
   return 1;
 }
@@ -207,22 +210,9 @@ channel_clear(void)
 static int
 radio_send(const void *payload, unsigned short payload_len)
 {
-  int radiostate = simRadioHWOn;
+  int result;
+  int radio_was_on = simRadioHWOn;
 
-  /* Simulate turnaround time of 2ms for packets, 1ms for acks*/
-#if COOJA_SIMULATE_TURNAROUND
-  simProcessRunValue = 1;
-  cooja_mt_yield();
-  if(payload_len > 3) {
-    simProcessRunValue = 1;
-    cooja_mt_yield();
-  }
-#endif /* COOJA_SIMULATE_TURNAROUND */
-
-  if(!simRadioHWOn) {
-    /* Turn on radio temporarily */
-    simRadioHWOn = 1;
-  }
   if(payload_len > COOJA_RADIO_BUFSIZE) {
     return RADIO_TX_ERR;
   }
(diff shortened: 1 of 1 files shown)


arurke:
On the random init change:

Are you sure this is necessary? When I start a sim with all default values I get a different value of `simRandomSeed` at every mote, e.g. mote 1 gets 123457, mote 2 get 123458, etc.
In fact, when ORing with the mote-id it could result in the same seed, e.g. if I do "Reload simulation with new random seed" in the GUI it increments the "base-seed". This results in mote 1 getting seed 123458 which after OR is `123459`, mote 2 would get 123459 which after OR is also `123459`.

Perhaps adding the mote-id would be more robust, but again, it does not seem necessary.

-----

For what it's worth I did a sanity check of the Energest values and they seem fine.

atiselsts:
Looking at the Cooja code, you're right!

I did this change because someone complained there being no way how to initialize the RNG differently on each mote. Apparently they were mistaken.

atiselsts:
The second commit has been removed and now the PR is ready for review.

arch/platform/cooja/dev/cooja-radio.c
```diff
@@ -233,24 +223,47 @@ radio_send(const void *payload, unsigned short payload_len)
...
+  simProcessRunValue = 1;
+  cooja_mt_yield();
+  if(payload_len > 3) {
+    simProcessRunValue = 1;
     cooja_mt_yield();
   }
+#endif /* COOJA_SIMULATE_TURNAROUND */
+
+  /* Transmit on CCA */
+  if(COOJA_TRANSMIT_ON_CCA && send_on_cca && !channel_clear()) {
```
yatch, line 244: This condition should be `if(send_on_cca && channel_clear()) {`, shouldn't it?

arch/platform/cooja/dev/cooja-radio.c
```diff
@@ -233,24 +223,47 @@ radio_send(const void *payload, unsigned short payload_len)
...
-  /* Transmit on CCA */
-#if COOJA_TRANSMIT_ON_CCA
-  if(send_on_cca && !channel_clear()) {
-    return RADIO_TX_COLLISION;
+  if(radio_was_on) {
+    ENERGEST_SWITCH(ENERGEST_TYPE_LISTEN, ENERGEST_TYPE_TRANSMIT);
+  } else {
+    /* Turn on radio temporarily */
+    simRadioHWOn = 1;
+    ENERGEST_ON(ENERGEST_TYPE_TRANSMIT);
```
yatch, line 231: I think, at this point, the radio is in the listening state. Then, it will perform RX-TX turnaround.

arch/platform/cooja/dev/cooja-radio.c
```diff
@@ -233,24 +223,47 @@ radio_send(const void *payload, unsigned short payload_len)
...
   }
-#endif /* COOJA_TRANSMIT_ON_CCA */
 
-  /* Copy packet data to temporary storage */
-  memcpy(simOutDataBuffer, payload, payload_len);
-  simOutSize = payload_len;
-
-  /* Transmit */
-  while(simOutSize > 0) {
+#if COOJA_SIMULATE_TURNAROUND
```
yatch, line 234: I know this block is placed at the beginning of this function in the original code. But, now, it seems this block should be under the following `else` branch (L247). `ENERGEST_SWITCH(ENERGEST_TYPE_LISTEN, ENERGEST_TYPE_TRANSMIT)` will follow this block.

atiselsts: I see your point that the Cooja radio operation energy consumption simulation should be improved. At least I think I do have an idea how to make it more realistic, I'll give it a shot.

arch/platform/cooja/dev/cooja-radio.c
atiselsts: I'm not following you here. The goal here is not to change the behavior of the code, just refactor it.

Do you think the proposed patch changes the behavior?
Or do you think `COOJA_TRANSMIT_ON_CCA` check shoould be removed and the `!channel_clear()` check inverted for some other reason?

yatch: Ah, ok. Thanks. You put `COOJA_TRANSMIT_ON_CCA` in the condition instead of keeping `#if COOJA_TRANSMIT_ON_CCA` in order to make the code simpler (we don't want to return with `RADIO_TX_COLLISION` immediately).

It was a bit confusing to me that both of  `COOJA_TRANSMIT_ON_CCA` and `send_on_cca` appear in a single condition since `COOJA_TRANSMIT_ON_CCA` is used to initialize `send_on_cca`. 

> Do you think the proposed patch changes the behavior?

No.

yatch:
@atiselsts Thanks for your feedback. I understand the scope of this PR, and think it ready for merge. 

> I see your point that the Cooja radio operation energy consumption simulation should be improved. At least I think I do have an idea how to make it more realistic, I'll give it a shot.

This sounds great. And, I presume you will do that after v4.5. I don't think you will update this PR (branch `feature/cooja-mote-improvements`) any more, but wait until later today just in case.

PR #1223, opened 2020-04-10
tunslip6 provide waiting periodic poll for connection to server

alexrayne:
introduce tunslip6 option -w - provide waiting for server connection by periodic polling,

+ option -c - provide loop mode, after server disconnect.

* this options allow some what like daemon mode for tunslip. so tunslip always
   ready for connect, handy for debug.

### DIFF
--- a/tools/serial-io/Makefile
+++ b/tools/serial-io/Makefile
@@ -4,7 +4,14 @@ DEPEND = tools-utils.h
 
 all: $(APPS)
 
-CFLAGS += -Wall -Werror -O2
+CFLAGS += -Wall -Werror
+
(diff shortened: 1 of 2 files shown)


atiselsts:
We don't see a big need to change this in the .c code level.

Why not simply have a bash script that restarts the tunslip continously? Or run tunslip from the supervisor daemon that automatically restarts the tunslip once it exits?

alexrayne:
Came reasons as for ping, imho. tunslip can check it`s arguments; bash scrit with passing arguments and check exit code cames not trivial; and there is no such bash script in serial-io yet.
i alredy done it, you cat accept, or reject.

PR #1226, opened 2020-04-15
Don't build Cooja unless a travis test actually needs it

### LINKED ISSUE #1088: Build cooja inside docker, conditionally
Currently we build cooja on the travis VM, outside docker. This is causing today's travis failures.

In the longer term, we want to move the cooja build to take place inside the container, but only for jobs that actually need cooja (`$BUILD_COOJA` is true).

g-oikonomou:
Our testing system already builds Cooja when needed, so this pull removes the step to build it from `.travis.yml`.

Closes #1088

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,11 +46,6 @@ before_install:
       fi
     fi
 
-  # Build Cooja conditionally
-  - if [ ${BUILD_COOJA:-false} = true ] ; then
-      ant -q -f $CNG_HOST_PATH/tools/cooja/build.xml jar ;
-    fi
-
   # Create a directory for out of tree tests and clone the test repo therein
   # The directory will need created unconditionally so we can always chgrp and
   # mount it, even if empty. Checkout a pre-defined version.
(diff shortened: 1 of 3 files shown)


g-oikonomou:
So as per our discussions earlier, fixing the java class version mismatch is pretty straightforward. Our testing system does in fact to build Cooja when it's not there, but as we see here the test will still fail and I am now out of ideas.

.travis.yml
```diff
@@ -46,11 +46,6 @@ before_install:
       fi
     fi
 
-  # Build Cooja conditionally
-  - if [ ${BUILD_COOJA:-false} = true ] ; then
```
simonduq, line 50: Remove variable BUILD_COOJA altogether?

g-oikonomou: Good point. Done now

g-oikonomou:
Do you mean this contiki-ng/cooja#34 ?

simonduq:
Yes I do! :) Merging this then

PR #1228, opened 2020-04-15
provide Makefile system with SOURCE_EXCLUDE feature.

alexrayne:
SOURCE_EXCLUDE now cleanup from builded objects.

* this feature i use to cleanup cooja library from unwanted dependences - stack-check.c
  not implemented by cooja. But it passed into library, and if library build with
  --gc-section, it crash loading, since no symbol _stack.


atiselsts:
I see that this might cause a problem - but is hacking the build system the right way to address it?

It's possible that in the file stack-check.c `#if STACK_CHECK_ENABLED ... #endif` guards need to be added for the whole file, so the code does not gets compiled if stack check is not enabled.

alexrayne:
> I see that this might cause a problem - but is hacking the build system the right way to address it?
> 
> It's possible that in the file stack-check.c `#if STACK_CHECK_ENABLED ... #endif` guards need to be 
yep, i see. looks that i have somewhere broken build, that infers stack_check into cooja

PR #1233, opened 2020-04-19
New neighbor table replacement policy

simonduq:
The current replacement policy for RPL had the issue that it only kicked in when the table was full; letting any new entry in until then. This would result in a failure to keep the intended maximum number of children: for instance, as a node receives new DAOs from direct children, it will add them to the table blindly. These entries are locked by the ùip-ds6-routes` module, and as a result might end up never removed. In certain scenarios a node can end up with only one single slot for the preferred parent, and left with switching option at all.

For RPL-Lite, we adopt a much simpler neighbor policy: simple remove the neighbor that offers the worst path rank (only non-storing mode, children don't require locked entries).

This is contributed by Yanzi Networks AB (www.yanzi.se).


os/net/routing/rpl-classic/rpl-nbr-policy.h
```diff
@@ -0,0 +1,24 @@
+/*
+** ===================================================================
+** Project  : mermaid
+**
+** Copyright (c) 2020 Yanzi Networks. All Rights reserved. This
+** file is the confidential and proprietary property of Yanzi Networks
+** and the possession, use, duplication or transfer in any way of this
+** file requires a written agreement from Yanzi Networks.
+**
+**===================================================================
```
arurke, line 10: These limitations might be a problem 😅

simonduq: Ahah very sorry, I forgot to update the header!
This and some more of our embedded code is green-lighted for open-sourcing.

simonduq: Fixed now!

atiselsts:
I feel now could be a good time to review and merge, to get a sufficient duration of user testing before including in the next release. What do you think?

simonduq:
Yes definitely!

joakimeriksson:
Totally agree on that!

atiselsts:
So - any updates? Should merge this soon? ;)

nvt:
Thanks for the reminder -- I'll take a look at this soon.

os/net/routing/rpl-classic/rpl-nbr-policy.c
```diff
@@ -27,231 +28,112 @@
...
-const linkaddr_t *
-rpl_nbr_policy_find_removable(nbr_table_reason_t reason,void * data)
+bool
+rpl_nbr_can_accept_new(const linkaddr_t *new, const linkaddr_t *candidate_for_removal,
+                       nbr_table_reason_t reason, void *data)
 {
-  /* When we get the DIO/DAO/DIS we know that UIP contains the
-     incoming packet */
+  bool accept;
   switch(reason) {
```
nvt, line 104: Shouldn't there be breaks in the cases where accept is set?

os/contiki-default-conf.h
```diff
@@ -97,9 +97,20 @@
...
+#ifndef NBR_TABLE_CONF_GC_GET_WORST
+#define NBR_TABLE_CONF_GC_GET_WORST            rpl_nbr_gc_get_worst
+#endif /* NBR_TABLE_CONF_GC_GET_WORST */
+#if ROUTING_CONF_RPL_CLASSIC
+/* Only Classic handles a max number of children */
+#ifndef NBR_TABLE_CONF_CAN_ACCEPT_NEW
+#define NBR_TABLE_CONF_CAN_ACCEPT_NEW          rpl_nbr_can_accept_new
+#endif /* NBR_TABLE_CONF_CAN_ACCEPT_NEW */
+/* Leave 3 spots for candidate parents incl. default route. */
+#if RPL_NBR_POLICTY_MAX_NEXTHOP_NEIGHBORS
```
nvt, line 110: Typo: "POLICTY"

os/net/nbr-table.c
```diff
@@ -179,101 +168,148 @@ nbr_set_bit(uint8_t *bitmap, nbr_table_t *table, nbr_table_item_t *item, int val
...
-#endif /* NBR_TABLE_FIND_REMOVABLE */
+  }
+  return count;
+}
+/*---------------------------------------------------------------------------*/
+static int
+used_count(const linkaddr_t *lladdr)
+{
+  int item_index = index_from_lladdr(lladdr);
+  int used = used_map[item_index];
```
nfi, line 211: Should there not be a check if lladdr is not found (`index_from_lladdr()` returns `-1`)?

os/net/nbr-table.c
```diff
@@ -179,101 +168,148 @@ nbr_set_bit(uint8_t *bitmap, nbr_table_t *table, nbr_table_item_t *item, int val
...
+                           nbr_table_reason_t reason, void *data)
+{
+  return entry_is_allowed(table, lladdr, reason, data, NULL);
+}
+/*---------------------------------------------------------------------------*/
+static nbr_table_key_t *
+nbr_table_allocate(nbr_table_reason_t reason, void *data, const linkaddr_t *to_be_removed_lladdr)
+{
+  if(memb_numfree(&neighbor_addr_mem) > 0) {
+    return memb_alloc(&neighbor_addr_mem);
```
nfi, line 303: Why the change to search for free memory twice instead of simply calling `memb_alloc()` and see if it returns `NULL` or not?

nfi (changes requested):

os/net/routing/rpl-classic/rpl-nbr-policy.c
simonduq: Oops well yes, absolutely. Done

os/contiki-default-conf.h
simonduq: Fixed

os/net/nbr-table.c
simonduq: I think it should not happen as this is only called on existing entries by `nbr_table_gc_get_worst()`, but I added the check anyway, for good measure.

os/net/nbr-table.c
simonduq: Yes good point, done now.

os/net/nbr-table.c
nfi: Yes, I agree but `nbr_table_gc_get_worst()` is a non-static function that potentially could be called by an application. Thanks for fixing.

nfi (approved):

nvt (approved):

PR #1234, opened 2020-04-19
Make CI test suite faster

simonduq:


simonduq:
This also removed a large number of compile tests; the idea being that we run `compile-all` before releases, and soon-ish periodically as a Travis cron job.

PR #1239, opened 2020-04-21
Move TSCH neighbors to a neighbor table

### LINKED ISSUE #1054: Neighbor table consistency

atiselsts:
Reduces memory usage and partially fixes #1054 .

### DIFF
--- a/os/net/mac/tsch/tsch-queue.c
+++ b/os/net/mac/tsch/tsch-queue.c
@@ -95,7 +95,6 @@ tsch_queue_add_nbr(const linkaddr_t *addr)
         /* Initialize neighbor entry */
         memset(n, 0, sizeof(struct tsch_neighbor));
         ringbufindex_init(&n->tx_ringbuf, TSCH_QUEUE_NUM_PER_NEIGHBOR);
-        linkaddr_copy(&n->addr, addr);
         n->is_broadcast = linkaddr_cmp(addr, &tsch_eb_address)
           || linkaddr_cmp(addr, &tsch_broadcast_address);
         tsch_queue_backoff_reset(n);
@@ -132,6 +131,12 @@ tsch_queue_get_time_source(void)
   return NULL;
 }
 /*---------------------------------------------------------------------------*/
+linkaddr_t *
+tsch_queue_get_nbr_address(const struct tsch_neighbor *n)
+{
+  return nbr_table_get_lladdr(tsch_neighbors, n);
+}
+/*---------------------------------------------------------------------------*/
 /* Update TSCH time source */
 int
 tsch_queue_update_time_source(const linkaddr_t *new_addr)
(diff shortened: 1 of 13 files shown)


atiselsts:
@yatch Some of the 6top tests are failing. I think because `13-ieee802154/code-6tisch/project-conf.h` configures `NBR_TABLE_CONF_MAX_NEIGHBORS         2` and relies on the neighbor table having exaclty 2 free spaces. Now that TSCH is also using neighbor table this is not always true.

yatch:
@atiselsts I will handle the failures.

yatch:
@atiselsts I found `tsch_queue_init()` now consumes two `nbr` entries: one for `tsch_eb_address`, and the other for `tsch_broadcast_address`. Increased `NBR_TABLE_CONF_MAX_NEIGHBORS` to `4` and modified the code a little bit. Please find attached a file which is a patch resoloving the failures. Tested on my machine locally.

[pr-1239.txt](https://github.com/contiki-ng/contiki-ng/files/4510867/pr-1239.txt)

os/net/mac/tsch/tsch-types.h
```diff
@@ -107,9 +107,7 @@ struct tsch_packet {
 
 /** \brief TSCH neighbor information */
 struct tsch_neighbor {
-  /* Neighbors are stored as a list: "next" must be the first field */
-  struct tsch_neighbor *next;
-  linkaddr_t addr; /* MAC address of the neighbor */
+//  linkaddr_t addr; /* MAC address of the neighbor */
```
yatch, line 110: I guess you're aware of this; but just in case...

atiselsts:
Now ready for review.

os/net/mac/tsch/tsch-queue.c
```diff
@@ -211,16 +213,14 @@ tsch_queue_remove_nbr(struct tsch_neighbor *n)
...
-
       tsch_release_lock();
 
       /* Flush queue */
       tsch_queue_flush_nbr_queue(n);
 
-      /* Free neighbor */
-      memb_free(&neighbor_memb, n);
+      /* Unlock and free neighbor */
+      nbr_table_unlock(tsch_neighbors, n);
```
simonduq, line 222: No need to unlock really, remove will do the job ("Locked" only means "not GC-able")

atiselsts: Thanks, updated the code.

atiselsts:
> How tested is this?

I checked that adding and removing neighbors work fine.

The regression tests pass, so apparently the new approach of using the nbr-table API to get neighbor's address works as well.

PR #1243, opened 2020-04-22
Re-organise the arch/dev directory

g-oikonomou:
In this PR I am proposing an additional level in the directory hierarchy under `arch/dev`. Hopefully we'll be adding more sensor and perhaps even radio drivers in the future and the top-level directory listing may become too long.


simonduq (changes requested):
Looks good! Can we revert to full(er) include paths rather than just the header file name?

g-oikonomou:
I reverted to full include paths and I also introduced full paths at various places that were just including with the filename alone. Hope I've not missed anything, let's see what actions thinks.

Ready for review and merge

simonduq (approved):

PR #1247, opened 2020-04-22
Add support for github actions

g-oikonomou:
This pull request adds support for github actinos. The intention is to run in parallel with travis CI or to replace it altogether.

This is WIP. We are partially setting up the GH actions framework, but there is still a lot left to be done:

* Need to add the ability to push containers to dockerhub - I may need some help with the secrets here
* Need to make sure we can run docker the way we want it to run
* Need to correctly communicate success/failure of the various steps to GH actions so we can get correct test outcomes.

I thought I'd throw this out here for review, ideas and thoughts.


g-oikonomou:
Removing draft status here!

PR #1250, opened 2020-04-25
Add support for MQTT 5

alexandruioanp:
This PR adds support for MQTT 5.0 (protocol level 5) as summarised [here](https://github.com/mqtt/mqtt.github.io/wiki/Differences-between-3.1.1-and-5.0).

Most of these new features are implemented via **Properties**. A large part of this diff is comprised of code to handle the encoding and decoding of these properties. I was considering pulling out the property handling code to a separate set of files (`mqtt-prop.c/h`?)

EDIT:
See [here](https://github.com/eclipse/mosquitto/issues/1416) for how to print properties using the mosquitto client. You should get something like this (see the **properties** field):
```
# mosquitto_sub  --verbose -t "#" -F %J -V 5

{"tst":1587821578,"topic":"iot-2/evt/status/fmt/json","qos":0,"retain":0,"payloadlen":93,"properties":{"user-properties":{"Contiki":"v4.5+"}},"payload":{"d":{"Platform":"native","Seq #":3,"Uptime (sec)":418,"Def Route":"fd00::1","RSSI (dBm)":0}}}
```


g-oikonomou:
!

> I was considering pulling out the property handling code to a separate set of files (mqtt-prop.c/h?)

Sounds like a reasonable idea. I see you have wrapped them around a cpp #if for conditional compilation - I don't think you need to necessarily do that, perhaps you can always compile and let the linker garbage collect those symbols.

I'm not sure why the diff is showing up with all those file renames.

alexandruioanp:
Regarding the renames, I made some space for contiguous MQTT test numbering (02 through 07, though I see I missed `07-native-ping.sh`).

The compiler was throwing errors for unused functions, so I had to wrap some of them in `ifdef`s, but I think it might be my container misbehaving.

g-oikonomou:
> The compiler was throwing errors for unused functions, so I had to wrap some of them in ifdefs, but I think it might be my container misbehaving.

Hmmm I this this will happen for unused static functions. But if you were to move them to a separate file and make them non-static it would go away.

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+
+typedef struct {
+  uint16_t len;
+  uint8_t data[MQTT_MAX_PROP_LENGTH];
+} mqtt_bin_data_t;
+
+typedef struct {
+  struct mqtt_string auth_method;
+  mqtt_bin_data_t auth_data;
+} mqtt_auth_event_t;
```
g-oikonomou, line 98: Here and for the data types above we seem to be mixing and matching `typedef struct {...} foo_t` with `struct foo_t {}`. Suggest consistency - is `foo_t` a struct or is it a typedef. Surely the latter?

I personally often go with `typedef struct foo_s {} fot_t` but that is just my style.

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+                      struct mqtt_out_property_t *prop,
+#endif
+                      mqtt_msg_type_t msg,
+                      mqtt_vhdr_prop_t prop_id, ...);
+
+void create_prop_list(struct mqtt_prop_list_t **prop_list_out);
+
+void print_props(struct mqtt_prop_list_t *prop_list, mqtt_vhdr_prop_t prop_id);
+
+void props_init();
```
g-oikonomou, line 123: PITA, but all of that will need renamed to `mqtt_prop_init()` , `_print`, `_create_list` etc

os/net/app-layer/mqtt/mqtt.h
```diff
@@ -544,14 +746,45 @@ void mqtt_set_username_password(struct mqtt_connection *conn,
...
 /*---------------------------------------------------------------------------*/
+void encode_var_byte_int(uint8_t *vbi_out,
+                         uint8_t *vbi_bytes,
+                         uint32_t val);
+/*---------------------------------------------------------------------------*/
+uint8_t decode_var_byte_int(const uint8_t *input_data_ptr,
+                            int input_data_len,
+                            uint32_t *input_pos,
+                            uint32_t *pkt_byte_count,
+                            uint16_t *dest);
```
g-oikonomou, line 770: Those two functions will need renamed too to be prefixed by the module name.

os/net/app-layer/mqtt/mqtt.h
```diff
@@ -544,14 +746,45 @@ void mqtt_set_username_password(struct mqtt_connection *conn,
...
+ *        re-authentication).
+ * \param prop_list Output properties.
+ * \return MQTT_STATUS_OK or some error status
+ *
+ * This function send an MQTT authentication message.
+ */
+mqtt_status_t mqtt_auth(struct mqtt_connection *conn,
+                        mqtt_auth_type_t auth_type,
+                        struct mqtt_prop_list_t *prop_list);
+#endif
```
g-oikonomou, line 786: OK functionality only supported in v5, but perhaps the prototype does not have to be conditional?

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+#define MQTT_PROP_H_
+/*---------------------------------------------------------------------------*/
+#include "mqtt.h"
+
+#include <stdarg.h>
+/*---------------------------------------------------------------------------*/
+/* If not using memb, you must provide a pointer to
+ * statically-allocated memory to register_prop()
+ */
+#ifdef MQTT_CLIENT_CONF_PROP_USE_MEMB
```
g-oikonomou, line 41: `MQTT_PROP_CONF_USE_MEMB`

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+#define MQTT_PROP_USE_MEMB MQTT_CLIENT_CONF_PROP_USE_MEMB
+#else
+#define MQTT_PROP_USE_MEMB 1
+#endif
+/*---------------------------------------------------------------------------*/
+/* Number of output property lists */
+#define MQTT_MAX_OUT_PROP_LISTS 1
```
g-oikonomou, line 48: `MQTT_PROP_MAX_OUT_LISTS`

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+
+/* Number of output properties that will be declared, regardless of
+ * message type
+ */
+#define MQTT_MAX_OUT_PROPS 2
```
g-oikonomou, line 53: `MQTT_PROP_...` etc for the defines

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+struct mqtt_prop_list {
+  /* Total length of properties */
+  uint32_t properties_len;
+  uint8_t properties_len_enc[MQTT_MAX_PROP_LEN_BYTES];
+  uint8_t properties_len_enc_bytes;
+  LIST_STRUCT(props);
+};
+
+/* This struct represents output packet Properties (MQTTv5.0). */
+struct mqtt_out_property {
```
g-oikonomou, line 75: `mqtt_prop_out_property`

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+   * (the range of 1 VBI byte is 0x00 - 0x7F)
+   */
+  mqtt_vhdr_prop_t id;
+  /* Property length */
+  uint32_t property_len;
+  /* Property value */
+  uint8_t val[MQTT_MAX_PROP_LENGTH];
+};
+
+struct mqtt_bin_data {
```
g-oikonomou, line 90: `mqtt_prop_bin_data` etc, so those need prefixed with `mqtt_prop_`

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+};
+
+struct mqtt_auth_event {
+  struct mqtt_string auth_method;
+  struct mqtt_bin_data auth_data;
+};
+/*----------------------------------------------------------------------------*/
+void mqtt_print_input_props(struct mqtt_connection *conn);
+
+uint32_t mqtt_encode_prop(struct mqtt_out_property **prop_out, mqtt_vhdr_prop_t prop_id,
```
g-oikonomou, line 102: `mqtt_prop_encode` etc for functions

os/net/app-layer/mqtt/mqtt-prop.h
```diff
@@ -0,0 +1,126 @@
...
+                           struct mqtt_out_property **prop_out,
+#if !MQTT_PROP_USE_MEMB
+                           struct mqtt_out_property *prop,
+#endif
+                           mqtt_msg_type_t msg,
+                           mqtt_vhdr_prop_t prop_id, ...);
+
+void mqtt_create_prop_list(struct mqtt_prop_list **prop_list_out);
+
+void mqtt_print_props(struct mqtt_prop_list *prop_list, mqtt_vhdr_prop_t prop_id);
```
g-oikonomou, line 121: `mqtt_prop_print`...

Issue #1038, opened 2019-08-31
Add TSCH support to the SimpleLink platform


g-oikonomou:
The SimpleLink platform's radio drivers do not support all the features required by TSCH. This is proving to be a popular platform and some users have been bitten by trying to use TSCH.

I have changed the platform's wiki page to document this lack of support.

This is something we want to be adding, so a roadmap item here.

https://github.com/contiki-ng/contiki-ng/wiki/Platform-simplelink

Yagoor:
Are you aware if Texas Instrument is working on this, since they are the maintainers of this platform port? 
If not I can work on it because I have the physical board at home.

Yagoor:
@g-oikonomou

Mano2Code:
> Are you aware if Texas Instrument is working on this, since they are the maintainers of this platform port?
> If not I can work on it because I have the physical board at home.

@Yagoor  Do you have a solution for this?

jeandudey:
What needs to be done specificly to get this working?

hseokman:
I have somewhat working version implemented for CC1312R
https://github.com/hseokman/contiki-ng/tree/tsch-cc1312r
Please provide feedback.
I had to change rtimer interrupt priority so that the radio API work in the interrupt handler and RTC and RAT was drifting away that I had to sync periodically. Also printing in the interrupt handler will reset the device that I had to disable framer and MAC logging.

stainlessinoxx:
@hseokman thanks! I’m trying your code on a CC1352. So far I’ve got 5-25% packet transfer success rate, using a scope there’s about a 10 factor missing in your sending time.

Someone also needs to point me to the TSCH algorithm documentation. I’ll get it working, it’s a good start. Understanding exactly what’s missing in the “gluing” layer and addressing it will be a pleasure for me, since I’m paid to do it.

hseokman:
I tried to match CC1310 implementation, and here are what I did to make CC1312R to work. It's been running pretty good. CC1312R is sub Ghz radio, so I don't know what may not work in 2.4Ghz simplelink platform.

1. Added missing cc13xx-50kbps-tsch.c so that it uses the 50kbps timing
2. Match radio parameters to CC1310 in cc13xx-cc26xx-def.h
3. Lowed RTC interrupt priority so that radio command can be executed in the interrupt handler
4. Changed preamble bytes and added timestamp in the radio setup to match CC1310
[...]

PR #648, opened 2018-09-09
TSCH support for CC13xx

atiselsts:
This allows to use TSCH in the sub-GHz range on CC13xx based devices.

The PR adds the missing radio API used by TSCH to the `prop-mode.c` radio driver, unifies the radio timer handling between `ieee-mode.c` and `prop-mode.c` drivers, and adds TSCH timing template for longer timeslots suitable for the sub-GHz networks. There is also a an update in CC13xx/CC26xx platform header files configures the platform for TSCH timings with the current 50 kbps data rate.

It has been tested on custom CC1310 hardware for several months.

PR #1255, opened 2020-05-04
TSCH support for the Simplelink platform

atiselsts:
This PR is based on the initial work of @hseokman (#1038) It adds support of TSCH for the IEEE mode. 

**Changes:**
The code is quite long as it separates RAT (radio timer) stuff in a new .C file and makes sure RTC and RAT are always synchronized. There is some refactoring of the `ieee_mode` and `prop_mode` data structures to allow to access them from a base class ( `simplelink_radio_mode_t`) pointer.

**Testing**
CC2652R1 nodes, CC2650 nodes, CC1312R1 nodes and a CC1310 node have been tested.

*People are welcome to try it out and report problems or lack of them!* 

Depends on #1295


atiselsts:
Now verified that basic functionality also works on CC2650 and that transmission works on CC1310 (had only single CC1310 node that could be flashed via JTAG).

hseokman:
I tried your branch, and it works on CC1312R. I guess your change worked to keep RAT and RTC in sync. However, there must be some default timeout set for 2.4Ghz that is faster than expected in the sub Ghz network. The device joins the network and falls off network after some time and it continues the same behavior. 
Do you think RAT timer still needs manual start and stop? I thought the TI radio driver automatically handles that.

atiselsts:
@hseokman - thanks for trying out. Do you have logs that you could share?

I have a couple ideas what could go wrong:
1) The `receiving_packet` function in IEEE mode is now behaving better, but I did not touch the prop mode code. It would be one of the first things to look.
2) The maximum waiting time in `transmit` code  might not be sufficient to transmit the whole packet in 50 kbps mode.

I have a question about the change you did in `netstack_sched_fs()`. The previous code tried to start the FS multiple times. After your change, just a single attempt is made. Should we add back the loop?

I have another question about `EVENTS_CMD_DONE` - you removed this macro from a couple places and replaced with checking the `CMD_STATUS`. Are you sure the change does not break anything?

I'm also on gitter, feel free to write there for a quicker turnaround.

hseokman:
I attached log from the device.
I have not had chance to look at why it is leaving the network. I think one of default time out period is short that the device is leaving the network.
[tsch_log.txt](https://github.com/contiki-ng/contiki-ng/files/4594722/tsch_log.txt)

I copied receiving_packet from cc26xx-cc13xx prop mode to simplelink prop mode.

I don't understand why netstack_sched_fs() required a loop. I don't know if syncth_error is expected that cmd_fs needs to be passed until the error is cleared.
The main reason I changed this and other areas is that RF_runCmd cannot run in the interrupt handler. TSCH calls radio layer from rtimer handler which makes them called in the interrupt handler. I guess this is for the scheduling at the precise time, so I changed them to RF_scheduleCmd and wait for the command is executed.

szilardkantor:
Could you explain please why is this neccessary when there has already been this pull?https://github.com/contiki-ng/contiki-ng/pull/648

atiselsts:
The PR you linked to is for the CC26x0/CC13x0 platform. TI SimpleLink [is a newer platform](https://github.com/contiki-ng/contiki-ng/wiki/Platform-simplelink ) that adds support for CC2652R1, CC1312R1 and other devices, as well as runs on the older ones.

szilardkantor:
I couldn't make it associate with 2 cc1310 launchpads. 
The node logs:
[INFO: TSCH      ] scan: received packet (20 bytes) on channel 25
[WARN: TSCH      ] scan: dropping packet, timestamp too far from current time 66520498 49579540
Is it supposed to send 20 bytes?

atiselsts:
@szilardkantor thanks for the report, the `bAppendTimestamp` was not included in the RF settings for CC1310 radios. I have updated the PR.

atiselsts:
Testing on my side is done. I'll keep watch on this for a few weeks to see if anyone tries this out reports any remaining problem. If all is good, then I'll remove the "work-in-progress" label and consider is a candidate for merging.

andresblanco06:
@atiselsts What is the status on this? any help needed?

atiselsts:
@andresblanco06
All four main board types have been tested (cc13x0, cc13x2, cc26x0, cc26x2) and interoperability with the cc26x0-cc13x0 platform was tested as well.

Just fixed the remaining outstanding issues I was aware of - in IEEE mode, 124 and 125 byte payload packets were not received correctly, and IEEE mode with CSMA was not working well.

Doing larger scale tests would be good, but it isn't normally required before merging PRs in the Contiki-NG develop branch. Anyway, if you can help / are aware of suitable testbeds let me know.

os/net/mac/tsch/tsch.c
```diff
@@ -928,7 +928,7 @@ PROCESS_THREAD(tsch_send_eb_process, ev, data)
         }
       }
     }
-    if(tsch_current_eb_period > 0) {
+    if(0 && tsch_current_eb_period > 0) {
```
alexrayne, line 931: you turn off EB here. How do you compensate it? don't see in other code, how EB sends.

atiselsts:
Removed the work-in-progress status on this one. Ready for review, depends on #1295 which should be handled before this can be merged.

sunilassao:
Have you verified the 200Kbps mode?

atiselsts:
@sunilassao the PR only adds support to 250 kbps mode in the 2.4 GHz range and 50 kbps mode in the 868/915 MHz range.
All other radio modes are untested and, and modes with different data rates (5 kbps, 2.5 kbps etc) are going to need different TSCH timings.

PR #1259, opened 2020-05-11
Simplelink radio fixes

atiselsts:
Before the TSCH driver can be developed, there are some fixes needed in the Simplelink platform. Discovered based on IEEE mode testing.


arch/cpu/simplelink-cc13xx-cc26xx/rf/sched.c
```diff
@@ -261,7 +261,7 @@ netstack_sched_fs(void)
    *
    * For Prop-mode, the synth is always manually calibrated with CMD_FS.
    */
-#if (RF_MODE == RF_CORE_MODE_2_4_GHZ)
+#if (RF_MODE == RF_MODE_2_4_GHZ)
```
g-oikonomou, line 264: Should probably also edit the comment next to `#endif` below

PR #1266, opened 2020-05-16
Update the motelist submodule

g-oikonomou:
This commit updates the motelist submodule to the latest version (v0.2).

This update brings a number of improvements:

* We fix mote detection on macOS Catalina
* We suppress all output when running with `--brief`
* We improve handling of `--version`:
  * The correct version will now be printed irrespective of the working dir that was used to run motelist from.
  * The git error message is suppressed if motelist is not inside a git-versioned dir
* motelist.py becomes executable


g-oikonomou:
This PR has now been updated to update the motelist submodule to the latest version (v0.3).

This update brings a number of improvements:

* Even further detection improvements on macOS Catalina
* We can now search for additional filename patterns under `/dev` through the `--paterns` command line argument
* Detection logic will persist in future macOS versions

nfi (approved):
Nice new feature set for motelist. Looks good!

Issue #1249, opened 2020-04-23
Orchestra rule prototypes are missing extern keyword


hseokman:
When I use native gcc tool, it complains about orchestra rule prototypes.
I see these in the orchestra.h

struct orchestra_rule eb_per_time_source;
struct orchestra_rule unicast_per_neighbor_rpl_storing;
struct orchestra_rule unicast_per_neighbor_rpl_ns;
struct orchestra_rule default_common;

Shouldn't these have extern keyword in front of all of these since it is not declaring these variables here?

alexrayne:
+1

g-oikonomou:
You are right. Flagging this as a bug.

atiselsts:
This is trivial to "fix", but I suspect there was some intention behind this lack of `extern`, so I never did change it. @simonduq should comment.

PR #1269, opened 2020-05-18
a minor trivial fixes set

alexrayne:
here provided a set of fixes tat i need to apply to build native and cooja motes with WERROR strict.

also included fix on Issue #1249

### DIFF
--- a/arch/platform/cooja/Makefile.cooja
+++ b/arch/platform/cooja/Makefile.cooja
@@ -72,5 +72,7 @@ CLEAN += COOJA.log
 CFLAGSNO = $(EXTRA_CC_ARGS) -Wall -g -I/usr/local/include -DCLASSNAME=$(CLASSNAME)
 ifeq ($(WERROR),1)
 CFLAGSNO += -Werror
+CFLAGSNO += -Wno-error=unused-function -Wno-error=format-security
+CFLAGSNO += -Wno-unused-function -Wno-format-security
 endif
 CFLAGS   += $(CFLAGSNO)
--- a/os/net/mac/tsch/tsch-slot-operation.c
+++ b/os/net/mac/tsch/tsch-slot-operation.c
@@ -93,9 +93,9 @@
  * of the guard time (one fourth of TSCH_DEFAULT_TS_RX_WAIT) */
 #define SYNC_IE_BOUND ((int32_t)US_TO_RTIMERTICKS(tsch_timing_us[tsch_ts_rx_wait] / 4))
 
-/* By default: check that rtimer runs at >=32kHz and use a guard time of 10us */
-#if RTIMER_SECOND < (32 * 1024)
-#error "TSCH: RTIMER_SECOND < (32 * 1024)"
+/* By default: check that rtimer runs at >=8 kHz and use a guard time of 10us */
+#if RTIMER_SECOND < (8 * 1024)
+#error "TSCH: RTIMER_SECOND < (8 * 1024)"
 #endif
 #if CONTIKI_TARGET_COOJA
 /* Use 0 usec guard time for Cooja Mote with a 1 MHz Rtimer*/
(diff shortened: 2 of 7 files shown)


os/net/mac/tsch/tsch-slot-operation.c
```diff
@@ -93,9 +93,9 @@
  * of the guard time (one fourth of TSCH_DEFAULT_TS_RX_WAIT) */
 #define SYNC_IE_BOUND ((int32_t)US_TO_RTIMERTICKS(tsch_timing_us[tsch_ts_rx_wait] / 4))
 
-/* By default: check that rtimer runs at >=32kHz and use a guard time of 10us */
-#if RTIMER_SECOND < (32 * 1024)
-#error "TSCH: RTIMER_SECOND < (32 * 1024)"
+/* By default: check that rtimer runs at >=8 kHz and use a guard time of 10us */
```
atiselsts, line 96: Why did you change this?

os/net/routing/rpl-lite/rpl.c
```diff
@@ -163,7 +163,7 @@ rpl_set_prefix_from_addr(uip_ipaddr_t *addr, unsigned len, uint8_t flags)
   }
 
   /* Try and initialize prefix */
-  memset(&curr_instance.dag.prefix_info.prefix, 0, sizeof(rpl_prefix_t));
```
atiselsts, line 166: I think it's the `sizeof` argument that needs to be changed, no the first argument of memset. The goal appears to be set the IPv6 address to zero, as the other members are initialized explicitly later in the code.

alexrayne: this is code from `prl-classic/rpl-dag.c` :

```
int
rpl_set_prefix(rpl_dag_t *dag, uip_ipaddr_t *prefix, unsigned len)
{
  rpl_prefix_t last_prefix;
  uint8_t last_len = dag->prefix_info.length;

  if(len > 128) {
    return 0;
  }
  if(dag->prefix_info.length != 0) {
    memcpy(&last_prefix, &dag->prefix_info, sizeof(rpl_prefix_t));
  }
  memset(&dag->prefix_info.prefix, 0, sizeof(dag->prefix_info.prefix));    //<-  looks like rpl-lite need this line same?
  memcpy(&dag->prefix_info.prefix, prefix, (len + 7) / 8);

```

os/net/mac/tsch/tsch-slot-operation.c
alexrayne: This comes with my first experiments with cooja polling precise resolu resolution. I wanna use RTIMER_SECOND as resolution, but later i go another way.
Anyway, imho, this change not bad, since code shouldnt stricts it`s conditions more then realy need.
This can be droped with no any suspicions.

os/net/routing/rpl-lite/rpl.c
alexrayne: yep, refixed it.

os/net/mac/tsch/tsch-slot-operation.c
atiselsts: Yes, please drop this commit - I think we need more justification than that!

arch/platform/cooja/Makefile.cooja
```diff
@@ -72,5 +72,7 @@ CLEAN += COOJA.log
 CFLAGSNO = $(EXTRA_CC_ARGS) -Wall -g -I/usr/local/include -DCLASSNAME=$(CLASSNAME)
 ifeq ($(WERROR),1)
 CFLAGSNO += -Werror
+CFLAGSNO += -Wno-error=unused-function -Wno-error=format-security
```
atiselsts, line 75: Why the change? Especially `-Wno-format-security` and `-Wno-error=format-security`. I think it should be left as is.
Which code caused you problems to build without these flags?

alexrayne: as i remember printf with only one fmt, whitout arguments, causes this warning

alexrayne:
close for cleanup PR

PR #1283, opened 2020-05-29
Various fixes for Simplelink/sensortag sensors 

g-oikonomou:
This PR starts addressing some of the sensortag sensor flaws identified for platform Simplelink in #1282

* We fix an error in a CPP conditional which resulted in the platform startup code failing to initialise the sensors module correctly.
* We fix the I2C acquire/release logic for the OPT3001 sensor driver
* We fix the I2C acquire/release logic for the BMP280 sensor driver
* We fix the I2C acquire/release logic for the TMP007 sensor driver
* We fix the I2C acquire/release logic for the HDC1000 sensor driver
* We pass `-D<board>=1` instead of `-D<board>` at compile time. This is done to make this platform consistent with platform cc26x0-cc13x0.

@jsolderitsch any help with testing appreciated.

PR #1285, opened 2020-05-30
Add I2C manipulation module for platform Simplelink

g-oikonomou:
All Simplelink sensortag sensing element drivers implement I2C transactions and bus manipulation internally using static functions. This creates a lot of code duplication.

This work in progress moves all those functions to a new module and makes them re-usable. It also changes all affected drivers to use those new functions.

This will be very handy once we have finalised our I2C HAL (#1242), since the new module implemented here is where we'd put all of the HAL's arch-specific functions.

~Requires #1283~

### DIFF
--- a/arch/cpu/simplelink-cc13xx-cc26xx/dev/i2c-arch.c
+++ b/arch/cpu/simplelink-cc13xx-cc26xx/dev/i2c-arch.c
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2018, Texas Instruments Incorporated - http://www.ti.com/
+ * Copyright (c) 2020, George Oikonomou - http://www.spd.gr
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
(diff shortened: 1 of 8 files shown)


arch/cpu/simplelink-cc13xx-cc26xx/dev/i2c-arch.c
```diff
@@ -0,0 +1,96 @@
...
+{
+  I2C_Transaction i2cTransaction = {
+    .writeBuf = wbuf,
+    .writeCount = wcount,
+    .readBuf = rbuf,
+    .readCount = rcount,
+    .slaveAddress = slave_addr,
+  };
+
+  if(i2c_handle) {
```
alexstanoev, line 65: Should be `if(!i2c_handle) {`

arch/platform/simplelink/cc13xx-cc26xx/sensortag/tmp-007-sensor.c
```diff
@@ -251,37 +206,50 @@ enable_sensor(bool enable)
...
 
   uint8_t status_data[] = { REG_STATUS };
   uint16_t status_value = 0;
 
-  spi_ok = i2c_write_read(status_data, sizeof(status_data),
-                          &status_value, sizeof(status_value));
-  if(!spi_ok) {
+  i2c_handle = i2c_arch_acquire(Board_I2C0);
+
+  if(i2c_handle) {
```
alexstanoev, line 216: Should be `if(!i2c_handle) {`

alexstanoev (changes requested):

g-oikonomou:
Rebased on top of `develop`

arch/platform/simplelink/cc13xx-cc26xx/sensortag/tmp-007-sensor.c
g-oikonomou: Fixed by rebase

arch/cpu/simplelink-cc13xx-cc26xx/dev/i2c-arch.c
g-oikonomou: Fixed with new commit

alexstanoev (approved):

PR #1286, opened 2020-05-30
Delete the obsolete very-sleepy-demo

g-oikonomou:
~Just a documentation update here.~

Edit following discussion:

This example was first developed for the original Contiki OS and was designed under the assumption that the underlying radio duty cycling is controlled by ContikiMAC.

This example has not been updated since the release of Contiki-NG and was only provided here for reference. In hindsight, its presence in the source tree is confusing users and overall does more harm than it does good.

This PR removes it.


g-oikonomou:
As recommended by @simonduq, I have changed this pull to now remove the example altogether. I'll also update the OP and title.

simonduq (approved):

g-oikonomou:
@simonduq shall we merge this one?

ahmedmibrahim:
We need a low power example. Currently Contiki-ng is not sufficient for Battery powered devices.
The MAC ON period should be controllable and less than the old 10 seconds period.

g-oikonomou:
Thank you for leaving a comment, but I am not sure I understand what you mean? Surely with TSCH we have good power savings?

ahmedmibrahim:
Thanks for following up. I've asked about this several time on gitter with no responses.
Here is what I meant:
- MAC on uses a lot of current, say ~7mA on CC1310 for example.
- To use Contiki-ng on a battery powered application, the developer has to turn the MAC off to save power (NETSTACK_MAC.off();)
- However, the MAC needs to stay on for at least 10 seconds (according to the very sleepy demo note)
_/*
 * If our preferred parent is not NBR_REACHABLE in the ND cache, NUD will send
 * a unicast NS and wait for NA. If NA fails then the neighbour will be removed
 * from the ND cache and the default route will be deleted. To prevent this,
 * keep the MAC on until the parent becomes NBR_REACHABLE. We also keep the MAC
 * on if we are about to do RPL probing.
 *
 * In all cases, the radio will be locked on for KEEP_MAC_ON_MIN_PERIOD secs
 */_


- These 10 seconds are too much and are enough to drain a coin cell. We need to have a way to keep MAC for a shorter period. I'm not an expert how this work, but this KEEP_MAC_ON_MIN_PERIOD should somehow be decreased or controllable.

PR #1287, opened 2020-05-30
Fix sensortag reed relay logic

g-oikonomou:
The button HAL is configured to use negative logic (active: low) for the sensortag's reed relay. This is incorrect, the reed is actually active high. This commit fixes the behaviour for both platforms that support sensortags.


alexstanoev (approved):
Tested on a CC2650 sensortag

PR #1288, opened 2020-05-30
Fix buffer overflow in the cc26x0-cc13x0 BMP280 sensor driver

g-oikonomou:
The BMP280 sensor's calibration data structure's size is 28 bytes, but we were only allocating 24 bytes for the buffer used for calibration I2C transactions. I2C transactions would write 4 bytes beyond the end of the buffer.

This PR fixes this bug.


atiselsts:
Looks good! In principle, I can merge it now. But for completeness, I also suggest getting rid of `MEAS_DATA_SIZE` define and replacing that name with `SENSOR_DATA_BUF_SIZE` in the code.

g-oikonomou:
I've done the recommended renaming + some other minor improvements. Thanks

PR #1290, opened 2020-05-30
Make the CC13xx/CC26xx base demo work for platform simplelink too

g-oikonomou:
This pull changes the baseline `cc26xx-demo` under `examples/platform-specific/cc26x0-cc13x0` to work for both CC13xx/CC26xx platforms (the older `cc26x0-cc13x0` but now also the newer `simplelink` platform).

Firstly, we move the demo to its own subdirectory and we rename it to `base-demo`, since it was never CC26xx-specific anyway, it always also supported all CC13xx flavours. This allows us to add platform-specific subdirs without making the top-level dir structure confusing.

Secondly, we adjust the example such that it supports both intended platforms.

Lastly, we add tests to make sure the example builds correctly for both supported platforms. In compliance with the paperwork reduction act, we remove a test that is now redundant ;)

We currently don't have any simplelink-specific example projects. This PR will provide simplelink platform users with an example that demonstrates the platform's hardware-specific features.


jsolderitsch:
The base-demo is NOT working for SensorTags and the SImplelink platform. If you build it with all sensors enabled and deploy to either a cc1350 or a cc2650, the devices enter a repeating cycle of watchdog resets. This behavior occurs when the source file changes for PRs 1283, 1287 and 1288 have been applied to a build environment after creating an environment based on this PR. I had worked with @g-oikonomou  in early Summer on trying to get the SimpleLink platform supported and this PR is what resulted.

jsolderitsch:
I have made some changes to a few files so that HDC sensor readings are now appearing to be accurate. See the branch jsolderitsch:contrib/simplelink-base-demo-enhanced for the details.

alexstanoev:
Tested and working on the CC2650 LaunchPad under both `simplelink` and `cc26x0-cc13x0` platforms.

This brings out some SensorTag driver issues as mentioned by @jsolderitsch, which we can start resolving. 
Merging #1283 and #1285 (when extended to the MPU-9250) will fix the watchdog resets, so I see no issues merging this.

g-oikonomou:
Changed base to branch `release-4.6`

g-oikonomou:
Following a discussion with @alexstanoev, we are deferring this to v4.7. We have changed the base branch for this PR back to `develop`.

alexstanoev (approved):

PR #1293, opened 2020-06-01
Improve ENERGEST support for platform Simplelink

### LINKED ISSUE #1292: ENERGEST_TYPE_DEEP_LPM not supported by platform simplelink

g-oikonomou:
As reported in #1292, platform simplelink does not currently toggle ENERGEST sleep/deep sleep states correctly. Unfortunately, we have no way to distinguish between LPM and DEEP_LPM without messing around with CoreSDK, which is something we should not do. But we can at least mark every entry to a low power state as entering `ENERGEST_TYPE_DEEP_LPM`.

Fixes #1292


atiselsts:
Why `ENERGEST_TYPE_DEEP_LPM`? If we're not confident that it actually drop to deep sleep, generic "sleep" may be better.

PR #1297, opened 2020-06-04
fix tsch update_current_bacokff for overlaped links

alexrayne:
provide update tsch current_bacokff  for links tha are overlaped by other active links.

   this fixes situation when backed-off link newer activates, due hide by other activities

* the scenario of a bug:
1) have pair of links on same slot configuration:
    - linkA - tx|shared at slot1
    - linkB - rx at slot1
2) linkA have interference, and therefore failes transmit. Due to shared, It increase some backoff - window =2
3) every time slot1 choosed linkA and backup link B. But since backoff - works linkB.
    So have that linkA never updates it's backoff

### DIFF
--- a/os/net/mac/tsch/tsch-slot-operation.c
+++ b/os/net/mac/tsch/tsch-slot-operation.c
@@ -370,6 +370,18 @@ get_packet_and_neighbor_for_link(struct tsch_link *link, struct tsch_neighbor **
 
   return p;
 }
+
+static
+void tsch_slot_operation_update_current_bacokff(void){
+    if(current_link != NULL
+        && (current_link->link_options & LINK_OPTION_TX)
+        && (current_link->link_options & LINK_OPTION_SHARED) )
+    {
+      /* Decrement the backoff window for all neighbors able to transmit over
+       * this Tx, Shared link. */
+      tsch_queue_update_all_backoff_windows(&current_link->addr);
+    }
+}
 /*---------------------------------------------------------------------------*/
 uint64_t
 tsch_get_network_uptime_ticks(void)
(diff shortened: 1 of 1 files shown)


atiselsts:
A good catch. I wonder why this hasn't been noticed earlier.

Just please, please follow the Contik-NG code style: https://github.com/contiki-ng/contiki-ng/wiki/Code-style
You can use uncrustify to fix / check your code. Avoid unnecessary changes such as in line 1029.

alexrayne:
> Avoid unnecessary changes such as in line 1029.
i not see in diff any changes about line 1029?

alexrayne:
I see another problem, that may be a reason for another PR here: imho, bacoff check should be moved from stage of get_packet _from queue to a stage where current link choose - since backoff state not depends on queue packets state, and therefore can be evaluated at any time. 
And more, we can have a situation when a pair of TX links wich is backoff ocupy evaluated current link and backup. Therefore both are rejects at slot execution time, and other links - RX or not backoff, are losse, and therefore slot miss with no work.

atiselsts:
> I see another problem

Yes, there's also this issue. I can more easily explain why it was not noticed previously - all Contiki-NG scheduling mechanisms (6tisch minimal and Orchestra) add links only to broadcast addresses, which means that `tsch_queue_update_all_backoff_windows` tries to decrement the backoff windows for *all* links. If you have two or more unicast links at the same timeslot, sure, this is a problem.

We need to think more carefully about the backoff functionality in Contiki-NG TSCH.  There's also the issue (not strictly a bug) of the Orchestra Non-storing rule, which tries to decrement the backoff-window on *every* slot, leading to backoff effectively almost not working.

Adding some tests regression tests which test the performance under high packet collisions would be very desirable.

> i not see in diff any changes about line 1029?

See here: https://github.com/contiki-ng/contiki-ng/pull/1297/commits/909e772c23f5c24c13cf494d1cc33b4a79c90ba5#diff-e9c374e1c1b686e77587ece3c4939e49L1029

alexrayne:
reopen for clean pr

PR #1300, opened 2020-06-08
Add LOG_*_BYTES to simplify logging a string of bytes at different log levels

MBradbury:
This PR adds macros to support logging bytes at different log levels. When investigating new protocols, it is useful to be able to log the raw data received while debugging and to then disable this logging by changing the log level.

### DIFF
--- a/os/sys/log.c
+++ b/os/sys/log.c
@@ -160,6 +160,16 @@ log_lladdr_compact(const linkaddr_t *lladdr)
 }
 /*---------------------------------------------------------------------------*/
 void
+log_bytes(const void *data, size_t length)
+{
+  const uint8_t *u8data = (const uint8_t *)data;
+  size_t i;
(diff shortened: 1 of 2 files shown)


os/sys/log.c
```diff
@@ -160,6 +160,16 @@ log_lladdr_compact(const linkaddr_t *lladdr)
 }
 /*---------------------------------------------------------------------------*/
 void
+log_bytes(const void *data, size_t length)
+{
+  const uint8_t *u8data = (const uint8_t *)data;
+  size_t i;
+  for (i = 0; i != length; ++i) {
```
g-oikonomou, line 167: Can you delete the space between for and the opening paren?

MBradbury: Done.

g-oikonomou:
I'm happy with that. I'll wait for Simon to have a look before we merge. Thanks! 👍

PR #1301, opened 2020-06-08
Support color coding log prefixes based on log level

MBradbury:
When debugging it is useful for the log prefix to be coloured based on the log level. This helps spot error messages when printing a large number of messages at other levels.


g-oikonomou:
I wonder whether we should make this default for platform native.

g-oikonomou:
I'm happy with that too. Again, I'll wait for Simon to have a look before we merge. Thanks! 👍

simonduq:
I'm also thinking this should be on by default on native, but merging this PR as it is anyway

PR #1271, opened 2020-05-20
minor trivial fixes set

alexrayne:
here provided a set of fixes tat i need to apply to build native and cooja motes with WERROR strict.

also included fix on Issue #1249

PR #1304, opened 2020-06-08
rpl-lite: fix out-of-bounds write warning in rpl_set_prefix_from_addr()

LDVG:
The previous call to `memset` wrote zeroes after the `prefix` member,
causing `-Warray-bounds` to issue warnings in GCC. This changes the
`memset` to only write zeroes to the `prefix` member before using `memcpy` to
copy the `prefix` from the `addr` argument. The remaining members of the
`prefix_info` struct are initialized using assignment expressions after
the `memcpy`.

### DIFF
--- a/os/net/routing/rpl-lite/rpl.c
+++ b/os/net/routing/rpl-lite/rpl.c
@@ -163,7 +163,7 @@ rpl_set_prefix_from_addr(uip_ipaddr_t *addr, unsigned len, uint8_t flags)
   }
 
   /* Try and initialize prefix */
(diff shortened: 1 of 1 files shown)


atiselsts:
This is already proposed in PR #1271 
We are currently waiting for @alexrayne to clean that one up.

LDVG:
> This is already proposed in PR #1271
> We are currently waiting for @alexrayne to clean that one up.

Cool! I guess this can be closed as a duplicate then. :slightly_smiling_face:

PR #1305, opened 2020-06-09
Add support for 'z' precision prefix to format strings

jakeru:
The 'z' precision prefix is very useful when formatting values of type
size_t and it is part of the C99 standard.

So add support for this to the function format_str_v(). This function is
used for example when formatting log output.

### DIFF
--- a/os/lib/dbg-io/strformat.c
+++ b/os/lib/dbg-io/strformat.c
@@ -373,6 +373,17 @@ format_str_v(const strformat_context_t *ctxt, const char *format, va_list ap)
       } else {
         flags |= SIZE_SHORT;
       }
+    } else if(*pos == 'z') {
+      if(sizeof(size_t) == sizeof(short)) {
+        flags |= SIZE_SHORT;
+      } else if(sizeof(size_t) == sizeof(long)) {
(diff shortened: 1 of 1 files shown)


os/lib/dbg-io/strformat.c
```diff
@@ -373,6 +373,17 @@ format_str_v(const strformat_context_t *ctxt, const char *format, va_list ap)
       } else {
         flags |= SIZE_SHORT;
       }
+    } else if(*pos == 'z') {
+      if(sizeof(size_t) == sizeof(short)) {
+        flags |= SIZE_SHORT;
+      } else if(sizeof(size_t) == sizeof(long)) {
+        flags |= SIZE_LONG;
+#ifdef HAVE_LONGLONG
+      } else if (sizeof(size_t) == sizeof(long long)) {
```
g-oikonomou, line 382: Remove space between if and opening paren

jakeru: Well spotted! Perhaps if I setup my environment better some automated tool would have helped me realizing this before pushing. Anyway. Now it is fixed.

Issue #1302, opened 2020-06-08
Backoff corner cases are broken with TSCH


atiselsts:
This is based on the discussion in PR #1297 as reported by @alexrayne

The following issues are identified as of now:

* When Rx and Tx links overlap, the Tx link may be ignored forever because the backoff is not decremented. This will be fixed in PR #1299
* When Tx and Tx links overlap with explicitly set unicast addresses, the backoff logic is probably not correct - all of them should have backoff decremented, not just one.
* When Orchestra NS rule is used, the backoff window is decremented in every slot, which is almost certainly not what the users expect. As the NS rule is the default, this should be given high importance!
* Regression tests that check these corner cases and backoff usage in general are missing.

atiselsts:
The first bullet point is fixed by #1307.

The rest are open for discussion. Not strictly under bugs, more like "counter-intuitive behaviors".

PR #1307, opened 2020-06-11
TSCH: also update backoff for Tx links that are overlapped by Rx links

atiselsts:
This fixes situation where a backed-off link never activates, due to being hidden by other links.

The scenario of the bug:
1) have pair of links on same slot configuration:
    - linkA - tx|shared at slot1
    - linkB - rx at slot1
2) linkA have interference, and therefore fails to transmit. backoff window is increased.
3) now, every time linkB is chosen over linkA. So linkA never updates its backoff


atiselsts:
Fixes #1302

simonduq (approved):
Oh I see, interesting bug, and very neat fix. 👍

PR #1311, opened 2020-06-13
propose remove static from process declaration for better debug

alexrayne:
Propos remove `static` from process pt-thread, since it is not affects code optimisation, but provides debug symbols.

* stack-trace on crash does not show names of a static symbols. statics shown only by routine adress. Therefore it is not a trivial to resolve - what process executes at crush.

* Also static declaration prevents declare process-thread visible to other units. Most cases it is not a problem, but it is unnesasary limitation.

* It is still not a problem declare process-thread with static explicit.


simonduq:
Hi!

Issue #292, opened 2018-02-14
Documentation on Unit-Testing


yatch:
There is no documentation on unit-testing.

It'd be nice to have explanation or introduction about how unit-tests found under `tests` directory work. In addition, example code for unit-testing under `examples` directory would be great.

PR #1340, opened 2020-07-27
Example unittest project

kam193:
This PR contains a simple example how to make own unittests project and run it. It partially covers #292 and


Yagoor:
Hi @yatch.
Could you please validate this PR?
It would be really good to have an example on how to use the unit test library. :D

nvt:
Thanks, this is a nice thing to have. There used to be a small unit test example in the old Contiki OS (https://github.com/contiki-os/contiki/blob/master/apps/unit-test/example-test.c), but it somehow got lost in the move to Contiki-NG.

kam193:
I searched and I found this `example-test.c` is still in the repo (https://github.com/contiki-ng/contiki-ng/blob/98c6580e6f1249737235a07caebe68caed250606/os/services/unit-test/example-test.c)  - maybe I should merge it with my example?

I think `example` dir is better for it. Anyway, without a whole example how to use unittests together with Cooja, it's still hard to start

nvt:
OK, in that case you can either just remove the old example file, or merge whatever content you find useful into the new file in this PR. The difference between them is pretty small, however.

kam193:
Ok, I used comments from old example because they have some information I missed

tests/02-compile-arm-ports/Makefile
```diff
@@ -45,6 +45,7 @@ libs/shell/openmote \
 libs/simple-energest/openmote \
 libs/timers/zoul \
 libs/trickle-library/zoul \
+libs/unit-tests/zoul \
```
atiselsts, line 48: I think it would be much better to add execution test rather than just a compilation test. A new test could be added in 07-simulation-base, and then this change removed.

tests/Makefile.simulation-test
```diff
@@ -35,7 +35,7 @@ BASESEED ?= 1
 # The number of runs (with different random seeds)
 RUNCOUNT ?= 1
 
-CONTIKI=../..
+CONTIKI ?= ../..
```
atiselsts, line 38: Is this change necessary?

atiselsts (commented):
Don't have objections to this, just a few comments. If you address them I'm personally fine with merging.

tests/Makefile.simulation-test
kam193: Yes, file `examples/libs/unit-tests/Makefile` is on other deep in files tree than this default path expects. It cannot be compiled without this change.

tests/02-compile-arm-ports/Makefile
atiselsts: No new commits, but I think we should merge anyway.

atiselsts (approved):

PR #1344, opened 2020-08-03
Orchestra: upate the ORCHESTRA_LINKADDR_HASH2 function

atiselsts:
The current implementation is particularly bad for the default unicast slotframe size of 17. Since `256 % 17` is 1, it means that `ORCHESTRA_LINKADDR_HASH2(a1, a2) == ORCHESTRA_LINKADDR_HASH2(a2, a1)`. This leads to slot allocation that is clearly not what the link-based rule was meant to provide, as both directions of the link now get the same timeslot.

The value of 264 is a good choice for the default size 17. It also a good choice for most other slotframe sizes: there are no prime numbers between 2 and 101 (including) that produce modulo 1 when used to divide 264.


simonduq (approved):
Ah great :) Maybe add a comment similar to the PR description? People might end up staring at this and really wonder why 264

atiselsts:
Updated with comment

PR #1345, opened 2020-08-03
TCSH: improve TSCH and Orchestra logging

atiselsts:
Additions:
* log slotframe sizes when initializing Orchestra;
* log the reason why EB packet is not generated.


simonduq (approved):
Nice.

PR #1356, opened 2020-08-23
Proposal: setup and teardown functions for unit tests

kam193:
This is a proposal of extending unit test library. Setup and teardown functions can be defined to run before/after each test (to e.g. prepare/clean data), which is common pattern when writing unit tests.

### DIFF
--- a/os/services/unit-test/unit-test.h
+++ b/os/services/unit-test/unit-test.h
@@ -72,7 +72,7 @@ typedef struct unit_test {
 /**
  * Run setup function when defined
  *
- * Use #define UNIT_TEST_SETUP_FUNCTION to set setup function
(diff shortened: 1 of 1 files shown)


atiselsts:
I don't think we need this functionality. The PR itself does no provide any usage examples. I'd vote against unless it is better justified.

kam193:
Right, I should add an example. I created it based on my experience when I used Contiki-NG UT toolkit. If you think it could be useful - I will add an example.

nvt:
I agree with @atiselsts. The benefit of the introduced functionality should be shown either in this thread or in example code added to the PR.

kam193:
Unfortunately, I've stopped working with Contiki some time ago. As so, I'm not planning to finish this in the near future.

PR #1361, opened 2020-08-27
TSCH: disable HW frame filtering for cc2420 radios

### LINKED ISSUE #1358: TSCH needs frame filtering disabled on msp430, at least with mspsim
I can't see any obvious reasons for this, however, the behavior is there, at least with the current version of mspsim.
Its not something recently broken in Contiki-NG, as it goes back to at least release 4.3.

Perhaps just adding this flag for the msp430 platforms would be the simplest solution:

    #define TSCH_CONF_HW_FRAME_FILTERING 0

atiselsts:
Fixes #1358

### DIFF
--- a/arch/platform/z1/contiki-conf.h
+++ b/arch/platform/z1/contiki-conf.h
@@ -64,6 +64,8 @@
 #define TSCH_LOG_CONF_QUEUE_LEN 4
 #endif
 
+#define TSCH_CONF_HW_FRAME_FILTERING 0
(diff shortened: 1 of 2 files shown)


simonduq (commented):
Hmm, apparently this was already defined for sky in `arch/platform/sky/sky-def.h`

atiselsts:
Updated

simonduq (approved):

PR #1362, opened 2020-08-27
viewconf: extend the viewconf info with RPL MOP (mode of operation)

atiselsts:


tools/viewconf/viewconf.c
```diff
@@ -16,6 +16,10 @@
 
 ##### "CONTIKI_VERSION_STRING": ________________ == CONTIKI_VERSION_STRING
 
+#undef FRAME802154_IEEE802154_2003
+#undef FRAME802154_IEEE802154_2006
+#undef FRAME802154_IEEE802154_2015
+
```
g-oikonomou, line 22: I cannot immediately understand why that is here. Is it required as part of this PR?

atiselsts: This is not connected to RPL MOP, but aims to prettify the output, in the same way that `#undef`'ing RPL modes of operation makes the output more user friendly. I feel its too small change to deserve own commit.

Output before the patch; FRAME802154_IEEE802154_2006 is found by the preprocessor and replaced by "(0x01)":

     ##### "FRAME802154_CONF_VERSION":_______________ == (0x01)

Output after the patch, FRAME802154_IEEE802154_2006 is not found by the preprocessor:

     ##### "FRAME802154_CONF_VERSION":_______________ == FRAME802154_IEEE802154_2006

g-oikonomou (approved):

PR #1363, opened 2020-08-29
Simulation Makefile: target for running single simulation & clean build

kam193:
This PR proposes useful changes to the `Makefile.simulation` system:

1. Makefile target that allow running single *.csc simulation (just by `make simulation.csc`)
2. `clean` target for simulation also removes all `build` dirs in the simulation

### DIFF
--- a/tests/Makefile.simulation-test
+++ b/tests/Makefile.simulation-test
@@ -48,10 +48,13 @@ endif
 	@echo "========== Summary =========="
 	@cat summary
 
-%.testlog: %.csc cooja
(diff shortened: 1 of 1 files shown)


atiselsts:
1. I think it goes against the tradiotional usage patter of make, where the name of the target usually is the name of the file to be built. Anyway, there already is the %.testlog target, which provides the same functionality, the proposed change is just a "syntactic sugar". 

2. Perhaps adding a new target, for example distclean, would be a better approach. I'm not 100% sure that modifying existing target in this way will not break someone's workflow.

joakimeriksson:
Thanks for the efforts and the PR. On the basis of the feedback above from @atiselsts we decided to close this for now. We are happy to re-open it if you feel that there is good motivation that we missed or if you have improvements!

PR #1373, opened 2020-09-05
CBOR Library implementation

Yagoor:
In this pull request I'm proposing a CBOR library that is meant to be generic enough to be used in any application.
At the moment that this PR is being created the entire decoding engine is implemented with tests and examples.
I'm still working in the encoding engine (It is partially finished) and  the documentation doxygen is non-existent, yet.

I'm just creating this PR to get some feedback in the API definition because I want it to be easy to use but also not so heavy in terms of MIPS, RAM and ROM.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,8 @@ summary
 tests/[0-9][0-9]-*/org/
 tests/18-coap-lwm2m/Californium.properties
 tests/18-coap-lwm2m/leshan-server-demo*.jar
+
+# Visual Studio Code
+.vscode/*
+*.code-workspace
+.history/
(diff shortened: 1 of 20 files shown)


MBradbury:
How does this compare to [NanoCBOR ](https://github.com/bergzand/NanoCBOR)or [TinyCBOR](https://github.com/intel/tinycbor)? I know that NanoCBOR is pulled into RIOT as a library https://riot-os.org/api/group__pkg__nanocbor.html

Yagoor:
I must confess that I didn't knew NanoCBOR. 
However, I knew the other two and I believe my implementation aims to be simpler.

I can run some memory analysis to see if I have a smaller code footprint around these options above.

PR #1377, opened 2020-09-09
TSCH: fix filling and resetting of last_eb_nbr_addr

atiselsts:
The fix in eb_input was already discussed and agreed in #890 - somehow it slipped through the cracks.

The fix in the function `resynchronize` adds resetting of the `last_eb_nbr_addr`  variable. Otherwise, after trying to resync on an address it once, the variable will be equal to the time source address. If there are no other EB sending neighbor, the variable will always remain equal to the time source address, and subsequent resync calls will just try to resync on the same neighbor.


simonduq (approved):
Good catch, this now seems to implement what the original var was intended for :)

PR #1381, opened 2020-09-13
Support uni-directional UART for cc26x0-cc13x0

arurke:
Removes the requirement for both UART TX and RX to be set, thus adding support for boards which only have one or the other. (Same PR as https://github.com/contiki-os/contiki/pull/2259)

### DIFF
--- a/arch/cpu/cc26x0-cc13x0/dev/cc26xx-uart.c
+++ b/arch/cpu/cc26x0-cc13x0/dev/cc26xx-uart.c
@@ -194,7 +194,7 @@ lpm_drop_handler(uint8_t mode)
    * between MCU and AON when we drop to deep sleep. This here is essentially a
    * workaround
    */
-  if(accessible() == true) {
+  if(accessible()) {
     while(ti_lib_uart_busy(UART0_BASE));
   }
 
@@ -213,7 +213,7 @@ lpm_drop_handler(uint8_t mode)
    *
    * Only touch UART registers if the module is powered and clocked
    */
-  if(accessible() == true) {
+  if(accessible()) {
     /* Disable the module */
     ti_lib_uart_disable(UART0_BASE);
 
@@ -299,7 +299,7 @@ cc26xx_uart_write_byte(uint8_t c)
     return;
   }
 
-  if(accessible() == false) {
+  if(!accessible()) {
     enable();
   }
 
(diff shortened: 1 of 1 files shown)


g-oikonomou:
Can we have a single function `usable()` that accepts `BOARD_IOID_UART_xx` as an argument and compares with `IOID_UNUSED`? Not sure what the argument data type would be, off the top of my head I think `uint32_t`?

arurke:
> Can we have a single function usable() that accepts BOARD_IOID_UART_xx as an argument and compares with IOID_UNUSED?

Sure. Updated and added some minor cleaning.

> I think uint32_t

That's correct

arch/cpu/cc26x0-cc13x0/dev/cc26xx-uart.c
```diff
@@ -297,12 +294,12 @@ cc26xx_uart_init()
...
 {
-  /* Return early if disabled by user conf or if ports are misconfigured */
-  if(usable() == false) {
+  /* Return early if disabled by user conf or if TX pin is misconfigured */
+  if(!usable(BOARD_IOID_UART_TX)) {
     return;
   }
 
-  if(accessible() == false) {
+  if(accessible()) {
```
g-oikonomou, line 302: That seems wrong

arch/cpu/cc26x0-cc13x0/dev/cc26xx-uart.c
```diff
@@ -346,13 +343,13 @@ cc26xx_uart_set_input(int (*input)(unsigned char c))
...
-  /* Return early if disabled by user conf or if ports are misconfigured */
-  if(usable() == false) {
+  /* Return early if disabled by user conf or if TX pin is misconfigured */
+  if(!usable(BOARD_IOID_UART_TX)) {
     return UART_IDLE;
   }
 
   /* If the UART is not accessible, it is not busy */
-  if(accessible() == false) {
+  if(accessible()) {
```
g-oikonomou, line 352: Ditto

arch/cpu/cc26x0-cc13x0/dev/cc26xx-uart.c
arurke: 🤦  Quite impressive. Fixed.

arch/cpu/cc26x0-cc13x0/dev/cc26xx-uart.c
arurke: 🤦  Quite impressive. Fixed.

g-oikonomou:
ReAdy with that?

g-oikonomou (approved):

PR #1383, opened 2020-09-13
Enabled stack check on sky

Yagoor:
In this PR I enabled the Stack check on the sky port.

I believe that this was an old setting based on the comment.

If the comment is valid some unit test must fail.

### DIFF
--- a/arch/platform/sky/contiki-conf.h
+++ b/arch/platform/sky/contiki-conf.h
@@ -35,11 +35,6 @@
 #ifndef AES_128_CONF
 #define AES_128_CONF cc2420_aes_128_driver
 #endif /* AES_128_CONF */
-
-/* Disable the stack check library by default: .rom overflow otherwise */
-#ifndef STACK_CHECK_CONF_ENABLED
-#define STACK_CHECK_CONF_ENABLED 0
-#endif
 /*---------------------------------------------------------------------------*/
 #include "msp430-conf.h"
 /*---------------------------------------------------------------------------*/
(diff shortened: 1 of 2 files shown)


atiselsts:
If the old policy is not necessary anymore, I'm fine with enabling it.

The patch as its now does have a couple of issues though. The platform should not force this check to be *enabled* - it can force something to be disabled, if there is a good reason. If a user does not want to use this feature, they should be able to do disable it in application's `project-conf.h`. Also, the comment after your change becomes misleading!

I suggest simply removing all lines starting with the comment and ending with `#endif`. Then check if all examples still build for sky by running the compile-all script:

```
$ cd tests/compile-all/
$ PLATFORMS=sky ./build.sh
```

Yagoor:
Thank you for the feedback.
I did the changes you suggested and I ran the script and the following exampled failed.
I checked and this example also fails on develop.

```
Failed examples: ; ../../examples/storage/cfs-coffee for sky (default)
```

atiselsts:
Can you rebase and squash the commits in a single one? I will merge then. Thanks.

Yagoor:
HI @atiselsts.
Sorry for the late response. It is done. :)

atiselsts (approved):

atiselsts:
@g-oikonomou I approved the PR. We need for Travis to pass. Can you restart the build?

g-oikonomou:
I restarted and its passed. Good to merge? 😄

PR #1385, opened 2020-09-15
Make sure sensors-process runs when needed in simplelink platform

arurke:
As in title.

Disclaimer: I don't have hardware to test this, but seems obvious enough (although surprising no one has noticed... 🤔 )


g-oikonomou:
Someone had in fact noticed: #1283 😺

g-oikonomou (approved):

PR #1387, opened 2020-09-17
TSCH: Enable TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS by default and for nRF52840

toshiba-bril:
This pull request enables `TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS ` by default, thus defaulting any new platforms to hardware SFD timestamps for TSCH everywhere, as discussed on Gitter. To avoid changing the behaviour of existing platforms, their original setting has been maintained.

On the nRF52840 platform this PR enables the setting to fix occasional TSCH desync issues, as well as the inability to use older nRF52840 preview development kits (PDK) in TSCH networks without them dropping out of sync almost immediately.


g-oikonomou (approved):
I'm approving this, as discussed. @atiselsts any thoughts?

atiselsts:
Sorry, I don't agree with this at all. This PR disables hardware timestamps on all platforms that did not explicitly enable it before, effectively ignoring and overriding any user config.

In my opinion, the only valid reason to override `TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS` in platform's config is that the platform is not capable of accurate timestamping in HW (but I don't know any Contiki-NG platforms that can run TSCH and have this problem).

What could be two valid options:
- change tsch-conf.h, and change the default on all platforms (which also requires retesting on all platforms)
- change the default just for nrf52840, keeping `#ifndef TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS` guards around it to allow user overrides - this is what I had in mind yesterday

Eventually we should aim for the former anyway.

g-oikonomou:
Do you think we should be allowing users to override this? I was thinking of this as more of a maintainer config, hence my recommendation to put this in `-def.h`.

Effectively this PR tries to do two things in one go. We can factor out the nrf fix if that's more comfortable

atiselsts:
We can make this non-user configurable if there are good reasons, but then it should be removed from `tsch-conf.h`, no? I assume things that are user-configurable are in `tsch-conf.h`, while things that are not  (the various `DELAY_BEFORE_`) are only in platform's files.

g-oikonomou:
For Simplelink and cc13x0-cc26x0 we do this in the CPU `-def.h`:

```
#ifndef TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS
#define TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS 1
#define TSCH_CONF_TIMESYNC_REMOVE_JITTER 0
#endif
```

So for both CPUs we disable `TSCH_CONF_TIMESYNC_REMOVE_JITTER`, which defaults to the value of `TSCH_RESYNC_WITH_SFD_TIMESTAMPS`. So for the nRF52840 element of this PR, we should perhaps be doing the same and this can be done in a separate PR to move it forward, since it clearly improves TSCH performance very significantly for this platform.

> In my opinion, the only valid reason to override TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS in platform's config is that the platform is not capable of accurate timestamping in HW (but I don't know any Contiki-NG platforms that can run TSCH and have this problem).

The only one I am not 100% certain about is Zoul sub-ghz with the CC1200. I _think_ (this has always confused me), time use SFD timestamps for packets we need the GPIO2 interrupt functionality, which is being disabled by the Zoul configuration.

Regarding the change to the default in TSCH, we can do it at the same time as changing platform behaviour (which as you say needs testing, and I am not sure who would do such testing for legacy platforms like sky, z1), or we can change the default, but leave platform behaviour unchanged (which is what I had in mind).

These things of course need to stay configurable in `tsch-conf.h`, but I was thinking that it would be unlikely for an end-user / application-developer to want to change this. I was thinking that a platform developer/maintainer would provide the correct configuration for each individual platform and then end-users would not need to worry about it.

atiselsts:
I agree with copy-pasting the four lines from TI platforms' code. That's the simplest approach for this PR.

In the future, if we do a bigger PR, I can test for sky and z1 (a couple nodes on the table, nothing large scale). Would be nice to hear some input from Simon what made him to avoid setting HW timestamps as the default.

simonduq:
Hi!

PR #1391, opened 2020-09-21
Include LWM2M protocol version in registration message

### LINKED ISSUE #1389: lwm2m build does not report its protocol version
Problem is at line 729 of `os/services/lwm2m/lwm2m-rd-client.c`:

`snprintf(query_data, sizeof(query_data) - 1, "?ep=%s&lt=%d&b=%s",
session_info->ep, session_info->lifetime, session_info->binding);`

Notice the endpoint name and the binding mode are included, but not the protocol version.  The wakaama LWM2M server thus refuses to accept a connection from this contiki-ng client.  Changing that line to:

'snprintf(query_data, sizeof(query_data) - 1, "?lwm2m=1.1&ep=%s&lt=%d&b=%s", session_info->ep, session_info->lifetime, session_info->binding);`

fixes this problem.  Of course a real fix should separate the protocol version, e.g., including it as a preprocessor definition.

Robert-Lasater:
Adds a preprocessor definition LWM2M_PROTOCOL_VERSION; set it to "1.0".  Extend query_data to include this with the key "lwm2m=".

Fixes #1389

### DIFF
--- a/os/services/lwm2m/lwm2m-rd-client.c
+++ b/os/services/lwm2m/lwm2m-rd-client.c
@@ -726,7 +726,8 @@ periodic_process(coap_timer_t *timer)
         coap_init_message(session_info->request, COAP_TYPE_CON, COAP_POST, 0);
         coap_set_header_uri_path(session_info->request, "/rd");
 
-        snprintf(query_data, sizeof(query_data) - 1, "?ep=%s&lt=%d&b=%s", session_info->ep, session_info->lifetime, session_info->binding);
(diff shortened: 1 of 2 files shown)


g-oikonomou:
Thanks for this. Some feedback since it's your first one. You can make the title shorter and the first post more detailed. (Perhaps you can edit).

So as a title I would have used something like "Include LWM2M protocol version in registration message"

Adding a "Fixes #1389" (or "Closes" or some other keywords) in the PR description or commit message will automatically close the respective issue when this gets merged.

joakimeriksson (approved):
I think this looks ok to merge!

nfi (approved):
Looks good.

PR #1395, opened 2020-09-24
Add a result visualization example with logfile analysis and plotting script

atiselsts:
This adds an often requested feature - an example that shows how various metric can be obtained and plotted from log files.

Documentation is provided in the README file. If the PR is accepted, I will publish a more extensive documentation as a tutorial in the Contiki-NG wiki.


g-oikonomou:
This is VERY desirable! 🥳

rajeev1986:
A much-needed feature. Although IoTBench consortium has been running for the past 2years, open-source scripts for analyzing and plotting data has never been released, which can lead to discrepancies in results published out there. 
Very cool @atiselsts .

atiselsts:
@simonduq 

1. I have changed the location to examples/benchmarks/result-visualization

2. I looked into `rpl-req-resp` example and I'm not convinced they should be merged. You have a data query application, while my example has the more common data collection traffic pattern. I think there is room for both, in different .c files.

3. I don't know about adding the examples as a regression test as it is - it takes relatively long time to execute. Alternatively, a shorter test script could be added to `tests` folder (just to check if the basic running+plotting functionality is working, not to get useful results - for example, just 5 minute execution in a small network). What do you think? Is it fine to just use the cooja.csc file as it is? It takes 1min 20sec to run the current example on fast CPU (10th gen i7).

simonduq:
Great! I thought we aimed for building all examples on at least one target. Generally not great to add code that isn't even built by CI.

Adding a test script sure! Even run in CI I'd say.

atiselsts:
New regression test added, with its own .csc file.

simonduq (approved):

PR #1396, opened 2020-09-25
Tunslip help update

andraskalmar:
This PR updates the helper for tunslip to show and to describe all
the possible input command line parameters.


g-oikonomou:
I can understand the addition for `-I` and `-P`. 

`-CiRUVl` I do not understand at all. These seem to be non-existing options

`-L` is documented, but it doesn't seem to be doing anything by the way.

g-oikonomou:
No activity here, so #1415 and closing.

PR #1402, opened 2020-10-02
framer15.4 has_panid() optimisation

alexrayne:
Here provided optimised frame802154_has_panid( ... )  on code size and speedup.


atiselsts:
Not sure if there are significant and measurable benefits from this, do you have benchmarks?

Code style is off, it would need a lot of work before being ready to merge even if there is a good motivation for this change.

nvt:
The comments from @atiselsts have not been addressed in about half a year. Closing.

PR #1404, opened 2020-10-02
Prevent out-of-bounds write when processing external headers in sicslowpan

nvt:
When processing external headers in sicslowpan, a long chain of headers could possibly overflow the uip_buf variable.

In the code below, which follows the now inserted check, the writes to the exthdr struct could result in an out-of-bounds write.

<code>
    exthdr = (struct uip_ext_hdr *)ip_payload;
    exthdr->len = (2 + len) / 8 - 1;
    exthdr->next = next;
</code>

A packet test that should trigger the bug is included in the PR.


nvt:
I realize now that the fix can be improved, so I will close this and prepare a new one.

PR #1410, opened 2020-10-09
Prevent infinite loop in ND6

nvt:
The 8-bit nd6_opt_offset variable is used as a counter in a loop in the ns_input() function. Unfortunately, the line <code>nd6_opt_offset += (ND6_OPT_HDR_BUF(nd6_opt_offset)->len << 3);</code> can cause it to wrap around to the same value, resulting in an infinite loop.

This pull request changes it to be a 16-bit integer.


nfi (approved):
Looks good.

PR #1411, opened 2020-10-10
Native USB support for nRF52840 platforms

alexstanoev:
This PR adds support for the native USB interface found on the nRF52840 SoC.
Both SLIP and debug IO over USB are supported. 
This provides a large performance increase for border routers running on the nRF and makes the PCA10059 dongle a great choice for cheap border routers.

This PR also brings support for the Nordic DFU Trigger interface, which allows the nRF dongle to be programmed automatically by `nrfutil` without having to press the bootloader button on the board.
Note that this requires permissions that can be granted with the following udev rule, which would have to be added to the wiki:
```
# Set /dev/bus/usb/*/* as read-write for all users (0666) for Nordic Semiconductor devices
SUBSYSTEM=="usb", ATTRS{idVendor}=="1915", MODE="0666"
```

This has been tested on an nRF52840 DK (PCA10056) and nRF52840 Dongle (PCA10059), each with `examples/rpl-border-router`, but any additional testing is greatly appreciated.

Passing `NRF52840_NATIVE_USB=1` when building for the dk switches everything over to the native USB port (J3 labelled nRF USB).
Passing `NRF52840_USB_DFU_TRIGGER=0` when building for the dongle will disable the DFU trigger interface.


arch/cpu/nrf52840/Makefile.nrf52840
```diff
@@ -23,13 +23,17 @@ endif
 ifeq ($(BOARD), dongle)
   LDSCRIPT ?= $(CONTIKI_CPU)/ld/nrf52840-bootloader.ld
   CFLAGS += -DBOARD_PCA10059
+  NRF52840_NATIVE_USB ?= 1
+  NRF52840_USB_DFU_TRIGGER ?= 1
 else
   LDSCRIPT ?= $(CONTIKI_CPU)/ld/nrf52840.ld
   CFLAGS += -DBOARD_PCA10056
+  NRF52840_NATIVE_USB ?= 0
+  override NRF52840_USB_DFU_TRIGGER = 0
```
g-oikonomou, line 32: Suggest throwing a warning instead of overriding silently, like so

```
  ifeq ($(NRF52840_NATIVE_USB)$(NRF52840_USB_DFU_TRIGGER), 11)
    $(warning "Building for BOARD=dk")
    $(warning "with NRF52840_NATIVE_USB=1 NRF52840_USB_DFU_TRIGGER=1")
    $(warning "This is invalid. Overriding with NRF52840_USB_DFU_TRIGGER=0")
    override NRF52840_USB_DFU_TRIGGER = 0
  endif  
endif
```

g-oikonomou (approved):
Tested with DK on macOS.

Just a small comment and then happy to merge

PR #1412, opened 2020-10-14
TSCH: implement packet filtering by backup link

atiselsts:
In my opinion, in a simple network with just 2 nodes and 100% link quality 100% of packets should be successfully received. This is currently not the case with Orchestra, as there are occasional conflicts between the different slotframes, leading to around %95 packet ACK rate in these perfect conditions.

The IEEE 802.15.4 standard mandates that packet transmission should take precedence over reception. However, it does not prohibit being more conservative and postpone sending packets in case we know that another node might transmit a packet to us within this slot. This PR implements this approach, by using the backup link to quickly check for potential receptions. The backup link is required to belong to another slotframe in order to avoid blocking Tx in case of a single-slotframe scheduling conflicts.

A regression test is added which checks whether 100% packet ACK rate is achieved.

I have tested this approach in simulated larger network and in the FIT-IoT lab with 30 nodes. The results were not statistically different between the previous approach and after adding this new feataure.

### DIFF
--- a/os/net/mac/tsch/tsch-queue.c
+++ b/os/net/mac/tsch/tsch-queue.c
@@ -409,7 +409,8 @@ tsch_queue_is_empty(const struct tsch_neighbor *n)
 /*---------------------------------------------------------------------------*/
 /* Returns the first packet from a neighbor queue */
 struct tsch_packet *
-tsch_queue_get_packet_for_nbr(const struct tsch_neighbor *n, struct tsch_link *link)
+tsch_queue_get_packet_for_nbr(const struct tsch_neighbor *n, struct tsch_link *link,
+                              struct tsch_link *backup_link)
 {
   if(!tsch_is_locked()) {
     int is_shared_link = link != NULL && link->link_options & LINK_OPTION_SHARED;
@@ -421,8 +422,14 @@ tsch_queue_get_packet_for_nbr(const struct tsch_neighbor *n, struct tsch_link *l
 #if TSCH_WITH_LINK_SELECTOR
         int packet_attr_slotframe = queuebuf_attr(n->tx_array[get_index]->qb, PACKETBUF_ATTR_TSCH_SLOTFRAME);
         int packet_attr_timeslot = queuebuf_attr(n->tx_array[get_index]->qb, PACKETBUF_ATTR_TSCH_TIMESLOT);
-        if(packet_attr_slotframe != 0xffff && packet_attr_slotframe != link->slotframe_handle) {
-          return NULL;
+        if(packet_attr_slotframe != 0xffff){
+          if (packet_attr_slotframe != link->slotframe_handle) {
+            return NULL;
+          }
+          if(backup_link != NULL && packet_attr_slotframe != backup_link->slotframe_handle) {
+            /* Filter out by backup link: prefer receiving packets from another slotframe to sending packets */
+            return NULL;
+          }
         }
         if(packet_attr_timeslot != 0xffff && packet_attr_timeslot != link->timeslot) {
           return NULL;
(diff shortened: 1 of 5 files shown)


simonduq:
Can you explain exactly what happens?
95% sounds like a collision with the RDV slot with default slotframe length of 17, right?

os/net/mac/tsch/tsch-queue.c
```diff
@@ -421,8 +422,14 @@ tsch_queue_get_packet_for_nbr(const struct tsch_neighbor *n, struct tsch_link *l
 #if TSCH_WITH_LINK_SELECTOR
         int packet_attr_slotframe = queuebuf_attr(n->tx_array[get_index]->qb, PACKETBUF_ATTR_TSCH_SLOTFRAME);
         int packet_attr_timeslot = queuebuf_attr(n->tx_array[get_index]->qb, PACKETBUF_ATTR_TSCH_TIMESLOT);
-        if(packet_attr_slotframe != 0xffff && packet_attr_slotframe != link->slotframe_handle) {
-          return NULL;
+        if(packet_attr_slotframe != 0xffff){
+          if (packet_attr_slotframe != link->slotframe_handle) {
+            return NULL;
+          }
+          if(backup_link != NULL && packet_attr_slotframe != backup_link->slotframe_handle) {
```
alexrayne, line 429: why not place this `if` in #if/#endif config condition? This allow easy backoff to standart behaviour?

also there is possible introduce new link option for "strong" rx-links, that are win vs tx?

alexrayne (commented):
just a trivial question

os/net/mac/tsch/tsch-queue.c
atiselsts: I would rather not have yet another configurable option.

atiselsts:
@simonduq as you know, the different slotframes in Orchestra occasionally conflict by design.

For example, with the default settings of unicast slotframe size 17, default common 31, EB slotframe 397 slots. The unicast slotframe conflicts with the default common slotframe every 31 active unicast slots, and with the EB every 397 slots active unicast slots. If node A has a queued packet assigned to the unicast slotframe to send, and node B has a packet assigned to the default common slotframe, they will miss each others packets.

simonduq:
> If node A has a queued packet assigned to the unicast slotframe to send, and node B has a packet assigned to the default common slotframe, they will miss each others packets.

Ah I see, B is also busy sending (I thought you had a scenario in mind where it listening on the wrong slotframe; also possible but almost not in the default settings).

simonduq:
So, I think I'm following now, and I agree it would be nice to address this!

On the default Orchestra config, isn't what you propose here identical in behavior to simply ignoring the "tx takes precedence over rx" rule, and simply always taking the lowest slotframe handle?

I have to admit I never really liked the backup link thing (I know, it was me adding it in the first place!). Always felt like a bit of a hack, with limitations, for example there's only one single backup link. I'd be happy if we can avoid it spreading more and more in the codebase.

atiselsts:
> On the default Orchestra config, isn't what you propose here identical in behavior to simply ignoring the "tx takes precedence over rx" rule, and simply always taking the lowest slotframe handle?

For the default Contiki-NG config the behaviour would the same (I guess), but not in the general case.

Anyway, I will close this PR. I did a closer inspection of whats actually going on, using TSCH-Sim. It turned out that most of the PDR drops are coming from a different problem - as you mention, the backup link is a hack and I think there are a couple of bugs in the implementation. I think its not worth breaking the standard compatibility over what would probably be a very minor PDR gain. If only the initially described problem remains, I saw just one packet lost of around 1500 packets.

Let me describe the more severe problem. There are two issues that I found using the simulator. First is that backup links are not prioritized by slotframe handle. Second is that if the best link has no packets to send, the final decision to use either of those also does not take into account the slotframe handle. I have implemented a fix which actually lets the regression test pass even when the problem described initially remains as was before.

simonduq:
Wow, fantastic. Sounds like TSCH-Sim is a really useful tool BTW :)

simonduq:
So, where's the new PR? :)

PR #1413, opened 2020-10-14
SimpleLink and CC26x0-CC13x0 platforms: timining adjustments to make SCH more usable on SimpleLink

atiselsts:
This PR implements a number of small timing changes to improve TSCH operation on SimpleLink. Some outstanding issues remain, for example CC1352P1 support.

Changes:
* Increase guard time. This is needed for the SimpleLink platform.
* SimpleLink and CC26x0-CC13x0: change the slot timing to make the transmission happen at a later point of time, in order to accomodate for the increased guard time on SimpleLink and remain interoperable. 3800 microseconds are selected to be compatible with CC1200. Note that CC1200 and the newer radios are still not interoperable, as the slot timing is different.
* Fix IEEE mode timestamp offset, detected to be incorrect when running IEEE mode on CC13x2 devices.


arch/cpu/simplelink-cc13xx-cc26xx/rf/ieee-mode.c
```diff
@@ -105,7 +105,11 @@
 #define TIMEOUT_ENTER_RX_WAIT   (RTIMER_SECOND >> 10)
 /*---------------------------------------------------------------------------*/
 /* XXX: don't know what exactly is this */
```
g-oikonomou, line 107: Do we still need that XXX here? 😁

g-oikonomou (approved):

PR #1415, opened 2020-10-14
Tidy up tunslip6 usage and help messages

g-oikonomou:
This PR changes tunslip6's help message the way I think #1396 should have been changing it.

It also tidies up the usage message such that it too documents everything, in the same order as `-h`

### DIFF
--- a/tools/serial-io/tunslip6.c
+++ b/tools/serial-io/tunslip6.c
@@ -848,14 +848,21 @@ fprintf(stderr," -B baudrate    9600,19200,38400,57600,115200 (default),230400,4
 #else
 fprintf(stderr," -B baudrate    9600,19200,38400,57600,115200 (default),230400\n");
 #endif
-fprintf(stderr," -H             Hardware CTS/RTS flow control (default disabled)\n");
-fprintf(stderr," -I             Inquire IP address\n");
-fprintf(stderr," -X             Software XON/XOFF flow control (default disabled)\n");
-fprintf(stderr," -L             Log output format (adds time stamps)\n");
-fprintf(stderr," -s siodev      Serial device (default /dev/ttyUSB0)\n");
-fprintf(stderr," -M             Interface MTU (default and min: 1280)\n");
-fprintf(stderr," -T             Make tap interface (default is tun interface)\n");
-fprintf(stderr," -t tundev      Name of interface (default tap0 or tun0)\n");
+fprintf(stderr, " -C channel     Use channel\n");
+fprintf(stderr, " -i panID       Use panID\n");
+fprintf(stderr, " -R             Reset border router\n");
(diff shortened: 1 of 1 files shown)


tools/serial-io/tunslip6.c
```diff
@@ -848,14 +848,15 @@ fprintf(stderr," -B baudrate    9600,19200,38400,57600,115200 (default),230400,4
...
+fprintf(stderr, " -H             Hardware CTS/RTS flow control (default disabled)\n");
+fprintf(stderr, " -I             Inquire IP address\n");
+fprintf(stderr, " -X             Software XON/XOFF flow control (default disabled)\n");
+fprintf(stderr, " -L             Log output format (adds time stamps)\n");
+fprintf(stderr, " -s siodev      Serial device (default /dev/ttyUSB0)\n");
+fprintf(stderr, " -M             Interface MTU (default and min: 1280)\n");
+fprintf(stderr, " -T             Make tap interface (default is tun interface)\n");
+fprintf(stderr, " -t tundev      Name of interface (default tap0 or tun0)\n");
 #ifdef __APPLE__
 fprintf(stderr," -v level       Verbosity level\n");
```
simonduq, line 861: `-v` only shows here if `__APPLE__` is defined, but is always listed in the usage

g-oikonomou: On Linux we can do `-v N` as well as `-v`. On macOS we have to provide the level, `-v` without a level throws an error.

`[-v level]` is correct in the usage IMHO. Unless we want to change to `[-v [level]]` for non-macOS

g-oikonomou: So if there is a problem here, I would argue it's with the docs for `-d`, not `-v`

simonduq: Ah, I got fooled by the diff. You're correct.
What's the problem with `-d`?

g-oikonomou: The help suggests that the `basedelay` param is optional, which is only the case for Linux. Same problem on macOS, we can't just pass `-d`, the param is mandatory. Basically optional params are a GNU thing.

simonduq: OK I'll approve, up to you whether you want to fix in this PR.

simonduq (approved):

tools/serial-io/tunslip6.c
g-oikonomou: * I have fixed the code style across the entire help text.
* I have fixed the help text for `-d` on macOS
* I have differentiated the usage text for macOS vs other OSs. In the latter case we now show `[-v [level]]`, similar for `-d`

simonduq: Awesome 👍

simonduq (approved):

PR #1417, opened 2020-10-16
TSCH: fix backup link selection 

atiselsts:
There are two issues fixed by this PR:
1. in backup link selection, they are not prioritized by slotframe handle.
2. if the best link has no packets to send, the final decision to use either the best link of the backup link also does not take into account the slotframe handle. 

This lets the regression test pass with almost 100% packet ACK rate. The only loss appears to be very occasional (1 packet of 1500), because of the conflict by design between multiple Orchestra slotframes.


simonduq (commented):
Simple yet great improvement! Only minor comments :)

simonduq (approved):

PR #1420, opened 2020-10-22
GH Actions: keep docker image with tag 'latest' up-to-date

simonduq:


.github/workflows/build.yml
```diff
@@ -80,11 +80,15 @@ jobs:
         MERGE_BRANCH_REF=$(echo ${{ github.ref }} | sed -e 's|refs/heads/||g')
         echo This is a build for branch $MERGE_BRANCH_REF and it updates the docker container
         echo Push images to Dockerhub
-        echo Will push $DOCKER_IMG as well as $DOCKER_BASE_IMG:$MERGE_BRANCH_REF
         echo ${{ secrets.DOCKERHUB_PASSWD }} | docker login --username contiker --password-stdin
-        docker tag $DOCKER_IMG $DOCKER_BASE_IMG:$MERGE_BRANCH_REF
         docker push $DOCKER_IMG
+        docker tag $DOCKER_IMG $DOCKER_BASE_IMG:$MERGE_BRANCH_REF
```
g-oikonomou, line 85: Looks good, but why are we re-ordering the tag and the push here?

simonduq: Was to get more consistent in the order: push img, tag branch, push branch, tag latest, push latest.
I don't feel strongly at all; shall I revert the re-ordering?

simonduq: It's tagged when we build it with `docker build tools/docker -t $DOCKER_IMG`.
Or directly pulled as `$DOCKER_IMG` if already exists on dockerhub.

I think we should ideally only run stage "Push images to dockerhub" if `env.DOCKER_NEED_BUILD == '1'` but maybe let's fix that as a separate PR so we make sure `latest` gets updated first?

g-oikonomou: > I think we should ideally only run stage "Push images to dockerhub" if env.DOCKER_NEED_BUILD == '1' but maybe let's fix that as a separate PR so we make sure latest gets updated first?

I couldn't agree more.

g-oikonomou (approved):

PR #1422, opened 2020-10-22
Extend documentation of the radio API

g-oikonomou:
This pull recommends extensive radio API documentation improvements based on the proposal in #197 

Once we adopt these semantics for our radio API (with or without changes), we'll need to go back to our TSCH and CSMA implementations to make sure that the radio API is being used consistently between the two. And #197 renamed to reflect that.

## Proposal
Here is what I am proposing to improve the situation without making any changes to the API. We will almost certainly have to modify some radio drivers and also CSMA. Perhaps some changes required for TSCH here and there, but not 100% certain.

* `transmit()`: should be kept as simple as possible:
  * It should not try to guess what the MAC layer may need to do next. It may, but should not be expected to put the radio in RX after TX. What I am proposing is that the MAC layer make no assumptions about the state of the radio after TX. If it wants to revert to RX it should explicitly call `on()`. If it needs to power down the radio it should call `off()`. In the case of some drivers this will prevent unnecessary time spent in transition to RX.
  * It should expect to be called from a powered down state. In this case it should power up and go from the disabled state directly to TX, bypassing the RX state. The only exception is if the driver has to first perform CCA, unless we get rid of this as above.
* `on()` should be conceptually equivalent to "Do everything required to enter RX from any state". If the radio is on but idle it should simply perform a transition to RX. If powered down, it should first power on. The radio driver will not expect a call to `on()` before a call to `transmit()`.

Furthermore, I am proposing clarifications about the following:

* `init()`: Should only bother with one-off initialisation of the radio. Anything that does not persist across on/off cycles should be done in `on()` instead. `init()` may, but does not have to put the radio in RX mode.
* `off()` must power down the radio, but must not discard any previously received frames
* `on()` must not discard any previously received frames
* `channel_clear()` must be preceded by a call to `on()`, or fail with default return value if called while the radio is not in RX.
* `receiving_packet()` must be preceded by a call to `on()`, or simply return 0 without making any attempt of entering RX.
* `pending_packet()` may be called while the radio is powered down. In this case, the driver must be able to return the correct value _without_ manipulating the radio hardware.
* `read()`: Ditto
* `prepare()` may be called while the radio is powered down.

### DIFF
--- a/os/dev/radio.h
+++ b/os/dev/radio.h
@@ -62,69 +62,126 @@
 /**
  * Each radio has a set of parameters that designate the current
  * configuration and state of the radio. Parameters can either have
- * values of type radio_value_t, or, when this type is insufficient, a
+ * values of type `radio_value_t`, or, when this type is insufficient, a
  * generic object that is specified by a memory pointer and the size
  * of the object.
  *
- * The radio_value_t type is set to an integer type that can hold most
+ * The `radio_value_t` type is set to an integer type that can hold most
  * values used to configure the radio, and is therefore the most
  * common type used for a parameter. Certain parameters require
- * objects of a considerably larger size than radio_value_t, however,
+ * objects of a considerably larger size than `radio_value_t`, however,
  * and in these cases the documentation below for the parameter will
  * indicate this.
  *
  * All radio parameters that can vary during runtime are prefixed by
- * "RADIO_PARAM", whereas those "parameters" that are guaranteed to
- * remain immutable are prefixed by "RADIO_CONST". Each mutable
+ * `RADIO_PARAM_`, whereas those "parameters" that are guaranteed to
+ * remain immutable are prefixed by `RADIO_CONST_`. Each mutable
  * parameter has a set of valid parameter values. When attempting to
  * set a parameter to an invalid value, the radio will return
- * RADIO_RESULT_INVALID_VALUE.
+ * `RADIO_RESULT_INVALID_VALUE`.
  *
  * Some radios support only a subset of the defined radio parameters.
  * When trying to set or get such an unsupported parameter, the radio
- * will return RADIO_RESULT_NOT_SUPPORTED.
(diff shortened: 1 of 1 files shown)


simonduq:
Great 👍
Two comments:
* Let's not get rid of optional CCA before Tx, or else we'll be unable to use hardware-supported CCA
* How about we enforce that `init()` does not ever turn on the radio? It's a bit of the same argument as for not having `transmit` restore RX mode implicitly. Can save power in some cases.

g-oikonomou:
> How about we enforce that init() does not ever turn on the radio? It's a bit of the same argument as for not having transmit restore RX mode implicitly. Can save power in some cases.

By "turn" on, do you mean put in RX?

simonduq:
> By "turn" on, do you mean put in RX?

Yes sorry that's what I meant :)

g-oikonomou:
Thanks Simon. I just had a look at various docs, as far as I can see CC2538, CC26xx/CC13xx and nRF52840 all have some "powered on but otherwise idle state". So we want to enforce that init puts the radio in exactly that state?

simonduq:
I think that would make sense, but maybe we don't enforce it yet, only something like you did with what state you end up with after `transmit()`. Like "ìnit()` may, but should not be expected to put the radio `on()`."

g-oikonomou:
OK thanks yes that makes sense. How about your comment wrt the CCA before TX? Are you point to any specific place in the doc?

simonduq:
> OK thanks yes that makes sense. How about your comment wrt the CCA before TX? Are you point to any specific place in the doc?

The doc is fine, this was only about the PR description: "unless we get rid of this as above". I think we shouldn't get rid of the ability to let the driver perform CCA.

g-oikonomou:
OK gotcha thanks

g-oikonomou:
I've made some minor modifications to reflect what we've discussed here.

This is ready for review/merge.

simonduq (approved):

PR #1442, opened 2020-11-20
Update Github Actions: replace set-env with environment files

jeppenodgaard:
Github Actions seem to fail due to https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


g-oikonomou:
We were discussing exactly that 10' ago. Thanks! This looks like the correct approach, will review asap

g-oikonomou (approved):

PR #1441, opened 2020-11-20
Update submodule cc2538-bsl to latest commit

jeppenodgaard:
Adds support for python that ships with newest Ubuntu LTS and ability to flash cc1312r1

PR #1448, opened 2020-11-23
BSL: Change PYTHON value to python3

### LINKED ISSUE #1444: Build problems for openmote-cc2538 (inconsistent Python version)
Hello,

I tried to build `hello-world.c` (located in `examples/hello-world`) for Openmote CC2538 Rev. A1.

Step to reproduce (assuming that the working directory is `examples/hello-world`) :
```bash
make TARGET=openmote BOARD=openmote-cc2538 BOARD_REVISION=REV_A1 savetarget
make hello-world.upload TARGET=openmote BOARD_REVISION=REV_A1
```

Output :

```python
  [...]
  CC        ../../arch/cpu/cc2538/./ieee-addr.c
  CC        ../../arch/cpu/cc2538/cc2538.lds
  CC        ../../arch/cpu/cc2538/./startup-gcc.c
  CC        hello-world.c
  LD        build/openmote/openmote-cc2538/hello-world.elf
  OBJCOPY   build/openmote/openmote-cc2538/hello-world.elf --> build/openmote/openmote-cc2538/hello-world.bin
python ../../tools/cc2538-bsl/cc2538-bsl.py --bootloader-invert-lines -e -w -v -b 450000 -p /dev/ttyUSB0 -a 0x00202000 build/openmote/openmote-cc2538/hello-world.bin
  File "../../tools/cc2538-bsl/cc2538-bsl.py", line 82
    print(message, end=attr, file=sys.stderr)
                      ^
SyntaxError: invalid syntax
../../arch/platform/openmote/Makefile.openmote:44: recipe for target 'hello-world.upload' failed
make: *** [hello-world.upload] Error 1
rm hello-world.o build/openmote/openmote-cc2538/obj/startup-gcc.o
```

I went to see the [JelmerT cc2538-bsl GitHub repository](https://github.com/JelmerT/cc2538-bsl/commits/master) and I found that `tools/cc2538-bsl/cc2538-bsl.py` now only supports **Python 3** according to recent commits.
However it seems that `contiki-ng/arch/platform/openmote/Makefile.openmote` uses **Python 2** (line 31) :
```
PYTHON = python
```

I changed the above line to
```
PYTHON = python3
```

and installed `pyserial` with `pip3` and now it works.

Is this the best way to alleviate the problem?

Thank you in advance.

Best regards,

Guillaume Nibert

jeppenodgaard:
Should fix #1444


g-oikonomou:
Thanks.

We are excluding #1441 from the upcoming release (but keeping it in develop), it looks like there is a little bit of work to be done in order to transition cleanly to Python 3, but it's good that we've started this already.

jeppenodgaard:
I consider removing the `python ...` from the BSL lines because the script uses shebang to choose the correct interpreter. Would this be a better approach?

g-oikonomou:
So you are saying we should make the script executable and run it in itself?

jeppenodgaard:
That is my suggestion yes. The file already seems to have executable permissions.
This allows removing the PYTHON variable and the need to maintain it.
I'm not sure if there's any cons to this approach..

g-oikonomou:
@atiselsts how do you feel about this?

atiselsts:
I prefer the shebang approach (just learned that is the name of the thing!). It reduces the number of times python version is referenced in the codebase and is consistent with the approach already used in older platforms (Sky and Z1).

g-oikonomou:
OK, I too would be happy with that. @jeppenodgaard do you have time to work on it?

jeppenodgaard:
@g-oikonomou Yes. I should be able to work on it tomorrow.

jeppenodgaard:
Ready for review

atiselsts (approved):

PR #1461, opened 2020-11-29
Updated arm-gcc to 9

Yagoor:
# Description

In this PR we are updating our old arm-gcc 5 to the latest version available at the moment arm-gcc 9

# Consequences

I expect this PR to break a few things. 
Example: The mqtt-client was not compiling with the following: "make BOARD=firefly DEFINES=MQTT_CONF_VERSION=5 TARGET=zoul"
This happened due to a new warning introduced on the GCC 9. "address-of-packed-member"

I expect that new warning will appear and they'll be fixed.

# Reason

In order to support the new arm boards we need to update the arm-gcc.
The new nRF53 uses the Cortex-M33 that is not suppported by the current arm-gcc.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,7 +67,7 @@ jobs:
         echo $DOCKER_IMG does not exist on dockerhub or pull failed
         echo This is normal for PR builds and for builds on forks
         echo Building from dockerfile
-        docker build tools/docker -t $DOCKER_IMG
+        docker build tools/docker -t $DOCKER_IMG --no-cache --pull
 
     # If i) the previous step built an image and ii) we are on the main
     # contiki-ng repo and iii) this is a push (merge commit) to one of the
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -52,9 +52,9 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E03280
   && apt-get -qq clean
 
 # Install ARM toolchain
-RUN wget -nv https://launchpad.net/gcc-arm-embedded/5.0/5-2015-q4-major/+download/gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2 && \
-  tar xjf gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2 -C /tmp/ && \
-  cp -f -r /tmp/gcc-arm-none-eabi-5_2-2015q4/* /usr/local/ && \
+RUN wget -nv https://github.com/Yagoor/gcc-arm-none-eabi/releases/download/9-2020-q2-update/gcc-arm-none-eabi-9-2020-q4-major-i386-linux.tar.bz2 && \
+  tar xjf gcc-arm-none-eabi-9-2020-q4-major-i386-linux.tar.bz2 -C /tmp/ && \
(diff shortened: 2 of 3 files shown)


atiselsts:
Thanks for this.

Can you use the build-all script to check which other programs fail to compile when the compiler is upgraded?

To do that for all ARM platforms, go to `tests/compile-all` and run this:

    PLATFORMS="cc26x0-cc13x0 simplelink cc2538dk openmote zoul nrf52840 nrf52dk" ./build.sh

I suggest submitting any fixes in a separate PR.

Yagoor:
Hi @atiselsts. I'll run it and fix the issues, if any.

The fix in this PR was necessary because it's part of the gate checks, that is why I had to push it in this PR. :) 

Thank you.

Yagoor:
HI @atiselsts.

I ran the compile-all and saw no problem related to the compiler.
Most of the problems is memory limitation that also happens with arm gcc 5. :)

.github/workflows/build.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         echo $DOCKER_IMG does not exist on dockerhub or pull failed
         echo This is normal for PR builds and for builds on forks
         echo Building from dockerfile
-        docker build tools/docker -t $DOCKER_IMG
+        docker build tools/docker -t $DOCKER_IMG --no-cache --pull
```
atiselsts, line 70: Why do we need to add these parameters? Judging by the comment above, this code is hit when a pull failed. Why try `--pull` again?

tools/docker/Dockerfile
```diff
@@ -52,9 +52,9 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E03280
   && apt-get -qq clean
 
 # Install ARM toolchain
-RUN wget -nv https://launchpad.net/gcc-arm-embedded/5.0/5-2015-q4-major/+download/gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2 && \
-  tar xjf gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2 -C /tmp/ && \
-  cp -f -r /tmp/gcc-arm-none-eabi-5_2-2015q4/* /usr/local/ && \
+RUN wget -nv https://github.com/Yagoor/gcc-arm-none-eabi/releases/download/9-2020-q2-update/gcc-arm-none-eabi-9-2020-q4-major-i386-linux.tar.bz2 && \
```
atiselsts, line 55: Why not use the ARM site here instead your personal github?

atiselsts (commented):
I also tried it, and think we could merge it. Left just a couple comments on the PR.

.github/workflows/build.yml
Yagoor: I just wanted to ensure no caching and force a pull from the base image too.
The pull was to enforce to get the latest image from the dockerhub.
The nocaching was to force the layer with the apt-get install to be ran. The layer will only run if the command in it or before was changed, however, this can leave the image with outdated libs and binaries.

This just ensures that we get the latest of everything. It might be an overkill but I see no harm in enforcing it.

tools/docker/Dockerfile
Yagoor: ARM doesn't support i386 anymore.
I had to compile it locally.
In this github the files to compile are available using docker. (Documentation is not existent.)

atiselsts (approved):

PR #1462, opened 2020-11-29
Warn when using old GCC versions with platform nRF52840

### LINKED ISSUE #1449: Execution problems with old compilers on nrf52840-dongle
Hello, 
I had some issue in executing two examples (specifically, rpl-border-router and example-coap-server) on two nrf52840 dongles compiled on Ubuntu 16.04. The distro ships with an old compiler arm-none-eabi-gcc (15:4.9.3+svn231177-1) 4.9.3.
In particular the code run, however, it seemed to have issues in writing into the NRF_RADIO structure. For instance the channel on the example-coap-server was not properly set. 

Upgrading to a newer compiler,  arm-none-eabi-gcc 5.3.1 20160307 in my case, solved the issue.

alexstanoev:
Older GCC versions generate broken executables for the nRF52840 platform. 
The resulting binaries appear functional but are unable to send or receive any packets.

#1449 has seen this happen with arm-none-eabi-gcc 4.9.3, and I can reproduce it with arm-none-eabi-gcc 5.2.1.
This PR warns anyone using a version of gcc older than 6.x with the nRF52840 platform and prompts them to update their toolchain.

Closes #1449.

### DIFF
--- a/Makefile.gcc
+++ b/Makefile.gcc
@@ -16,6 +16,17 @@ ifeq ("$(CC)","arm-none-eabi-gcc")
   endif
 endif
 
+# Warn if using anything older than version 6.x of arm-none-eabi-gcc on nrf52840
+ifeq ("$(TARGET)","nrf52840")
+  ifeq ("$(CC)","arm-none-eabi-gcc")
+    ifeq ($(shell test $(GCC_MAJOR_VERSION) -lt 6; echo $$?),0)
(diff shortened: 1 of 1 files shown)


atiselsts:
We need to change the Wiki: https://github.com/contiki-ng/contiki-ng/wiki/Toolchain-installation-on-Linux#install-arm-compiler

Othewise, looks fine.

atiselsts (approved):

PR #1463, opened 2020-12-01
Fix overflow bug in the HeapMem module

nvt:
Add overflow guards to avoid having very large allocations appear as if they succeeded.


nfi (approved):
Looks good!

PR #1467, opened 2020-12-02
add C++ support to the Contiki-NG build system 

atiselsts:


Makefile.include
```diff
@@ -319,6 +327,7 @@ endif
 ifneq ($(MAKECMDGOALS),clean)
 -include ${addprefix $(OBJECTDIR)/,$(CONTIKI_SOURCEFILES:.c=.d) \
                                    $(PROJECT_SOURCEFILES:.c=.d)}
+-include ${addprefix $(OBJECTDIR)/,$(PROJECT_SOURCEFILES:.cpp=.d)}
```
g-oikonomou, line 330: So we expect Contiki-NG to stay C only, and we only allow project source files to be CPP.

No problem with me at all, for now at least. Perhaps in the future we can revisit if required!

g-oikonomou:
👌 Sweeeet! Looks fairly straightforward too.

g-oikonomou (approved):

PR #1469, opened 2020-12-03
nRF SoCs port

Yagoor:
# Description

In this PR there is a proposal for a nRF single port, the TARGET nrf is introduced for this porpuse.

The port is being tested on the nRF52840 DK and the nRF5340 DK therefore only these two BOARDS were introduced.


Yagoor:
@g-oikonomou: Could you please put Alex in this Pull Request? 

I believe we can have a single port for all boards and CPU's from the Nordic, but I remember you mentioned he has some experience, maybe he can assist me on getting this port and maybe we can work together into getting all boards and cpu's under the same port.

alexstanoev:
Thanks for this!

Support for the nRF5340 and a single port for all Nordic platforms are both things we're really interested in.

One problem with having a single port I can see is that our current nRF52840 support relies on some bits that are exclusive to the nRF5 SDK (e.g. `app_usbd`).

To go around this we could:
- Conditionally compile bits of the nRF5 SDK when building for the nRF52840 or other nRF52 parts with USB.
- Use the existing cdc-acm implementation in `arch/cpu/arm/common/usb/` and implement all the hardware-specific bits.
- Have an `nrf52` cpu/platform backed by the nRF5 SDK, as well as `nrf53` backed solely by `nrfx`.

In any case USB is not a priority for the nRF53 (and might not be accessible by the network core?), so this is purely if we want to combine the current nRF52 support into one platform.

A couple of questions:

- I'm not sure if the PPI differences are abstracted well enough to not require a different version of the radio driver.
Have you had a change to take a look at how portable the current radio implementation is to the nRF53?

- Are you planning on this being a port exclusively to the network core on the nRF5340, or will this be a compile-time decision?


I have a bunch of nRF5340 PDKs with me and would be happy to assist with this work!

Yagoor:
Thank you for the comments.

I don't think we need the app_usbd, we should be able to have this functionality using the nrfx drivers.
The nrfx abstracts a lot based on the cpu, if the CPU supports it, we should have no problem.
In the nrfx documentation there is a table with compatibilities this should help us in the future.

I haven't looked into the radio implementation, yet. I'm trying to get the board up and running with traces over USB in order to progress, it's hard to progress without being able to trace things. :) 

I'm only using the application core for now for simplicity. We need to have a firmware in the application core in order to boot the network core, I haven't reached this point yet. But I think we should let the user decide if he want low power or performance. Even better, in the best case we have the host running on the application and the controller running on the network. (This would be amazing.)

I'll keep pushing changes into this branch as I progress and I'll let you know once I have it in a workable stage than we can take it from there. I just wanted to signal to you that I'm working on it and to get insights since you have more experience. 

Thank you again. :)

g-oikonomou:
That looks really cool. One thing I can say already is, the CMSIS update can go in a separate PR which we can test and merge while this is being developed.

Yagoor:
Hi @alexstanoev.

I believe to have a basic nRF target with the nRF52 PDK and nRF53 PDK ports.
No radio port was done yet, but it would be great to get feedback on the structure that I adopted.

Portions of the code were taken from the two current implementations nRF52840 and nRF52832.
However, a lot also had to be changed since the sdk has changed as well.

I'll start working on the radio ports.

alexstanoev:
Hi @Yagoor 

I just tried this out on an nRF5340 and it does compile, run and hit `lpm_drop()` :)

The overall structure seems fine, I have a few comments:

- CMSIS as a submodule is a great idea.
- I'd put `nrfx` under a `lib` subfolder to stay consistent with the structure of most other NG ports (nrf52840, simplelink, cc26x-cc13x0).
- For the comment you have in `linker-symbols.c`, I'd suggest just copying the linker scripts to a `ld` folder and modifying them with the stack checker symbols the same way nrf52840 does.
- The semantics of the RNG implementation when using the hardware RNG are problematic, since a user could call `random_rand()` in a tight loop and get non-random values back. I would suggest seeding the software RNG once from the hardware RNG, the way the nrf52840 port does it.
- Even though nrfx does abstract the USB hardware bits, we still need some kind of implementation of the higher layer CDC-ACM class, but this is a low priority.
- (Minor one) pdk stands for Preview devkit, so the nrf52 target would technically just be dk.

I'd be very interested to see how you get along with the radio port, feel free to send any questions here or on gitter!

Yagoor:
Hi @alexstanoev,

1 - I managed to port the radio ALMOST one-to-one. I haven't tested it, but my next goal will be to get some traces out of the board.
2 - I'll move it to a lib directory as mentioned for consistency
3 - I'm aware of the PDK, but I only have PDK boards and I know that the GPIO setup changes between PDK and DK, that is why I made it a different board. :) 
4 - Implemented a basic application core app that just forwards a few GPIO's and starts the network core.

I'll see what is the best way and easier way to get traces out.

Would you be able to create the DK board and test the radio communication on your setup?
If not, I'll do it after I have traces. :)

alexstanoev:
The latest changes look good!

I'll be able to test the radio communication later this week, I'll let you know how it goes :)

Yagoor:
Thank you,
I'll do my best to get some traces working.
I suggest that we add on the documentation a list of todo's and one should be to remove the nrfx dependencies and only use the hal layer. This should allow us to have a smaller code footprint and maybe better performance since the nrfx is very heavy and we don't really need it's functionalities.

Yagoor:
Hi @alexstanoev,

I'm having troubles porting the radio.
I managed to copy the nrf52840 implementation and make it compile however I'm having trouble with timing.
I think it's related to the PPI timing stamping.
I took a look into the documentation and apparently the high precision timer is only running when the radio is active (not sleeping)
Would you be able to assist in this?
The radio driver is also not working on the nrf 52 board which could be some driver that changed.

arch/cpu/nrf/nrf-ieee-driver-arch.c
```diff
@@ -0,0 +1,1008 @@
...
+
+  timestamps.sfd = 0;
+  timestamps.framestart = 0;
+  timestamps.end = 0;
+  timestamps.mpdu_duration = 0;
+  timestamps.phr = 0;
+
+  /* Request the HF clock */
+  nrf_clock_event_clear(NRF_CLOCK, NRF_CLOCK_EVENT_HFCLKSTARTED);
+  nrf_clock_task_trigger(NRF_CLOCK, NRF_CLOCK_TASK_HFCLKSTART);
```
alexstanoev, line 515: You need to wait for the clock to start up - see the recent fix in #1478.
Could this be the source of your timer issues @Yagoor?

alexstanoev:
Hi @Yagoor,

I just tested this on both an nRF5340 PDK and an nRF52840 DK and I can ping both of them through a CSMA border router on an nRF52840 dongle (running the current nrf52840 port) :)

There are some issues still, as the latency is around 3-4 times worse compared to an nRF52840 DK running the nrf52840 target, and the nRF53 board stops replying after a while. I haven't had time to look into this in detail yet.

I copied over `populate_link_address()` from the nrf52840 port so the boards don't start up with a zero address, and applied the changes from #1478:
```
   nrf_clock_event_clear(NRF_CLOCK, NRF_CLOCK_EVENT_HFCLKSTARTED);
   nrf_clock_task_trigger(NRF_CLOCK, NRF_CLOCK_TASK_HFCLKSTART);
+  while(!nrf_clock_event_check(NRF_CLOCK, NRF_CLOCK_EVENT_HFCLKSTARTED));
+  nrf_clock_event_clear(NRF_CLOCK, NRF_CLOCK_EVENT_HFCLKSTARTED);
```

Yagoor:
Hi @alexstanoev,

I'll give a try with those changes.
However, I was trying TSCH and not CSMA.

Does the nrf52840 port supports the TSCH?

alexstanoev:
Hi @Yagoor, did you manage to try these changes with CSMA, or do you want me to try something out on my boards?

The nRF52840 port does support TSCH, but CSMA would be the easier MAC to get running on a new port first.

Yagoor:
Hi @alexstanoev,

Thank you for the reply.
I'll try CSMA and verify why you have the reboot experience.
I have a hunch on why.
I'll use the nrf52840 port on one side and the nrf53 port on the other.
This will ensure that one node is stable which will make easier to debug.

Yagoor:
Hi @alexstanoev,

I fixed one issue in the radio driver. I ported the ppi wrong from the nrf52840 driver.
However I can't figure out why the port just hangs.
I debugged it to being a HardFault but I can't figure out why the HardFault happens.
I know that the HardFault does not happen if I enabled the debugger flags (-g)

I took the hard fault handler from the nrf-sdk and added to this port, the output is as below.

```bash
[INFO: NRF HARDFAULT] HARD FAULT at 0x01002C2E
[INFO: NRF HARDFAULT]   R0:  0x21001E22  R1:  0x21001D6C  R2:  0x00000000  R3:  0xAC3C25E5
[INFO: NRF HARDFAULT]   R12: 0x21001E22  LR:  0x01002C2F  PSR: 0x69000000
[INFO: NRF HARDFAULT] Cause: Data bus error (PC value stacked for the exception return points to the instruction that caused the fault).
[INFO: NRF HARDFAULT] Bus Fault Address: 0x253CACF4
```

We can see that the Hard Fault happened at 0x01002C2E which point to the ctimer process around the call back call.
Now knowing that it is a Hard Fault would you happen to know a better way to debug it? 

```c
PROCESS(ctimer_process, "Ctimer process");
PROCESS_THREAD(ctimer_process, ev, data)
{
  struct ctimer *c;
  PROCESS_BEGIN();

  for(c = list_head(ctimer_list); c != NULL; c = c->next) {
    etimer_set(&c->etimer, c->etimer.timer.interval);
  }
  initialized = 1;

  while(1) {
    PROCESS_YIELD_UNTIL(ev == PROCESS_EVENT_TIMER);
    for(c = list_head(ctimer_list); c != NULL; c = c->next) {
      if(&c->etimer == data) {
        list_remove(ctimer_list, c);
        PROCESS_CONTEXT_BEGIN(c->p);
        if(c->f != NULL) {
          c->f(c->ptr);
        }
        PROCESS_CONTEXT_END(c->p);
        break;
      }
    }
  }
  PROCESS_END();
}
```

alexstanoev:
Hi @Yagoor,

Great work with the port so far!

The crash you're describing sounds like memory corruption to me, but that's a complete guess.
Which combination of examples are you getting the hardfault on? I can try to reproduce it here.

I tried running `platform-specific/nrf/empty-application` on the application core and `hello-world` on the network core on a nRF5340-DK (not PDK; using gcc 9.3.1) and everything seems to be working!
I've had ping running through a nRF52840 border router for at least 15 minutes without crashing.

This is without any changes done to your branch, and I don't see `-g` being defined by default.

Yagoor:
Hi @alexstanoev,

Do you have the UART logs out from the nrf53 board?
I have extended the HardFault handler to in fact reboot the board, so this might explain why you think it's working fine.

I have been using the same examples you mentioned above.

If you don't have FTDI cables, you can comment out the 
```c
NVIC_SystemReset();
```
from the hardfault-handler.c

I also think it's some memory corruption but I've been looking into the code and cannot find anything wrong there. At least not obviously wrong :)

alexstanoev:
Hi @Yagoor,

Yes I had the UART up via the JLink, I could see it printing hello world without resetting, and a reset would have caused it to drop off the network and lose packets, neither of which happened.

Are you using the same gcc 9.3.1 version?
I'm also assuming you're using a PDK instead of a DK, I'll try on a PDK later just in case.

Yagoor:
Hi @alexstanoev,

That is correct. I'm using the PDK and not DK.
On the GCC, I'm using the one from docker. GCC 9, I'm not sure on the other sub versions.
I have also tried with GCC 10, same problem there.

alexstanoev:
After a private discussion between myself and Yago we narrowed down the crash to errata on the nRF5340 QKAA-AB0 preview silicon.

I can confirm that d0a1cd2 fixes things for the PDK and works fine on the DK :)

examples/platform-specific/nrf/empty-application/project-conf.h
```diff
@@ -0,0 +1,36 @@
...
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+/*---------------------------------------------------------------------------*/
+#ifndef PROJECT_CONF_H_
+#define PROJECT_CONF_H_
+/*---------------------------------------------------------------------------*/
+#endif /* PROJECT_CONF_H_ */
+/*---------------------------------------------------------------------------*/
```
g-oikonomou, line 36: Does this file actually serve a purpose?

Yagoor: No. This is a leftover.
Initially I was setting the radio driver here but then I moved it to the cpu conf.
I'll remove it. :)

Yagoor: Done :)

alexstanoev:
Hi @Yagoor,

I just had a quick look at TSCH, but the port seems to crash when transmitting.
I've tracked it down to line 624 in nrf-ieee-driver-arch.c:

```
  /* When we reach here we are in state RX. Send a STOP to drop to RXIDLE */
  nrf_radio_task_trigger(NRF_RADIO, NRF_RADIO_TASK_STOP);
  while(nrf_radio_state_get(NRF_RADIO) != NRF_RADIO_STATE_RXIDLE);
```
where the radio seems to stay in the RX state forever until the watchdog kicks the board. 

This is quite odd as nothing seems wrong here, but changing this to:
```
  nrf_radio_task_trigger(NRF_RADIO, NRF_RADIO_TASK_DISABLE);
  while(nrf_radio_state_get(NRF_RADIO) != NRF_RADIO_STATE_DISABLED);
```
does work around that hang.
This definitely needs more investigation and could be related to the next issue.

Doing this, I didn't have much luck with the nRF53 DK as a root, however having an nRF52840 (running the nrf52840 port) board as the root gets it a bit further.
The nRF53 manages to receive EBs and syncs but never actually joins the DAG properly - it has issues receiving packets after syncing, even though the root receives its KA packets.

An interesting observation I made is that adding this:
```
  radio_set_power(false);
  radio_set_power(true);
  configure();
```
to the start of `enter_rx()` completely breaks reception, which it shouldn't do.

Something's not right when transitioning from off to RX, but I haven't been able to track down what yet.
If you add `if(power == 0) return;` to the start of `radio_set_power` then TSCH associates, joins the DAG and seems to stay in sync. This works both when the nrf53 is a client and root.

As an aside, it might be worth looking into anomaly 117 from the rev. 1 errata ([117] RADIO: Changing MODE requires additional configuration).

You should also include the following defines (taken from `nrf52840-def.h` in the nrf52840 port) over to `nrf-def.h` to use hardware TSCH timestamping:
```
#ifndef TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS
#define TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS 1
#define TSCH_CONF_TIMESYNC_REMOVE_JITTER 0
#endif
```

Yagoor:
Hi @alexstanoev,

Thank you for the feedback.

I took a quick look into the state machines from nrf52840 and nrf5340. It seems that there is some small changes in the transitions. It might just be the drawing that confused me. 
I'll investigate more this and update here once I have news. :)

Yagoor:
Hi @alexstanoev,

I don't see any issue in my PDK which means it is related to the radio power register.
When the power register is set to false it resets all the other registers.
I don't have this behaviour which is making my board work.
I'll try to get my hands in a Development Board but since this is a short week in Denmark (We have two holidays), it might take some time.
In the meantime, I'll see if I can figure out which register is the culprit.

alexstanoev:
Hi @Yagoor,

This has been a really tricky issue to debug, but I've managed to narrow it down a bit more.

If you set `TSCH_CONF_RADIO_ON_DURING_TIMESLOT` to 1 then TSCH will work fine.

I saw at least one instance where the radio driver assumes it's on but would actually be turned off by TSCH, on line 823-825 in `tsch-slot-operation.c`:
```
      tsch_radio_off(TSCH_RADIO_CMD_OFF_WITHIN_TIMESLOT);
      if(NETSTACK_RADIO.pending_packet()) {
```

`pending_packet()` ends up calling `nrf_radio_event_check(NRF_RADIO, NRF_RADIO_EVENT_CRCOK)` which won't work, as turning off the radio zeroes all registers. 

I don't know why this works on the nRF52, but I suppose this register remains set even after setting power to 0?

To test this out I made `tsch-slot-operation.c` call `pending_packet()` before turning off the radio, but that didn't seem to fix the issue, so there might be more cases where we access registers that have been zeroed.

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
```diff
@@ -0,0 +1,1065 @@
...
+/*---------------------------------------------------------------------------*/
+/*
+ * Powering off the peripheral will reset all registers to default values
+ * This function here must be called at every power on to set the radio in a
+ * known state
+ */
+static void
+configure(void)
+{
+  nrf_radio_mode_set(NRF_RADIO, NRF_RADIO_MODE_IEEE802154_250KBIT);
```
alexstanoev, line 390: You might want to handle [nRF5340 anomaly 117](https://infocenter.nordicsemi.com/index.jsp?topic=%2Ferrata_nRF5340_Rev1%2FERR%2FnRF5340%2FRev1%2Flatest%2Fanomaly_340_117.html&anchor=anomaly_340_117) here.

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
```diff
@@ -0,0 +1,1065 @@
...
+    *value = 11;
+    return RADIO_RESULT_OK;
+  case RADIO_CONST_CHANNEL_MAX:
+    *value = 26;
+    return RADIO_RESULT_OK;
+  case RADIO_CONST_TXPOWER_MIN:
+    *value = (radio_value_t)RADIO_TXPOWER_TXPOWER_Neg40dBm; /* XXX */
+    return RADIO_RESULT_OK;
+  case RADIO_CONST_TXPOWER_MAX:
+    *value = (radio_value_t)RADIO_TXPOWER_TXPOWER_0dBm; /* XXX */
```
alexstanoev, line 866: Why the XXX comments?
This is also different for the nRF52840 where the maximum TX power is 8dBm.

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
```diff
@@ -0,0 +1,1065 @@
...
+  case RADIO_PARAM_CHANNEL:
+    if(value < NRF_CHANNEL_MIN ||
+       value > NRF_CHANNEL_MAX) {
+      return RADIO_RESULT_INVALID_VALUE;
+    }
+    rf_config.channel = value;
+
+    /* If we are powered on, apply immediately. */
+    if(radio_is_powered()) {
+      set_channel(value);
```
alexstanoev, line 916: You need to power the radio off and on when changing the channel during RX, i.e.:
```     if(radio_is_powered()) {
+      off();
       set_channel(value);
+      on();
```

arch/platform/nrf/Makefile.nrf
```diff
@@ -0,0 +1,20 @@
+ifndef CONTIKI
+  $(error CONTIKI not defined! You must specify where CONTIKI resides!)
+endif
+
+### Board and BSP selection
+BOARD ?= nrf53/pdk/application
+BOARDS = nrf53/pdk/application nrf53/pdk/network nrf52/pdk
```
alexstanoev, line 7: You forgot to update the board names here

tests/02-compile-arm-ports/Makefile
```diff
@@ -59,6 +67,10 @@ platform-specific/cc26x0-cc13x0/cc26x0-web-demo/cc26x0-cc13x0:BOARD=sensortag/cc
 platform-specific/cc26x0-cc13x0/base-demo/cc26x0-cc13x0:BOARD=sensortag/cc2650 \
 platform-specific/cc26x0-cc13x0/base-demo/simplelink:BOARD=sensortag/cc1350 \
 platform-specific/cc26x0-cc13x0/base-demo/simplelink:BOARD=sensortag/cc1352r1 \
+platform-specific/nrf/empty-application/nrf:BOARD=nrf52840/dk \
```
alexstanoev, line 70: What's the idea behind testing this for the nRF52840?

arch/cpu/nrf/doxygen-group.txt
```diff
@@ -0,0 +1,9 @@
+/**
+ * \defgroup nrf The nRF's System-on-Chips
```
alexstanoev, line 2: I'd suggest something like `The nRF5340 and nRF52840 SoCs (using nRF MDK)` to make it more clear which SoCs we support here.

arch/cpu/nrf/doxygen-group.txt
```diff
@@ -0,0 +1,9 @@
...
+ * \ingroup cpu
+ */
+
+/**
+ * \defgroup nrf-platforms Nordic Semiconductor nRF-powered platforms
```
alexstanoev, line 7: I'd add `(using nRF MDK)` here too to avoid confusion with the existing nRF52840 platform.

alexstanoev (commented):
Hi @Yagoor,

Thanks again for working on this, we are very keen to get it merged soon! 
I had a look through the code and have left a few comments there, plus a few general thoughts:

* It would be nice if you could clean up the git history and squash commits into logical chunks.
* I can see that nrfx 2.5.0 is out now, it might be worth updating the submodule before we merge this.
* You might want to add a `doxygen-group.txt` in `arch/platform/nrf` to give the board modules friendlier names (like the existing nRF52840 platform).
* The `empty-application` example could do with a better name to make it more obvious that it's responsible for booting the network core. 
Perhaps something like `examples/platform-specific/nrf/nrf5340-application/start-network-core` (or a shorter version)?
* I noticed that some files have a trailing newline while others don't. Ideally they would all have one to stick to the C standards.

arch/cpu/arm/Makefile.arm
```diff
@@ -32,7 +32,7 @@ ASFLAGS += $(CFLAGS)
 ### Specify '-c' option to assemble only and not link
 ASFLAGS += -c 
 
-LDFLAGS += -mthumb -mlittle-endian
+LDFLAGS += -mthumb -mabi=aapcs -mlittle-endian
```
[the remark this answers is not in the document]
Yagoor: I noticed that it seems to be inconsistent here.
On line 21 we have
```c
CFLAGS += -mthumb -mabi=aapcs -mlittle-endian
```

But the mabi was missing in the LDFLAGS.
I can move it to the Makefile.nrf if this is on purpose. :)

arch/cpu/nrf/Makefile.nrf
```diff
@@ -0,0 +1,131 @@
...
+EXTERNALDIRS += $(addprefix $(NRFX_ROOT)/, $(NRFX_INC_PATHS))
+
+# Upload via J-Link using nrfjprog 
+
+.PHONY: nrf-upload-sequence
+
+NRFJPROG ?= nrfjprog
+NRFUTIL ?= nrfutil
+
+ifdef NRF52_UPLOAD_SN
```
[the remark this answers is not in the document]
Yagoor: Agree, will do it.

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
Yagoor: Agree, will do it.

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
Yagoor: Ups. Thank you.
The XXX is exactly a place holder for me to verify it, for some reason I missed those. 
It is related to different SoC's having different MAX and MIN, I'll take a look and get back to you here.

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
Yagoor: Okay. I guess when I ported the radio driver, this change was not yet there on the nrf52840. 
Will port it.

arch/platform/nrf/Makefile.nrf
Yagoor: Thank you.

tests/02-compile-arm-ports/Makefile
Yagoor: The idea behind is to make sure it's skipped in the Makefile and if it is not, it should fail the unittest.
Maybe it's not needed, I can remove if it is too much.

arch/cpu/nrf/Makefile.nrf
```diff
@@ -0,0 +1,131 @@
...
+CFLAGS += -DNRFX_RTC_ENABLED
+CFLAGS += -DNRFX_RTC0_ENABLED
+CFLAGS += -DNRFX_CLOCK_ENABLED
+CFLAGS += -DNRFX_POWER_ENABLED
+CFLAGS += -DNRFX_GPIOTE_ENABLED
+CFLAGS += -DNRFX_ATOMIC_USE_BUILT_IN
+CFLAGS += -DNRFX_UARTE_ENABLED
+CFLAGS += -DNRFX_UARTE0_ENABLED
+
+# LDFLAGS += --specs=nano.specs
```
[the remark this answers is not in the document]
Yagoor: I think one target was giving me issues when enabled, I'll revisit it and check it again.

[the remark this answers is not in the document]
Yagoor: I think one target was giving me issues when enabled, I'll revisit it and check it again.

tests/02-compile-arm-ports/Makefile
alexstanoev: Makes sense, that's a good test to have.

arch/cpu/nrf/Makefile.nrf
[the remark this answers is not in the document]
Yagoor: HI Alex,
I have tried it again and all boards won't boot if I enabled Newlib nano.
I'll look into why that is the case, even the nrf52840 DK won't boot...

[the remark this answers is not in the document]
Yagoor: The heap and eheap symbols were flipped. :)

arch/cpu/nrf/doxygen-group.txt
Yagoor: Done

arch/cpu/nrf/doxygen-group.txt
Yagoor: Done

Yagoor:
Hi @alexstanoev,

I believe, I did work on all your comments. Please, let me know if I missed something.
I did not work on the rebase one, because it would destroy the history so it could affect your review.
I propose we leave the rebase for last. Once you approve all the changes, I'll do the rebase and squash. :)

alexstanoev (approved):
Everything looks great to me!

Yagoor:
Hi @alexstanoev,
Following up on our chat yesterday.
I'll make the squash and add the slip. :)

PR #1470, opened 2020-12-04
CMSIS updated to v5

Yagoor:


g-oikonomou:
Thinking aloud: Do we want to make this a submodule?

Yagoor:
I think this makes more sense, making it a submodule.
I'll give it a try. :)

g-oikonomou:
They are distributing it through github, right?

g-oikonomou:
We are currently building CMSIS docs in our doxygen, but if it's being temperamental I would not oppose adding to the exclude list (`EXCLUDE_PATTERNS`)

Yagoor:
Yes, it's on github. 

It'll be easier to keep track as a submodule, we also don't need to worry about the documentation build if it's a submodule.
I like the idea more than copying the files when we need a new version. :)

g-oikonomou:
I support the submodule idea. We can probably add ARM-software/CMSIS_5 as a submodule without forking.

g-oikonomou:
I'd recommend setting the submodule to their [tag 5.7.0](https://github.com/ARM-software/CMSIS_5/releases/tag/5.7.0)

.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "arch/cpu/nrf52840/lib/nrf52-sdk"]
 	path = arch/cpu/nrf52840/lib/nrf52-sdk
 	url = https://github.com/contiki-ng/nrf-sdk-contiki-ng.git
+[submodule "arch/cpu/arm/CMSIS"]
+	path = arch/cpu/arm/CMSIS
```
g-oikonomou, line 35: We could put the submodule under `arch/cpu/arm`? And submodule initialisation would create a `CSMSIS` sub-dir. Right?

arch/cpu/arm/cortex-m/Makefile.cortex-m
```diff
@@ -1,4 +1,4 @@
-CONTIKI_ARM_DIRS += cortex-m cortex-m/CMSIS
+CONTIKI_ARM_DIRS += cortex-m CMSIS/CMSIS/Core/Include
```
g-oikonomou, line 1: and if we do the above then this becomes `CMSIS/Core/Include`, which is a bit prettier

tools/doxygen/Makefile
```diff
@@ -1,8 +1,9 @@
...
-basedirs	:= os arch
-empty		:=
-space		:= $(empty) $(empty)
-pwd			:= $(shell cd $(contiking); pwd)
+contiking    := ../..
+basedirs	 := os arch
+empty		 :=
+space		 := $(empty) $(empty)
+pwd			 := $(shell cd $(contiking); pwd)
+exclude_dirs := "$(contiking)/arch/cpu/arm/CMSIS"
```
g-oikonomou, line 6: Any reason not to use `EXCLUDE_PATTERNS` in `tools/doxygen/Doxyfile` instead?

```
EXCLUDE_PATTERNS       = */cpu/cc26x0-cc13x0/lib/* \
                         */cpu/cc26x0-cc13x0/rf-core/api/* \
                         */cpu/simplelink-cc13xx-cc26xx/lib/* \
                         */cpu/simplelink-cc13xx-cc26xx/rf-settings/* \
                         */cpu/simplelink-cc13xx-cc26xx/cc13x0-cc26x0/driverlib/* \
                         */cpu/simplelink-cc13xx-cc26xx/cc13x0-cc26x0/rf_patches/* \
                         */cpu/nrf52840/lib/* \
                         */platform/stm32nucleo-spirit1/stm32cube-lib/* \
                         */os/net/security/tinydtls/*
```

g-oikonomou (changes requested):
Thanks! See some change requests that you can of course alternatively rebut!

Yagoor:
Hi @g-oikonomou   
I answered the change requests but I'm not sure how the notifications work. 😁

g-oikonomou:
Hi @Yagoor have you had a chance to look at the most recent review? Couple of things I'd recommend changing and then we can get on with this one.

.gitmodules
g-oikonomou: But the submodule will always be under a CMSIS dir, right? So this structure will create `arch/cpu/arm/CMSIS/CMSIS`. What have I missed?

Yagoor: That is correct. So at the moment submodule is under CMSIS (arch/cpu/arm/CMSIS).
However the submodule also contains another folder called CMSIS in which the files are located.

ARM root folder:
```console
rosario@yagos-macbook-pro arm % pwd
/Users/rosario/contiki-ng/arch/cpu/arm
rosario@yagos-macbook-pro arm % ls -lah
total 40
drwxr-xr-x  10 rosario  staff   320B Feb  9 18:32 .
drwxr-xr-x  11 rosario  staff   352B Feb  9 18:32 ..
-rw-r--r--@  1 rosario  staff   6.0K Feb  9 18:32 .DS_Store
drwxr-xr-x  16 rosario  staff   512B Feb  9 18:32 CMSIS
-rw-r--r--   1 rosario  staff   2.3K Feb  9 18:30 Makefile.arm
-rw-r--r--   1 rosario  staff   3.2K Feb  9 18:30 arm-def.h
drwxr-xr-x   4 rosario  staff   128B Feb  9 18:30 common
drwxr-xr-x   8 rosario  staff   256B Feb  9 18:30 cortex-m
-rw-r--r--   1 rosario  staff    62B Feb  9 18:30 doxygen-group.txt
drwxr-xr-x   4 rosario  staff   128B Feb  9 18:30 openocd
rosario@yagos-macbook-pro arm % 

```

submodule root folder
```console
rosario@yagos-macbook-pro CMSIS % pwd
/Users/rosario/contiki-ng/arch/cpu/arm/CMSIS
rosario@yagos-macbook-pro CMSIS % ls -lah
total 3920
drwxr-xr-x  16 rosario  staff   512B Feb  9 18:32 .
drwxr-xr-x  10 rosario  staff   320B Feb  9 18:32 ..
-rw-r--r--@  1 rosario  staff   6.0K Feb  9 18:32 .DS_Store
-rw-r--r--   1 rosario  staff    52B Feb  9 18:32 .git
-rw-r--r--   1 rosario  staff   676B Feb  9 18:32 .gitattributes
-rw-r--r--   1 rosario  staff   224B Feb  9 18:32 .gitignore
-rw-r--r--   1 rosario  staff   263K Feb  9 18:32 ARM.CMSIS.pdsc
drwxr-xr-x  15 rosario  staff   480B Feb  9 18:32 CMSIS
-rw-r--r--   1 rosario  staff   1.2M Feb  9 18:32 CMSIS_Review_Meeting_2020.pdf
drwxr-xr-x   5 rosario  staff   160B Feb  9 18:32 Device
-rw-r--r--   1 rosario  staff    11K Feb  9 18:32 LICENSE.txt
-rw-r--r--   1 rosario  staff    11K Feb  9 18:32 README.md
-rw-r--r--   1 rosario  staff   431K Feb  9 18:32 Scalable Software Stack.pdf
drwxr-xr-x  16 rosario  staff   512B Feb  9 18:32 docs
-rw-r--r--   1 rosario  staff    12K Feb  9 18:32 linter.py
-rw-r--r--   1 rosario  staff   2.1K Feb  9 18:32 manifest
rosario@yagos-macbook-pro CMSIS % 

```

g-oikonomou: Duh. I'm with you now.

tools/doxygen/Makefile
Yagoor: Did you understand the issue I faced here? :)

g-oikonomou: Not really, no 😁 But I have not tried either

Yagoor: :D 
So the bash script collects the files to be passed to the Doxygen and saves it a env. variable. 
However these variables have a length limit.
I was unlucky and with the CMSIS we hit the ceiling. :) 
So I had to filter in the bash script.

Yagoor:
Hi @g-oikonomou, 
Is there something else that I need to work on this PR in order to get it in? :)

g-oikonomou:
We want this. Alex and I are trying to figure out what we want to do with this exclude patterns thing!

Yagoor:
Great. Thanks. 
One possible solution is to not use an env variable but instead write the files to a temporary file and them the Doxygen get the sources from the file and not from the env variable. 
If you like this proposal, I can try to implement it.

alexstanoev:
Hi @Yagoor 

We had a look at this, the `docsrc` variable is getting too long due to the large amount of .txt files in CMSIS:
`find $(dir) -type f -name "*.txt"` for CMSIS has 1801 results alone, up from 36 for the entire contiki-ng repo before this change.

Considering every .txt file in the repo for doxygen is perhaps not the best solution, we're still discussing the right way to sort this out long-term.
One way around this is to only consider files named `doxygen-group.txt`, `modules.txt` or `mainpage.txt`.
We'd then also need to move `tools/doxygen/pt-doc.txt` and  `arch/platform/sky/doc/sky.txt` over to their relevant .h files.

g-oikonomou:
@Yagoor can you try on top of #1530?

Yagoor:
Hi @g-oikonomou,

It worked, I reverted the workaround I had done and I added CMSIS path in the EXCLUDE_PATTERNS.

g-oikonomou:
We're ready here, right?

g-oikonomou (approved):

PR #1478, opened 2020-12-14
nRF52840 RF init HFCLK fix

schuschu:
This PR adds a busy wait for the HFCLK during init to avoid hardware dependent race conditions (nrf boards seem to not be affected, custom boards with less reputable crystals are).


alexstanoev:
Thanks Markus!

This was a tricky one to find, and from my testing it is also the root cause of #1449 and solves our issue with old compilers.

alexstanoev (approved):

PR #1482, opened 2020-12-23
Added buffer boundary checks for 6LoWPAN decompression

nvt:
Unlike for the 6LoWPAN compression function, there are no boundary checks in the decompression function. Additionally, packets that are not decompressed correctly are passed on the to the upper layer because the IPHC decompression function does not return a value indicating whether the operation succeeded.

This PR addresses these two issues. Moreover, a packet that demonstrates the out-of-bounds read is added to the packet test, but the problem might be noticed only when compiled with an address sanitizer.


nfi (approved):
Looks good!

PR #1501, opened 2021-02-09
nrf52840: Implement USB power events for hotplug support

alexstanoev:
Enabling the USB stack on boot works for the nRF52840 dongle, however, it can fail to enumerate on custom boards or on the DK when the native USB port is plugged in after boot.

This commit handles USB power events to enable and disable the nRF USB stack when the USB power state changes. This is necessary to make USB hotplugging work in any order.

Tested on both a DK and a dongle.

### DIFF
--- a/arch/cpu/nrf52840/usb/usb-serial.c
+++ b/arch/cpu/nrf52840/usb/usb-serial.c
@@ -181,14 +181,14 @@ usbd_user_ev_handler(app_usbd_event_type_t event)
 void
 usb_serial_init(void)
 {
+  static const app_usbd_config_t usbd_config = {
+    .ev_state_proc = usbd_user_ev_handler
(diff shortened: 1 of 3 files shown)


arch/cpu/nrf52840/usb/usb-serial.c
```diff
@@ -146,7 +186,11 @@ usb_serial_init(void)
 
   app_usbd_serial_num_generate();
 
-  ret = app_usbd_init(NULL);
+  static const app_usbd_config_t usbd_config = {
+    .ev_state_proc = usbd_user_ev_handler
+  };
```
g-oikonomou, line 191: Can this definition be moved to the top of the func?

g-oikonomou (approved):

PR #1502, opened 2021-02-09
Misc. static-code-analysis fixes

arurke:
This PR contains proposed fixes to misc. complaints from cppcheck and clang scan-build. Some might be a bit pedantic so we can cherry-pick whatever we want to keep after discussion.

The commit on coffee might warrant some discussion as I am not too familiar with the usage of that struct, but to me it seems you would want the the directory-descriptor name to be the same size as the file-header name which stores it.

### DIFF
--- a/os/lib/newlib/syscalls.c
+++ b/os/lib/newlib/syscalls.c
@@ -76,7 +76,7 @@ _sbrk(int incr)
   static uint8_t *heap_end = &_heap;
   uint8_t *prev_heap_end = heap_end;
 
-  if(heap_end + incr > &_eheap) {
+  if((uintptr_t)heap_end + incr > (uintptr_t)&_eheap) {
     PRINTF("Out of heap space!\n");
     errno = ENOMEM;
     return (caddr_t)-1;
--- a/os/sys/stack-check.c
+++ b/os/sys/stack-check.c
@@ -44,6 +44,7 @@
 #include "sys/stack-check.h"
 #include "dev/watchdog.h"
 #include <string.h>
+#include <stdint.h>
 #include <inttypes.h>
 
 #include "sys/log.h"
@@ -67,7 +68,7 @@ PROCESS(stack_check_process, "Stack check");
 #else
 /* use the value provided by the linker script */
 extern int _stack_origin;
-#define GET_STACK_ORIGIN() (&_stack_origin)
+#define GET_STACK_ORIGIN() (uintptr_t)(&_stack_origin)
 #endif
 /*---------------------------------------------------------------------------*/
 void
@@ -77,13 +78,13 @@ stack_check_init(void)
   volatile uint8_t *p;
 
   /* Make this static to avoid destroying it in the while loop */
-  static void *stack_top;
+  static uintptr_t stack_top;
   /* Use address of this local variable as a boundary */
-  stack_top = &p;
+  stack_top = (uintptr_t) &p;
 
   /* Note: this is expected to be called before the WDT is started! */
   p = &_stack;
-  while(p < (uint8_t *)stack_top) {
+  while((uintptr_t)p < stack_top) {
     *p++ = STACK_FILL;
   }
 
@@ -104,30 +105,30 @@ stack_check_get_usage(void)
   /* Skip the bytes used after heap; it's 1 byte by default for _stack,
    * more than that means dynamic memory allocation is used somewhere.
    */
-  while(*p != STACK_FILL && p < (uint8_t *)GET_STACK_ORIGIN()) {
+  while(*p != STACK_FILL && (uintptr_t)p < GET_STACK_ORIGIN()) {
     p++;
   }
 
   /* Skip memory-region reserved for the stack not used yet by the program */
-  while(*p == STACK_FILL && p < (uint8_t *)GET_STACK_ORIGIN()) {
+  while(*p == STACK_FILL && (uintptr_t)p < GET_STACK_ORIGIN()) {
     p++;
   }
 
   /* Make sure WDT is not triggered */
   watchdog_periodic();
 
-  if(p >= (uint8_t*)GET_STACK_ORIGIN()) {
+  if((uintptr_t)p >= GET_STACK_ORIGIN()) {
     /* This means the stack is screwed. */
     return -1;
   }
 
-  return (uint8_t *)GET_STACK_ORIGIN() - p;
+  return GET_STACK_ORIGIN() - (uintptr_t)p;
 }
 /*---------------------------------------------------------------------------*/
 int32_t
 stack_check_get_reserved_size(void)
 {
-  return (uint8_t *)GET_STACK_ORIGIN() - &_stack;
+  return GET_STACK_ORIGIN() - (uintptr_t)&_stack;
 }
 /*---------------------------------------------------------------------------*/
 #if STACK_CHECK_PERIODIC_CHECKS


os/lib/newlib/syscalls.c
```diff
@@ -76,7 +76,7 @@ _sbrk(int incr)
   static uint8_t *heap_end = &_heap;
   uint8_t *prev_heap_end = heap_end;
 
-  if(heap_end + incr > &_eheap) {
+  if((uintptr_t)heap_end + incr > (uintptr_t)&_eheap) {
```
atiselsts, line 79: Can you explain the need to cast pointer values to `uintptr_t`, here and in other places?

os/services/ip64/ip64-arp.c
```diff
@@ -285,7 +285,7 @@ ip64_arp_check_cache(const uint8_t *nlhdr)
     /* Multicast. */
     return 1;
   } else {
-    uip_ip4addr_t ipaddr;
+    uip_ip4addr_t ipaddr = {0};
```
atiselsts, line 288: I don't think we need to initialize `ipaddr` in this line - it is never used before a value is assigned to it later.

os/services/ip64/ip64-arp.c
```diff
@@ -342,7 +342,7 @@ ip64_arp_create_ethhdr(uint8_t *llhdr, const uint8_t *nlhdr)
     ethhdr->dest.addr[4] = ipv4_hdr->destipaddr.u8[2];
     ethhdr->dest.addr[5] = ipv4_hdr->destipaddr.u8[3];
   } else {
-    uip_ip4addr_t ipaddr;
+    uip_ip4addr_t ipaddr = {0};
```
atiselsts, line 345: Same comment as above

os/storage/cfs/cfs-coffee.c
```diff
@@ -75,6 +75,10 @@
 #error "Cannot have COFFEE_APPEND_ONLY set when COFFEE_MICRO_LOGS is set."
 #endif
 
+#if COFFEE_NAME_LENGTH != CFS_DIR_ENTRY_NAME_LENGTH
```
atiselsts, line 78: why not simply use `COFFEE_NAME_LENGTH` instead of introducing the new constant `CFS_DIR_ENTRY_NAME_LENGTH`?

atiselsts (commented):
I think most of this is good, just a few comments.

os/services/ip64/ip64-arp.c
arurke: Agreed, I will remove (I guess this is not right place for debate on initializing variables 😄 )

os/services/ip64/ip64-arp.c
arurke: as above

os/storage/cfs/cfs-coffee.c
arurke: That would create a dependency from CFS to Coffee, which I assume would not survive sunlight since Coffee implements CFS(?).

However, I see this break at least z1 and sky. Leaving out the constant and add only the memcpy()-change is an option. Users would have to be aware that their directory-names would be limited by `COFFEE_NAME_LENGTH` and not the hard-coded `32` in cfs.h. With that said I notice this comment in coffee: 
```
  /*
   * Coffee is only guaranteed to support the directory names "/" and ".",
   * but it does not enforce this currently.
   */
```
So, it seems the directory-naming-support is dubious already. Leave as-is and fix the memcpy?

os/lib/newlib/syscalls.c
arurke: Cppcheck v2.3 says:
```
    "severity": "critical",
    "check_name": "comparePointers",
    "description": "[cppcheck] Comparing pointers that point to different objects (CWE-570)",
```

With the point being that doing relational comparison on pointers to different memory spaces is undefined behavior - while comparing `uintptr_t`/`intptr_t` (which is intended to be used for integer presentation of pointers) is not. However, I see two problems now: 1) After a bit more reading it seems we must actually first cast to `void*` because that is the only type the standard guarantees may be casted to `uintptr_t`/`intptr_t`, and 2) Sky and z1 now complains. Not sure why, but these types are optional...

g-oikonomou:
Hi Andreas @arurke any idea why the tests are failing?

arurke:
yea, because of 

1. Some platforms use different length for directory names and the struct that holds them. The problem is discussed here: https://github.com/contiki-ng/contiki-ng/pull/1502#discussion_r628521101. As mentioned there I suggest we just fix the overflow and leave the rest as is.
2. Z1 and sky does not like the cast to uintptr_t. The problem is discussed here: https://github.com/contiki-ng/contiki-ng/pull/1502#discussion_r628598750 . Was hoping some platform maintainers would jump in 😅

g-oikonomou:
@arurke do you have time to split this in two PRs? We are happy to merge the 3 commits: 4f918ed, 52645b1, 955edd2 right now.

The other two will need a little bit more work but we don't want to hold the entire PR on those two cause there is some good stuff to merge already.

Please let me know if you don't have time and I'll split

os/storage/cfs/cfs-coffee.c
nvt: The warning could be removed, and then just keep the fix of the memcpy call to the shorter name. Additionally, record->name should be null-terminated with an offset that is the shorter of the two lengths.

arurke: :+1:

> Additionally, record->name should be null-terminated with an offset that is the shorter of the two lengths.

Indeed! fixed

arurke:
@g-oikonomou :+1:  https://github.com/contiki-ng/contiki-ng/pull/1585

Following the discussion and fixes in https://github.com/contiki-ng/contiki-ng/pull/1502#discussion_r645089921, it might be 4e07c47f20057827bec40a916a49584027e2e161 is ready as well.

os/lib/newlib/syscalls.c
pjonsson: The failure on Sky/Z1 was because they were configured to compile for an older standard than C99, that has been fixed.

I would have expected the pointers to be char * rather than uint8_t * in this code, and upstream newlib seems to both do that and have updated the code to also remove the non-standard caddr_t type: 

https://sourceware.org/git/?p=newlib-cygwin.git;a=blob_plain;f=libgloss/arm/syscalls.c;hb=refs/heads/master

Does cppcheck still warn for the updated upstream function?

arurke: Thank you for the input @pjonsson . I will re-visit this in the not too distance future.

pjonsson:
Doesn't this work fine if you use `char *` as upstream (instead of `void *` and `uint8_t *`)?

As a bonus, that should avoid strict aliasing issues.

arurke:
I tried the following starting from develop-branch:

```
--- a/os/lib/newlib/syscalls.c
+++ b/os/lib/newlib/syscalls.c
@@ -70,11 +70,11 @@ _sbrk(int incr)
   extern int errno;
 
   /* Heap boundaries from linker script. */
-  extern uint8_t _heap;
-  extern uint8_t _eheap;
+  extern char _heap;
+  extern char _eheap;
 
-  static uint8_t *heap_end = &_heap;
-  uint8_t *prev_heap_end = heap_end;
+  static char *heap_end = &_heap;
+  char *prev_heap_end = heap_end;
```

Is this what you had in mind? Cppcheck (now at v2.9.0) still complains. As we [discussed](https://github.com/contiki-ng/contiki-ng/pull/1502#discussion_r628598750) earlier, I believe the issue is that the pointers are pointing to different objects in memory, and we can thus not do relational operations on them (I found [this](https://stackoverflow.com/questions/59516346/how-does-pointer-comparison-work-in-c-is-it-ok-to-compare-pointers-that-dont-p) stackoverflow answer helpful). With that in mind I did not expect changing to `char*` would help. But I am learning as I am going, so please feel free to chime in.

I have not had time to dig deeper yet but my earlier thoughts on what are the remaining issues: 
> However, I see two problems now: 1) After a bit more reading it seems we must actually first cast to void* because that is the only type the standard guarantees may be casted to uintptr_t/intptr_t, and 2) Sky and z1 now complains. Not sure why, but these types are optional...

pjonsson:
My point with `char *` was just to remove the strict aliasing issues.

Upstream has also changed the signature to `void * _sbrk (ptrdiff_t incr)`, so the function gets a difference between two pointers. My intuition is to try to stick with pointer arithmetic if it's possible, but I don't have any pointers to the standard for why pointer arithmetic would be preferred. Here's another link where someone describes some funky behavior on a Cray:

https://stackoverflow.com/questions/43418618/c-why-cast-to-uintptr-t-vs-char-when-doing-pointer-arithmetic

simonduq:
Hi!

PR #1506, opened 2021-02-19
Add buffer boundary check in sicslowpan

nvt:
This PR adds a buffer boundary check before reading from hc06_ptr in a loop condition in the <code>uncompress_hdr_iphc</code> function. The behavior is visible when injecting certain packets into Contiki-NG when running in native mode with ASAN enabled.


nfi (approved):
Looks good!

PR #1511, opened 2021-02-25
Bugfix: remote dual rf

tinstructor:
The current implementation of the board header files for zoul platforms prevents enabling/disabling dual radio usage via an application's project-conf.h file as [explained here](https://github.com/Zolertia/Resources/wiki/RE-Mote-2.4GHz-dual-antenna#enabling-both-radios-in-contiki) (which should be standard behavior IMO). A simple preprocessor check solves this issue. 

~Second (and unrelated), vscode keeps adding an annoying hidden config folder and I really think this belongs in the gitignore. However, it's so minor of a change that I didn't really think dedicating an entire pull request to it was warranted.~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@
 *.ihex
 *.pyc
 *~
-*.vscode
 build/*
 Makefile.target
 Makefile.*.defines
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 *.ihex
 *.pyc
 *~
+*.vscode
 build/*
 Makefile.target
 Makefile.*.defines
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -12,6 +12,7 @@
 *.ihex
 *.pyc
 *~
+*.vscode
```
g-oikonomou, line 15: The recommended way to deal with this is through your local `.git/info/exclude`. The top-level `.gitignore` is for files directly associated to the project (e.g. output binaries, temporary files generated by the build system)

g-oikonomou (changes requested):

.gitignore
tinstructor: Reverted changes to gitignore as requested

g-oikonomou (approved):

PR #1514, opened 2021-03-01
add UIPBUF_ATTR_RSSI and UIPBUF_ATTR_LINK_QUALITY attibutes to uipbuf

atiselsts:
Adds these fields to the uipbuf, and replaces the function `sicslowpan_get_last_rssi()` which provided similar functionality, but was not as clean in API sense (also, I could not find it when required, so made this change before discovering it exists :P)

### DIFF
--- a/examples/6tisch/timesync-demo/node.c
+++ b/examples/6tisch/timesync-demo/node.c
@@ -75,10 +75,12 @@ udp_rx_callback(struct simple_udp_connection *c,
 
     LOG_INFO("Received from ");
     LOG_INFO_6ADDR(sender_addr);
-    LOG_INFO_(", created at %lu, now %lu, latency %lu clock ticks\n",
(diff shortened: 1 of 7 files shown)


alexstanoev:
This does make the API much cleaner, however, would it be worth indicating to users with legacy code that `sicslowpan_get_last_rssi()` has been deprecated, as this method still shows up on web searches?

Perhaps something like this in sicslowpan.h:
```
int __attribute__ ((deprecated("Use uipbuf_get_attr(UIPBUF_ATTR_RSSI) instead"))) sicslowpan_get_last_rssi(void);
```

Which ends up looking like:
 `error: 'sicslowpan_get_last_rssi' is deprecated: Use uipbuf_get_attr(UIPBUF_ATTR_RSSI) instead [-Werror=deprecated-declarations]`

If we do actually want to do this it would probably need to go through `cc.h` as a deprecated macro though.

atiselsts:
Restored back `sicslowpan_get_last_rssi`.

alexstanoev (approved):

PR #1516, opened 2021-03-01
RPL: do not assume that a neighbor is unreachable if IPv6 neighbor discovery status is not REACHABLE

### LINKED ISSUE #1498: switch to parent NULL when UIP_CONF_ND6_SEND_NS=1 with RPL-lite
I have a cooja setup where RPL lite is enabled and
`UIP_CONF_ND6_SEND_NS=1`
When a node is set as root, the other nodes join the rpl instance as expected but after some time the network stops working. Nodes loose the preferred parent for some time and re-join again. 

I used the hello-world example and enabled the shell, rpl lite and the UIP_CONF_ND6_SEND_NS flag. 5 nodes, linear setup each in range of one neighbor on each side. Root node on the far left. The mote output log is attached here (ID 3 is 1 hop from the root and has RPL logging enabled):

[loglistener_append.txt](https://github.com/contiki-ng/contiki-ng/files/5905684/loglistener_append.txt)

Then it cascades and gets worse and worse the deeper in the network, because DAO-ACK's are not received due to broken links on the way.

The interesting parts are switch to parent NULL, e.g.

```
[INFO: RPL       ] parent switch: fe80::201:1:1:1 -> (NULL IP addr)
```

I have a branch reproducing this issue: https://github.com/damien-v/contiki-ng/tree/issue-rpl-lite-with-send-ns

atiselsts:
Also, removes the function `rpl_neighbor_is_reachable` of RPL lite, as it was not used in the code.

Fixes #1498.

### DIFF
--- a/os/net/routing/rpl-lite/rpl-neighbor.c
+++ b/os/net/routing/rpl-lite/rpl-neighbor.c
@@ -270,23 +270,6 @@ rpl_neighbor_is_fresh(rpl_nbr_t *nbr)
 }
 /*---------------------------------------------------------------------------*/
 int
-rpl_neighbor_is_reachable(rpl_nbr_t *nbr) {
-  if(nbr == NULL) {
-    return 0;
-  } else {
-#if UIP_ND6_SEND_NS
-    uip_ds6_nbr_t *ds6_nbr = rpl_get_ds6_nbr(nbr);
-    /* Exclude links to a neighbor that is not reachable at a NUD level */
-    if(ds6_nbr == NULL || ds6_nbr->state != NBR_REACHABLE) {
-      return 0;
-    }
-#endif /* UIP_ND6_SEND_NS */
-    /* If we don't have fresh link information, assume the nbr is reachable. */
-    return !rpl_neighbor_is_fresh(nbr) || curr_instance.of->nbr_has_usable_link(nbr);
-  }
-}
-/*---------------------------------------------------------------------------*/
-int
 rpl_neighbor_is_parent(rpl_nbr_t *nbr)
 {
   return nbr != NULL && nbr->rank < curr_instance.dag.rank;
@@ -378,12 +361,9 @@ best_parent(int fresh_only)
     }
 
 #if UIP_ND6_SEND_NS
-    {
-    uip_ds6_nbr_t *ds6_nbr = rpl_get_ds6_nbr(nbr);
     /* Exclude links to a neighbor that is not reachable at a NUD level */
-    if(ds6_nbr == NULL || ds6_nbr->state != NBR_REACHABLE) {
+    if(rpl_get_ds6_nbr(nbr) == NULL) {
       continue;
-    }
     }
 #endif /* UIP_ND6_SEND_NS */
 
--- a/os/net/routing/rpl-classic/rpl-dag.c
+++ b/os/net/routing/rpl-classic/rpl-dag.c
@@ -223,9 +223,8 @@ rpl_parent_is_reachable(rpl_parent_t *p) {
     return 0;
   } else {
 #if UIP_ND6_SEND_NS
-    uip_ds6_nbr_t *nbr = rpl_get_nbr(p);
     /* Exclude links to a neighbor that is not reachable at a NUD level */
-    if(nbr == NULL || nbr->state != NBR_REACHABLE) {
+    if(rpl_get_nbr(p) == NULL) {
       return 0;
     }
 #endif /* UIP_ND6_SEND_NS */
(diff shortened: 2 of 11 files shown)


joakimeriksson:
What was the specific issues with using the not reachable in the IPv6 neighbour entry. Do we often get a case where they are "not reachable" in ND6 but still reachable?

atiselsts:
One thing I'm confused about is the multiple states of the neighbors. Not sure if states DELAY, PROBE and STALE should be treated somewhat differently from each?

joakimeriksson:
The states are defined in the RFC (ND - 4861).

`   INCOMPLETE  Address resolution is in progress and the link-layer
                  address of the neighbor has not yet been determined.

      REACHABLE   Roughly speaking, the neighbor is known to have been
                  reachable recently (within tens of seconds ago).

      STALE       The neighbor is no longer known to be reachable but
                  until traffic is sent to the neighbor, no attempt
                  should be made to verify its reachability.

      DELAY       The neighbor is no longer known to be reachable, and
                  traffic has recently been sent to the neighbor.
                  Rather than probe the neighbor immediately, however,
                  delay sending probes for a short while in order to
                  give upper-layer protocols a chance to provide
                  reachability confirmation.

      PROBE       The neighbor is no longer known to be reachable, and
                  unicast Neighbor Solicitation probes are being sent to
                  verify reachability.`

joakimeriksson:
The only possible issue is that if we do have a case when we do not have a ds6-nbr in the tables but still have a link-metric that is fresh which means that we still would treat this as a "solid" and reachable neighbour even If the IPv6 stack can not send anything to it (packets are dropped if we do not have a ipv6-nbr entry if I remember correctly). One possible fix would probably be to just remove the check on REACHABLE on the nbr state check?!

atiselsts:
Added some regression tests using the (so far unused) `tests/07-simulation-base/code-ipv6/` test applications.

atiselsts:
@joakimeriksson let me see if understood the comment - did you mean change the check `if(nbr == NULL || nbr->state != NBR_REACHABLE)` to `if(nbr == NULL)`, and revert the change that checks for `fresh_only`?

joakimeriksson:
@atiselsts - yes - that would make the change very small - but still solve the problem - since the NBR entry should still be there for a probe - and if that succeeds it will just continue to work fine. But if the NBR entry is removed and we do not check for it we can still assume connectivity while dropping all the packets.

atiselsts:
@joakimeriksson changed the code as discussed!

joakimeriksson:
@atiselsts - I guess we should also keep the RPL-classic is reachable part e.g. 
```
#if UIP_ND6_SEND_NS
     uip_ds6_nbr_t *nbr = rpl_get_nbr(p);
     /* Exclude links to a neighbor that is not reachable at a NUD level */
     if(nbr == NULL || nbr->state != NBR_REACHABLE) {
       return 0;
     }
#endif /* UIP_ND6_SEND_NS */

```
=> 
```
#if UIP_ND6_SEND_NS
     uip_ds6_nbr_t *nbr = rpl_get_nbr(p);
     /* Exclude links to a neighbor that is not reachable at a NUD level */
     if(nbr == NULL) {
       return 0;
     } 
#endif /* UIP_ND6_SEND_NS */
```
or maybe even skipping the variable there:
```
#if UIP_ND6_SEND_NS
     /* Exclude links to a neighbor that is not reachable at a NUD level */
     if(rpl_get_nbr(p) == NULL) {
       return 0;
     } 
#endif /* UIP_ND6_SEND_NS */
```

os/net/routing/rpl-lite/rpl-neighbor.c
```diff
@@ -378,13 +361,11 @@ best_parent(int fresh_only)
     }
 
 #if UIP_ND6_SEND_NS
-    {
     uip_ds6_nbr_t *ds6_nbr = rpl_get_ds6_nbr(nbr);
     /* Exclude links to a neighbor that is not reachable at a NUD level */
-    if(ds6_nbr == NULL || ds6_nbr->state != NBR_REACHABLE) {
```
joakimeriksson, line 384: Would it even make sense to skip the variable? And just check if the function returns non-null?

os/net/routing/rpl-lite/rpl-neighbor.c
```diff
@@ -378,13 +361,11 @@ best_parent(int fresh_only)
...
+    if(ds6_nbr == NULL) {
```
joakimeriksson, line 366: Would it even make sense to skip the variable and just check the function return value directly?
      if(rpl_get_ds6_nbr(nbr) == NULL)
?

joakimeriksson (approved):

PR #1517, opened 2021-03-02
Fix the length field check for SNMP input packets

nvt:
The <code>*length</code> value in  can be controlled through an input packet, but it is not validated correctly. An attacker can thus cause an overflow of the <code>used</code> variable. This PR changes the check to a comparison instead of a subtraction.

Thanks to Tobias Scharnowski from Ruhr-University Bochum for the initial reporting of a missing length check.

### DIFF
--- a/os/net/app-layer/snmp/snmp-ber.c
+++ b/os/net/app-layer/snmp/snmp-ber.c
@@ -416,7 +416,7 @@ snmp_ber_decode_string_len_buffer(snmp_packet_t *snmp_packet, const char **str, 
 
   *str = (const char *)snmp_packet->in;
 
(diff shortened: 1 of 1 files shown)


Yagoor:
Hi @nvt,
Thank you for spotting this issue and for the fix.

Would it be possible to have a unittest to cover this?
I made several tests where there is a running server and the client issues commands to it and compare the return value. 
This is in the string decoder, so from the top of my head only the community name should use it.
Maybe a test case with a very long community name which we would expect to fail but in wasn't (before your fix).

I can do the test if you prefer and push it in another PR.

yatch (approved):
It looks good to me. I understand the change also fixes a bug returning `0` when `snmp_packet->used` is equal to `*length`.

PR #1518, opened 2021-03-04
Orchestra root rule

atiselsts:
This PR updates the Orchestra scheduler with a new kind of rule. This rule is active on the root node, and on all direct neighbors to the root node.

The assumption is that the root node can afford to keep Rx always on. The code exploits this and adds a special TSCH slotframe on the root node that has all slots set to Rx. Nodes that are direct neighbors to the root node send their packets to the root using this special root rule; each node selects its own slot from ORCHESTRA_ROOT_PERIOD slots (by default 7). This allows to reduce the number of packet collisions on the root, and massively improves performance for data collection applications, where the root node is often the bottleneck for the whole network.

For more info see the paper https://ieeexplore.ieee.org/document/9305207/


yatch:
> The assumption is that the root node can afford to keep Rx always on. 

This is a sensible assumption. We have a similar approach in some project. Thanks for the reference as well. Congrats!

atiselsts:
Thanks @yatch !

I have updated the PR to better match the existing Orchestra API approach, which was using callbacks. There is now a new callback definition `TSCH_CALLBACK_ROOT_NODE_UPDATED` that is used from TSCH instead of calling the Orchestra functions directly.

examples/6tisch/simple-node/Makefile
```diff
@@ -15,28 +15,45 @@ MAKE_WITH_PERIODIC_ROUTES_PRINT ?= 0
...
+      ORCHESTRA_EXTRA_RULES +=,&special_for_root
     endif
   else
     ifeq ($(MAKE_WITH_LINK_BASED_ORCHESTRA),1)
-      $(error "Inconsistent configuration")
+      $(error "Inconsistent configuration: link-based Orchestra requires routing info")
+    endif
+
+    ifeq ($(MAKE_WITH_ORCHESTRA_ROOT_RULE),1)
+     $(error "Inconsistent configuration: NS rule and root rule conflicts!")
```
simonduq, line 50: Wouldn't it be possible to run the root rule along with NS, at least in theory?
Send following receiver schedule, but if receiver is root send anytime?

os/net/mac/tsch/tsch-roots.c
```diff
@@ -0,0 +1,180 @@
+/*
+ * Copyright (c) 2018, Amber Agriculture
```
simonduq, line 2: Is this the desired copyright and date?

os/net/mac/tsch/tsch-roots.c
```diff
@@ -0,0 +1,180 @@
...
+  struct tsch_root_info *next;
+  clock_time_t now;
+
+  now = clock_seconds();
+  root = list_head(tsch_roots);
+  while(root != NULL) {
+    next = root->next;
+    if((int32_t)(root->last_seen_seconds + ROOT_ALIVE_TIME_SECONDS - now) < 0) {
+      /* the root info has become obsolete; remove its scheduled link */
+      LOG_INFO_("remove root address ");
```
simonduq, line 136: I think you want `LOG_INFO` not `LOG_INFO_`

os/net/mac/tsch/tsch-roots.c
```diff
@@ -0,0 +1,180 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ */
+/**
```
simonduq, line 29: A few lines of comment on what the module is about would be helpful :)

simonduq: So btw does everything work just fine with multiple TSCH roots?
Not something I've ever tested so I'm curious :)

os/services/orchestra/orchestra-rule-special-for-root.c
```diff
@@ -0,0 +1,202 @@
...
+static uint8_t is_root_rule_used;
+static uint16_t timeslot_tx;
+
+static void set_self_to_root(uint8_t is_root);
+/*---------------------------------------------------------------------------*/
+static inline uint8_t
+self_is_root(void)
+{
+  /* If this is changed to look at the RPL status instead of tsch_is_coordinator,
+   * multiple roots become possible in a single TSCH network. */
```
simonduq, line 55: But ok now reading this comment I'm confused: how does a network with one coordinator but multiple roots work?
What does the time-source tree/dag look like in that case?

simonduq (approved):
Cool stuff! I only have a few questions I'd really want an answer on, but that's an approve.
If the nightly testbed ever gets up again we should add a config with this!

examples/6tisch/simple-node/Makefile
atiselsts: Good comment - I had to think for a while to remember what the issue was.
I think the issue is in the prioritization of the different rules - NS rule is prioritized over the root rule, because it has a lower slotframe handle. And the NS rule all slots potentially active, with all-except-one slots marked as Tx. This leads to the NS slot is always being selected over the root rule slot. Meaning that the root rule never get used for transmission, and the queue of packets towards the root just fills up and never gets cleared.

os/net/mac/tsch/tsch-roots.c
atiselsts: ok!

examples/6tisch/simple-node/Makefile
atiselsts: For the NS rule, there is a related PR: https://github.com/contiki-ng/contiki-ng/pull/1520
It's basically ready from my side, but I'm waiting for this one to be merged before moving it out of the draft status, because its based on top of this one. It would be great if you could also look there and give quick feedback about the idea.

os/services/orchestra/orchestra-rule-special-for-root.c
simonduq: Hmm ok. So, you have a time-source tree with corresponding routeCost then, independent from the multiple RPL DAGs?

PR #1528, opened 2021-03-11
cc13xx prop-mode: make payload len configurable to reduce max rx time

lafka:
Radio may enters sync due to random rf noise or by packets from other networks.
This may lead to long RXs periods where radio is attempting to read packets
which are too large for us to handle, blocking other valid receptions.
In areas with high noise and/or operating with low bitrate this may cause radio
to miss valid reception for several 100 milliseconds.

When Pinging a different device in a loop prop-mode.c will print "RF: Too short"
and radio will miss valid receptions.

Setting max frame length to same as rxbuf size increased reliability in certain
environments from 98% to 99.9%. Test was done with transmitting 255
requests in sequence then waiting a random delay and repeating for a total of
100 000 requests.

Same reliability could also be reached by setting a longer syncword indicating this
was due to random noise (just changing to a different syncword of same size had no effect).

### DIFF
--- a/arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
+++ b/arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
@@ -133,7 +133,9 @@ static int off(void);
 static rfc_propRxOutput_t rx_stats;
 /*---------------------------------------------------------------------------*/
 /* Defines and variables related to the .15.4g PHY HDR */
-#define DOT_4G_MAX_FRAME_LEN    2047
+#ifndef DOT_4G_MAX_FRAME_LEN
+  #define DOT_4G_MAX_FRAME_LEN    2047
+#endif
 #define DOT_4G_PHR_LEN             2
 
 /* PHY HDR bits */
@@ -219,6 +221,12 @@ static const prop_mode_tx_power_config_t *tx_power_current = &TX_POWER_DRIVER[1]
 #define DATA_ENTRY_LENSZ_NONE 0
 #define DATA_ENTRY_LENSZ_BYTE 1
 #define DATA_ENTRY_LENSZ_WORD 2 /* 2 bytes */
+
+#ifndef PROP_MODE_CONF_RX_WAIT_TIME
+  #define PROP_MODE_RX_WAIT_TIME (RTIMER_SECOND / 50)
+#else
+  #define PROP_MODE_RX_WAIT_TIME PROP_MODE_CONF_RX_WAIT_TIME
+#endif
 
 /* The size of the metadata (excluding the packet length field) */
 #define RX_BUF_METADATA_SIZE \
@@ -861,7 +869,7 @@ read_frame(void *buf, unsigned short buf_len)
   /* wait for entry to become finished */
   rtimer_clock_t t0 = RTIMER_NOW();
   while(entry->status == DATA_ENTRY_STATUS_BUSY
-      && RTIMER_CLOCK_LT(RTIMER_NOW(), t0 + (RTIMER_SECOND / 50)));
+      && RTIMER_CLOCK_LT(RTIMER_NOW(), t0 + PROP_MODE_RX_WAIT_TIME));
 
 #if MAC_CONF_WITH_TSCH
   /* Make sure the flag is reset */


atiselsts:
Thanks for the investigation and the proposal, for sure this is a nice find and impressive reliability improvement.

But I'm still confused about some things here - in particular, how the description and the title of the PR related to the actual code changes.

> Setting max frame length to same as rxbuf size 

From my reading of this, the correct approach would be to define `DOT_4G_MAX_FRAME_LEN` to be equal to `MAX_PAYLOAD_LEN` summed with some metadata (probably `RADIO_CONST_PHY_OVERHEAD`, but need to check the manual to confirm).

Another comment is that it would be best to make waiting time in `read_frame()` dependent on `RADIO_BYTE_AIR_TIME`  multiplied by the sum of `RADIO_CONST_PHY_OVERHEAD` and `MAX_PAYLOAD_LEN` (round the result up in rtimer ticks, and add some small guard time). Then you don't need to introduce a new configuration constant for `PROP_MODE_RX_WAIT_TIME`.

lafka:
I was a little too quick in my description.

Testing was done with values that matched my use case, they are more or less what you describe but including that would not be backwards compatible. I though it be better to not change the current defaults and let the user define what best matches his setup.

Roughly i'm using (from top of my head):

```c
#define  DOT_4G_MAX_FRAME_LEN (RX_BUF_DATA_OFFSET + MAX_PAYLOAD_LEN + (CRC_LEN * RX_CORE_RX_BUF_INCLUDE_CRC))
#define PROP_MODE_CONF_RX_WAIT_TIME  US_TO_RTIMERTICKS(RADIO_BYTE_AIR_TIME * DOT_4G_MAX_FRAME_LEN)`
```
I can update to include those but I don't have any clue about that will affect TSCH MAC or setups with different RF bitrate (i guess default is 250kbps, we're using 50kbps). Probably the same should  be done for simplelink platform, i've not looked into that.

arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
```diff
@@ -133,7 +133,9 @@ static int off(void);
 static rfc_propRxOutput_t rx_stats;
 /*---------------------------------------------------------------------------*/
 /* Defines and variables related to the .15.4g PHY HDR */
-#define DOT_4G_MAX_FRAME_LEN    2047
+#ifndef DOT_4G_MAX_FRAME_LEN
+  #define DOT_4G_MAX_FRAME_LEN    2047
```
atiselsts, line 137: Do you feel that changing this constant leads to any difference in performance? In any case, the value 2047 is defined by the standard, it feels wrong to make it user-configurable.

atiselsts:
I propose to replace this PR with https://github.com/contiki-ng/contiki-ng/pull/1557.

arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
lafka: Yes, the radio will attempt to receive maximum number of bytes before failing. It's clear that it's not an issue if all neighboring devices are well behaving 802.15.4 devices. There is some other device(s) at the site which sends data with the same PHY layer which triggers this. 

I get your point, i'm not sure if it holds true for all variations of 15.4 spec, in our cases there will never be frames larger than 128bytes. From previous projects we've seen just doubling the frame size and you get a higher of corruption/data loss (using 2047 will be downright impossible with low data rate and dense networks).

I think the practical solution would be to use the maximum supported payload size: `MIN(MAX_PAYLOAD_LEN_WITH_META, 2047)` or making in user configurable.

I'm open for any solution that can make the radio terminate early on packets it knows can't be processed.

atiselsts: I'm thinking we should set the value of `maxPktLen` equal to the maximum of actually supported packet size in terms of rx buffer space - basically what you propose. The value 2047 probably should just go in a comment next to the `MAX_PAYLOAD_LEN` constant definition.

lafka:
> I propose to replace this PR with #1557.

That would not really solve the issue. This PR ensures the max packet size in `CMD_PROP_RX_ADV` can be configured to match RX buffer size. This is what causes the higher success rate by letting the radio discard the packet (when to large/invalid) immediately and then continue looking for sync word.

> I'm thinking we should set the value of `maxPktLen` equal to the maximum of actually supported packet size in terms of rx buffer space - basically what you propose. The value 2047 probably should just go in a comment next to the `MAX_PAYLOAD_LEN` constant definition.

That would work.

arch/cpu/cc26x0-cc13x0/rf-core/prop-mode.c
lafka: That works out for me.

atiselsts: In the IEEE mode, we do not set this field at all - does this mean unlimited max packet?! This should be fixed then as well.

lafka: AFAIK the IEEE mode has no parameter to set the maximum supported length.

atiselsts:
I have updated 1557. Can you test this, if you got the time?

atiselsts:
closing as the changes are now incorporated in #1557

lafka:
> I have updated 1557. Can you test this, if you got the time?

Yeah, that should do it! Thanks. I'll give it a try ASAP (most likely next week or so).

atiselsts:
@lafka were you able to test this?

PR #1530, opened 2021-03-12
Change what text files we consider for the doxygen build

g-oikonomou:
Our current doxygen build assumes that any file ending in `*.txt` is a candidates of containing doxygen documentation. This list of files is getting unnecessarily long, especially as a result of various text files shipped as part of our various submodules. 

This pull changes our doxygen build to only consider files with a name of a `doxygen*.txt` pattern.

In order for this to work we move some existing documentation from text files to header files, and we also do some file renaming.

Motivated by the discussion in #1470


alexstanoev (approved):
Tested, does not affect the generated docs.

PR #1537, opened 2021-03-30
MAC: add MAC_TX_QUEUE_FULL error code, and account it in link stats separately

atiselsts:
Allows to collect statistics on queue losses, which are important for network performance evaluation.

### DIFF
--- a/os/net/link-stats.c
+++ b/os/net/link-stats.c
@@ -137,7 +137,7 @@ link_stats_packet_sent(const linkaddr_t *lladdr, int status, int numtx)
   uint8_t ewma_alpha;
 #endif /* !LINK_STATS_ETX_FROM_PACKET_COUNT */
 
-  if(status != MAC_TX_OK && status != MAC_TX_NOACK) {
+  if(status != MAC_TX_OK && status != MAC_TX_NOACK && status != MAC_TX_QUEUE_FULL) {
     /* Do not penalize the ETX when collisions or transmission errors occur. */
     return;
   }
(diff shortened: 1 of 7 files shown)


atiselsts:
In order to see a demonstration of this functionality, you can check the `examples/benchmarks/result-visualization` example - it has `LINK_STATS_CONF_PACKET_COUNTERS` configuration enabled, so it now collects statistics about the queue losses and prints them. `run-cooja.py` followed by `run-analysis.py` should print the queue losses (among other stats).

os/net/link-stats.c
```diff
@@ -166,6 +166,14 @@ link_stats_packet_sent(const linkaddr_t *lladdr, int status, int numtx)
   stats->last_tx_time = clock_time();
   stats->freshness = MIN(stats->freshness + numtx, FRESHNESS_MAX);
 
+  if(status == MAC_TX_QUEUE_FULL) {
```
nfi, line 169: Full queues were earlier reported as `MAC_TX_ERR` which did not update last timestamp and freshness (`stats->last_tx_time` and `stats->freshness`). What is the motivation of changing this behaviour?

atiselsts: A good point - this PR is not meant to change that logic. Don't remember exactly why it was made this way, probably an oversight on my part.

Moving this `if` statement to line 166 should be sufficient to get the original behavior, right?

atiselsts: changed the code by moving the `if` block before the freshness update logic

nfi (approved):

joakimeriksson:
We need to remember to add this to the change-list.

PR #1541, opened 2021-04-01
Fix out-of-bounds read in SNMP

nvt:
When looping over the <code>varbinds</code> array, the element index is checked against the maximum value only after reading the array from that index. This PR fixes the problem by moving the checks before the read.


yatch (approved):
It looks good to me.

PR #1543, opened 2021-04-07
Correctly identify the root when using MRHOF

mahboobkarimian:
Using MRHOF, the root rank is 128, otherwise 256.
To correctly identify the root (by rank) we need to adjust it depending on using / not using the MRHOF as an objective function.


atiselsts (approved):
Look good to me.

PR #1555, opened 2021-04-30
Remove python, pip and kept python3.9 and pip3

Yagoor:
# Description

Python2 and pip2 are completely removed
Python3.8 is installed instead of python3.6 from package manager.
Pip3 is installed using get-pip and locked to version 21.0.1

# Warning

**This PR completely removed python2 from the docker environment!**

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -9,6 +9,26 @@ USER root
 RUN apt-get -qq update && \
     apt-get -qq -y --no-install-recommends install gnupg ca-certificates software-properties-common > /dev/null
 
+# This must be ran before installing other programs.
+# They might install python3 from package manager which is python3.6
+# Install python3.9 from deadsnakes
+# Use get-pip to install pip instead of using package manager
+# Install pip 21.0.1 to avoid warning introduced in 21.1
+# https://pip.pypa.io/en/stable/news/#v21-1
+RUN apt-get -qq update \
+    && apt-get -qq install -y software-properties-common > /dev/null \
+    && add-apt-repository -y ppa:deadsnakes/ppa \
+    && apt-get -qq update \
+    && apt-get -qq install -y \
+        wget \
+        python3.9-dev \
+        python3.9-distutils \
+        > /dev/null \
+    && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.9 1 \
+    && wget -nv https://bootstrap.pypa.io/pip/3.5/get-pip.py \
+    && python3 get-pip.py pip==21.0.1 \
+    && rm get-pip.py
+
 RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF && \
   echo "deb http://download.mono-project.com/repo/ubuntu stable-bionic main" | tee /etc/apt/sources.list.d/mono-xamarin.list && \
   add-apt-repository ppa:mosquitto-dev/mosquitto-ppa && \
@@ -32,13 +52,6 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E03280
     mosquitto-clients \
     net-tools \
     openjdk-11-jdk \
-    python-dev \
-    python3-dev \
-    python-pip \
-    python3-pip \
-    python3-setuptools \
-    python-serial \
-    python3-serial \
     rlwrap \
     sudo \
     screen \
@@ -99,15 +112,15 @@ RUN wget -nv https://www.nordicsemi.com/-/media/Software-and-other-downloads/Des
 
 # Install sphinx and sphinx_rtd_theme, required for building and testing the
 # readthedocs API documentation
-RUN pip -q install --upgrade pip
-RUN pip -q install setuptools && pip -q install sphinx_rtd_theme sphinx
 # Install matplotlib for result visualization
-RUN pip3 -q install --upgrade pip
-RUN pip3 -q install matplotlib
-
-# Install nrfutil and work around broken pc_ble_driver_py dependency
-RUN pip3 -q install nrfutil && \
-  pip3 -q install --no-deps -t /usr/local/lib/python3.6/dist-packages --python-version 3.6 --ignore-requires-python --upgrade nrfutil
+# Install nrfutil
+RUN pip3 -q install \
+      setuptools \
+      sphinx_rtd_theme \
+      sphinx \
+      matplotlib \
+      nrfutil \
+      pyserial
 
 # Make sure we're using Java 11 for Cooja
 RUN update-java-alternatives --set /usr/lib/jvm/java-1.11.0-openjdk-i386


g-oikonomou:
Hi @Yagoor we're just having a chat about this. What is the motivation behind installing python 3.9 as opposed to an earlier 3.x?

Yagoor:
HI @g-oikonomou.

The reason behind was based on the fact that the nrfutil pip package was not compatible with the python3.6 (There was a workaround for it in the Dockerfile). This led me to choose a newer version 3.7+. The latest version is version 3.10, but I choose 3.9 just to be safe (I hope it's more stable than 3.10).

I can try with an older version but since python is constantly updating it might be a problem later on.

atiselsts:
I thin we should aim to have this in 3.8.

There are some conflicts in the Dockerfile now, can you solve them? Thanks!

Yagoor:
@atiselsts: When you say 3.8 ,do you mean that I should use Python 3.8 or do you mean Contiki-NG 4.8?
I'll fix the conflicts later today.

Yagoor:
Alright. I'll fix the conflicts and squash the commits later, :)

Yagoor:
Some link is broken. I'll need to fix that too.

tools/docker/Dockerfile
```diff
@@ -9,6 +9,26 @@ USER root
...
+# They might install python3 from package manager which is python3.6
+# Install python3.9 from deadsnakes
+# Use get-pip to install pip instead of using package manager
+# Install pip 21.0.1 to avoid warning introduced in 21.1
+# https://pip.pypa.io/en/stable/news/#v21-1
+RUN apt-get -qq update \
+    && apt-get -qq install -y software-properties-common > /dev/null \
+    && add-apt-repository -y ppa:deadsnakes/ppa \
+    && apt-get -qq update \
+    && apt-get -qq install -y \
```
pjonsson, line 22: Does the install need to be done before the installation of all the other packages, or could the regular install just contain "python3.9-dev python3-distutils"? If it needs to happen before, that information is probably worth putting in a comment.

pjonsson:
The initial pull request used Python 3.8 which is the default in Ubuntu 20.04 LTS. If the merge-team is considering to upgrade the docker image to Ubuntu 20.04 sometime in the future it might be worth to stick to Python 3.8.

Yagoor:
> The initial pull request used Python 3.8 which is the default in Ubuntu 20.04 LTS. If the merge-team is considering to upgrade the docker image to Ubuntu 20.04 sometime in the future it might be worth to stick to Python 3.8.

I can try version 3.8, I believe there was some package incompatible, but I'm not 100% sure. 
Ubuntu 22.04 LTS uses Python 3.10, but I know for a fact that nrfutil is not compatible with it, yet.

@atiselsts: Do you have any preference on the version here?

pjonsson:
> > The initial pull request used Python 3.8 which is the default in Ubuntu 20.04 LTS. If the merge-team is considering to upgrade the docker image to Ubuntu 20.04 sometime in the future it might be worth to stick to Python 3.8.
> 
> I can try version 3.8, I believe there was some package incompatible, but I'm not 100% sure. Ubuntu 22.04 LTS uses Python 3.10, but I know for a fact that nrfutil is not compatible with it, yet.

32-bit PPA binaries aren't built for Ubuntu 19 and later, so upgrading past 18.04 requires going to 64-bit image. My understanding is that a 64-bit image would be beneficial for Apple M1 as well, but going to a 64 bit image requires completing the 64-bit support in Cooja. I have a JVM crash on the Contiki-NG regression tests on my Linux computer that @nfi does not have on his (older) Linux computer.

I looked into my Cooja 64-bit issue today and it's reading and writing in unexpected places which causes getMemory/setMemory to eventually crash, and this does not happen in 32-bit mode. I'm not sure why, there might be differences with object alignment between 32 and 64-bit binaries that Cooja doesn't cater for.

Cooja also does not work out of the box with GCC 10+ and Clang 11+ because they enable -fno-common and that causes the map file to not have the size information for symbols that Cooja needs. The gcc package in Ubuntu 22.04 is version 11.2 and the Clang package is 14.0.

Since the map file reading in Cooja will become obsolete, I decided to start in that end rather than trying to figure out alignment differences between 32- and 64-bit binaries and possible compensations for those. I ran into other issues in that code so I've put up a pull request for those.

This isn't trying to discourage going to 64-bit, I think it's the right thing to do and I have dockerfile updates in my local repository for doing that. Realistically though, that requires merging quite a few fixes in other places.

Yagoor:
> > > The initial pull request used Python 3.8 which is the default in Ubuntu 20.04 LTS. If the merge-team is considering to upgrade the docker image to Ubuntu 20.04 sometime in the future it might be worth to stick to Python 3.8.
> > 
> > 
> > I can try version 3.8, I believe there was some package incompatible, but I'm not 100% sure. Ubuntu 22.04 LTS uses Python 3.10, but I know for a fact that nrfutil is not compatible with it, yet.
> 
> 32-bit PPA binaries aren't built for Ubuntu 19 and later, so upgrading past 18.04 requires going to 64-bit image. My understanding is that a 64-bit image would be beneficial for Apple M1 as well, but going to a 64 bit image requires completing the 64-bit support in Cooja. I have a JVM crash on the Contiki-NG regression tests on my Linux computer that @nfi does not have on his (older) Linux computer.
> 
> I looked into my Cooja 64-bit issue today and it's reading and writing in unexpected places which causes getMemory/setMemory to eventually crash, and this does not happen in 32-bit mode. I'm not sure why, there might be differences with object alignment between 32 and 64-bit binaries that Cooja doesn't cater for.
> 
> Cooja also does not work out of the box with GCC 10+ and Clang 11+ because they enable -fno-common and that causes the map file to not have the size information for symbols that Cooja needs. The gcc package in Ubuntu 22.04 is version 11.2 and the Clang package is 14.0.
> 
> Since the map file reading in Cooja will become obsolete, I decided to start in that end rather than trying to figure out alignment differences between 32- and 64-bit binaries and possible compensations for those. I ran into other issues in that code so I've put up a pull request for those.
> 
> This isn't trying to discourage going to 64-bit, I think it's the right thing to do and I have dockerfile updates in my local repository for doing that. Realistically though, that requires merging quite a few fixes in other places.

Oh, I see. I didn't know about the 32-bit PPA support.
It turns out that it is no longer simpler to install different  python on Ubuntu 18.04. 
_software-properties-common_ requires python, which installs python 3.6. I cannot simple purge it after installing python 3.9 (I tried)

I don't think this PR is useful anymore, in fact, it is good that it was not merged else we would have to figure a way out of this situation. 

Bottom line is, I cannot avoid installing Python3.6 anymore which is why all checks failed.

atiselsts:
Should we close this PR then? Or postpone it?

pjonsson:
Is this issue fixed in the latest docker image after the update to Ubuntu 20?

pjonsson:
The docker image runs Ubuntu 22.04 so we have a fairly recent Python version now, can this be closed @Yagoor?

Yagoor:
@pjonsson for sure. I'll close it.

PR #1557, opened 2021-05-05
CC radios: make waiting time in read_frame dependent on the maximum packet size instead of using hardcoded values and fix maxPktLen for CC13xx

atiselsts:
The discussion and motivation for this is in #1528 .

TL;DR: 
1. Make the waiting time in read_frame() dependent on RADIO_BYTE_AIR_TIME multiplied by the sum of RADIO_CONST_PHY_OVERHEAD and MAX_PAYLOAD_LEN (and round the result up in rtimer ticks, and add some small guard time). It will work correctly and efficiency in different datarates, unlike the current approach which uses a constant time.
2. Set `cmd_rx_adv->maxPktLen` to the actual max frame size supported by the current config of the Contiki-NG network stack. Previously the value 2047 from the standard was used. This lead to receivers occasionally being on for a very long time while up to 2047 bytes were received, even when the packet would be dropped anyway because of the network stack not being able to process it.

### DIFF
--- a/arch/cpu/cc26x0-cc13x0/cc13xx-cc26xx-def.h
+++ b/arch/cpu/cc26x0-cc13x0/cc13xx-cc26xx-def.h
@@ -71,6 +71,9 @@
 
 #define RADIO_BYTE_AIR_TIME  (1000000 / (RADIO_BIT_RATE / 8))
 
+#define RADIO_FRAME_DURATION(payload_len)                               \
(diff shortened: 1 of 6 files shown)


alexrayne:
What the purpose of it?

atiselsts:
@alexrayne the discussion is here: https://github.com/contiki-ng/contiki-ng/pull/1528

g-oikonomou (approved):

PR #1563, opened 2021-05-07
Add spinlock on LDREXB/STREXB-pair for cortex mutex lock

### LINKED ISSUE #1544: Mutex lock fail on ARM cortex
Hi,
on our cc13xx platform we are seeing a larger amount of mutex-lock failures than expected.

It seems this happens not because the mutex is already locked, but because an interrupt happens after LDREXB and before STREXB in https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/arm/cortex-m/mutex-cortex.h#L60. In our case, the interrupt does not touch the mutex, but ARM manual A3.4.4 states that a context switch may cause STREX to fail. To avoid this I was looking at adding a spinlock on the LDREXB + STREXB procedure (not on the entire lock), something ala:

```
static inline bool
mutex_cortex_try_lock(volatile mutex_t *mutex)
{
  do {
    if(__LDREXB(mutex) != 0) {
      return false;
    }
  } while(__STREXB(1, mutex) != 0);
  
  __DMB();

  return true
}
```

This way the lock-attempt still immediately returns if the lock is actually taken, yet it spins at attempts to lock an available lock.

I could not immediately imagine any problems or possible eternal spinning (except when continuously interrupted) from this, but I am seldom at this low level so any input appreciated.

arurke:
This PR fixes #1544 by adding a spinlock to the LDREXB+STREXB procedure. This avoids mutex lock fail in cases where we are  interrupted after LDREXB but ISR does not actually lock the mutex (see details in #1544). Note that this does not add a spinlock to the lock itself, if it is actually taken, we still immediately return.

### DIFF
--- a/arch/cpu/arm/cortex-m/mutex-cortex.h
+++ b/arch/cpu/arm/cortex-m/mutex-cortex.h
@@ -59,15 +59,15 @@ typedef uint8_t mutex_t;
 static inline bool
 mutex_cortex_try_lock(volatile mutex_t *mutex)
 {
-  int status = 1;
-
-  if(__LDREXB(mutex) == 0) {
-    status = __STREXB(1, mutex);
-  }
+  do {
+    if(__LDREXB(mutex) != 0) {
+      return false;
+    }
+  } while(__STREXB(1, mutex) != 0);
 
   __DMB();
 
-  return status == 0 ? true : false;
+  return true;
 }
 /*---------------------------------------------------------------------------*/
 static inline void


g-oikonomou:
Hi @arurke we are looking at this with @alexstanoev 

> Note that this does not add a spinlock to the lock itself, if it is actually taken, we still immediately return.

But if we get interrupted and the lock _is_ actually taken, then we will spin. But of course if a handler wants a mutex, one would expect the handler to also release the mutex...

arurke:
Hmmm, not following you:

1. We find mutex is unlocked
2. We get interrupted and ISR locks
3. _ISR does not unlock_
4. `__STREXB` fails because we were interrupted
5. Spin. We now find mutex is locked. **Return**.

------

1. We find mutex is unlocked
2. We get interrupted and ISR locks
3. _ISR does unlock_
4. `__STREXB` fails because we were interrupted.
5. Spin. We again find mutex is unlocked.
6. Lock successful. **Return**.

Some scenario I am missing here? Obviously if we continuously get interrupted we will spin, but at that point we have other concerns imo :)

g-oikonomou:
Yes, thanks. I think Alex and I are reaching the same conclusion here.

https://community.arm.com/developer/ip-products/processors/f/cortex-m-forum/10361/ldrex-strex-on-the-m3-m4-m7

* `__LDREXB`
* interrupt
* `__LDREXB` /* Does this succeed, or does it not? */

I _think_ the 2nd `__LDREXB` above will succeed because an exception clears the exclusive state.

So if we get interrupted at the first pass, the situation will get resolved at the 2nd pass: We will either succeed or fail, but in both case we will return. Unless as you say we get interrupted again, in which case again as you say we should be having a different conversation 😃

alexstanoev (approved):

g-oikonomou:
As @alexstanoev pointed out, this change also makes the semantics of the function more robust. Returning false means "I failed", as opposed to "either I failed or I got unlucky"

PR #1577, opened 2021-05-24
DHT11 Sensor

Yagoor:
# Description

In this PR I implemented the DHT11 sensor which is a low cost humidity and temperature sensor.

Product: https://www.mouser.dk/ProductDetail/OSEPP-Electronics/HUMI-01?qs=wNBL%252BABd93P6g2NY71n2BA%3D%3D
Datasheet: https://www.mouser.com/datasheet/2/758/DHT11-Technical-Data-Sheet-Translated-Version-1143054.pdf

There is an implementation of the DHT22 in the Contiki-NG code base but this implementation is very bound to the Zoul platform.

Since the sensor requires granularity of 1us and the rtimer granularity is not so small, a few tricks were used to get the timing right. I used an nrf 52840 DK to test this but since I'm using the rtimer functions to calculate ticks it should work in any platform.

### DIFF
--- a/arch/dev/sensor/dht11/dht11-sensor.c
+++ b/arch/dev/sensor/dht11/dht11-sensor.c
@@ -1,0 +1,370 @@
+/*
+ * Copyright (C) 2021 Yago Fontoura do Rosario <yago.rosario@hotmail.com.br>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


g-oikonomou (approved):

atulp201:
How do I apply it for reading the sensor from CC2650 Launchpad, while running cc26x0-web-demo example, your guidance is highly appreciated.

PR #1584, opened 2021-06-01
result visualization example: fix the run-cooja.py script

atiselsts:


nfi (approved):
Looks good. Thanks for fixing!

PR #1589, opened 2021-06-04
Verify DIO prefix info lengths in RPL-Classic

nvt:
By sending a DIO with a too large length in the prefix info option, it is possible to cause a buffer overflow when copying the prefix in the <code>set_ip_from_prefix</code> function. In this PR, we insert a validation check that discards the DIO if this value is too large.

### DIFF
--- a/os/net/routing/rpl-classic/rpl-icmp6.c
+++ b/os/net/routing/rpl-classic/rpl-icmp6.c
@@ -447,6 +447,14 @@ dio_input(void)
           goto discard;
         }
         dio.prefix_info.length = buffer[i + 2];
+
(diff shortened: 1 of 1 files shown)


joakimeriksson (approved):

sei-vsarvepalli:
Hello @joakimeriksson  and @nvt 

Thank you for a critical update and release. 

Are these in any way related to the bugs reported by Forescout? 
https://www.kb.cert.org/vuls/id/815128 

Were there any CVE's to identify the vulnerabilities so they can be tracked?

Thanks 
Vijay
CERT/CC

joakimeriksson:
Yes, some of the fixes in the release do fix some of the reported bugs. We have started to publish some of the CVEs here in our repo also. Take a look at the security section.

PR #1593, opened 2021-06-05
Fix DHT-11 compilation on all targets

Yagoor:
DHT11 needs some macros and the gpio-hal.h which are not implemented in all targets


atiselsts (approved):

atiselsts:
Can you squash the 2 commits in a single one here, before we merge this?

PR #1596, opened 2021-06-09
Update README.md

atiselsts:
1) Add a link to the GitHub releases / changelog
2) Remove the link to testbed runs, as the testbed has not been active since January 2020


joakimeriksson (approved):
Lets get the testbed runs back - but agree on removing the reference!

PR #1598, opened 2021-06-15
Avoid division by 0 due to lifetimes and units of zero in RPL Classic

nvt:
If DIO messages with 0 lifetimes are accepted, it can lead to a division by zero when calculating the interval timer length. This PR simply increases the values to 1 in those cases.


nfi (approved):
Looks good!

PR #1607, opened 2021-07-01
Fix hexconv_hexlify()

simonduq:
That one has bothered me for a while but somehow never took time to do the PR :)


joakimeriksson (approved):
I agree with you totally here - there is no need for two zeros!

PR #1609, opened 2021-07-03
Integrated tinyusb as USB stack for the nrf port

Yagoor:
# TODO


Yagoor:
Hi @alexstanoev,

I think we have a problem using the tinyusb as a submodule.
The original repo has too many submodules, it increase the checkout time from 25 seconds to 1.45 minutes. It also breaks the documentation build because of the too many subdirectories.

I propose we make a fork of the tinyusb under the Contiki-ng user and we remove the submodules from it.

alexstanoev:
Hi @Yagoor,

I recently came across this source-only mirror of tinyusb (no vendor SDKs), which seems perfect for linking as a submodule:
https://github.com/hathach/tinyusb_src 

Do you see any issues with that approach?

mbaddeley:
Hi @Yagoor, are you interested in taking this forward? I'd be happy to help out.

Yagoor:
Hi @mbaddeley,
The only missing functionality is the DFU on the dongle because nordic does not use the USB DFU standard, so we need to implement it.
Else, this requires a rebase and use the new submodule suggested by @alexstanoev , it should be enough for us. I'll do the rebasing and the new submodule. I don't believe I'll have time for the DFU part soon.

joakimeriksson:
Could we do a removal of work-in-progress state and merge this soon? Would it make sense to add the DFU support in a another PR?

alexstanoev:
I agree, we should handle DFU trigger in a separate PR. 
This looks good for review to me!

Yagoor:
Alright. Then let's gets this in. :) 
I'll add the DFU support to my list.

pjonsson:
@Yagoor I think the build system is better at giving warnings now than it was when you pushed this change, could you rebase  against the latest develop branch?

Yagoor:
> @Yagoor I think the build system is better at giving warnings now than it was when you pushed this change, could you rebase against the latest develop branch?

Alright. Sorry for the delay, I just rebased it.

arch/cpu/nrf/Makefile.nrf
```diff
@@ -73,9 +75,67 @@ NRFX_INC_PATHS += templates
...
+  TINYUSB_C_SRCS += $(TINYUSB_ROOT)/src/portable/nordic/nrf5x/dcd_nrf5x.c
+
+  TINYUSB_INC_PATHS += src/portable/nordic/nrf5x/
+
+  #source common
+  TINYUSB_C_SRCS += $(TINYUSB_ROOT)/src/tusb.c
+  TINYUSB_C_SRCS += $(TINYUSB_ROOT)/src/common/tusb_fifo.c
+  TINYUSB_C_SRCS += $(TINYUSB_ROOT)/src/device/usbd.c
+  TINYUSB_C_SRCS += $(TINYUSB_ROOT)/src/device/usbd_control.c
+  TINYUSB_C_SRCS += $(TINYUSB_ROOT)/src/class/audio/audio_device.c
```
alexstanoev, line 95: Is there a reason for building unused classes?

arch/cpu/nrf/dev/slip-arch.c
```diff
@@ -31,17 +31,26 @@
...
 #include "dev/slip.h"
-#include "dev/uarte-arch.h"
+#include "uarte-arch.h"
+#include "usb.h"
 /*---------------------------------------------------------------------------*/
 void
 slip_arch_writeb(unsigned char c)
 {
+#if PLATFORM_DBG_CONF_USB
+  usb_write(&c, sizeof(c));
```
alexstanoev, line 41: This, as well as dbg-arch.c, will end up calling flush for every single byte.

I suggest handling this in the same way as the cc2538 and nrf52840 platforms:
https://github.com/contiki-ng/contiki-ng/blob/343e25144228fc6997cf6d02f4ffd4b0b180764f/arch/cpu/nrf52840/slip-arch.c#L57-L60

arch/cpu/nrf/nrf-def.h
```diff
@@ -77,21 +84,10 @@
...
-#endif /* TSCH_CONF_RADIO_ON_DURING_TIMESLOT */
-
-#ifndef TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS
-#define TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS 1
-#endif /* TSCH_CONF_RESYNC_WITH_SFD_TIMESTAMPS */
-
-#ifndef TSCH_CONF_TIMESYNC_REMOVE_JITTER
-#define TSCH_CONF_TIMESYNC_REMOVE_JITTER 0
-#endif /* TSCH_CONF_TIMESYNC_REMOVE_JITTER */
+#define TSCH_CONF_HW_FRAME_FILTERING          0
```
alexstanoev, line 87: What's the reason for removing the ifndef checks?

arch/cpu/nrf/os/dbg-arch.c
```diff
@@ -48,31 +48,41 @@
 #include "contiki.h"
 
 #include "uarte-arch.h"
+#include "usb.h"
 /*---------------------------------------------------------------------------*/
 unsigned int
 dbg_send_bytes(const unsigned char *s, unsigned int len)
 {
+#if PLATFORM_DBG_CONF_USB
+  usb_write((uint8_t *)s, len);
```
alexstanoev, line 57: I suggest handling the flush separately from the write here as well, similar to:
https://github.com/contiki-ng/contiki-ng/blob/629b374eafb850caba7dc6d0330caa336b2ab037/arch/cpu/nrf52840/dbg.c#L83

arch/cpu/nrf/os/dbg-arch.c
```diff
@@ -48,31 +48,41 @@
...
+
+
 }
 /*---------------------------------------------------------------------------*/
 int
 dbg_putchar(int c)
 {
-#if PLATFORM_HAS_UARTE
+#if PLATFORM_DBG_CONF_USB
+  usb_write((uint8_t *)&c, sizeof(c));
```
alexstanoev, line 82: This will call flush on every byte.

arch/cpu/nrf/usb/usb.c
```diff
@@ -0,0 +1,127 @@
...
+
+  /* Initialize the usb process */
+  process_start(&usb_arch_process, NULL);
+}
+/*---------------------------------------------------------------------------*/
+void
+usb_write(uint8_t *buffer, uint32_t buffer_size)
+{
+  tud_cdc_write(buffer, buffer_size);
+  tud_cdc_write_flush();
```
alexstanoev, line 86: `tud_cdc_n_write_flush` returns the number of bytes written - this can be less than what's queued in the internal FIFO and might cause data loss if not handled - I haven't checked if tinyusb handles draining TX buffers internally.

arch/cpu/nrf/usb/usb.c
```diff
@@ -0,0 +1,127 @@
...
+  usb_arch_init();
+
+  /* Initialize the usb process */
+  process_start(&usb_arch_process, NULL);
+}
+/*---------------------------------------------------------------------------*/
+void
+usb_write(uint8_t *buffer, uint32_t buffer_size)
+{
+  tud_cdc_write(buffer, buffer_size);
```
alexstanoev, line 85: This call could write less than `buffer_size` if the underlying FIFO is full.

Flushing when full seems to be partly handled internally, but it might still leave partial data from the last write?
https://github.com/hathach/tinyusb/blob/4f6e770eda9f8c605b238f35def1edaff93a8fb8/src/class/cdc/cdc_device.c#L168-L180

arch/cpu/nrf/usb/usb.c
```diff
@@ -0,0 +1,127 @@
...
+  tud_cdc_write_flush();
+
+  /* Call the task to handle the events immediately */
+  tud_task();
+}
+/*---------------------------------------------------------------------------*/
+void
+cdc_task(void)
+{
+  unsigned char usb_buffer[64];
```
alexstanoev, line 95: I suggest marking this as static to move 64 bytes of stack usage into bss.

arch/cpu/nrf/usb/usb.h
```diff
@@ -0,0 +1,94 @@
...
+ * @brief Initialize the architecture specific USB driver
+ *
+ */
+void usb_arch_init(void);
+/*---------------------------------------------------------------------------*/
+/**
+ * @brief Boot device into DFU
+ * 
+ */
+void usb_arch_reboot_to_dfu(void);
```
alexstanoev, line 87: This isn't defined anywhere.

arch/cpu/nrf/usb/usb_descriptors.c
```diff
@@ -0,0 +1,204 @@
...
+// String Descriptors
+//--------------------------------------------------------------------+
+
+// array of pointer to string descriptors
+char const* string_desc_arr [] =
+{
+  (const char[]) { 0x09, 0x04 }, // 0: is supported language is English (0x0409)
+  "TinyUSB",                     // 1: Manufacturer
+  "TinyUSB Device",              // 2: Product
+  "123456",                      // 3: Serials, should use chip ID
```
alexstanoev, line 162: This should be populated from DEVICEADDR in FICR to allow nrfutil to follow the device as it resets into DFU mode:

https://github.com/contiki-ng/nrf-sdk-contiki-ng/blob/66a3b16198ffad356fdf5cf8699c51c0a0b363ac/components/libraries/usbd/app_usbd_serial_num.c#L71-L84

alexstanoev (commented):
Thanks again for working on this! :)
I've left some comments, mostly around handling flush on writes.

This PR creates another use case for IPC support in Contiki-NG, as the nRF53 USB controller is an app core peripheral, so SLIP over USB won't work there. 
That won't be trivial and is definitely a problem for another day! :)

arch/cpu/nrf/Makefile.nrf
Yagoor: I think this was just leftover. I'll remove all classes except cdc which is the only one being used.

arch/cpu/nrf/nrf-def.h
Yagoor: Thanks for the catch. I don't know why I did this.

arch/cpu/nrf/dev/slip-arch.c
Yagoor: Thank you. Done.

arch/cpu/nrf/nrf-def.h
Yagoor: Reverted

arch/cpu/nrf/os/dbg-arch.c
Yagoor: Thank you. Done.

arch/cpu/nrf/os/dbg-arch.c
Yagoor: Fixed. :)

arch/cpu/nrf/usb/usb.c
Yagoor: As you said, it handles partially. So I did the handling in this function too.

arch/cpu/nrf/usb/usb.c
Yagoor: Fixed.

arch/cpu/nrf/usb/usb.h
Yagoor: Yes, I forgot to remove it. Done.

arch/cpu/nrf/usb/usb_descriptors.c
Yagoor: Done. I made the descriptors configurable and the Serial should be done by the arch.

arch/cpu/nrf/usb/usb.c
Yagoor: Good point. I did that.

Yagoor:
> Thanks again for working on this! :) I've left some comments, mostly around handling flush on writes.

Thank you for the feedback. I believe I addressed all of them. I kept them into multiple commits to make your review easier. I'll squash once the review is over.

> This PR creates another use case for IPC support in Contiki-NG, as the nRF53 USB controller is an app core peripheral, so SLIP over USB won't work there. That won't be trivial and is definitely a problem for another day! :)

I didn't even try to address IPC. :D 
I think this it would deserve a good design before implementation. 
I know FreeRTOS supports SMP (only for ExpressIF) and zephyr supports IPC, so there are some good references out there.

arch/cpu/nrf/dev/usb-arch.c
```diff
@@ -76,8 +77,8 @@ power_event_handler(nrfx_power_usb_evt_t event)
 void
 usb_arch_init(void)
 {
-  const uint16_t serial_num_high_bytes = (uint16_t)NRF_FICR->DEVICEADDR[1] | 0xC000; // The masking makes the address match the Random Static BLE address.
-  const uint32_t serial_num_low_bytes  = NRF_FICR->DEVICEADDR[0];
+  const uint16_t serial_num_high_bytes = nrf_ficr_deviceid_get(NRF_FICR, 1) | 0xC000;
```
alexstanoev, line 80: Note DEVICEID and DEVICEADDR are different registers in FICR (https://infocenter.nordicsemi.com/pdf/nRF52840_PS_v1.1.pdf#%5B%7B%22num%22%3A160%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C85.039%2C774.623%2Cnull%5D).

There doesn't seem to be an nrfx method to read DEVICEADDR.

alexstanoev:
Aside from the FICR HAL detail, everything looks good to me! :) 
I'll be able to test this on actual hardware in a few days, I'll let you know.

I was mostly thinking out loud about IPC, it would definitely require a solid design first.
Zephyr's OpenAMP RPMsg approach seems like a good candidate though!

arch/cpu/nrf/dev/usb-arch.c
Yagoor: Yes, but `DEVICEADDR` only exists in in boards with radio, so the application core does not have it, while `DEVICEID` exists in all boards. `DEVICEID` is also an unique identifier according to the nrf5340 data sheet (https://infocenter.nordicsemi.com/pdf/nRF5340_OPS_v0.5.pdf). Is there a reason to specifically use `DEVICEADDR`? I know that it is used in the SDK but the SDK does not support nrf5340.

alexstanoev: Good point about non-radio cores!

On the nRF52840, if I recall correctly, the DFU bootloader will generate the SN from DEVICEADDR. The matching serial numbers are used by nrfutil to find the bootloader port after a DFU reset.

Yagoor: I'll do some research on this. I believe nordic came up with something for this. :)

Yagoor: It seems that you are correct, based on the nrf52840 port I got the following:

```
Bus 020 Device 020: ID 1915:520f Nordic Semiconductor ASA nRF52 USB CDC on Contiki-NG  Serial: E46F5CD692A5
```

```
Bus 020 Device 021: ID 1915:521f Nordic Semiconductor ASA Open DFU Bootloader  Serial: E46F5CD692A5
```

and we know that the nRF5 SDK uses the `DEVICEADDR` to generate the serial.

May I suggest to leave it using the `DEVICEID` until we add DFU support? 
I'm curious on how nordic intends to handle this for the nrf5340. 
I'd say this is a design mistake on their side to make the USB serial depend on the `DEVICEADDR` which is available on boards with radio.

pjonsson:
Is testing on hardware the only thing that remains, or what is the status for this PR?

Yagoor:
> Is testing on hardware the only thing that remains, or what is the status for this PR?

Correct.

alexstanoev:
All good on the second hardware test too! 
Using `DEVICEID` for serial numbers is reasonable until nRF52840 DFU trigger support gets added, at which point we'd have to handle it the same as the nRF5 SDK for that family.

My last comment is that `TINYUSB` should default to 0 to maintain consistency between platforms and decrease resource usage. 
This could be exposed as a compile flag such as `NRF_NATIVE_USB` which can automatically enable `PLATFORM_DBG_CONF_USB` too.
The option should then be mentioned in the docs under `doc/platforms/nrf.md`.

I'm happy to merge as-is and sort out these changes in a new PR if there's interest in getting this merged quicker :)

joakimeriksson:
I would vote on merging now and then try to do the fixes later!

Yagoor:
@alexstanoev Thank you for the feedback.
I'll do that change in another PR and I'll also update the documentation.
At the time of this PR the docs was not yet in the repository. :)

Issue #1510, opened 2021-02-24
simple-udp.c: client_process is never set


kluchsinger:
In simple-udp.c, the process context is set to client_process right before calling the receive_callback:
[see here](https://github.com/contiki-ng/contiki-ng/blob/48a3799e2d2c52e91cb7153b163bb26dfd9a4b4f/os/net/ipv6/simple-udp.c#L163)

However, client_process is not set anywhere.

If I understand it correctly, it should be set in simple_udp_register():
`c->client_process = PROCESS_CURRENT();`

Or am I misunderstanding something here?

damien-v:
I think you are correct. Maybe submit a pull request and hopefully someone with sufficient rights will have look at it?

nvt:
Thanks for reporting this issue -- it has now been fixed.

PR #1612, opened 2021-07-06
simple-udp: set the client process at registration

damien-v:
fixes issue #1510 

In simple_udp_register(), the client process is never set, this PR fixes this.


nfi (approved):
Good catch!

PR #1615, opened 2021-07-09
Check RPL Target prefix length and buffer boundary.

nvt:
RPL-Classic accepts RPL Target options in DAO messages with invalid prefix lengths. Additionally, it does not check whether there are enough bytes in the packets to copy a prefix of this length. This PR adds checks for these conditions so that the DAO message can be discarded.


nfi (approved):
Important check to have. Thanks for fixing!

PR #1616, opened 2021-07-10
fix process exiting - vs immediate restart process after exiting.

alexrayne:
this patch inspirited by a couple of hours debuging nontrivial aplication bug - servicing protothreads are strangely hungs, and later i see that them not even starts as planed. process_start fails without any side-effects. and that was fails, since protothread try start service just processing event PROCESS_EVENT_EXITED, on same service exited.

What happend in current implementation:
1) all running-processes notifyes about exiting process. If some of them wants to restart exiting-process, `process_start` fails, since exiting-process still in running-list
2) exiting-process removes from runnig-list. 

With this patch have:
1) exiting-process notifyes with PROCESS_EVENT_EXITED.
2) exiting-process removes from runnig-list. 
3) all running-processes notifyes about exiting process. And there any of them can restart exited process.

### DIFF
--- a/os/sys/process.c
+++ b/os/sys/process.c
@@ -137,18 +137,6 @@ exit_process(struct process *p, struct process *fromprocess)
 
   if(process_is_running(p)) {
     /* Process was running */
-    p->state = PROCESS_STATE_NONE;
-
-    /*
-     * Post a synchronous event to all processes to inform them that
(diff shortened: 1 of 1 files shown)


alexrayne:
fixed trivial bug

joakimeriksson (approved):

nfi (approved):

Issue #1610, opened 2021-07-05
using LLSEC802154_CONF_ENABLED is unsecure


madsw:
When using link layer security on "non-TSCH" mode (e.g. cc1352 prop mode) packets are secured and decrypted, but unsecure communication is allowed.

There is two challenges:
1:

in csma_security_parse_frame() unsecure frames are not dropped:

`
  if(packetbuf_attr(PACKETBUF_ATTR_SECURITY_LEVEL) == 0) {
    /* No security - no more processing required */
    return hdr_len;
  }
`
This should result in a FRAMER_FAILED instead.

2:
The first DIO message is sent without security.
This will allow a unsecure node picking up the DIO and joining un-securely because of the first issue.
this is because output() in sicslowpan.c sets security level to 0 first time for some reason.

PR #1632, opened 2021-09-06
Reject unencrypted frames in csma_security_parse_frame() (Issue 1610)

David-Wallis-Seneco:
When using link layer security on "non-TSCH" mode (e.g. cc1352 prop mode) packets are secured and decrypted, but unsecure communication is allowed.

Issue 1 in #1610 states:

in csma_security_parse_frame() unsecure frames are not dropped:

if(packetbuf_attr(PACKETBUF_ATTR_SECURITY_LEVEL) == 0) { /* No security - no more processing required */ return hdr_len; }
This should result in a FRAMER_FAILED instead.

Code has been modified with a #define to opt to reject insecure packets, or leave as the original contiki-ng default. 

    #ifdef LLSEC802154_REJECT_INSECURE
    return FRAMER_FAILED;
    #else
    /* No security - no more processing required */
    return hdr_len;
    #endif

### DIFF
--- a/os/net/mac/csma/csma-security.c
+++ b/os/net/mac/csma/csma-security.c
@@ -238,8 +238,14 @@ csma_security_parse_frame(void)
   }
 
   if(packetbuf_attr(PACKETBUF_ATTR_SECURITY_LEVEL) == 0) {
+    /* From https://github.com/contiki-ng/contiki-ng/issues/1610 */
+    /* This should reject unencrypted packets */
+    #ifdef LLSEC802154_REJECT_INSECURE
+    return FRAMER_FAILED;
+    #else
     /* No security - no more processing required */
     return hdr_len;
+    #endif
   }
 
   LOG_INFO("LLSEC-IN: ");


alexrayne:
Why `This should result in a FRAMER_FAILED` ? 
applications have an packet attributes, that show security of received packet, so application can decide - reject it or accept.
how you intended implement mix of secured and unsecured transmitions with `return FRAMER_FAILED` ?

David-Wallis-Seneco:
> how you intended implement mix of secured and unsecured transmitions with `return FRAMER_FAILED` ?
The intended behaviour is to reject all unsecured messages, using the suggested implementation in #1610. The application will not use a mix of secured and unsecured transmissions. 

The change is implemented with an option (LLSEC802154_REJECT_INSECURE). Applications using a mix of secure and unsecured messages will not set this option and allow both.

dvseneco:
@alexrayne It is not only at user application that this occurs, also standard protocol messages are handled. I stumbled upon this issue with devices supposedly only joining my private network but anyone coming by with a unit set as root and encryption disabled would cause my devices to join. Since this behavior is standard contiki implementation it needs this fix to completely prevent unintended interaction with unencrypted networks.

damien-v:
@alexrayne It is not only at user application that this occurs, also standard protocol messages are handled. I stumbled upon this issue with devices supposedly only joining my private network but anyone coming by with a unit set as root and encryption disabled would cause my devices to join. Since this behavior is standard contiki implementation it needs this fix to completely prevent unintended interaction with unencrypted networks.

os/net/mac/csma/csma-security.c
```diff
@@ -238,8 +238,14 @@ csma_security_parse_frame(void)
   }
 
   if(packetbuf_attr(PACKETBUF_ATTR_SECURITY_LEVEL) == 0) {
+    /* From https://github.com/contiki-ng/contiki-ng/issues/1610 */
+    /* This should reject unencrypted packets */
+    #ifdef LLSEC802154_REJECT_INSECURE
```
nvt, line 243: Thanks for the contribution. Perhaps this should be named CSMA_CONF_LLSEC_REJECT_INSECURE, however, and be put in csma-security.h among the other configuration parameters for the CSMA implementation?

atiselsts: I agree, but that could be a different PR? This one seems fairly critical to get in, and the define naming is just a cosmetic thing with lower priority IMO.

atiselsts (approved):

os/net/mac/csma/csma-security.c
nvt: Yes, I think that's fine.

PR #1634, opened 2021-09-07
tcp-socket: Retain input buffer contents

nproth:
When receiving data over a TCP socket, an application can return the number of bytes that are retained in the TCP input buffer until the next call  in `tcp_socket_data_callback_t`, according to the callback documentation in `tcp-socket.h`.
This pull request implements the documented behaviour where previously the return value was ignored and the TCP input buffer contents discarded.

The implementation uses the existing field `input_data_len` from `struct tcp_socket`. I have not seen it in use anywhere else and assume this was its original purpose.


simonduq:
Hi!

PR #1643, opened 2021-09-13
Prevent integer wrapping in snmp-ber

nvt:
The <code>str</code> variable is incremented by <code>length - 1</code>, which will wrap if <code>length</code> equals 0.


nfi (approved):
Looks good.

PR #1648, opened 2021-09-20
uipbuf: Do not assume that there is enough space for a following extension header

nvt:
When processing IPv6 extension headers in uipbuf.c, we need to check the offset before casting to a <code>uip_ext_hdr</code> pointer.


nfi (approved):
Looks good.

PR #1653, opened 2021-09-28
Platform-independent SHA-256 API

kkrentz:
This PR introduces a platform-independent SHA-256 API in the style of `aes-128.ch. It includes a software implementation of SHA-256, a generic implementation of HMAC as per RFC 2104, and a generic implementation of HMAC-based key derivation as per RFC 5869.


simonduq:
Hi!

os/lib/sha-256.h
```diff
@@ -0,0 +1,161 @@
...
+
+  struct {
+    /* used by CC2538 implementation */
+    uint32_t state[SHA_256_DIGEST_LENGTH / sizeof(uint32_t)];
+    size_t hashed_len;
+    size_t buf_len;
+    uint8_t buf[SHA_256_BLOCK_SIZE];
+    bool was_crypto_enabled;
+  } cc2538;
+} sha_256_state_t;
```
arurke, line 73: This seems out-of-place being in the API of a platform-independent library in `os`. I haven't had time to do a proper review, but isn't it possible to avoid this? Are by any chance those fields needed by all hardware-implementations?

kkrentz: Good proposal. Now, both of my drivers use the same context variables.

I was also thinking about using static variables inside each implementation instead of an externally managed state. This would simplify the API a lot, but it would no longer be possible to do `PROCESS_PAUSE` between `SHA_256.init`, `SHA_256.update`, and `SHA_256.finalize`.

kkrentz:
UPDATE: SHA-256 drivers are now assumed to keep an internal state, which simplifies the API at the cost of about 100 bytes of heap memory. If a `PROCESS_PAUSE` shall happen between `init`, `update`, and `finalize`, one has to call `create_checkpoint` beforehand and `restore_checkpoint` afterwards. These changes have the added benefit that the crypto hardware no longer has to be enabled and disabled every time `update` is called.

tests/08-native-runs/14-sha-256.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
```
pjonsson, line 1: I think this script can use /bin/sh.

tests/08-native-runs/14-sha-256/Makefile
```diff
@@ -0,0 +1,9 @@
+CONTIKI_PROJECT = test-sha-256
+all: $(CONTIKI_PROJECT)
+
+TARGET = native
+
+MODULES += os/services/unit-test
+
+CONTIKI = ../../..
```
pjonsson, line 8: Makefile.include defines CONTIKI nowadys, so this definition isn't required.

tests/08-native-runs/14-sha-256/test-sha-256.c
```diff
@@ -0,0 +1,469 @@
...
+  }
+
+  UNIT_TEST_END();
+}
+/*---------------------------------------------------------------------------*/
+UNIT_TEST_REGISTER(sha_256_hash_with_checkpoint, "SHA-256 with checkpoint");
+UNIT_TEST(sha_256_hash_with_checkpoint)
+{
+  uint8_t sha256[SHA_256_DIGEST_LENGTH];
+  int i, j;
```
pjonsson, line 342: I think both these can be declared in the head of the for-loop they belong to, and they should probably be `size_t` to avoid signed/unsigned comparisons.

Same comment for the loop counters in the other tests.

pjonsson:
Is the second commit somehow specific for the Platform-independent SHA-256 API, or is that a fix that could/should be split out in a separate PR?

kkrentz:
The second commit could/should be separated.

arch/cpu/cc2538/dev/udma.c
```diff
@@ -52,6 +53,13 @@ struct channel_ctrl {
 
 static volatile struct channel_ctrl channel_config[UDMA_CONF_MAX_CHANNEL + 1]
   __attribute__ ((section(".udma_channel_control_table")));
+
+/*---------------------------------------------------------------------------*/
+bool
+udma_is_valid_source_address(uint32_t source_address)
```
pjonsson, line 59: The source_address should be `uintptr_t`.

os/lib/sha-256.c
```diff
@@ -0,0 +1,461 @@
...
+      S[(68 - i) % 8], S[(69 - i) % 8], \
+      S[(70 - i) % 8], S[(71 - i) % 8], \
+      W[i + ii] + K[i + ii])
+
+/* Message schedule computation */
+#define MSCH(W, ii, i) \
+  W[i + ii + 16] = \
+      s1(W[i + ii + 14]) + W[i + ii + 9] + s0(W[i + ii + 1]) + W[i + ii]
+
+static sha_256_checkpoint_t ctx;
```
pjonsson, line 155: There are many different contexts in Contiki-NG, so a slightly more unique name than `ctx` would be beneficial.

Does prepending `sha_256_` (or `sha_`, or some other suitable prefix/suffix) make the lines too long and make the file unreadable?

pjonsson:
I think the first commit can go in a separate PR, that should be fairly easy to review and merge.

tests/08-native-runs/14-sha-256.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -e
+
+./run-one.sh 14-sha-256
```
pjonsson, line 3: A fairly recent change is that it now also requires a line in `tests/08-native-runs/Makefile` for the test to be compiled and run. (Can copy the 13-coffee line and modify.)

tests/08-native-runs/14-sha-256/test-sha-256.c
```diff
@@ -0,0 +1,465 @@
...
+    42
+  }
+};
+
+/*---------------------------------------------------------------------------*/
+UNIT_TEST_REGISTER(sha_256_hash_stepwise, "SHA-256 stepwise hashing");
+UNIT_TEST(sha_256_hash_stepwise)
+{
+  uint8_t sha256[SHA_256_DIGEST_LENGTH];
+  size_t i, j;
```
pjonsson, line 314: I realize the code was written a long time ago, but since Contiki-NG is compiled as C11 now, it would be nice if the loop counters and similar variables were moved into the scope where they are required.

It's only the subset of C11 that GCC 4.7 handles that can be used in general code for now, but static_assert() is part of that subset.

os/lib/sha-256.c
kkrentz: Renamed this variable to `checkpoint`.

kkrentz:
Update: Added some special handling to the CC2538 driver since it freezed on the empty string.

Making the first commit stand-alone would require adapting the examples twice, which is I think not worth the effort.

kkrentz:
Added some more doxygen.

nvt (approved):

nfi (approved):

nvt:
Thanks for the nice contribution. It would be good if you can fix the remaining conflict also.

PR #1654, opened 2021-09-28
uip-nd6: Check buffer space for ND6 option headers.

nvt:
When processing incoming IPv6 neighbor solicitation packets, it is possible for an out-of-bounds read to occur because there are insufficient checks for space with regards to ND option headers.

This PR adds a check for first reading the generic 2-byte option header, and then one check for reading the source-link-layer address option (SLLAO). Note that we are currently only supporting SLLAOs using UIP_LLADDR_LEN bytes for the address.

### DIFF
--- a/os/net/ipv6/uip-nd6.c
+++ b/os/net/ipv6/uip-nd6.c
@@ -123,7 +123,8 @@ static uip_ds6_prefix_t *prefix; /**  Pointer to a prefix list entry */
 /*------------------------------------------------------------------*/
 /* Copy link-layer address from LLAO option to a word-aligned uip_lladdr_t */
 static int
-extract_lladdr_from_llao_aligned(uip_lladdr_t *dest) {
+extract_lladdr_from_llao_aligned(uip_lladdr_t *dest)
+{
   if(dest != NULL && nd6_opt_llao != NULL) {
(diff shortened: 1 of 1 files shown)


nfi (approved):
Looks good.

os/net/ipv6/uip-nd6.c
```diff
@@ -202,6 +204,11 @@ ns_input(void)
 #endif /* UIP_CONF_IPV6_CHECKS */
     switch (ND6_OPT_HDR_BUF(nd6_opt_offset)->type) {
     case UIP_ND6_OPT_SLLAO:
+      if(uip_l3_icmp_hdr_len + nd6_opt_offset +
+         UIP_ND6_OPT_DATA_OFFSET + UIP_LLADDR_LEN > uip_len) {
+        LOG_ERR("Insufficient space for NS SLLAO option\n");
```
joakimeriksson, line 209: Maybe this comment should be Insufficient data for ... since it is a read and not a write (write needs space and read needs data).

joakimeriksson (approved):

os/net/ipv6/uip-nd6.c
nvt: Yes, it would be better to say that. I've changed it now.

PR #1660, opened 2021-10-09
cooja:make build CLASS-specified platform.c so that make can easy check it's fresh

alexrayne:
cooja make now builds JNI_OBJ from platform.c for cooja-specified library CLASSNAME, to allow safe and fast make rebuilds.

* Cooja need compiled version platform.c specified for every library CLASSNAME. So, to allow make can normaly check that this object is obsolete and need rebuild, it should be different files.  So, here provide making JNI_OBJ object wich is now platform_\<CLASSNAME\>.o
        This inermediate keeps with compiled objects, so make can check freshness, and rebuild for new simulation.

* with this patch no more need make clean before simulation build. 
* this should fix [such](https://github.com/contiki-ng/contiki-ng/pull/1661/checks?check_run_id=3957178259) cooja builds failures

### DIFF
--- a/arch/platform/cooja/Makefile.customrules-cooja
+++ b/arch/platform/cooja/Makefile.customrules-cooja
@@ -28,9 +28,31 @@ endif ## REDEF_PRINTF
 .PHONY: $(CONTIKI_APP).cooja
 $(CONTIKI_APP).cooja: $(JNILIB)
 	cp $(JNILIB) $@
-	rm $(CONTIKI_APP_OBJ)
+ 
+# rm $(CONTIKI_APP_OBJ)
+
+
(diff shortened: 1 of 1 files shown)


pjonsson:
I just saw this PR. Cooja now builds mtype193.o directly from platform.c after #1915 was merged. Does that fix the problem?

pjonsson:
I just saw this PR. Cooja now builds mtype193.o directly from platform.c after #1915 was merged. Does that fix the problem?

alexrayne:
> Does that fix the problem?

yes, looks #1915 make similar job

joakimeriksson:
Closing this (assuming that #1915 fixes this problem).

PR #1661, opened 2021-10-09
fix cooja-radio RADIO_PARAM_LAST_RSSI and RADIO_PARAM_LAST_LQI

alexrayne:
here fixed cooja-radio RADIO_PARAM_LAST_RSSI and RADIO_PARAM_LAST_LINK_QUALITY - now it reports rssi and lqi captured at packet start.

* there was used current rssi(cca) for last packet, but accidentaly, cooja updates it right after packet end, and if mote code not luck, it miss rssi at packet end, and got current strength (after packet) on air.

### DIFF
--- a/arch/platform/cooja/dev/cooja-radio.c
+++ b/arch/platform/cooja/dev/cooja-radio.c
@@ -58,9 +58,23 @@
 
 const struct simInterface radio_interface;
 
+
+
+/* There radio driver can provide cooja it's nosignal value.
+ * But at present, cooja ignore and override it.
+ * */
 enum {
-    RSSI_NO_SIGNAL = -120,
-    LQI_NO_SIGNAL  = 120,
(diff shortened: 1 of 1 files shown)


alexrayne:
> This is clearly desirable, but there are some code style issues. Are you willing to change to code to conform with the code style guidelines? If so, we can provide more detailed comments.

Well, lets show what you want - i try work on it.

But as for function result on separate line - this rule annoing me, i can't accept it. Even dont imagine - what benefits it give you. So maybe somebody just fix it for you. 
At least, i've spot a problem, hope it's some help you .

alexrayne:
done with all above notes

alexrayne:
> CC1310 is a better choice. -110 would be good.

changed with this value

atiselsts (approved):

PR #1673, opened 2021-10-15
Orchestra: optimize channel offsets of unicast slotframes

atiselsts:
**The problem:** there is an intention to logically separate the channel offsets that are used for EB packets, for broadcast packets, and for unicast packets. However, the max unicast channel offset is set to 255. The real channel is calculated as `offset % hopping_sequence_size`. As typical hopping sequences are much shorter than 256 elements, in many cases unicast packet are going to use the same physical channels as EB and broadcast packets, leading to potential collisions between these types of packets.

**The solution:** the PR makes the range of channel offsets dependent on the hopping sequence size. This will not work as intended if the hopping sequence is changed at the runtime to have a smaller number of channels. However, this is not typical for most networks.

Of course, keeping the three groups of packets is separate channels is not possible if the number of channels is less than 3. In that special case we set `ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET` to 0 and allow the ranges to overlap.

The practical impact of this is very small for most networks, however, it makes it easier to reason about the causes of the collisions in the networks.

This is based on work done with Cysca Technologies.

### DIFF
--- a/os/services/orchestra/orchestra-conf.h
+++ b/os/services/orchestra/orchestra-conf.h
@@ -126,18 +126,11 @@
 #define ORCHESTRA_COLLISION_FREE_HASH             0 /* Set to 1 if ORCHESTRA_LINKADDR_HASH returns unique hashes */
 #endif /* ORCHESTRA_CONF_COLLISION_FREE_HASH */
 
-/* Channel offset for the EB rule, default 0 */
-#ifdef ORCHESTRA_CONF_EB_CHANNEL_OFFSET
-#define ORCHESTRA_EB_CHANNEL_OFFSET               ORCHESTRA_CONF_EB_CHANNEL_OFFSET
-#else
-#define ORCHESTRA_EB_CHANNEL_OFFSET               0
-#endif
-
-/* Channel offset for the default common rule, default 1 */
+/* Channel offset for the default common rule, default 0 */
 #ifdef ORCHESTRA_CONF_DEFAULT_COMMON_CHANNEL_OFFSET
 #define ORCHESTRA_DEFAULT_COMMON_CHANNEL_OFFSET   ORCHESTRA_CONF_DEFAULT_COMMON_CHANNEL_OFFSET
 #else
-#define ORCHESTRA_DEFAULT_COMMON_CHANNEL_OFFSET   1
+#define ORCHESTRA_DEFAULT_COMMON_CHANNEL_OFFSET   0
 #endif
 
 /* Min channel offset for the unicast rules; the default min/max range is [2, sizeof(HS)-2].
--- a/os/services/orchestra/orchestra-conf.h
+++ b/os/services/orchestra/orchestra-conf.h
@@ -140,18 +140,34 @@
 #define ORCHESTRA_DEFAULT_COMMON_CHANNEL_OFFSET   1
 #endif
 
-/* Min channel offset for the unicast rules; the default min/max range is [2, 255] */
+/* Min channel offset for the unicast rules; the default min/max range is [2, sizeof(HS)-2].
+   If the HS has less then 3 channels [1, 1] is used instead.
+*/
 #ifdef ORCHESTRA_CONF_UNICAST_MIN_CHANNEL_OFFSET
 #define ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET       ORCHESTRA_CONF_UNICAST_MIN_CHANNEL_OFFSET
 #else
-#define ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET       2
+#define ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET       (sizeof(TSCH_DEFAULT_HOPPING_SEQUENCE) > 2 ? 2 : 1)
 #endif
 
 /* Max channel offset for the unicast rules */
 #ifdef ORCHESTRA_CONF_UNICAST_MAX_CHANNEL_OFFSET
 #define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       ORCHESTRA_CONF_UNICAST_MAX_CHANNEL_OFFSET
 #else
-#define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       255
+#define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       \
+  (MAX(ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET, sizeof(TSCH_DEFAULT_HOPPING_SEQUENCE) - 1))
+#endif
+
+/* Channel offsets for the EB rule, default: 1 */
+#ifdef ORCHESTRA_CONF_EB_MIN_CHANNEL_OFFSET
+#define ORCHESTRA_EB_MIN_CHANNEL_OFFSET ORCHESTRA_CONF_EB_MIN_CHANNEL_OFFSET
+#else
+#define ORCHESTRA_EB_MIN_CHANNEL_OFFSET 1
+#endif
+
+#ifdef ORCHESTRA_CONF_EB_MAX_CHANNEL_OFFSET
+#define ORCHESTRA_EB_MAX_CHANNEL_OFFSET ORCHESTRA_CONF_EB_MAX_CHANNEL_OFFSET
+#else
+#define ORCHESTRA_EB_MAX_CHANNEL_OFFSET 1
 #endif
 
 #endif /* __ORCHESTRA_CONF_H__ */
--- a/os/services/orchestra/orchestra-rule-eb-per-time-source.c
+++ b/os/services/orchestra/orchestra-rule-eb-per-time-source.c
@@ -29,11 +29,11 @@
  */
 /**
  * \file
-
  *         Orchestra: a slotframe dedicated to transmission of EBs.
  *         Nodes transmit at a timeslot defined as hash(MAC) % ORCHESTRA_EBSF_PERIOD
  *         Nodes listen at a timeslot defined as hash(time_source.MAC) % ORCHESTRA_EBSF_PERIOD
  * \author Simon Duquennoy <simonduq@sics.se>
+ *         Atis Elsts <atis.elsts@edi.lv>
  */
 
 #include "contiki.h"
@@ -43,6 +43,7 @@
 static uint16_t slotframe_handle = 0;
 static uint16_t channel_offset = 0;
 static struct tsch_slotframe *sf_eb;
+static struct tsch_link *timesource_link;
 
 /*---------------------------------------------------------------------------*/
 static uint16_t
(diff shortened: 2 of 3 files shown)


os/services/orchestra/orchestra-conf.h
```diff
@@ -126,32 +126,38 @@
...
 
 /* Max channel offset for the unicast rules */
 #ifdef ORCHESTRA_CONF_UNICAST_MAX_CHANNEL_OFFSET
 #define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       ORCHESTRA_CONF_UNICAST_MAX_CHANNEL_OFFSET
 #else
-#define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       255
+#define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       (sizeof(TSCH_DEFAULT_HOPPING_SEQUENCE) - 1)
+#endif
+
+/* Channel offsets for the EB rule, default: same as unicast */
```
arurke, line 150: Are we silently hurting some setups with this default?

The clearest example I can think of: If running sender-based with collison-free hash, default 4x4 hopping sequence, and have set `ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET` to e.g. 3. Would I not be getting increased EB and unicast collisions after this PR? Previously I would have EB on offset 0, common on 1, and unicast on 2-3. After the PR, EB and unicast would share 1-3.

arurke (changes requested):

os/services/orchestra/orchestra-rule-eb-per-time-source.c
```diff
@@ -74,47 +91,51 @@ select_packet(uint16_t *slotframe, uint16_t *timeslot, uint16_t *channel_offset)
...
-    uint8_t link_options = LINK_OPTION_RX;
-    if(new_ts == get_node_timeslot(&linkaddr_node_addr)) {
-      /* This is also our timeslot, add necessary flags */
-      link_options |= LINK_OPTION_TX;
-    }
     /* Listen to the time source's EBs */
-    tsch_schedule_add_link(sf_eb, link_options, LINK_TYPE_ADVERTISING_ONLY,
-      &tsch_broadcast_address, new_ts, ORCHESTRA_EB_CHANNEL_OFFSET, 1);
+    tsch_schedule_add_link(sf_eb, LINK_OPTION_RX, LINK_TYPE_ADVERTISING_ONLY,
+      &tsch_broadcast_address, new_ts, new_channel_offset, 0);
```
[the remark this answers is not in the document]
atiselsts: Old code had `do_remove` parameter of `tsch_schedule_add_link` set to 1. New code has it set to 0. The intention behind the PR was to simplify the logic in the file by making this change. Do you think there is a possible problem? I will look this into depth if you still think its suspicious.

atiselsts:
Updated the code style here.

os/services/orchestra/orchestra-rule-eb-per-time-source.c
```diff
@@ -55,6 +56,21 @@ get_node_timeslot(const linkaddr_t *addr)
 #endif
 }
 /*---------------------------------------------------------------------------*/
+static uint16_t
+get_node_channel_offset(const linkaddr_t *addr)
+{
+#if ORCHESTRA_EBSF_PERIOD > 0
+  if(ORCHESTRA_EB_MAX_CHANNEL_OFFSET >= ORCHESTRA_EB_MIN_CHANNEL_OFFSET) {
```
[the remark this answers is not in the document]
atiselsts: Haha, now I remember why this does not work. `sizeof` cannot be used in preprocessor's conditions.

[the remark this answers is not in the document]
arurke: Aaaah :+1:

os/services/orchestra/orchestra-conf.h
arurke: The example is true for networks of any size no? As long as the the hash is still collision free?

os/services/orchestra/orchestra-rule-eb-per-time-source.c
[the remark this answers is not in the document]
arurke: When seeing this the first time I was thinking that `tsch_schedule_add_link()` with `do_remove` set to 0 would simply update an existing link with new options. But when inspecting the code it seems like it rather adds a second link in addition to the existing one. And with that in mind I do not see this working. Am I reading it wrong?

os/services/orchestra/orchestra-rule-eb-per-time-source.c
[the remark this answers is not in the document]
arurke: When we get to c11 we can use `static_assert` :)

os/services/orchestra/orchestra-conf.h
atiselsts: Previously get_node_channel_offset could return something like 4, even if the hash is collision free, and 4 mod 4 is just zero, which is the same as ORCHESTRA_EB_CHANNEL_OFFSET default value.

os/services/orchestra/orchestra-rule-eb-per-time-source.c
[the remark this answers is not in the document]
atiselsts: If the timeslots match, then there will be two links at the same time offset. One with Tx flag, for transmissions from the local node, one with Rx flag for receptions from the parent. The channel offsets may be same of different. Why do you think it is the issue?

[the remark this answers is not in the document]
arurke: Ah, my bad, I was misunderstanding how links with identical ts/ch was handled.

Let's see if my last open item is valid:
Is there anything stopping `tsch_schedule_remove_link_by_timeslot()` at line 119 from inadvertently deleting both the TX and RX link? E.g. if we lose the time-source, we call `new_time_source(old_addr, NULL)`, which would call the mentioned function and potentially remove our TX as well(?).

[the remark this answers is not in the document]
atiselsts: Yes, you're right that there is a problem! If both channel offsets match, it will remove both.
I changed the code to fix this, thanks for spotting it.

os/services/orchestra/orchestra-conf.h
atiselsts: Yes, under these assumptions the collisions will be worse after the PR.

I still think that it's better to apply it than not, because by default the max unicast channel offset is set to 255, as said in the first message, and I suspect very few users change this default.

os/services/orchestra/orchestra-rule-eb-per-time-source.c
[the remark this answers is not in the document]
arurke: :+1:

os/services/orchestra/orchestra-conf.h
arurke: Is it possible to get the best of both worlds by defaulting the EB min/max to [1, 1], and have unicast at [2, MAX(2, (HS-1)]?

atiselsts: I think that would be better, yes. What do you think about the situation when we have just 2 channels? Then have 0 for default common channel offset, and 1 for both EB and unicast - does that sound the best?

arurke: Finding the definitive answer seems like a miniature research project :) I think your suggestion is good, and should regardless be better than the existing default.

atiselsts: Updated the PR. For both corner cases (there are 1 or 2 channels in channel hopping sequence) I set the unicast channel offset to 1, to simplify the macro. In case of 1 channel if will be effectively identical to 0 anyway.

os/services/orchestra/orchestra-conf.h
```diff
@@ -126,32 +126,40 @@
...
-#define ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET       2
+#define ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET       (sizeof(TSCH_DEFAULT_HOPPING_SEQUENCE) > 2 ? 2 : 1)
 #endif
 
 /* Max channel offset for the unicast rules */
 #ifdef ORCHESTRA_CONF_UNICAST_MAX_CHANNEL_OFFSET
 #define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       ORCHESTRA_CONF_UNICAST_MAX_CHANNEL_OFFSET
 #else
-#define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       255
+#define ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET       (sizeof(TSCH_DEFAULT_HOPPING_SEQUENCE) - 1)
```
arurke, line 149: On `ORCHESTRA_UNICAST_MAX_CHANNEL_OFFSET`:  If the HS size is 1, the max. would now be 0, which would be smaller than the min. offset (1) - which would be considered faulty by the code. Maybe something like `MAX(1, sizeof(TSCH_DEFAULT_HOPPING_SEQUENCE) - 1)`?

os/services/orchestra/orchestra-conf.h
arurke: :+1:

os/services/orchestra/orchestra-conf.h
atiselsts: yes, a good point. changed it to `MAX(ORCHESTRA_UNICAST_MIN_CHANNEL_OFFSET, sizeof(TSCH_DEFAULT_HOPPING_SEQUENCE) - 1)`

os/services/orchestra/orchestra-conf.h
```diff
@@ -126,32 +126,40 @@
...
+/* Channel offset for the default common rule, default 0 */
 #ifdef ORCHESTRA_CONF_DEFAULT_COMMON_CHANNEL_OFFSET
 #define ORCHESTRA_DEFAULT_COMMON_CHANNEL_OFFSET   ORCHESTRA_CONF_DEFAULT_COMMON_CHANNEL_OFFSET
 #else
-#define ORCHESTRA_DEFAULT_COMMON_CHANNEL_OFFSET   1
+#define ORCHESTRA_DEFAULT_COMMON_CHANNEL_OFFSET   0
 #endif
 
-/* Min channel offset for the unicast rules; the default min/max range is [2, 255] */
+/* Min channel offset for the unicast and EB rules; the default min/max range is [2, sizeof(HS)-2].
```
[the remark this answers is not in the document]
atiselsts: changed

arurke:
A note for future reference:

This PR changes the default channel offsets for Orchestra. An overview for cases when there are more than 2 channels in the hopping sequence (HS):
* Was: 
    * EB: 0, Common: 1, Unicast: 2 - 255. 
* After this PR:
    * EB: 1, Common: 0, Unicast: 2 - <sizeof(HS) - 1>

Mixed deployment will thus not work (or work poorly depending on HS) and adjustments to the different `ORCHESTRA_*_CHANNEL_OFFSET` defines  in orchestra-conf.h might be needed.

arurke (approved):

PR #1674, opened 2021-10-15
link stats: fix initialization of RSSI and improve initialization of ETX

### LINKED ISSUE #1629: Invalid link_stats RSSI value
If a neighbor is added automatically (UIP_ND6_AUTOFILL_NBR_CACHE needs to be enabled) because it is not in the nbr table, the RSSI value seems to set to 0.

This is not an issue at first because link_stats_from_lladdr() returns NULL.

However when the neighbor replies the RSSI value is updated and link_stats_from_lladdr() returns a pointer and the RSSI value is averaged with the invalid RSSI value 0.

The issue can be reproduced by checking out [this repo/branch](https://github.com/jeppenodgaard/contiki-ng/tree/bug/invalid-link-stats-rssi-value) and open and run tests/test-link-stats-rssi.csc in cooja.

atiselsts:
This fixes #1629 as well as changes the initialization of ETX. Instead of starting from the default ETX in case a link stats entry is created on sending a packet, we start with the packet's ETX. This should help the EWMA of the ETX to converge faster.
I also added a special value `LINK_STATS_RSSI_UNKNOWN` to be used instead of 0, as 0 potentially (theoretically) could be a valid RSSI value.

### DIFF
--- a/os/net/link-stats.c
+++ b/os/net/link-stats.c
@@ -28,6 +28,7 @@
  *
  *
  * Authors: Simon Duquennoy <simonduq@sics.se>
+ *          Atis Elsts <atis.elsts@edi.lv>
  */
 
 #include "contiki.h"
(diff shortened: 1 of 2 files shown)


fdojurado:
I had an issue with the initial value of the RSSI in the link-stats. If this NB was added during the packet sent, the initial value was set to 0. With this fix, the initial value is modified to the first received RSSI. This  PR allows the EWMA to converge faster.

arurke (approved):

PR #1684, opened 2021-11-02
Fixed minor style and logging issues in the resolver.

nvt:


os/services/resolv/resolv.c
```diff
@@ -871,21 +872,21 @@ newdata(void)
...
-#if LOG_LEVEL == LOG_LEVEL_DBG
-    char debug_name[40];
-    decode_name(queryptr, debug_name, uip_appdata);
-    LOG_DBG("Answer %d: \"%s\", type %d, class %d, ttl %"PRIu32", length %d\n",
-            ++i, debug_name, uip_ntohs(ans->type),
-            uip_ntohs(ans->class) & 0x7FFF,
-            (uint32_t)((uint32_t)uip_ntohs(ans->ttl[0]) << 16) |
-            (uint32_t)uip_ntohs(ans->ttl[1]), uip_ntohs(ans->len));
-#endif /* LOG_LEVEL == LOG_LEVEL_DBG */
+    if(LOG_DBG_ENABLED) {
```
arurke, line 875: Two comments/observations on `if` instead of `#if`:
1. TIL that the following does not generate a warning:
```
static void unused_function(void){
  return;
}

main() {
  if(0) {
    unused_function();
  }
}
```
So with this PR, the compiler will not pick up on `decode_name()` being unused. Showstopper? I guess not.

2. On native this PR generates larger binaries. I haven't had time to look deeper into this, but a hunch is that native does not use linker-flags `ffunction-sections`, so that `decode_name()` is included. If so, the size increase would be true for other platforms without this as well (arm targets does set it, and on those there is no difference in size with this PR).

nvt: Thanks for the comments. In general, we do not give much concern to code size on the native platform, but relevant compiler flags can be enabled if it is a concern to someone. Many of the older modules from the inherited Contiki code base employed a large number of ifdefs to perform code size optimizations when disabling unneeded features, but unfortunately this led to low readability of the code. With such changes, we ensure also that deactivated code will also be processed by the compiler.

arurke: >  unfortunately this led to low readability of the code.

Indeed

> we do not give much concern to code size on the native platform

Indeed, but this would be true for any platform without the appropriate config. But a quick tally seems this is not a problem.

Related note from my tally: Seems https://github.com/contiki-ng/contiki-ng/pull/1657 broke compilation for sky and z1 (at least):
```
hello-world$ make TARGET=z1
  CC        ../../os/services/resolv/resolv.c
../../os/services/resolv/resolv.c: In function 'check_entries':
../../os/services/resolv/resolv.c:577:3: error: 'for' loop initial declarations are only allowed in C99 mode
../../os/services/resolv/resolv.c:577:3: note: use option -std=c99 or -std=gnu99 to compile your code
../../os/services/resolv/resolv.c: In function 'resolv_lookup':
../../os/services/resolv/resolv.c:1273:3: error: 'for' loop initial declarations are only allowed in C99 mode
../../os/services/resolv/resolv.c: At top level:
cc1: error: unrecognized command line option "-Wno-unused-const-variable" [-Werror]
../../Makefile.include:365: recipe for target 'build/z1/obj/resolv.o' failed
make: *** [build/z1/obj/resolv.o] Error 1
```

Maybe this PR is a good opportunity to address?

nvt: Yes, I will adjust the for loops to fix the mspgcc compilation.

nfi (approved):

PR #1705, opened 2021-12-02
Add consts suggested by cppcheck

arurke:
Latest version of cppcheck added some additional checks for missing consts, which this PR corrects.

* I noticed all the list libraries consistently avoid the suggested consts, so maybe there is a rationale I am missing
* I put nbr-table in a separate commit because 1) The semantic of several functions might become confusing: E.g. functions to lock, unlock, remove, etc. now take _only_ const parameters, 2) I am not sure if the `data` parameter in misc. nbr-table functions are intended to allow for modification, and 3) I myself added some missing consts not found by cppcheck
* Added commit 4e07c47f20057827bec40a916a49584027e2e161 from https://github.com/contiki-ng/contiki-ng/pull/1502 since that commit is ready, while the rest of the PR needs some more attention.

A quick test reveals a handful of bytes are saved on code-size

### DIFF
--- a/os/lib/circular-list.c
+++ b/os/lib/circular-list.c
@@ -56,13 +56,13 @@ circular_list_init(circular_list_t cl)
 }
 /*---------------------------------------------------------------------------*/
 void *
-circular_list_head(circular_list_t cl)
+circular_list_head(const circular_list_t cl)
 {
   return *cl;
 }
 /*---------------------------------------------------------------------------*/
 void *
-circular_list_tail(circular_list_t cl)
+circular_list_tail(const circular_list_t cl)
 {
   struct cl *this;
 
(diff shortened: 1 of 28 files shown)


alexrayne:
imho, this work can continue with attaching pure and const attributes to apropriate functions. for example - nbr-table ndex evaluations, list head/tail/length evaluations ...

pjonsson:
I ran into the cfs issue you have already fixed in this pull request. It was hard to find your fix, please put that in a separate pull request that isn't about adding consts.

arurke:
> I ran into the cfs issue you have already fixed in this pull request. It was hard to find your fix, please put that in a separate pull request that isn't about adding consts.

Apologies for this frustration @pjonsson. Opened https://github.com/contiki-ng/contiki-ng/pull/1829

PR #1729, opened 2022-01-21
heapmem: detection of double free bugs

kkrentz:
In the event of double free bugs, the double linked list maintained by `heapmem` gets corrupted and causes hard-to-track-down crashes. This PR adds an asssertion to detect such bugs early on.

### DIFF
--- a/os/lib/heapmem.c
+++ b/os/lib/heapmem.c
@@ -58,6 +58,7 @@
 
 #include "heapmem.h"
 
+#include "lib/assert.h"
 #include "sys/cc.h"
 
 /* The HEAPMEM_CONF_ARENA_SIZE parameter determines the size of the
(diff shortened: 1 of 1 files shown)


nvt:
Thanks for the effort, but I'm missing an actual description or exhibit of the bug that this PR claims to fix. Do you have a small example program?

kkrentz:
I was not able to reproduce the error. It might have been a double free bug somewhere in libcoap. We might add an assertion to detect such bugs.

nvt:
Yes, it would be good to add a check for double free calls.

Rather than using assertions (Contiki-NG has its own assertion module included from "lib/assert.h"), I think that it is preferable to insert a check like the following after line 384 to always avoid calling <code>free_chunk()</code> in the double-free case.

<code>
    if(CHUNK_FREE(chunk)) {
      PRINTF("%s: attempt to free unallocated memory at %p\n", __func__, ptr);
      return;
    }
</code>

I suggest renaming the PR and updating the description since the contents have changed. Thanks!

kkrentz:
I would prefer an assertion because a printf gets overlooked very easily (if it is enabled at all). Also, it is nice that assertions can be excluded from production code.

nvt:
Yes, we can go with the assertion added in this PR, but for a later revision it could be interesting to prevent the error when assertions are disabled also.

PR #1731, opened 2022-01-28
nrf52840: configure serial-line (and shell) for USB when native USB is enabled

nfi:
Configure serial-line to use USB instead of UART0 when native USB is enabled. dbg-io is already configured to output via USB when native USB is enabled.


alexstanoev (approved):
Works fine!

PR #1751, opened 2022-02-15
Enable nvmc flash storage for target nrf52840

whati001:
Adjusted makefile to include nvmc flash storage files for target nrf52840 if the value NRF52840_USE_NVMC_FLASH is set in the local project makefile. In addition, I have added an example to ./examples/platform-specific/nrf52840/nvmc_flash/ directory.

### DIFF
--- a/arch/cpu/nrf52840/Makefile.nrf52840
+++ b/arch/cpu/nrf52840/Makefile.nrf52840
@@ -58,6 +58,20 @@ LDFLAGS += -mabi=aapcs -mfloat-abi=hard -mfpu=fpv4-sp-d16
 LDFLAGS += --specs=nano.specs -lc -lnosys
 LDFLAGS += -L $(NRF5_SDK_ROOT)/modules/nrfx/mdk
 
+# enable nvmc flash driver for nrf52840
+ifeq ($(NRF52840_USE_NVMC_FLASH),1)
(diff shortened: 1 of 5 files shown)


alexstanoev:
Raw flash memory is quite tricky to work with, and has a very limited amount of erase cycles per page. 
If you take a look at section 4.3.10.1 of the nRF52840 PS, each flash page is rated for ~10000 erase cycles, so at one erase every 10 seconds this example will destroy that flash page in just under 28 hours.

However, [Coffee](https://github.com/contiki-ng/contiki-ng/wiki/Documentation%3A-Coffee) is a wear-leveling file system designed for this exact case and is relatively straightforward to implement for a new platform ([for reference here's the CC2538 port](https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/cc2538/cfs-coffee-arch.c)).
Would you be interested in extending this pull to include support for Coffee on the nRF52840?

whati001:
Hi @alexstanoev, 
thank you very much for your feedback. My current use case occupies the entire flash memory, thus using a file system does not make any sense for me. Therefore, I have simply sourced the necessary files and created a pr to not work with my own fork.

Unfortunately, I have currently not enough time to start porting Coffee to the nrf52850 chip. What about setting the pr on hold and I will come back if it later when I have more time?

simonduq:
Hi!

PR #1760, opened 2022-03-02
Fixes to simplelink GPIO HAL

arurke:
Some bug-fixes and improvements to the GPIO HAL:
1. Fixed a bug where the MCU does not wakeup from standby on DIO edges, and thus interrupts may be lost.
    * The PIN driver does indeed configure wakeup on DIO edges [here](https://github.com/contiki-ng/coresdk_cc13xx_cc26xx/blob/master/source/ti/drivers/pin/PINCC26XX.c#L350), but the same wakeup-selector WU1 is re-purposed to RTC_CH1 in [rtimer-arch.c:153](https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/simplelink-cc13xx-cc26xx/dev/rtimer-arch.c#L153). This PR switches the wake-up for RTC_CH1 to WU3 (WU2 is used by BATMON). Happy to hear any comments on this solution.
2. Fixed a bug in the PIN-to-HAL config conversion which caused IRQ and pull-config not to be included.
3. cc13xx/cc26xx pins have independent output and input buffers which should both be configured when switching a pin between input and output. The current implementation only manipulated the output config, causing e.g. interrupts to work only if the PIN had its input enabled either manually or via the PIN init table. This PR sets both output and input config, mimicking cc13xx-cc26xx platform, and newer TI GPIO drivers.
4. Added some comments about the Simplelink GPIO HAL quirks on interrupt enable/disable (fixing them requires some more work).

### DIFF
--- a/arch/cpu/simplelink-cc13xx-cc26xx/dev/gpio-hal-arch.c
+++ b/arch/cpu/simplelink-cc13xx-cc26xx/dev/gpio-hal-arch.c
@@ -54,6 +54,26 @@ static PIN_Config pin_config[] = { PIN_TERMINATE };
 static PIN_State pin_state;
 static PIN_Handle pin_handle;
 /*---------------------------------------------------------------------------*/
+/**< PIN standard input configuration mimicking IOC_STD_INPUT */
+#define GPIO_HAL_PIN_STD_INPUT      (PIN_INPUT_EN | PIN_GPIO_OUTPUT_DIS | \
+                                     PIN_NOPULL | PIN_DRVSTR_MIN | PIN_IRQ_DIS)
+
+/**< PIN standard input configuration bitmask */
+#define GPIO_HAL_PIN_STD_INPUT_BM   (PIN_BM_INPUT_EN | PIN_BM_GPIO_OUTPUT_EN | \
+                                     PIN_BM_PULLING | PIN_BM_INV_INOUT | \
+                                     PIN_BM_DRVSTR | PIN_BM_SLEWCTRL | \
+                                     PIN_BM_HYSTERESIS | PIN_BM_IRQ)
+
+/**< PIN standard input configuration mimicking IOC_STD_OUTPUT */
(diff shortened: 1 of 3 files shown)


arurke:
@alexstanoev Regarding setting pins as input and output: I added a commit where when setting a pin to input or output, a set standard configuration is applied as well (as opposed to only changing input/output setting). This will closer match the semantics in the cc13xx-cc26xx platform (e.g. interrupts will be disabled when changing pin direction). I will squash the two commits if you agree.

alexstanoev:
I agree, not keeping pull/irq settings makes sense in this case and it does match cc13xx-cc26xx - consistency for these two platforms is always a good thing!

`gpio_hal_arch_no_port_interrupt_enable` and `gpio_hal_arch_no_port_interrupt_disable` really do have odd semantics; it's not obvious behaviour that the two are not an inverse of eachother. I'll look into this further.

Everything else looks good with the rest of the fixes in your PR, thanks!

arurke:
> I agree, not keeping pull/irq settings makes sense in this case and it does match cc13xx-cc26xx - consistency for these two platforms is always a good thing!

👍 Rebased and squashed

> I'll look into this further.

In the newer versions of the Simplelink SDK (from v5.30 I believe), the PIN and GPIO drivers have been merged into one GPIO driver. I am looking at upgrading the SDK (but currently not all the way up to v5.30, due to, amongst other the PIN/GPIO change). So, trying now to further adapt the HAL to the PIN driver may not be a sensible task for the long-term.

joakimeriksson:
What about a rebase and then very soon merging this?

arurke:
Rebased

arurke:
Seems there is an unrelated error in the snmp-test. @pjonsson is this something you want to look at? Ref. https://github.com/contiki-ng/contiki-ng/pull/2089?

pjonsson:
@arurke Thanks for the heads up, I had a quick look and I don't see anything obvious. I'm swamped for the rest of the year so hopefully someone else takes a closer look.

Let's get this PR merged, the snmp issues are not caused by this PR.

arurke:
:+1: Restarted the failed checks.

nvt (approved):

PR #1761, opened 2022-03-10
Fix some warnings from -Wextra

pjonsson:


pjonsson:
The wiki (https://github.com/contiki-ng/contiki-ng/wiki/Contributing) states ISO C99, but some of these tests fail with the compiler complaining that it doesn't get -std=c99. Is there documentation for which parts of the code base that should use C89 and C99?

nfi (approved):

PR #1774, opened 2022-03-24
etimer: adapted to use a sorted list

kkrentz:
Presently, the etimer implementation maintains a list of etimers without leveraging the existing list API. This PR fixes this and also adds a function for updating the `next_expiration` variable more efficiently.

### DIFF
--- a/arch/cpu/arm/arm-def.h
+++ b/arch/cpu/arm/arm-def.h
@@ -50,9 +50,6 @@
  */
 #define CLOCK_CONF_SECOND 128
 
-/* Clock (time) comparison macro */
-#define CLOCK_LT(a, b)  ((signed long)((a) - (b)) < 0)
-
 /* Platform typedefs */
 typedef uint32_t clock_time_t;
 typedef uint32_t uip_stats_t;
--- a/arch/cpu/msp430/f1xxx/clock.c
+++ b/arch/cpu/msp430/f1xxx/clock.c
@@ -41,7 +41,7 @@
 
 #define MAX_TICKS (~((clock_time_t)0) / 2)
 
-#define CLOCK_LT(a, b) ((int16_t)((a)-(b)) < 0)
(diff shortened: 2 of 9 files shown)


alexrayne:
Your new code looks clear and fine. imho, you can reimplement list routines localy in etimer.c as a statics or inlines, and with it - you new code same speed as current.

kkrentz:
I changed this PR to retain the custom list implementation in `etimer.c`. This, indeed, avoids some defensive checks and procedures inside the list_* functions. I think one could end up with about the same performance if one added more offensive list_* functions. Anyway, what should speed things up is to sort the etimer list by expiration. This avoids the expensive `update_time` function and we do not need to cycle through all etimers when being polled. Three further tweaks I introduced are (i) to only call  `etimer_request_poll` when the head of the sorted etimer list had been replaced, (ii) to immediately exit `etimer_stop` if we know that a given etimer has been stopped already, and (iii) to compare expiration times without invoking `clock_time` in this process.

atiselsts:
Hi Konrad, changes in core parts of the OS are quite scary to review and merge!
What I would personally like to see is more motivation for the changes, perhaps a benchmark that shows if there are any performance differences. Improved tests for the etimer library would also be good.

kkrentz:
@atiselsts Thank you for your input. I ran a quick benchmark with a protocol stack consisting of IEEE 802.15.4 CSL, RPL, and CoAP. On average, the processing time of the `etimer_process` is reduced by 83%. Some dedicated tests are also now part of this PR.

alexrayne:
IMHO, commit <[Share macro CLOCK_LT](https://github.com/contiki-ng/contiki-ng/pull/1774/commits/ef6a9bb46786a96342edbf59f7be7b6177a23639)>  not related to this PR. Better remove from here.

On my opinion that commit - disaster.

PR #1778, opened 2022-03-30
Use dbg-io for printf, snprintf, etc., in Contiki motes for Cooja

nfi:
There is a problem with some modern Linux systems like Ubuntu 19+ where Cooja sometimes crashes during calls to libc when using Contiki motes via JNI. A workaround for this is to use the dbg-io implementation of these functions provided in Contiki-NG instead of calling corresponding libc functions. The dbg-io implementation is the same implementation as is used when building most of the embedded devices in Contiki-NG.


arch/platform/cooja/sys/cooja-log.c
```diff
@@ -124,41 +121,26 @@ putchar(int c)
...
 
-  va_start(ap, fmt);
-  res = vsnprintf(buf, MAX_LOG_LENGTH, fmt, ap);
-  va_end(ap);
-
-  //    simlog(buf);
-  for(i = 0; i < res; i++) {
-    putchar(buf[i]);
+  while(s && *s != 0) {
+    if(i >= len) {
```
pjonsson, line 136: Can this if-statement be merged into the while() condition on the previous line?

pjonsson:
I think this patch is great, without it, I get a consistent crash in Cooja on my local computer with Ubuntu 20.04.

arch/platform/cooja/sys/cooja-log.c
nfi: Good point. I have simplified the loop.

nvt (approved):

PR #1781, opened 2022-04-01
Makefile: allow overriding $(Q)

pjonsson:
Allow the user to call make with
a specific Q on the command line.

This enables the user to see all
the calls to the tools by specifying
Q="" when calling make. It also enables
a really quiet mode if the user combines
V=1 and Q=@.


pjonsson:
Niclas pointed out that I had mixed things up with how make behaves, second push is an identical code change as the first change but with an updated commit message to reflect what the change enables.

nfi (approved):

PR #1784, opened 2022-04-10
Updated gcc on Dockerfile to 10.3 2021.10

Yagoor:
Important remark: ARM does not support i386 anymore so I created a workflow that builds it following arms instructions.
https://github.com/Yagoor/gcc-arm-none-eabi
I didn't update to 11.2 because it has some severe bugs. Therefore I believe we should wait before updating the toolchain to the latest and greatest.


arurke:
For what it's worth, we have been using this version of arm-gcc on cc13xx-cc26xx and simplelink for a while without any issues.

nfi (approved):

PR #1785, opened 2022-04-10
Bumped CMSIS pointer from 5.7.0 to 5.8.0

Yagoor:
Updated CMSIS to latest and greatest. 
Release notes: https://github.com/ARM-software/CMSIS_5/releases/tag/5.8.0


atiselsts:
Hi Yagoor, is there any specific motivation for the update, or just a desire to bring in the latest stuff in general?

Yagoor:
It is mainly the desire to bring the latest stuff in general. However it is worth mentioning that this CMSIS version along side with GCC 10.3 allows the stack sealing feature available on the Cortex M33 which is the nrf53. I don't believe nrfx has added support for it in their linker files/startup code. But it is a pretty interesting feature. 😄

pjonsson:
I'll mention that I also have an nrfx version bump in my local tree, but I'm not sure if it's required and there are some other (unrelated) things entangled with it so I haven't made a PR out of it. I'm happy to put that higher on my list of priorities if it paves the way for other exciting features.

joakimeriksson (approved):
Approving now so that we get some testing done on it before the release.

PR #1805, opened 2022-04-24
Makefile: detect $(JAVA_HOME)

pjonsson:
Add detection of $(JAVA_HOME) in
the build system so Cooja does not
need to know about Contiki-NG
build-internal details.

Since this requires the build system
to know about $(JAVA), also update
platforms that use the plain string
"java" to use the new variable.

### DIFF
--- a/arch/platform/cooja/Makefile.cooja
+++ b/arch/platform/cooja/Makefile.cooja
@@ -16,6 +16,24 @@ COOJA_DIR = $(CONTIKI_NG_TOOLS_DIR)/cooja
 # Use dbg-io for IO functions like printf()
 MODULES += os/lib/dbg-io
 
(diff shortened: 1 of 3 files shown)


arch/platform/cooja/Makefile.cooja
```diff
@@ -16,6 +16,19 @@ COOJA_DIR = $(CONTIKI_NG_TOOLS_DIR)/cooja
 # Use dbg-io for IO functions like printf()
 MODULES += os/lib/dbg-io
 
+JAVA_INCDIR := $(shell $(JAVA) -XshowSettings:properties -version 2>&1 | grep java.home | awk -F"= " '{print $$2}')
```
joakimeriksson, line 19: This might not show the error if -XshowSetting:properties is not supported?!

joakimeriksson (approved):

joakimeriksson (approved):

nfi (approved):

Issue #1768, opened 2022-03-21
Safety of managing etimer elements within TSCH slot operation


mahboobkarimian:
As I understood, the tsch slot operation is called as a callback function of RTIMER and is called inside its ISR.
In Simplelink CC13xx platform RF driver implementation `arch/cpu/simplelink-cc13xx-cc26xx/rf/sched.c` [line 252](https://github.com/contiki-ng/contiki-ng/blob/705541797d5f3950a4af58c7e73938248d015cdf/arch/cpu/simplelink-cc13xx-cc26xx/rf/sched.c#L252), `etimer_stop()` is called in `rf_yield` which is called from `NETSTACK_RADIO.off()` within TSCH slot operation.
Since I am experiencing issues with processes that must be called by etimers (in particular EB transmission process) I am afraid adding or stopping etimer elements inside this interrupt operation (also other interupts) leads to an unpredictable behavior like a race condition.
Could anybody look at it and confirm, please?

atiselsts:
It's definitely not safe to use etimers in interrupt context. It's even documented in the code:
https://github.com/contiki-ng/contiki-ng/blob/develop/os/sys/etimer.h#L59

I had a look at the Simplelink platfrom's `rf` code and did not see any other potential causes for concern.
If you're using TSCH, you don't really need this period etimer-driven FS recalibration, as `netstack_sched_fs()` is called every time the radio's channel is changed. Periodic recalibration is only beneficial if the radio stays on the same channel for a long period of time. All this `synth_recal_timer` can be safely disabled for most TSCH users, as a quick fix.

We'll need to think how to change the simplelink's platform's code in a more robust way.

PR #1815, opened 2022-04-28
simplelink: avoid calling etimer_stop in interrupt context

atiselsts:
Radio off can be called in interrupt context, especially by the TSCH code. On the SimpleLink platform radio off is calling `rf_yield` which in turn calls `etimer_stop` for the frequency recalibration timer, but timer operations are not safe in interrupt context, as the timers are stored in linked lists, data structure vulnerable to race conditions.

This PR changes the logic to avoid calling the `etimer_stop`  from this interrupt context.

### DIFF
--- a/arch/cpu/simplelink-cc13xx-cc26xx/rf/sched.c
+++ b/arch/cpu/simplelink-cc13xx-cc26xx/rf/sched.c
@@ -117,6 +117,7 @@ static RF_Object rf_ble;
 static RF_CmdHandle cmd_rx_handle;
 
 static bool rf_is_on;
+static bool rf_start_recalib_timer;
 static volatile bool rx_buf_full;
 
 static rfc_CMD_SYNC_STOP_RAT_t netstack_cmd_stop_rat;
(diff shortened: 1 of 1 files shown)


arurke:
ping @mahboobkarimian , does this resolve the issues you experienced in #1768 ?

mahboobkarimian:
> ping @mahboobkarimian , does this resolve the issues you experienced in #1768 ?

@arurke , Indeed I just removed the etimer_stop because simply we don't need it when we use TSCH. It seems the re-calibration is done when the radio hops. and in TSCH we hop a lot, so we don't need it. Didn't tried Atis changes yet (Now I don't have hardware to test).

arch/cpu/simplelink-cc13xx-cc26xx/rf/sched.c
```diff
@@ -117,6 +117,7 @@ static RF_Object rf_ble;
 static RF_CmdHandle cmd_rx_handle;
 
 static bool rf_is_on;
+static bool rf_start_recalib_timer;
```
joakimeriksson, line 120: Maybe we should use volatile on this?

arurke:
I ran a sanity-test with two cc2650 sensortags on the desk. TSCH with minimal, rpl-udp example for ~24 hr, ~80000 packets - no issues.

joakimeriksson (approved):
I think it looks ok - and is now also tested.

PR #1830, opened 2022-05-06
sky-sensors: use ADC12MCTL from SDK

pjonsson:
This definition exists in the header files in both gcc 4.7.2 and gcc 9.3.1. Additionally, the definition in the header files is volatile.

### DIFF
--- a/arch/platform/sky/dev/sky-sensors.c
+++ b/arch/platform/sky/dev/sky-sensors.c
@@ -39,7 +39,7 @@
 #include "contiki.h"
 #include "lib/sensors.h"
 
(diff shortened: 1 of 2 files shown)


atiselsts:
I wonder should we change this old code? Is there a good reason to do so and risk breaking something, in some toolchain?

nfi:
This makes it possible to build with the newer compiler version that does not support the previous symbol. I have tested on Tmote Sky and verified the same behaviour before and after the change.

nfi (approved):

nvt (approved):

PR #1833, opened 2022-05-09
native: catch undefined syms in object files

pjonsson:


arch/cpu/native/Makefile.native
```diff
@@ -56,9 +56,12 @@ LDFLAGS += -Wl,-Map=$(CONTIKI_NG_PROJECT_MAP),-export-dynamic
 endif
 endif
 
+# Disallow undefined symbols in object files.
+LDFLAGS += -Wl,-zdefs
```
nfi, line 60: This fails when building under macOS due to unknown option.

nfi (approved):

PR #1851, opened 2022-05-12
native: do not use ?= for make predefines

pjonsson:
Make predefines variables such as CC, CXX,
AS and so on, so using ?= to define
them in a Makefile will have no effect.
Switch the Makefiles to using = instead.

### DIFF
--- a/arch/cpu/native/Makefile.native
+++ b/arch/cpu/native/Makefile.native
@@ -4,17 +4,17 @@ CONTIKI_SOURCEFILES += rtimer-arch.c watchdog.c eeprom.c int-master.c
 CONTIKI_SOURCEFILES += gpio-hal-arch.c
 
 ### Compiler definitions
-CC       ?= gcc
-CXX      ?= g++
+CC       = gcc
+CXX      = g++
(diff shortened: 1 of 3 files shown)


alexrayne:
that ?= means that CC and other can be provided by user environmet, or
other make includes.
what problem you solve with this patch? what benefits it provides?


пт, 13 мая 2022 г., 00:47 Peter A. Jonsson ***@***.***>:

alexrayne:
> Try changing "CC ?= gcc" in Makefile.native to "CC ?= false". Do you expect the Contiki-NG regression tests to still pass?

why you propose obviously invalid "CC ?= false" ? there other  compilers can provided on native platforms, clang for example.
makefiles designed not only for tests, it also builds release code...

pjonsson:
I suggested something obviously invalid because it shows that despite making an invalid choice, everything still builds exactly like before.

atiselsts:
Both in terms of consistency (with other Makefiles) and logic it makes sense.

atiselsts (approved):

PR #1854, opened 2022-05-13
Makefile: stop when rm fails

pjonsson:
The current behavior of "make clean" and
several other targets that use rm is to proceed
even if the removal failed, and eventually
return 0. This patch makes make fail and return
a non-zero exit code when rm fails, for example
if the user doesn't have permissions to remove
files/directories:

examples/hello-world$ make clean
TARGET not defined, using target 'native'
rm: cannot remove 'build/native/hello-world.native': Permission denied
rm: cannot remove 'build/native/contiki-ng-native.a': Permission denied
rm: cannot remove 'build/native/obj': Permission denied
make: *** [../../Makefile.include:358: clean] Error 1
examples/hello-world$ echo $?
2

Stopping on failure is important for reproducible
regression tests.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -355,9 +355,9 @@ CONTIKI_NG_PROJECT_MAP = $(BUILD_DIR_BOARD)/$(basename $(notdir $@)).map
 .PHONY: all clean distclean usage help targets boards savetarget savedefines viewconf
 
 clean:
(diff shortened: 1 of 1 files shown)


atiselsts:
Look reasonable. Do you have any guesses why the old behavior was there in the first place? I'm afraid of accidentally breaking some user's workflow.

PR #1862, opened 2022-05-15
Makefile: only include *.d for deps

pjonsson:
This adjusts the build rules to mostly match
the blog post that was linked in the initial
commit 7a228fea41f7. There are two adjustments
compared to the blog post:

1) Our SOURCEFILES can contain sources in any
   language, so we need to filter out the ones
   that can get automatic dependencies generated.
2) No dependency on $(DEPDIR)/%.d for the object
   build rules because that changes the build
   for Cortex-M due to a .SECONDEXPANSION.

Before this patch, make would try to include
$(OBJDIR)/%.c as Makefiles for dependencies.
There should never have been any c files in
$(OBJDIR), but this behavior might explain
why the CUSTOM_RULE_C_TO_O does not have
dependency generation.

The lack of dependency on $(DEPDIR)/%.d is an issue,
but things are no worse than before this patch.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -55,6 +55,7 @@ BUILD_DIR_FULL = $(BUILD_DIR_TARGET_BOARD:/=)/$(BUILD_DIR_CONFIG)
 # Ditto if BUILD_DIR_CONFIG was empty
 BUILD_DIR_BOARD = $(BUILD_DIR_FULL:/=)
 OBJECTDIR = $(BUILD_DIR_BOARD)/obj
+DEPDIR = $(OBJECTDIR)/.deps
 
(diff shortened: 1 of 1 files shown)


pjonsson:
Latest upload is a rebase against latest develop.

I compared the output of ```env RELSTR=test make NUMCORES=1 Q=''``` in 02-compile-arm-ports before/after the patch and the only differences I find are the order of arguments to rm for the object files that come from the chained rules. This patch does not cause that difference, and the rm commands still remove the same files.

nfi (approved):

joakimeriksson (approved):

PR #1845, opened 2022-05-11
Added support for Silicon Labs platform

Yagoor:
# Description

This PR is adds support for brd4162a and brd4166a from Silicon Labs. These boards use the efr32mg12p332f1024gl125 opn. This port supports TSCH and CSMA

PR #1863, opened 2022-05-15
Small improvements found while working on the Silabs port

Yagoor:
1. Doxygen script fails when there is a folder with empty space.
2. CMSIS cannot be overwritten forcing the usage of the CMSIS provided by contiki-ng
3. In our actions we are not checking out the lfs dependencies.

This PR contains the commits needed by #1845 which are stand alone and can be considered general improvements to the code.

### DIFF
--- a/tools/doxygen/Makefile
+++ b/tools/doxygen/Makefile
@@ -18,7 +18,7 @@ ifeq ($(HOST_OS),Windows)
   endif
 endif
 
-export docdirs = $(sort $(foreach dir,$(basedirs),${shell find $(contiking)/${dir} -type d}))
+export docdirs = $(sort $(foreach dir,$(basedirs),${shell find $(contiking)/${dir} -maxdepth 7 -type d}))
 export docsrc = $(docdirs) $(foreach dir,$(docdirs) .,${shell find $(dir) -type f -name "doxygen*.txt"})
 
 .PHONY: clean html
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,11 +34,15 @@ jobs:
     steps:
 
     # Checks out the repo with full history
-    - uses: actions/checkout@v2
+    - name: Checkout github repo (+ download lfs dependencies)
+      uses: actions/checkout@v2
       with:
           fetch-depth: '0'
           submodules: 'recursive'
           persist-credentials: false
+          lfs: true
+    - name: Checkout LFS objects
+      run: git lfs checkout
 
     # Construct the correct docker container image tag corresponding to this build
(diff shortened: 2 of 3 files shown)


joakimeriksson (dismissed):

tools/doxygen/Makefile
```diff
@@ -18,8 +18,8 @@ ifeq ($(HOST_OS),Windows)
   endif
 endif
 
-export docdirs = $(sort $(foreach dir,$(basedirs),${shell find $(contiking)/${dir} -type d}))
-export docsrc = $(docdirs) $(foreach dir,$(docdirs) .,${shell find $(dir) -type f -name "doxygen*.txt"})
+export docdirs = $(sort $(foreach dir,$(basedirs),${shell find $(contiking)/${dir} -type d -print0}))
```
joakimeriksson, line 21: It seems like the argument -print0 change the behavior - it will not produce the full documentation anylonger.

Yagoor: Ah alright. I'll take a look into that. :)

tools/doxygen/Makefile
```diff
@@ -18,7 +18,7 @@ ifeq ($(HOST_OS),Windows)
   endif
 endif
 
-export docdirs = $(sort $(foreach dir,$(basedirs),${shell find $(contiking)/${dir} -type d}))
+export docdirs = $(sort $(foreach dir,$(basedirs),${shell find $(contiking)/${dir} -maxdepth 7 -type d}))
```
Yagoor, line 21: @joakimeriksson Instead of using -print0. I set a limit to maxdepth of 7.
This should be more than enough, at this depth it is only submodules. At least, from what I could find in my investigations.

pjonsson:
The only comments so far were on the Doxygen patch, would it be worth your time to put the other two commits in a separate PR?

pjonsson:
Can the Makefile magic be replaced by setting `INPUT = ../../arch ../../os` and `RECURSIVE = YES` in Doxyfile and leaving the traversal of directories to Doxygen?

tools/doxygen/Makefile
joakimeriksson: Is the option suggested by @pjonsson below a possible solution - e.g. do a recursive config on doxygen?

SilabsYagoor:
@pjonsson: I can try that.
I'll do what you suggested above. I'll split the Doxygen patch to get the others merged in.
I'll also try the traversal to Doxygen and see the result.

Yagoor:
Created 3 PR's
#1896, #1897, #1898

PR #1872, opened 2022-05-20
java 11 for cooja and fix mosquitto filename

jbrusey:
Java 8 no longer seems to work (jni.h missing errors) however Java 11
seems ok now so switching to that.

mosquitto file in /etc/apt/sources.list.d causes errors on apt
commands and possibly never worked. Appending .list to name.

### DIFF
--- a/tools/vagrant/bootstrap.sh
+++ b/tools/vagrant/bootstrap.sh
@@ -2,7 +2,7 @@
 
 # Install i386 binary support on x64 system and required tools
 sudo dpkg --add-architecture i386
-echo "deb http://ppa.launchpad.net/mosquitto-dev/mosquitto-ppa/ubuntu bionic main" | tee /etc/apt/sources.list.d/mosquitto
+echo "deb http://ppa.launchpad.net/mosquitto-dev/mosquitto-ppa/ubuntu bionic main" | tee /etc/apt/sources.list.d/mosquitto.list
 sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 77B7346A59027B33C10CAFE35E64E954262C4500
 sudo apt update
 sudo apt install -y --no-install-recommends \
@@ -11,7 +11,7 @@ sudo apt install -y --no-install-recommends \
   default-jdk ant srecord python-pip iputils-tracepath uncrustify \
   mosquitto mosquitto-clients valgrind libcoap-1-0-bin \
   smitools snmp snmp-mibs-downloader \
-  python-magic linux-image-extra-virtual openjdk-8-jdk
+  python-magic linux-image-extra-virtual
 
 sudo apt-get clean
 sudo python2 -m pip install intelhex sphinx_rtd_theme sphinx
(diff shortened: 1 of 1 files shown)


tools/vagrant/bootstrap.sh
```diff
@@ -52,11 +52,11 @@ echo "export NRF52_SDK_ROOT=/usr/nrf52-sdk" >> ${HOME}/.bashrc
...
 sudo usermod -aG dialout vagrant
 
-# Make sure we're using Java 8 for Cooja
-sudo update-java-alternatives --set /usr/lib/jvm/java-1.8.0-openjdk-amd64
+# Make sure we're using Java 11 for Cooja
+sudo update-java-alternatives --auto
 
 # Environment variables
-echo "export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64" >> ${HOME}/.bashrc
+echo "export JAVA_HOME=/usr/lib/jvm/default-java" >> ${HOME}/.bashrc
```
pjonsson, line 59: Cooja does not need JAVA_HOME any longer so you should be able to remove this line.

tools/vagrant/bootstrap.sh
```diff
@@ -2,7 +2,7 @@
 
 # Install i386 binary support on x64 system and required tools
 sudo dpkg --add-architecture i386
-echo "deb http://ppa.launchpad.net/mosquitto-dev/mosquitto-ppa/ubuntu bionic main" | tee /etc/apt/sources.list.d/mosquitto
+echo "deb http://ppa.launchpad.net/mosquitto-dev/mosquitto-ppa/ubuntu bionic main" | tee /etc/apt/sources.list.d/mosquitto.list
```
pjonsson, line 5: I forgot this when I updated the docker, does replacing this line with:

add-apt-repository ppa:mosquitto-dev/mosquitto-ppa

work so this file is somewhat aligned with the dockerfile? I'm not sure what environment this script starts out with, the add-apt-repository command comes from the "software-properties-common" package.

pjonsson:
You're right that Java 8 no longer works. If you are running a recent develop, you should be getting a readable error message from ant that Java 11 is required rather than a mysterious missing jni.h failure. Did you run "git submodule update" after you did git pull?

tools/vagrant/bootstrap.sh
jbrusey: To be consistent with the rest of the script, I think I need
```
sudo add-apt-repository --yes ppa:mosquitto-dev/mosquitto-ppa
```
I need to make this change and check it.

Re Java 8 - no message (tested on two different machines) just the error during compile. Definitely had submodules updated.

tools/vagrant/bootstrap.sh
jbrusey: I can take this out and re-test.

tools/vagrant/bootstrap.sh
```diff
@@ -52,11 +52,7 @@ echo "export NRF52_SDK_ROOT=/usr/nrf52-sdk" >> ${HOME}/.bashrc
 
 sudo usermod -aG dialout vagrant
 
-# Make sure we're using Java 11 for Cooja
-sudo update-java-alternatives --auto
```
pjonsson: Sorry for being unclear, I think this is still required, it's just the JAVA_HOME line that shouldn't be required.

pjonsson:
The lack of error message for you is puzzling. I don't have Java 8 to test with but I changed release="11" to release="12" in tools/cooja/build.xml and ran ant compile and got this:

```
contiki-ng/tools/cooja$ ant compile
Buildfile: contiki-ng/tools/cooja/build.xml

init:

compile:
    [mkdir] Created dir: contiki-ng/tools/cooja/build
    [javac] Compiling 171 source files to contiki-ng/tools/cooja/build
    [javac] error: release version 12 not supported
    [javac] Usage: javac <options> <source files>
    [javac] use --help for a list of possible options

BUILD FAILED
contiki-ng/tools/cooja/build.xml:89: Compile failed; see the compiler error output for details.
```

tools/vagrant/bootstrap.sh
jbrusey: I tested out after removing the `sudo update-java-alternatives --auto` and it doesn't seem to be needed. The point is that only one version of java is installed under this revised script and the default is auto anyway. 

I should note that there are probably some other issues with the vagrant set-up - particularly around interaction with the vagrant-vbguest but that is probably for some other issue / pr.

jbrusey: According to the ant documentation https://ant.apache.org/manual/Tasks/javac.html, the release parameter is ignored when compiling on JDK 8 or earlier.

pjonsson:
I tested this change and it works on my computer.

I also made a pull request to Cooja so that ant gives an error message with Java < 11, thank you for figuring it all out for me!

jbrusey:
You're welcome! Are there any further changes needed prior to merging?

PR #1877, opened 2022-05-21
This fixes the issue when the LINKADDR_SIZE is not necessary 8 bytes …

fdojurado:
To force constant in this log print will raise an error when using a different link address size.


os/net/mac/tsch/tsch-schedule.c
```diff
@@ -546,7 +546,7 @@ tsch_schedule_print(void)
 
       while(l != NULL) {
         LOG_PRINT("* Link Options %02x, type %u, timeslot %u, channel offset %u, address %u\n",
-               l->link_options, l->link_type, l->timeslot, l->channel_offset, l->addr.u8[7]);
+               l->link_options, l->link_type, l->timeslot, l->channel_offset, l->addr.u8[LINKADDR_SIZE-1]);
```
nvt, line 549: Looks reasonable, but it would be good if you could fix the code style so it includes spaces around the operator: <code>LINKADDR_SIZE - 1</code>.

arurke:
Searching the codebase for `u8[7]` I believe there are a handful of places with the same issue, would this be a good opportunity to address those as well @fdojurado? (I found no suspects for `u16[3]`)

fdojurado:
I am happy to do that @arurke . I could only find those in the examples folder and one in RPL-classic. Do you mean those?

arurke:
Apologies for the late reply @fdojurado. Yes.

fdojurado:
I have fixed those in the "sixtop" example. The one in rpl-timers, I didn't change it as RPL relies on the 8-byte size address; therefore, it won't raise an error. However, this can be updated to be written more general. This also applies to the sixtop example.

Let me know your thoughts.

https://github.com/contiki-ng/contiki-ng/blob/7725f8234d5196c84546d3c0fe46f3390827cd3a/os/net/routing/rpl-classic/rpl-timers.c#L493-L498

arurke:
>However, this can be updated to be written more general.

Not sure if following, but I like what you did with using the log-macros, and to me that makes sense here as well (benefits also include respecting the deployment-id module)

nvt:
Thanks for the update. It would be nice to print the full LL address in rpl-timers.c also, but I don't think that it is necessary to add it in this PR. Still, a couple of minor changes could improve the PR a little bit:

1. The commit history can be reduced.
2. The "const" seems unnecessary added expressions <code>PRINTLLADDR((const uip_lladdr_t *)peer_addr)</code> in sf-simple.c.

Unfortunately, the network stack contains two different log interfaces and two different data types used for link addresses: net/net-debug.h handles printing of <code>uip_lladdr_t</code> values, whereas sys/log.h handles printing of <code>linkaddr_t</code> values. The sf-simple.c module uses the older style interface net-debug.h, but updating it to use sys/log.h is beyond the scope of this PR.

nvt:
Could you do an interactive rebase and squash some commits? There is some information here: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

fdojurado:
@nvt Thanks for the information. I have now squashed them.

nvt (approved):

PR #1880, opened 2022-05-23
unit-test: exclude example code for users

pjonsson:
$(CONTIKI_NG_TARGET_LIB) includes example-test.o
which contains the test_process symbol. That symbol
is also defined by the code for 02-ringbufindex.csc.

Move the unit-test example file to a subdirectory
so the example is not included for users of
the unit-test module.


nvt:
Thanks for the fix. The example module is moved to another place in PR #1340, which has been in the queue for a long time, but it is unclear whether that one will be merged.

PR #1881, opened 2022-05-23
heapmem: added unit tests and improved error checking

nvt:
This PR adds a set of unit tests for the heapmem module. Additionally, the heapmem_free() function now returns a boolean to indicate whether the operation succeeded. In addition to checking whether a chunk is allocated or not, we validate the pointer supplied to heapmem_realloc() and heapmem_free() to ensure that it is pointing to memory within the heapmem arena.


tests/08-native-runs/12-heapmem/test-heapmem.c
```diff
@@ -0,0 +1,216 @@
...
+  UNIT_TEST_BEGIN();
+
+  char *ptrs[TEST_CONCURRENT];
+  unsigned alloc_index;
+  unsigned count;
+  unsigned failed_allocations = 0;
+  unsigned failed_deallocations = 0;
+
+  /* Clear the allocation table. */
+  for(alloc_index = 0; alloc_index < TEST_CONCURRENT; alloc_index++) {
```
pjonsson, line 84: Can this loop be  replaced by

```
char *ptrs[TEST_CONCURRENT] = {0};
```
?

pjonsson:
Code looks good as far as I can tell. Is it worth switching to more locally scoped variables for the lines that this patch updates when possible?

PR #1885, opened 2022-05-24
native: remove unused build rule

pjonsson:
The rule for building shared libraries
seem to be unused so remove it.

The rule looks like a standard formulation
for shared libraries so it probably belongs
in Makefile.include if we need it.


alexrayne:
IMHO, shared libraries have meaning only on native platform, since it need full-functional objects loader. That loader provided by OS. Therefore that rule just in place where it have meaning.

nfi:
It does not look like the native platform is using this rule.

nfi (approved):

PR #1894, opened 2022-05-27
tests: use Cooja build file

pjonsson:
This depends on: https://github.com/contiki-ng/cooja/pull/131, https://github.com/contiki-ng/cooja/pull/129 and one more patch on top of those.

Remove the knowledge of Cooja build system
from Contiki-NG and just call ant on the run
target in Cooja instead.

This ensures that Cooja is rebuilt when
the sources have been updated.

Also reduce the number of calls to rm while
I'm touching these files.


pjonsson:
The `run_nogui` target in Cooja apparently changes working directory which causes the log to end up in the wrong place, will fix that first.

PR #1895, opened 2022-05-27
Improvements of the unit test library

nvt:
A few minor changes of the unit test module, including

1. Update the comments.
2. Switch unit test result type to boolean.
3. Switch timer to clock_t to avoid wrapping rtimer values.
4. Report the number of assertions executed in a unit test.

### DIFF
--- a/os/services/unit-test/unit-test.c
+++ b/os/services/unit-test/unit-test.c
@@ -29,11 +29,12 @@
 
 /**
  * \file
- *	A tool for unit testing Contiki software.
+ *	Default print function for unit tests.
  * \author
- * 	Nicolas Tsiftes <nvt@sics.se>
+ * 	Nicolas Tsiftes <nicolas.tsiftes@ri.se>
  */
 
+#include <inttypes.h>
 #include <stdio.h>
 
 #include "unit-test.h"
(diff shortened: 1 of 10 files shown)


os/services/unit-test/unit-test.h
```diff
@@ -29,86 +29,97 @@
...
- * The unit_test structure describes the results of a unit test. Each 
+ * The unit_test structure describes the results of a unit test. Each
  * registered unit test statically allocates an object of this type.
  */
 typedef struct unit_test {
   const char * const descr;
   const char * const test_file;
-  unit_test_result_t result;
+  uint32_t assertions;
+  bool result;
```
pjonsson, line 53: Is it worth calling this variable `passed` or something similar so the code checking it becomes `if(utp->passed) ..`?

os/services/unit-test/unit-test.c
```diff
@@ -48,12 +49,12 @@ void
...
-  printf("Start: %u\n", (unsigned)utp->start);
-  printf("End: %u\n", (unsigned)utp->end);
-  printf("Duration: %u\n", (unsigned)(utp->end - utp->start));
-  printf("Ticks per second: %u\n", (unsigned)RTIMER_SECOND);
+  printf("Result: %s\n", utp->result == false ? "failure" : "success");
+  printf("Exit point: %s:%u\n", utp->test_file, utp->exit_line);
+  printf("Assertions executed: %"PRIu32"\n", utp->assertions);
+  printf("Start: %"PRIu64"\n", (uint64_t)utp->start);
+  printf("End: %"PRIu64"\n", (uint64_t)utp->end);
+  printf("Duration: %u ticks\n", (unsigned)(utp->end - utp->start));
```
pjonsson, line 57: Should the computation for duration be done as `int64_t` so we don't have to worry about overflow and similar issues?

nvt:
Thanks for the comments. It has been updated now.

pjonsson (approved):

nfi (approved):

PR #1898, opened 2022-05-28
Moved recursive logic from Makefile into Doxyfile

Yagoor:


tools/doxygen/Doxyfile
```diff
@@ -753,7 +753,8 @@ WARN_LOGFILE           = doxygen.log
 # spaces.
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = $(docsrc)
+INPUT                  = $(docroot)/arch \
+                         $(docroot)/os
```
pjonsson, line 757: I don't know the Doxygen setup but the previous code also included `$(docroot)/tools/doxygen` in `INPUT`.

Yagoor: Good catch. It added in the second for each.
I'll fix that. :)

PR #1905, opened 2022-06-01
TSCH scanning: limit the etimer frequency to 1/CLOCK_SECOND to reduce CPU usage during the scanning

atiselsts:
The problem is that not only the scanning consumes 100% radio at the moment, but also 100% CPU because timers can be added with zero timeout on some platforms, and also potentially in the past (due to `etimer_reset` behavior). This PR aims to reduce that.

### DIFF
--- a/os/net/mac/tsch/tsch.c
+++ b/os/net/mac/tsch/tsch.c
@@ -786,7 +786,7 @@ PT_THREAD(tsch_scan(struct pt *pt))
 
   TSCH_ASN_INIT(tsch_current_asn, 0, 0);
 
-  etimer_set(&scan_timer, CLOCK_SECOND / TSCH_ASSOCIATION_POLL_FREQUENCY);
+  etimer_set(&scan_timer, MAX(1, CLOCK_SECOND / TSCH_ASSOCIATION_POLL_FREQUENCY));
   current_channel_since = clock_time();
 
   while(!tsch_is_associated && !tsch_is_coordinator) {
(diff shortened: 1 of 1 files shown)


nvt (approved):

alexrayne:
imho, this low limit should be also more then slot-time, may be even much
more. If you listen air about 1ms, but TSCH slot about 40ms - surely we
miss capture beacon packet.


ср, 1 июн. 2022 г. в 17:17, Nicolas Tsiftes ***@***.***>:

joakimeriksson:
> imho, this low limit should be also more then slot-time, may be even much more. If you listen air about 1ms, but TSCH slot about 40ms - surely we miss capture beacon packet. ср, 1 июн. 2022 г. в 17:17, Nicolas Tsiftes ***@***.***>:
> […](#)

@alexrayne Should we handle this in this PR or add anotherone later?

atiselsts:
Alex, I think the issue you're concerned about is not relevant to this PR. This only about the CPU usage, the radio is turned in 100% during the scanning cycle.

There is a potential issue with how long the scanning process waits for `pending_packet()`, which is 10 ms. This is not really an issue with the standard 250 kbps or 50 kbps settings, but could be a problem if someone configures 5 kbps or similar datarates. This could be improved in a different PR.

alexrayne:
with 50kbps you got 40ms slot time, 10ms not enouhg for normal purposes -
about 64-80 bytes data payload.


сб, 4 июн. 2022 г., 20:28 Atis Elsts ***@***.***>:

alexrayne:
concern to cpu usage - for cooja simulation you relax this problem. but for
normal aplication work - no, cpu still wastes time and energy. So, my
mention no far from this PR

сб, 4 июн. 2022 г., 20:28 Atis Elsts ***@***.***>:

atiselsts:
> with 50kbps you got 40ms slot time, 10ms not enouhg for normal purposes - about 64-80 bytes data payload. 

EB size inluding header is only 36 bytes or so. With 160 microseconds per byte 10 milliseconds are sufficient.

PR #1907, opened 2022-06-01
Add a Coffee test

nvt:
This test is based on the old test for Coffee in Contiki, but it is updated to use the unit-test module. Additionally, we make it possible to configure which CFS back end to use on the native platform, rather than always compiling with cfs-posix.

### DIFF
--- a/tests/08-native-runs/13-coffee/test-coffee.c
+++ b/tests/08-native-runs/13-coffee/test-coffee.c
@@ -1,0 +1,287 @@
+/*
+ * Copyright (c) 2008, Swedish Institute of Computer Science.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


tests/08-native-runs/13-coffee/test-coffee.c
```diff
@@ -0,0 +1,299 @@
...
+
+  int total_read = 0;
+  while((r = cfs_read(afd, buf2, sizeof(buf2))) > 0) {
+    for(int j = 0; j < r; j++) {
+      UNIT_TEST_ASSERT(buf2[j] == 1 + ((total_read + j) & 0x7f));
+    }
+    total_read += r;
+  }
+
+  UNIT_TEST_ASSERT(r >= 0);
```
pjonsson, line 187: Shouldn't this fail if `r > 0` at this point?

nfi (approved):

PR #1922, opened 2022-06-05
cooja: remove empty env variables

pjonsson:
COOJA_SOURCEFILES and COOJA_SOURCEDIRS
are always set to be empty by Cooja.


pjonsson:
Forget this, the compilation environment is apparently mutated elsewhere in the Cooja code.

PR #1931, opened 2022-06-08
tests: fix intermittent 17-tun-rpl tests

pjonsson:
The tests try to establish a connection with
tunslip6 before Cooja starts listening, increase
the delay from 20 to 30 seconds.


atiselsts:
I think we should get this in, I needed to manually restart failing 17-tun-rpl-br tests a couple times recently because of these random failures. The question is - are 30 seconds sufficient?

pjonsson:
All test output was piped to a file until yesterday so we don't know. The failure I saw needed 23 seconds sleep, so fluctuations of 2 more seconds isn't impossible. I added 5 more seconds on top of that as a safety margin.

This is already one of the slowest CI tests, so bumping from 20 to 30 seconds prolongs this test by 1 minute and 10 seconds.

nfi (approved):

joakimeriksson (approved):

PR #1946, opened 2022-06-11
Test caching docker images

pjonsson:


pjonsson:
It's possible Contiki-NG is not the target audience for the docker-cache action, making this a draft for now.

pjonsson:
The initial performance problems in the action have been fixed, so there is no longer any room for improvement there.

The `docker pull` is really quick, so until `docker load` becomes significantly faster, this approach is not beneficial for Contiki-NG.

PR #1966, opened 2022-06-18
simpelink: fix doxygen warnings

pjonsson:
The "===" makes Doxygen believe it is
in a header. The section looks fine
in the HTML without the delimiter,
so remove it.


arurke:
There are "===" in the equivalent files for the other simplelink boards as well, not getting a warning for them?

pjonsson:
These are the only 2 files I get these kind of warnings on. I get the warning on both Doxygen 1.8.17 and 1.9.1 though.

I don't really know doxygen so I'm more than happy to let someone else investigate this.

pjonsson:
I don't get the warning with doxygen 1.9.4 though. Anything more recent than Doxygen with Ubuntu 18 will give some kind of warnings on the code base, so I guess it's a matter of picking one version and sticking with that.

Besides the obviously broken doxygen 1.8.17 that comes with Ubuntu 20 that refuses to let our mqtt header past, I get the impression that a warning from doxygen only means "I'm at this line in that file, and I just detected some kind of inconsistency".

pjonsson:
Not required with Doxygen 1.9.4 that is now in the Docker image.

PR #1975, opened 2022-06-18
Print simple stats. in rpl-udp example

arurke:
I was looking to do a simple sanity on the simplelink radio (for #1815). However, simplelink (and many others) are excluded in `examples/benchmarks/rpl-req-resp`, so I added some simple stats. to `examples/rpl-udp` instead. I put my solution in this PR as it may be useful for others, yet I am not sure about merging as it slightly clutters a clean example.

### DIFF
--- a/examples/rpl-udp/udp-client.c
+++ b/examples/rpl-udp/udp-client.c
@@ -3,6 +3,8 @@
 #include "random.h"
 #include "net/netstack.h"
 #include "net/ipv6/simple-udp.h"
+#include <stdint.h>
+#include <inttypes.h>
 
 #include "sys/log.h"
 #define LOG_MODULE "App"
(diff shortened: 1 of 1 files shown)


atiselsts:
I suspect `examples/benchmarks/rpl-req-resp/` is disabled for the simplelink platform for historical reasons, when it still was not supporting TSCH at all. I will make a separate PR to enable it.
If we can use that benchmark, will that be sufficient, or you still think this PR is needed?

arurke:
:+1: on enabling the simplelink platform. 

> If we can use that benchmark, will that be sufficient, or you still think this PR is needed?

In my case I was using `sensortag/cc2650` which cannot fit the benchmark by default (and is excluded). But I still won't argue strongly to merge this PR as I mentioned. I guess the optimal solution would be to make a mode/config in rpl-req-resp which all boards can run - but I admit my motivation to do this is limited, especially as all the TI boards can be benchmarked via cc13xx-cc26xx platform (TIL, this may not be true, see https://github.com/contiki-ng/contiki-ng/pull/1977#issuecomment-1160830111), or with newer boards if wanting simplelink.

I can keep the PR here for a couple weeks and then close it if nobody feels strongly.

nvt:
I think that it is fine to add this change despite the minor clutter. The statistics can be interesting for anyone using the example, and shouldn't make the code harder to understand.

nfi (approved):

PR #1976, opened 2022-06-19
Use default mirror in docker build

Yagoor:
As we can see on #1845, the mirror://mirrors.ubuntu.com/mirrors.txt is not working.
I tried locally and it also does not work.
@pjonsson I can see that you added it. Is it okay to revert it?

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -8,7 +8,6 @@ USER root
 # Tools
 RUN apt-get -qq update && \
     apt-get -qq -y --no-install-recommends install ca-certificates > /dev/null && \
-    sed -i -e 's#http://archive.ubuntu.com/ubuntu/#mirror://mirrors.ubuntu.com/mirrors.txt#' /etc/apt/sources.list && \
     apt-get -qq update && \
     apt-get -qq -y --no-install-recommends install gnupg software-properties-common > /dev/null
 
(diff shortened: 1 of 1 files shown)


tools/docker/Dockerfile
```diff
@@ -7,9 +7,10 @@ USER root
 
 # Tools
 RUN apt-get -qq update && \
-    apt-get -qq -y --no-install-recommends install ca-certificates > /dev/null && \
-    apt-get -qq update && \
-    apt-get -qq -y --no-install-recommends install gnupg software-properties-common > /dev/null
+    apt-get -qq -y --no-install-recommends install \
+      gnupg \
```
pjonsson, line 11: While you're touching this, could you sort these packages alphabetically so they are consistent with the other large block of packages?

pjonsson:
The mirrors work wonderfully in Sweden, but I can't argue against the CI failures.

tools/docker/Dockerfile
Yagoor: Done. :)

Yagoor:
> The mirrors work wonderfully in Sweden, but I can't argue against the CI failures.

This is interesting. I'm just on the other side of Øresund and it does not work locally either.

pjonsson (approved):

pjonsson:
> > The mirrors work wonderfully in Sweden, but I can't argue against the CI failures.
> 
> This is interesting. I'm just on the other side of Øresund and it does not work locally either.

Even more strange, I get su.se, umu.se, c0urier.net, and a fourth mirror that I can't remember the name of, and they are all fast and stable.

nfi (approved):

PR #1977, opened 2022-06-19
examples/benchmarks/rpl-req-resp: enable building for Simplelink

atiselsts:
This PR enabled building the `rpl-req-resp` benchmark example on Simplelink by removing it from the PLATFORMS_EXCLUDE list.

### DIFF
--- a/examples/benchmarks/rpl-req-resp/Makefile
+++ b/examples/benchmarks/rpl-req-resp/Makefile
@@ -1,8 +1,8 @@
 CONTIKI_PROJECT = node
 all: $(CONTIKI_PROJECT)
 
-PLATFORMS_EXCLUDE = sky z1 nrf52dk native simplelink
-BOARDS_EXCLUDE = srf06/cc13x0 launchpad/cc1310 launchpad/cc1350 sensortag/cc2650 sensortag/cc1350
+PLATFORMS_EXCLUDE = sky z1 nrf52dk native
+BOARDS_EXCLUDE = launchpad/cc1350-4 launchpad/cc2640r2
 
 MODULES_REL += ../testbeds
 
--- a/examples/benchmarks/rpl-req-resp/node.c
+++ b/examples/benchmarks/rpl-req-resp/node.c
@@ -52,6 +52,19 @@
 #define UDP_PORT 8214
 #define SEND_INTERVAL (CLOCK_SECOND)
 
+/* Builds that are known to be failing */
+#if CONTIKI_TARGET_SIMPLELINK
+#if CONTIKI_BOARD_LAUNCHPAD_CC1310 \
+  || CONTIKI_BOARD_LAUNCHPAD_CC1350 \
+  || CONTIKI_BOARD_LAUNCHPAD_CC2650 \
+  || CONTIKI_BOARD_SENSORTAG_CC1350 \
+  || CONTIKI_BOARD_SENSORTAG_CC2650 \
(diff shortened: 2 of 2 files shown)


examples/benchmarks/rpl-req-resp/Makefile
```diff
@@ -1,8 +1,9 @@
 CONTIKI_PROJECT = node
 all: $(CONTIKI_PROJECT)
 
-PLATFORMS_EXCLUDE = sky z1 nrf52dk native simplelink
-BOARDS_EXCLUDE = srf06/cc13x0 launchpad/cc1310 launchpad/cc1350 sensortag/cc2650 sensortag/cc1350
+PLATFORMS_EXCLUDE = sky z1 nrf52dk native
+BOARDS_EXCLUDE = srf06/cc13x0 srf06/cc26x0 launchpad/cc1310 \
+	launchpad/cc1350 launchpad/cc1350-4 sensortag/cc2650 sensortag/cc1350
```
arurke, line 6: Can you add `launchpad/cc2650` to `BOARDS_EXCLUDE` - it can also not fit this example

arurke:
I just realized the `BOARDS_EXCLUDE` is independent of the `PLATFORMS_EXCLUDE`. Thus it is not possible to build this example for those boards with the cc26x0-cc13x0 either.

Those boards cannot fit the example when on simplelink, but they can when on cc26x0-cc13x0. Commit history does not give many hints - perhaps this was just a sacrifice forced by the limitation of `BOARDS_EXCLUDE`?

In any case, that issue is separate from this PR.

examples/benchmarks/rpl-req-resp/Makefile
atiselsts: We should keep `launchpad/cc2650` because as you say in the your other comment, it can built on cc26x0-cc13x0 with this board. There is no way how to limits specific TARGET/BOARD combination. The situation we have with simplelink and cc26x0-cc13x0 is not something this build exclusion system is prepared for.

arurke: > it can built on cc26x0-cc13x0 with this board

Just a note: This is true for the others boards as well, you thinking to add one to give a bit more of options to the platform? (There is only the `launchpad/cc2640r2` which is unique to cc26x0-cc13x0).

atiselsts:
This is my summary on how the example builds ("o" means success, "x" failure)

For the `simplelink` platfrom:

- o launchpad/cc1312r1
- o launchpad/cc1352p-2
- o launchpad/cc1352p-4
- o launchpad/cc1352p1
- o launchpad/cc1352r1
- o launchpad/cc26x2r1
- o sensortag/cc1352r1
- x launchpad/cc1310
- x launchpad/cc1350
- x launchpad/cc1350-4
- x launchpad/cc2650
- x sensortag/cc1350
- x sensortag/cc2650
- x srf06/cc13x0
- x srf06/cc26x0

For the `cc26x0-cc13x0` platform:

- o srf06/cc26x0
- o srf06/cc13x0
- o launchpad/cc2640r2 <- but should be successful as the chip is BLE only!
- o launchpad/cc2650
- o launchpad/cc1310
- o launchpad/cc1350
- o sensortag/cc2650
- o sensortag/cc1350

Only `launchpad/cc1350-4` and `launchpad/cc2640r2` can be safely excluded it looks like.
The other board we can deal with in the source code level, while there isn't a better system in place for filtering the out in Makefiles.

nvt (approved):

PR #1981, opened 2022-06-26
Fixed uncrustify.cfg errors after Docker upgrade

Yagoor:
The following errors appear when running uncrustify.

```
Option<NUM>: at tools/code-style/uncrustify.cfg:28: Expected number , for 'indent_comma_paren'; got 'false'
Option<NUM>: at tools/code-style/uncrustify.cfg:29: Expected number , for 'indent_bool_paren'; got 'false'
Option<UNUM>: at tools/code-style/uncrustify.cfg:83: Expected unsigned number , for 'mod_full_brace_if_chain'; got 'false'
```

These configs changed from boolean to numbers in the latest version of uncrustify which is being used in the new ubuntu 20.04 from our Docker image.

# Notice

I didn't run the upgrade tool from uncrustify because this would require a lot of validation work to make sure some new options are not violating the code style.


joakimeriksson:
I tried it but I did not see the issues - how did you run uncrustify to get these warnings?
(tested with what I got in docker ubuntu 20.04 - 0.69).

Yagoor:
@joakimeriksson: I got confused. I was running uncrustify locally (version 0.75.1_f) because the docker image does not come with uncrustify. You are right, running it with the version from inside docker (manually installing it), I don't see the errors. I'll close this PR, it does not make sense to have this fix.

PR #2005, opened 2022-07-02
examples: use UIPLIB_IPV6_MAX_STR_LEN

pjonsson:


arurke:
I am not familiar with this part of the code, but there is the `UIPLIB_IPV6_MAX_STR_LEN` in uiplib.h. However, for some reason it does not consider IPv4-mapped addresses, and thus have a size of only 40. Would it make sense to try to make one source of truth and for example change that existing one to 46?

On a related note, this macro might also be applicable here:
* https://github.com/contiki-ng/contiki-ng/blob/develop/examples/mqtt-client/mqtt-client.c#L169
* https://github.com/contiki-ng/contiki-ng/blob/develop/examples/mqtt-client/mqtt-client.c#L569
* https://github.com/contiki-ng/contiki-ng/blob/develop/examples/platform-specific/cc26x0-cc13x0/cc26x0-web-demo/mqtt-client.c#L638
* https://github.com/contiki-ng/contiki-ng/blob/develop/examples/platform-specific/nrf52dk/mqtt-demo/mqtt-demo.c#L119
* https://github.com/contiki-ng/contiki-ng/blob/develop/examples/platform-specific/nrf52dk/mqtt-demo/mqtt-demo.c#L483

(And some de-duplication might be possible in those examples by using `uiplib_ipaddr_snprint()`, but that is probably for another PR)

pjonsson:
I've updated the patch.

As a general remark, I think it would be beneficial for Contiki-NG in the long run if things that aren't uIP-specific were named the same thing as on other platforms, the barrier to entry would be reduced, and it might even enable unit-testing of code on host.

arurke:
> As a general remark, I think it would be beneficial for Contiki-NG in the long run if things that aren't uIP-specific were named the same thing as on other platforms, the barrier to entry would be lowered, and it might even enable unit-testing of code on host

Agreed. I will see if I can find time to make a PR in this direction.

nfi (approved):

joakimeriksson (approved):

PR #2012, opened 2022-07-03
list: put const in right place

pjonsson:
The intention of the const on the parameters
of the list functions is that they do not
modify that argument. Putting const on the list_t
type makes the type void **const rather than
the intended void *const *.

### DIFF
--- a/os/lib/list.c
+++ b/os/lib/list.c
@@ -79,7 +79,7 @@ list_init(list_t list)
  * \sa list_tail()
  */
 void *
-list_head(const list_t list)
+list_head(const_list_t list)
 {
   return *list;
 }
@@ -97,7 +97,7 @@ list_head(const list_t list)
  * \param src The source list.
  */
 void
-list_copy(list_t dest, const list_t src)
+list_copy(list_t dest, const_list_t src)
 {
   *dest = *src;
 }
@@ -114,7 +114,7 @@ list_copy(list_t dest, const list_t src)
  * \sa list_head()
  */
 void *
-list_tail(const list_t list)
+list_tail(const_list_t list)
 {
   struct list *l;
 
@@ -269,7 +269,7 @@ list_remove(list_t list, const void *item)
  */
 /*---------------------------------------------------------------------------*/
 int
-list_length(const list_t list)
+list_length(const_list_t list)
 {
   struct list *l;
   int n = 0;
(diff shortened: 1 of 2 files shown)


arurke:
I have not had time to look at this yet, but if true it might be so for more libraries as well, ref. the PR where it was introduced: https://github.com/contiki-ng/contiki-ng/pull/1705

nfi (approved):

arurke:
Apologies for no review, I am travelling and will be for the next two weeks as well

arurke:
The technical side looks good (and the same seems to be true for circular-list, dbl-circ-list, dbl-list, from #1705).

But I wanted to air a different perspective: It seems we are trading the benefits of consts (here mostly a potential gain in performance & code size), for reduced user-friendliness. From user perspective there would be two different datatypes when using these libraries. The typedefs themselves are not easy to read, however the comments are clear imo. I don't have a strong opinion either direction, but I wanted to raise the discussion. Thoughts @nfi ? @pjonsson ? (to be clear: I am thinking the alternative is to remove the consts altogether by reverting parts of #1705 )

pjonsson:
On a conceptual level, I don't see the distinction between the types `const_list_t` and `list_t` to be any different than the distinction between the types `const int *` and `int *`, and a Contiki-NG programmer is already familiar with the latter.

My experience is that a lot of time can be saved when things are ensured by tools, reviewers know that the source code would not compile if  the list_head function did something bad with the list, and the application programmer looking for a list corruption bug can stop looking at list_head after seeing the function declaration.

alexrayne:
IMHO, actualy nowhere in the code you see const list, always lists are vars - can changes on run-time. therefore, if compiler not allows casts list as const-list argument by design, programmer always have to do it explicit, manual, and everywhre. This code is a disaster. 

On other side, non-const list argument looks as a warning - it may be changed by called function. So programmer false-awared about list usage on design time. IMHO, this is not very much pay vs explisit cost-casts everywhere. 

At least, some finstions, that not change argue list, may be declaread with a const or pure atribute.

arurke (approved):

arurke:
Thanks for the discussion. Apologies @alexrayne, I am having trouble fully understanding your comment.

@pjonsson Would you mind rebasing and I'll see if we can't get this merged.

PR #2014, opened 2022-07-04
sixtop: return right enum value

pjonsson:
The function returns a sixp_trans_mode_t,
so use an enum value from that enum.


yatch (approved):

yatch:
Thanks!

Issue #1962, opened 2022-06-16
Wiki/Vagrant: bootstrap.sh should not be run with sudo


pjonsson:
The Vagrant documentation on the Wiki (https://github.com/contiki-ng/contiki-ng/wiki/Vagrant) lists the command:
```
$ sudo ./contiki-ng/tools/vagrant/bootstrap.sh
```
The individual commands inside bootstrap.sh that need root permissions use sudo, so the sudo when calling the script should be removed.

nfi:
Fixed via #2024.

PR #2024, opened 2022-07-04
Migrate Github wiki to Read the Docs

alexstanoev:
This PR migrates the existing Github wiki over to `doc/` inside the source tree, configures Sphinx and Read the Docs to render it, and updates the Docker image with any additional package requirements.
This will replace the current RTD page, which only hosts generated Doxygen, but will not modify any existing Doxygen URLs under `_api`.

Where required, Markdown heading levels have been made consistent, and titles have been added/modified, to make the build pass without warnings. 
All documentation files will be copied under `tools/readthedocs` as part of the build process, and any additional external directories can be added via `tools/readthedocs/conf.py`.

### Structure

The documentation has been organised into the following categories:

- Getting Started - under `doc/getting-started`
- Programming Contiki-NG - under `doc/programming`
- Tutorials - under `doc/tutorials`
- Platforms - under `doc/platforms`
- Organization - under `doc/project`
- Examples - contains all `.md` and `.rst` files under `examples/`
- API documentation - preexisting Doxygen API documentation (original index page)

### Motivation

This change allows us to:

- Have version-specific documentation online and as part of the source tree.
- Accept documentation pull requests using existing processes.
- Host our documentation using a service that can be indexed by search engines, as Github prohibits crawlers.
- Improve documentation organisation, discoverability, and search features.

Preview: https://contiki-ng--2024.org.readthedocs.build/en/2024/


tools/docker/Dockerfile
```diff
@@ -105,8 +105,10 @@ RUN wget -nv https://www.nordicsemi.com/-/media/Software-and-other-downloads/Des
 RUN python3 -m pip -q install --upgrade pip && \
     python3 -m pip -q install \
       setuptools \
-      sphinx_rtd_theme \
-      sphinx \
+      'sphinx_rtd_theme==1.0.0' \
+      'myst-parser==0.18.0' \
+      'linkify-it-py==2.0.0' \
+      'sphinx==5.0.2' \
```
pjonsson, line 111: Please leave a comment about why the latest version of everything doesn't work.

alexstanoev: This is the latest version of everything at the time of this PR - the goal is to keep the versions pinned and consistent between RTD (`tools/readthedocs/requirements.txt`) and the Docker image to prevent documentation builds from breaking in the future. 
We can bump as required, keeping `Dockerfile` and the RTD requirements file consistent.

.readthedocs.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+version: 2
+
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.8"
+  apt_packages:
+    - make
+    - doxygen
```
pjonsson, line 13: The Doxygen package in Ubuntu 20.04 gave unfounded warnings (mqtt got warnings because the preprocessor directives in Doxyfile had no effect, and some simplelink platform had `/**===============...` interpreted as `<h1>` with a followup failure). Can this use Doxygen 1.9.4 just like the docker image?

tools/readthedocs/requirements.txt
```diff
@@ -0,0 +1,4 @@
+sphinx==5.0.2
```
pjonsson, line 1: Do these need to be in sync with the Docker image versions?

alexstanoev: Yes, ideally - the CI pipeline builds readthedocs (`00-documentation`) in parallel with the newly enabled RTD hook on pull requests, and the two should use a consistent environment.

.readthedocs.yaml
alexstanoev: Can you please check if you can see wrongly generated doxygen at https://contiki-ng--2024.org.readthedocs.build/en/2024/_api/modules.html for those files?

The build output from RTD had the following warnings:
```
api-doc building API docs from "../doxygen"
api-doc executing "make -C ../doxygen > /dev/null"
warning: ignoring unsupported tag 'CREATE_SUBDIRS_LEVEL' at line 92, file Doxyfile
warning: ignoring unsupported tag 'PYTHON_DOCSTRING' at line 239, file Doxyfile
warning: ignoring unsupported tag 'NUM_PROC_THREADS' at line 477, file Doxyfile
warning: ignoring unsupported tag 'RESOLVE_UNNAMED_PARAMS' at line 547, file Doxyfile
warning: ignoring unsupported tag 'SHOW_HEADERFILE' at line 620, file Doxyfile
warning: ignoring unsupported tag 'WARN_IF_INCOMPLETE_DOC' at line 838, file Doxyfile
warning: ignoring unsupported tag 'WARN_LINE_FORMAT' at line 878, file Doxyfile
warning: ignoring unsupported tag 'CLANG_ADD_INC_PATHS' at line 1173, file Doxyfile
warning: ignoring unsupported tag 'DOCSET_FEEDURL' at line 1403, file Doxyfile
warning: ignoring unsupported tag 'FULL_SIDEBAR' at line 1614, file Doxyfile
warning: ignoring unsupported tag 'OBFUSCATE_EMAILS' at line 1645, file Doxyfile
warning: ignoring unsupported tag 'HTML_FORMULA_FORMAT' at line 1656, file Doxyfile
warning: ignoring unsupported tag 'MATHJAX_VERSION' at line 1704, file Doxyfile
warning: ignoring unsupported tag 'DOT_UML_DETAILS' at line 2495, file Doxyfile
warning: ignoring unsupported tag 'DOT_WRAP_THRESHOLD' at line 2504, file Doxyfile
warning: ignoring unsupported tag 'DIR_GRAPH_MAX_DEPTH' at line 2577, file Doxyfile
api-doc moving "../doxygen/html" to "/home/docs/checkouts/readthedocs.org/user_builds/contiki-ng/checkouts/2024/tools/readthedocs/_build/html/_api"
build succeeded.
```

Ideally we would stick to the version provided by Read the Docs unless it generates broken documentation, as they appear to discourage overriding parts of the build.

tools/docker/Dockerfile
pjonsson: It would be nice to have a single master for the version information, does something like the following work?
```
COPY ../../tools/readthedocs/requirements.txt /tmp
RUN python3 -m pip -q install --upgrade pip && \
    python3 -m pip -q install \
      setuptools \
      matplotlib \
      'protobuf<=4' && \
    python3 -m pip -q install \
      nrfutil && \
    python3 -m pip -q install -r /tmp/requirements.txt && \
    rm -rf /root/.cache /tmp/requirements.txt
```

alexstanoev: Yes, I wanted to do something like that too but the COPY trick never crossed my mind!
I'll update the PR in a bit with this.

.readthedocs.yaml
pjonsson: Yes, mqtt.h is not processed with MQTT_5 enabled so https://contiki-ng--2024.org.readthedocs.build/en/2024/_api/mqtt_8h.html is missing `prop_list` as argument to `mqtt_connect` (and a bunch of other functions).

You need to look in doxygen.log for the doxygen warnings if that is the same way as 00-documentation is set up.

alexstanoev: Would Doxygen 1.9.1 generate these correctly? 
That's the version in Ubuntu 22.04, which is a supported RTD build environment as well; everything else will be fine as it's just sphinx modules. 
I can easily try this change in the next commit before seeing if we can override the preinstalled system doxygen.

tools/docker/Dockerfile
pjonsson: The trick breaks stuff in our CI with calculating when docker images should be rebuilt, but it should work for your update, and the rest can be fixed after the release.

Please add a comment to requirements.txt that it is used by the Dockerfile, these kind of things are impossible to find for someone who does a drive-by and isn't looking in the right place by accident.

.readthedocs.yaml
pjonsson: Doxygen randomly breaks stuff between minor releases. I don't remember what was broken with 1.9.1, it's much better than the Ubuntu 20.04 version, but #1966 was for sure still broken with 1.9.1.

pjonsson:
I'm guessing this PR will go into the release, can you please add a commit that fixes #1962 into the PR?

.readthedocs.yaml
alexstanoev: I've got it building with 1.9.4 now, so this should be resolved.

tools/docker/Dockerfile
alexstanoev: This won't work it seems, Docker complains with `COPY failed: forbidden path outside the build context: ../../tools/readthedocs/requirements.txt ()`.

pjonsson: Put `tools/readthedocs/requirements.txt` in `tools/docker/files` instead, and change the Dockerfile to:
```
# Requirements referenced from tools/readthedocs.
COPY files/requirements.txt /tmp
# Matplotlib is required for result visualization.
# Nrfutil 6.1.3 does not work with protobuf 4, so install latest 3.x
# Keep the image size down by removing the pip cache when done.
RUN python3 -m pip -q install --upgrade pip && \
    python3 -m pip -q install \
      setuptools \
      matplotlib \
      'protobuf<=4' && \
    python3 -m pip -q install \
      nrfutil && \
    python3 -m pip -q install -r /tmp/requirements.txt && \
    rm -rf /root/.cache /tmp/requirements.txt
```
And put:
```
-r ../docker/files/requirements.txt
```
in `tools/readthedocs/requirements.txt`.

joakimeriksson (approved):

sungminlee114:
Names of two file: `doc/project/Where-to-report-issues,-ask-questions,-etc.?.md` and `doc/tutorials/Cooja:-simulating-a-border-router.md` seems creating errors when cloning the repository into Windows environment.

Here are the specifics.
When I clone by 
`git clone https://github.com/contiki-ng/contiki-ng.git`,
I get
`error: invalid path 'doc/project/Where-to-report-issues,-ask-questions,-etc.?.md'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'`.

I've solved this issue temporally by
[`git config core.protectNTFS false`](https://stackoverflow.com/questions/63727594/github-git-checkout-returns-error-invalid-path-on-windows)
then
`git checkout`,

but the best solution might be fixing the file names.
Could this be fixed?

joakimeriksson:
Ok, yes that is probably a problematic filename in windows. Is this the only file that will cause problem?
I would not mind the file to be renamed - but of course all links would need to be updated also. Would you do a PR on this - that would work in windows? (We are not testing in Windows environment therefore we have not seen this).

sungminlee114:
Yes. It's fine without those two. I'll make a PR on this ASAP.

PR #2025, opened 2022-07-05
Add a TCP socket test

nvt:
A basic test for TCP sockets involving a client and a server that communicate using link-local addresses. The test succeeds when the client has sent 100 kB.

Also changed some file references in other tests to distinguish between UDP and TCP code.


tests/07-simulation-base/28-ipv6-tcp-sockets.csc
```diff
@@ -0,0 +1,171 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<simconf>
+  <project EXPORT="discard">[APPS_DIR]/mspsim</project>
```
pjonsson, line 3: These two project lines are not required (and should not be generated by the latest Cooja).

nvt: The "project" lines have been removed now.

pjonsson (approved):

nfi (approved):

PR #2027, opened 2022-07-05
Fix storage examples to compile on the native platform.

nvt:
Additionally, this PR fixes an aliasing problem in the MaxHeap module and removes the deprecated test-coffee example.


pjonsson:
If these work on native, can they be enabled for Cooja too?

nvt:
> If these work on native, can they be enabled for Cooja too?

I don't think there is a Coffee port for the cooja platform available, but it could be interesting to add it in a separate PR.

pjonsson (approved):

nfi (approved):

PR #2030, opened 2022-07-06
Fix -Wempty-body warning

pjonsson:
Clang with -Wall gives a -Wempty-body
warning in these places, add a newline
to make it obvious that the loop does
nothing.

Uncrustify does not change this formatting.

### DIFF
--- a/arch/cpu/cc2538/lpm.c
+++ b/arch/cpu/cc2538/lpm.c
@@ -70,8 +70,8 @@ rtimer_clock_t lpm_stats[3];
   do { memset(lpm_stats, 0, sizeof(lpm_stats)); } while(0)
 #define LPM_STATS_ADD(pm, val)   do { lpm_stats[pm] += val; } while(0)
 #else
-#define LPM_STATS_INIT()
-#define LPM_STATS_ADD(stat, val)
+#define LPM_STATS_INIT() do {} while (0)
+#define LPM_STATS_ADD(stat, val) do {} while (0)
 #endif
 /*---------------------------------------------------------------------------*/
 /*
--- a/arch/cpu/cc26x0-cc13x0/rf-core/rf-switch.h
+++ b/arch/cpu/cc26x0-cc13x0/rf-core/rf-switch.h
@@ -90,10 +90,10 @@ void rf_switch_power_down(void);
  */
 void rf_switch_select_path(uint8_t path);
 #else
-#define rf_switch_init()
-#define rf_switch_power_up()
-#define rf_switch_power_down()
-#define rf_switch_select_path(p)
+#define rf_switch_init() do {} while (0)
+#define rf_switch_power_up() do {} while (0)
+#define rf_switch_power_down() do {} while (0)
+#define rf_switch_select_path(p) do {} while (0)
 #endif /* RF_SWITCH_ENABLE */
 /*---------------------------------------------------------------------------*/
 #endif /* RF_SWITCH_H_ */


arurke:
After reading a bit I get the impression clang should allow this. Any idea why it reacts? Which version of clang are you using?

pjonsson:
I'm using Clang-14 on Ubuntu 22.04.

Writing `while(cond);` is a compact way to represent a loop, but the semi-colon is easy to miss, and apparently leaving out the braces isn't aligned with the Contiki-NG coding style.  I don't think the additional line hurts the readability and there shouldn't be that many loops with empty bodies, so it seems easier to just do what clang wants.

arurke:
There are a lot these `while(expression);` spread throughout the code base. There is even an identical one to the one you catched in rf-core.c: https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/cc26x0-cc13x0/rf-core/rf-core.c#L295

And a quick test (and [discussion online](https://stackoverflow.com/questions/46502990/why-clang-wempty-body-doesnt-detect-this-case)) indicate that clang should not raise a warning on these. Perhaps this is related to how the defines in the expressions are expanded in your execution with clang? Or am I missing something here?

pjonsson:
You're right, most places don't give a warning, so I've removed a couple of while statements from the patch.

I'm not sure what you are trying to say though, I invoked clang with `make CC=clang-14 AS=clang-14 LD=clang-14`, who is going to invoke Clang in a different way and/or have different defines with the Contiki-NG code base?

arurke:
> I'm not sure what you are trying to say though, I invoked clang with make CC=clang-14 AS=clang-14 LD=clang-14, who is going to invoke Clang in a different way and/or have different defines with the Contiki-NG code base?

Sorry, I was grasping for straws trying to understand why clang was complaining, thinking perhaps you were doing something in the invocation.

In any case: The `while(expression);` idiom is very popular, there are hundreds of these throughout the code-base (also many more or less identical ones to those in this patch). And as mentioned, afaik clang is supposed to accept these, so there is something here causing false positives. It therefore doesn't seem ideal to start on the road suggested in this PR - potentially having us "pollute" a lot of "correct" code (as seen from clang intentions) with braces. I have not tried clang - is the lines in this PR the only occurrences which raise warnings? If so I guess the slippery slope is not so slippery and the PR makes more sense.

arurke:
I found why clang complains: The next line of code after the `while(expression);` is an empty define (such as `rf_switch_power_down()` which has a body only if there is RF switch on the board). The preprocessor yields an indented `;` on that line, and this causes clang to raise the warning. A quick example: https://godbolt.org/z/oG7nqPa8d.

I asked in the clang project if this is a bug, https://github.com/llvm/llvm-project/issues/56452.

arurke:
My thinking was that if its a bug we should rather suppress until they have fix instead of doing changes to the code. Especially since the issue is typically dependent on configuration (like in the case of `rf_switch_power_down()`) it would be hard catch all cases and avoid the warnings through code-changes. Happy to be convinced otherwise though.

pjonsson:
This PR fixes the warnings in 02-compile-arm-ports. Unless you have prodded someone you know to fix the issue in Clang, the chances are that the person who implemented the heuristics no longer works on LLVM, and your question is lost among the other 56k issues reported on LLVM.

arurke:
Apologies, I am probably way too pedantic here - I can't help thinking adding these braces "randomly" (from the view of future developers) seems like something we should avoid if possible. I'll abstain and tag this for additional review. An optional solution could be:
```
diff --git a/arch/cpu/cc26x0-cc13x0/rf-core/rf-switch.h b/arch/cpu/cc26x0-cc13x0/rf-core/rf-switch.h
index 8a701ba74..db89453f4 100644
--- a/arch/cpu/cc26x0-cc13x0/rf-core/rf-switch.h
+++ b/arch/cpu/cc26x0-cc13x0/rf-core/rf-switch.h
@@ -90,10 +90,10 @@ void rf_switch_power_down(void);
  */
 void rf_switch_select_path(uint8_t path);
 #else
-#define rf_switch_init()
-#define rf_switch_power_up()
-#define rf_switch_power_down()
-#define rf_switch_select_path(p)
+#define rf_switch_init()          do {} while (0)
+#define rf_switch_power_up()      do {} while (0)
+#define rf_switch_power_down()    do {} while (0)
+#define rf_switch_select_path(p)  do {} while (0)
 #endif /* RF_SWITCH_ENABLE */
 #endif /* RF_SWITCH_H_ */
```

This avoids the warning, but is also a known idiom, and already used a handful of places in the codebase. This makes the solution more maintainable in my view (has a clearer purpose, would need reasoning to be removed). It also gives additional benefits such as detecting if `;` is missing in the call.

nfi:
I think the suggested optional solution by @arurke looks nice and should generically solve the issues while keeping the code style.

nfi (approved):

PR #2033, opened 2022-07-06
Makefile: add source exclusion mechanism and use it

pjonsson:


examples/slip-radio/Makefile
```diff
@@ -5,6 +5,9 @@ all: $(CONTIKI_PROJECT)
 PLATFORMS_EXCLUDE = native nrf52dk
 BOARDS_EXCLUDE = nrf52840/dongle
 
+# The sky directory overrides the architecture putchar().
+CONTIKI_SOURCES_EXCLUDES += uart1-putchar.c
+
```
nfi, line 10: Perhaps it would be better to place this in `examples/slip-radio/sky/Makefile.sky` since it is only needed when building for Sky platform?

pjonsson: I assumed that somehow involved the magic around `$(TARGETDIRS)` in `Makefile.include`, but that is not a variable defined/known by GNU Make. Cooja doesn't mention TARGETDIRS either. Is TARGETDIRS leftovers from an earlier build system?

After 15 minutes of digging around I figured out that `Makefile.sky` would be included through `MODULES_REL`. I find the fact that I've spent 3 months on the build system and still can't figure out how it works a bit concerning.

I'm not sure if what you suggest works, but I think it further cements the architecture of the build system in a way that is not workable long-term without heroic efforts of someone. You're right that we probably shouldn't exclude the file for every target though, so I'll wrap the thing inside an ifeq on `$(TARGET)` tomorrow.

pjonsson: I realized we could avoid the if by just extending the filter-out to also include `*_EXCLUDES_$(TARGET)`. This makes it possible for applications to have build files such as:

```make
ifndef BIG_FEATURE
MODULES_SOURCES_EXCLUDES += file1.c file2.c
endif

# file1.c is never required for MSP430.
MODULES_SOURCES_EXCLUDES_sky += file1.c
MODULES_SOURCES_EXCLUDES_z1 += file1.c
```
so the developer can look in one (short) file and understand what is included and excluded.

nfi (approved):

PR #2034, opened 2022-07-06
6tisch/simple-node: disable LOG_CONF_LEVEL_FRAMER

pjonsson:
SimpleLink CC13xx/CC26xx prints a message in the regression
tests that it disables LOG_CONF_LEVEL_FRAMER. Stop
defining LOG_CONF_LEVEL_FRAMER for those configurations.


arurke:
@pjonsson Could we use the `CONTIKI_TARGET_SIMPLELINK` macro instead? That would be cleaner and more maintainable.

arurke:
@pjonsson Thinking some more about this, I worry about the scalability of this approach as we add more examples or platforms with other limitations. It also clutters the example-code which imo should be as clean as possible. Thinking aloud: Could we make a flag indicating we are running tests, and use that to squelch the warning?

joakimeriksson (approved):

PR #2040, opened 2022-07-08
Renamed get and remove tsch link functions

tinstructor:
@atiselsts Since there is a legitimate use for a function that checks whether a link exists based solely on slot offset (a.k.a., "timeslot"), it would be better to rename the functions `tsch_schedule_remove_link_by_timeslot` and `tsch_schedule_get_link_by_timeslot` to reflect the changes introduced in #1162 , where these functions were modified to operate on timeslot and channel offset, instead of just timeslot.

### DIFF
--- a/examples/6tisch/etsi-plugtest-2017/sf-plugtest.c
+++ b/examples/6tisch/etsi-plugtest-2017/sf-plugtest.c
@@ -247,7 +247,7 @@ delete_cell(const linkaddr_t *peer_addr, const sf_plugtest_cell_t *cell)
   channel_offset = cell->channel_offset[0] + (cell->channel_offset[1] << 8);
 
   if((slotframe = tsch_schedule_get_slotframe_by_handle(0)) == NULL ||
-     tsch_schedule_remove_link_by_timeslot(slotframe, timeslot, channel_offset) == 0) {
+     tsch_schedule_remove_link_by_offsets(slotframe, timeslot, channel_offset) == 0) {
     LOG_ERR("cannot delete a cell\n");
     return -1;
   }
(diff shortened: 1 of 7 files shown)


arurke:
This makes sense to me. A minor on code-style: I know the files are already teeming with them, but would you mind not adding new lines which are >80 chars?

tinstructor:
@arurke I've now limited the lines to <= 80 chars where possible (some lines weren't changed because of readability)

alexrayne:
>> it would be better to rename the functions tsch_schedule_remove_link_by_timeslot and tsch_schedule_get_link_by_timeslot

this functions alredy used by users of contiki. Would be care for thouse users, if declare old funstions as deprecated, and introduce new with better names, insteard replace old ones.

atiselsts:
1. Can you rebase this PR to squash the commits?

2. I've just closed and reopened it to force CI checks to run again, but it looks that the docs build is still failing. Anyone from the team want to look at this? @arurke @alexstanoev

tinstructor:
@atiselsts done

atiselsts (approved):

arurke:
I agree with @alexrayne's sentiment not to change APIs like this, but I can't really see a way to deprecate without raising warnings for users who don't use this function. We could keep the old functions and add a comment to that they are deprecated and will be removed in the next release. But I can't imagine this being noticed by existing users, and it is thus just kicking the rock/problem/surprise one release down the road (could add to release notes or similar to increase visibility). Thoughts @atiselsts? Any precedence I am not aware of perhaps?

When we reach C23 there is a [deprecated-attribute](https://en.cppreference.com/w/c/language/attributes/deprecated) available.

alexrayne:
> But I can't imagine this being noticed by existing users, and it is thus just kicking the rock/problem/surprise one release down the road (could add to release notes or similar to increase visibility).

i see no problem, if old API will co-exist with new one for 10 releases and more. Another aproach to get rid that deprecations, that used by CMSIS, and some another libraries - remove deprecated declarations to some another header called like 'xxx-legacy.h', that  include in origin header.

pjonsson:
@arurke there's a `CC_CONF_DEPRECATED(msg)` in cc-gcc.h.

PR #2043, opened 2022-07-08
Makefile: add more DEPFILES

pjonsson:
Add the CONTIKI_PROJECT-related dependency
files to DEPFILES.


pjonsson:
Some tests failed with the initial patch so split this into two, one part that can land now and improves the current situation, and the other part that triggered the failures.

nfi (approved):

PR #2051, opened 2022-07-11
Makefile: fix arch compilation dependencies

pjonsson:
Align the architecture specific build rules
with the rules in Makefile.include. This
prevents make from throwing in an extra
mkdir for a directory that already exists.


nfi (approved):

pjonsson:
Rebased to develop and shrunk to avoid causing conflicts with #2088.

PR #2057, opened 2022-07-12
CI: keep ccache size within limits

pjonsson:
The ccache configuration done by the action
is not available inside the docker container,
so perform a cleanup after completing the tests.


nfi (approved):

arurke:
@pjonsson Out of curiosity: What does the "2023" in `FIXME: (2023)` signify?

pjonsson:
Oh, when there's something blocking a FIXME to be implemented, I try to put something to indicate when it should/can be fixed, "FIXME: (GCC-10) .." for when things are compiled with GCC-10,  "FIXME: (2023) .." for next year.

Next year is pretty coarse grained, it's short to write and there shouldn't be many active pull requests that are branched in the window of when ccache was active in the CI but before this PR is merged.

PR #2059, opened 2022-07-13
cooja: wrap dbg-io funs and call them

pjonsson:
This removes the non-errorchecked symbol
renaming done through objcopy and uses
the --wrap parameter to ld instead.

MacOS X does not come with objcopy so
the symbol renaming has never been
active on that platform.


nfi (approved):

pjonsson:
Rebased on latest develop to fix the merge conflicts.

The `03-nullnet-broadcast.csc` test will fail if the printf redirection is broken. Niclas tested this PR on OS X before he went on vacation.

@arurke I have more changes on top of this PR and it's a bit awkward to work with that, could you take a look at this PR and possibly merge it?

PR #2069, opened 2022-07-14
CI: use hashFiles for docker image name

pjonsson:
Using the git commit hash on the Dockerfile
ignores the content of the files copied into
the docker image. Compute a hash of everything
and use that to identify the docker image
instead.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
     # Construct the correct docker container image tag corresponding to this build
     - name: Figure out correct docker image tag
       run:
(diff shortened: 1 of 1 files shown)


arurke:
A possible regression: This PR removes the link between the image versions and the corresponding commit in our repo. Thus it is no longer possible to pick images as described in https://github.com/contiki-ng/contiki-ng/wiki/Docker#using-a-different-docker-image, and it is in general difficult to learn which commit an image belongs to. Thoughts @pjonsson? Not to versed with dockerhub, am I missing something?

pjonsson:
@arurke The old way didn't pick up changes to the scripts, fixed with a new hash computation in #2107.

PR #2076, opened 2022-07-15
Stop defining reserved identifiers

pjonsson:
All identifiers that begin with an underscore
followed by a capital letter or by another
underscore are reserved in C. Use other names
for headerguards.


nvt:
It is good to avoid reserved identifiers, but could you remove one of the underscores in the end of the names also for consistency with other headers?

arurke (approved):

PR #2077, opened 2022-07-17
Makefile: add link-time-optimization option

pjonsson:
Add a build system option "LTO" so examples
can be built with "make <options> LTO=1".
Targets that do not support the option can
define NO_LTO_TARGET to have the PLATFORM_ACTION
set to skip when LTO is enabled.

The only targets that do not support
LTO are the msp430-based ones, the toolchain
recognizes the switch but crashes on every
program in the regression tests.

The size for hello-world in the docker container
without LTO:

  text	   data	    bss	    dec	    hex	filename
 118948	   2952	1110912	1232812	 12cfac	hello-world.native

and with LTO:

  text	   data	    bss	    dec	    hex	filename
  48579	   1728	  56568	 106875	  1a17b	hello-world.native

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -139,6 +139,17 @@ endif
 
 CP ?= cp
 
+# Configure link-time-optimization (LTO).
+FLAGS_LTO ?= -flto
+
+ifdef LTO
+  CFLAGS += $(FLAGS_LTO)
+  LDFLAGS += $(FLAGS_LTO)
+  ifdef NO_LTO_TARGET
+    PLATFORM_ACTION = skip
+  endif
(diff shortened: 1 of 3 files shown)


nfi (approved):

arurke:
@joakimeriksson This PR also contains a template for release notes. Seems like a good solution. Opinions before merging?

joakimeriksson:
> @joakimeriksson This PR also contains a template for release notes. Seems like a good solution. Opinions before merging?

i think it is a good idea to have that. Will merge - we have plenty of time to update the template before the next release.

PR #2078, opened 2022-07-18
Use renode deb from releases instead of building from source

Yagoor:
This is possible now that we are using a 64 bits docker image

Before and after. The git hash is the same, just the date is different.

```bash
❯ contiker
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

user@b08f5ae283d1:~/contiki-ng$ renode --version
Renode, version 1.13.0.15726 (e7f96b97-202207170844)
user@b08f5ae283d1:~/contiki-ng$ ^C
user@b08f5ae283d1:~/contiki-ng$ logout
❯ contiker
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

user@83d75c1fdb78:~/contiki-ng$ renode --version
Renode, version 1.13.0.13110 (e7f96b97-202204300716)
user@83d75c1fdb78:~/contiki-ng$
```

PR #2080, opened 2022-07-19
CI: increase the workflow timeout

pjonsson:
An unusually long docker image build will
make the tests timeout, so increase the timeout
slightly.


pjonsson:
Don't go too crazy with increasing the timeout, #2078 will reduce the time it takes to build the docker image.

nfi (approved):

PR #2084, opened 2022-07-21
arm: remove false dependencies

pjonsson:
The hex and bin files are not required
for the target according to the comment,
so stop building them.

This is preparation to remove the customrules
for Cortex-M, but it also saves IO load
on the IO-limited CI machines.


arurke (approved):

lafka:
This breaks when expecting bin/hex files to be outputted. AFAIK using HEX files is the most common way to distribute firmware for cc13xx/26xx. There should be a deprecation warning first such that users are warned, and can take action, before breaking existing development workflows and CI pipelines.

PR #2092, opened 2022-07-26
nrf: remove naked in hardfault-handler.c

pjonsson:
Naked functions are only supposed to contain
inline assembly, and this function contains
regular C statements. Remove the naked
attribute for the non-assembly version
of the function.

### DIFF
--- a/arch/cpu/nrf/arm/hardfault-handler.c
+++ b/arch/cpu/nrf/arm/hardfault-handler.c
@@ -215,7 +215,7 @@ HardFault_Handler(void)
  * @brief Hardfault handler
  * 
  */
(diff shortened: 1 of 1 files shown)


pjonsson:
@Yagoor hardfault-handler.c has your name in it, do you see any issues with removing the naked attribute for the C version of the function?

alexstanoev:
This is fine, the attibute is only needed to avoid function prologues as per the handler in the SDK:
https://github.com/contiki-ng/nrf-sdk-contiki-ng/blob/master/components/libraries/hardfault/nrf52/handler/hardfault_handler_gcc.c

alexstanoev (approved):

PR #2094, opened 2022-07-26
CI: reduce fetch-depth

pjonsson:
Reduce the fetch-depth from the entire
history down to 2 now that the Docker
image hash is computed based on file
content rather than the git hash.

A fetch-depth of 2 should be enough to
detect which files that were changed
in a PR, for example for running additional
checks such as code-formatting or similar
on them.


pjonsson:
This seems to save roughly 20 seconds on the typical 40-50s checkout time.

joakimeriksson (approved):

nfi (approved):

PR #2102, opened 2022-07-30
msp430: only pass -Wall once

pjonsson:
The flag is already added to CFLAGS
a few lines below this definition.


nvt (approved):

arurke:
@pjonsson We got a failure in `07-simulation-base/21-cooja-rpl-tsch-security` after this was merged, see https://github.com/contiki-ng/contiki-ng/runs/7638381912?check_suite_focus=true. Would you mind having a look?

PR #2128, opened 2022-08-11
Dockerfile: use graalvm instead of openjdk

pjonsson:
Use GraalVM as the JVM. This saves 50 Mb
of space for the Docker image and runtime
on the tests:

Runtime for 07-simulation-base before:

real 144.14
user 269.64
sys 16.18

After:

real 133.04
user 242.39
sys 19.22

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -10,7 +10,6 @@ RUN apt-get -qq update && \
     apt-get -qq -y --no-install-recommends install \
       ca-certificates > /dev/null && \
   apt-get -qq -y --no-install-recommends install \
-    ant \
     build-essential \
     gdb \
     git \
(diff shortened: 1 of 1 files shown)


joakimeriksson (approved):

nfi (approved):

alexrayne:
How to use this new docker image with cooja?

contiki builds with include-pathes from default Vm.  since graalvm - not a default VM, platform.c try builds with openjdk VM, that alredy have no include sources. Therefore gcc crushes to build, since absent <jni.h>

I have to reinstall openjdk gain to allow cotiki make builds project.

what i do wrong?

pjonsson:
I'm not sure I understand your question, the only JVM in the container after this change is GraalVM, so it must be the default.

If we're talking about the Contiki-NG build system, that queries `$(JAVA)` (which defaults to `java`) for the information to construct the include path to jni.h.

You can probably run into strange problems if you are trying to run an old version of Cooja/Contiki-NG with a more recent image (or vice versa). Use the image that corresponds to the version of Cooja/Contiki-NG you are working with.

alexrayne:
i`ve inspects your image again - lt have graalvm as default, as you say.

possibly, i have install and switches to openjdk when install visualvm.

Thank you

pjonsson:
GraalVM has its own package system for some reason, `sudo gu install visualvm` should install visualvm without replacing the default JDK/JVM in the container.

PR #2134, opened 2022-08-12
Avoid rebuilds of ieee-addr.o

pjonsson:


nfi (approved):

arurke:
@pjonsson Would you mind giving a short high-level description of the approach?

pjonsson:
@arurke always generate a new headerfile that reflects the nodeid, but write it all to a temporary name. If the contents of the new headerfile is different from the old headerfile, replace the old headerfile with the new headerfile. If the content is unchanged, make stops chasing that dependency chain since the timestamp of the old headerfile remains unchanged.

To avoid the nightmare of quoting and spaces, generate a define that isn't used in the headerfile for the case that no nodeid is provided to make.

I called the mechanism of only updating a file if it has changed `.diffupdate` because `.only_update_if_changed` is too long. The diffupdate mechanism can be applied to anything generated by the build system (such as linker scripts).

PR #2157, opened 2022-08-24
Update coap-example-observe-client.c

yolly98:
It's a possible solution to the compilation errors

### DIFF
--- a/examples/coap/coap-example-client/coap-example-observe-client.c
+++ b/examples/coap/coap-example-client/coap-example-observe-client.c
@@ -139,11 +139,7 @@ PROCESS_THREAD(er_example_observe_client, ev, data)
   PROCESS_BEGIN();
   
   coap_endpoint_parse(SERVER_EP, strlen(SERVER_EP), &server_ep);
-
   static struct etimer et;
-
-  /* store server address in server_ipaddr */
-  SERVER_NODE(server_ipaddr);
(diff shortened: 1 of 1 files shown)


nfi (changes requested):
Looks good, but two lines (line 145-146) with the old server_ipaddr need to be removed for the example to build. Also, it would also be good if you do not add whitespaces on empty lines.

nvt:
@yolly98 Could you please let us know if you intend to address @nfi's comments above?

yolly98:
I updated the code addressing the above requests

examples/coap/coap-example-client/coap-example-observe-client.c
```diff
@@ -115,15 +113,15 @@ notification_callback(coap_observee_t *obs, void *notification,
  * Toggle the observation of the remote resource
  */
 void
-toggle_observation(void)
+toggle_observation(coap_endpoint_t* server_ep)
```
pjonsson, line 116: Can the argument be `const`?

Formatting: the other pointers use the style `type *thing` rather than `type* thing`.

pjonsson:
If you rebase the PR on top of latest develop, the readthedocs test will pass.

arurke:
This looks good to me as well, @yolly98 would you be able to address the comment from pjonsson, and rebasing?

PR #2166, opened 2022-09-01
cooja: simplify Cooja quickstart rule

nfi:
Only run Cooja quickstart on Cooja simulation files. Also, avoid build warnings when, for example, running `make TARGET=cooja viewconf`.


arch/platform/cooja/Makefile.cooja
```diff
@@ -72,9 +72,8 @@ ifneq ($(MAKECMDGOALS),clean)
```
pjonsson, line 72: This is a really nice improvement, can the `ifneq ($(MAKECMDGOALS),clean)` also be removed?

nfi: Yes, that is no longer needed. Fixed now. Thanks for the suggestion!

pjonsson (approved):

nvt (approved):

PR #2173, opened 2022-09-07
Test HeapMem alignment

nvt:
This PR also adds a function to check the alignment used by the heapmem module.

### DIFF
--- a/os/lib/heapmem.h
+++ b/os/lib/heapmem.h
@@ -76,8 +76,6 @@
 
 #include <stdlib.h>
 
-#include "contiki.h"
-
 typedef struct heapmem_stats {
   size_t allocated;
   size_t overhead;
--- a/os/lib/heapmem.h
+++ b/os/lib/heapmem.h
@@ -164,6 +164,13 @@ bool heapmem_free(void *ptr);
 
 void heapmem_stats(heapmem_stats_t *stats);
 
+/**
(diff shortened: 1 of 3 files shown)


os/lib/heapmem.h
```diff
@@ -164,6 +162,13 @@ bool heapmem_free(void *ptr);
 
 void heapmem_stats(heapmem_stats_t *stats);
 
+/**
+ * \brief       Obtain the minimum alignment of allocated addresses.
+ * \return      The alignment value that is a power of two.
+ */
+
+unsigned heapmem_alignment(void);
```
pjonsson, line 170: The C11 `_Alignof` returns `size_t`, is there any reason to use a different type for heapmem alignment than what is used in the next C standard?

tests/08-native-runs/12-heapmem/test-heapmem.c
```diff
@@ -103,6 +111,9 @@ UNIT_TEST(do_many_allocations)
     if(ptrs[alloc_index] == NULL) {
       failed_allocations++;
     } else {
+      if((intptr_t)ptrs[alloc_index] & (min_alignment - 1)) {
```
pjonsson, line 114: Does this need to be signed, or could it use `uintptr_t` instead?

nfi (approved):

PR #2179, opened 2022-09-14
Enable the Cooja platform to build without using Cooja

nfi:
By moving the JNI functionality from platform.c to its separate file, the Cooja platform can be built similarly to other platforms, enabling test compilation for Cooja even if Cooja can not use the resulting binary.


nvt (approved):

pjonsson:
I think the patch looks fine, but I'm concerned that this way of compiling Cooja "binaries" could leave residue in the build tree that results in strange errors in Cooja.

The build system used to refuse to build anything for Cooja without LIBNAME set (I have worked around that several times, but I know that I'm doing something unsupported and won't complain when things break after I forget to clean up my leftovers). I'm not sure what broke that check, but #2177 should simplify reinstating that kind of unconditional check for the Cooja target without impeding the functionality to quickstart Cooja.

PR #2182, opened 2022-09-15
cooja: only allow building from Cooja

pjonsson:
Prevent building Cooja target binaries
from outside of Cooja.

Simulations can still be quickstarted
by using the native target (or any other
target than Cooja), or inside the docker
image:

cscmake file.csc


pjonsson:
This prevents Cooja users from getting stuck at the compilation dialog box because the build system considers everything finished.

PR #2191, opened 2022-09-30
Update script for building msp430 from source

jbrusey:
Also remove leading $ signs from commands as this impedes the user from cutting and pasting the commands into the command line.


jbrusey:
There was a conflict with another edit that I have now resolved. Hopefully this can now be reviewed and merged.

jbrusey:
@pjonsson - Hi there - would you be able to review this for me? Apologies if this is the wrong way of raising this to your attention. :)

arurke:
Hi @jbrusey unfortunately I don't work with the msp430 but I wanted to give a heads-up that the commits could use some cleanup: It seems the last three commits are unintentional / could be squashed with the first one. Also I would suggest that you put the updating of the script in one commit, and the `$`-sign doc. changes in another one (or probably even in a separate PR as it is completely unrelated to the msp changes + there are a lot other places in the doc. which uses the convention so it might be difficult to merge as-is regardless of its merits).

pjonsson:
@jbrusey sorry, it's been a bit hectic the last months before Christmas. I was hoping to get to testing this sooner, but I have this on my list for things before the end of the year unless someone beats me to it.

I agree with @arurke's comments, I hadn't seen the unintentional commits in this PR, and if you make a separate pull request for removing the dollar signs from the documentation, that should be easier to get reviewed and merged.

I don't really do git, so the way I deal with pull requests is to keep things on separate branches, and then rebase the branches against develop when conflicts arise. After a rebase, I do a `git push -f <place> <branch name>`. The workflow doesn't work for interdependent changes/feature branches, but it's simple enough for me to usually avoid shooting myself in the foot.

jbrusey:
@arurke thanks for the feedback and I agree with your comments - I will split into two PRs.

@pjonsson I'm also not so good with git but I appreciate the suggestions. Will convert this to draft, remove the $ sign changes, rebase and then resubmit.

PR #2192, opened 2022-10-01
Improve packet injector

SWW13:
These commits extend the packet injector with the following features:
- `ble-l2cap` protocol
- enable debug logging in all components
- print log when environment variable `DEBUG=1` is set even when all tests succeeded
- injecting multiple packets in one session (directory with files instead of a single file)


tests/20-packet-parsing/packet-injector.sh
```diff
@@ -42,12 +53,14 @@ done
 
 echo "Closing native node"
 
-if [ $((TIMEDOUT + FAILED)) -gt 0 ]; then
+if [ -n "$DEBUG" ] || [ $((TIMEDOUT + FAILED)) -gt 0 ]; then
```
pjonsson, line 56: The build system already uses `DEBUG` for building binaries with debug info.

The build system is not as verbose as it used to be, just remove the redirections (but do that in a separate commit). This will give us accurate time stamps in the CI tests as a bonus. Users who don't want the output can either redirect it to /dev/null themselves, or build with `QUIET=1`.

tests/20-packet-parsing/packet-injector/project-conf.h
```diff
@@ -0,0 +1,6 @@
+#define LOG_CONF_LEVEL_IPV6                        LOG_LEVEL_DBG
+#define LOG_CONF_LEVEL_RPL                         LOG_LEVEL_DBG
+#define LOG_CONF_LEVEL_6LOWPAN                     LOG_LEVEL_DBG
+#define LOG_CONF_LEVEL_TCPIP                       LOG_LEVEL_DBG
+#define LOG_CONF_LEVEL_MAC                         LOG_LEVEL_DBG
+#define LOG_CONF_LEVEL_FRAMER                      LOG_LEVEL_DBG
```
pjonsson, line 6: Seems to be a missing newline at the end of this file.

If this code is built on SimpleLink, add a workaround for `LOG_CONF_LEVEL_FRAMER` similar to commit afac9c18a8361 so the CI logs aren't filled with warnings.

tests/20-packet-parsing/packet-injector.sh
```diff
@@ -21,8 +21,19 @@ make -C $CODE_DIR TARGET=native > make.log 2> make.err
 
 for i in $PACKET_DIR/*
 do
-  export TEST_FILE=$i
-  echo Injecting file $TEST_FILE
+  if [ -d "$i" ]; then
+    export TEST_FILES=0
+    for f in $i/*
+    do
+      export TEST_FILE_$(printf "%02d" $TEST_FILES)=$f
```
pjonsson, line 28: Can't the existing `TEST_FILE` be renamed to `TEST_FILES` and generalized into a list of files separated by some suitable character instead of this dynamic environment variable creation? (That should give 100% coverage of the affected parts of the C code from a single test.)

nvt: Yes, a single environment variable would be cleaner. Another option is to make it possible to pass files as command line arguments using the contiki_argc and contiki_argv variables.

tests/20-packet-parsing/packet-injector.sh
SWW13: I removed the DEBUG environment variable.

tests/20-packet-parsing/packet-injector/project-conf.h
SWW13: Newline and SimpleLink work-around added.

tests/20-packet-parsing/packet-injector.sh
SWW13: I switched to `contiki_argc`/`argv` for test files.

pjonsson:
I don't know why the packet-parsing test fails, but if you rebase this PR on top of latest develop, the readthedocs-test will pass.

SWW13:
I squashed all the intermediate progress commits. Is this enough or do you prefer a single commit?

nvt (approved):

nfi (approved):

PR #2198, opened 2022-10-12
TSCH: Missing #include in tsch-packet.c

kkrentz:
This PR inserts a missing `#include` into `tsch-packet.c` so as to fix compilation errors.


nfi (approved):

nvt:
For which platform do you get compilation errors regarding this file?

kkrentz:
Actually, the compilation error only came up after I fixed something else in `llsec802154.h`. Maybe, it will be better if I create a PR that includes all these fixes at once.

PR #2205, opened 2022-10-16
cooja: use linker for section info

pjonsson:
Define helper symbols to inform Cooja about
start addresses and sizes of segments.

### DIFF
--- a/arch/platform/cooja/Makefile.cooja
+++ b/arch/platform/cooja/Makefile.cooja
@@ -5,7 +5,7 @@
 ## The purpose of this file is to compile a shared library that
 ## can be loaded into the Java part of COOJA.
 
-EXPECTED_COOJA_VERSION = 2022052601
+EXPECTED_COOJA_VERSION = 2022071901
 
 ifndef CONTIKI
   $(error CONTIKI not defined!)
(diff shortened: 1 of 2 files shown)


pjonsson:
This requires the linked Cooja PR to be merged at the same time.

alexrayne:
what problem with adress and sizes? generated .map (and ns too) provide
this ino in  listing.


вс, 16 окт. 2022 г., 12:41 Peter A. Jonsson ***@***.***>:

pjonsson:
Beyond the difficulty to parse the map file, the map file does not contain size information about all symbols (even when built with `-fcommon`). This patch removes the need to parse the map file.

alexrayne:
map ddfinitely have info about all sections - org and sizes. as for symbols
- there is problem, that nm can solve, and it looks used on old cooja.
that luck of symbol info can be filled by building with --fdata-section.
That solution used in old cooja too.
i have work version of map-parser in cooja that well loads -map files,
buildt with --fdata-section.


вс, 16 окт. 2022 г., 16:37 Peter A. Jonsson ***@***.***>:

nvt (approved):

nfi (approved):

PR #2209, opened 2022-10-17
Remove unsupported nrf52 platform

alexstanoev:
The nRF52832 port is partially broken and relies on an experimental SoftDevice release that has been unsupported since 2015.

Both the `nrf` and `nrf52840` platforms are capable of running on the nRF52832 with minimal changes.
The radio drivers in both platforms support the IEEE 802.15.4 mode exclusively, which is not available on the nRF52832 SoC, so official platform support is not provided.


tools/docker/Dockerfile
```diff
@@ -86,14 +86,6 @@ RUN wget -nv http://simonduq.github.io/resources/mspgcc-4.7.2-compiled.tar.bz2 &
   cp -f -r /tmp/msp430/* /usr/local/ && \
   rm -rf /tmp/msp430 mspgcc*.tar.bz2
 
-## Install nRF52 SDK
```
pjonsson, line 89: Should the same lines be removed from `tools/vagrant/bootstrap.sh` too?

pjonsson:
Does `arch/cpu/nrf52840/lib/nrf52-sdk` belong to this platform as well, or is that a different NRF52-SDK? If it is different, it would be nice to have an update of that submodule (but that is obviously for a different PR).

alexstanoev:
No, `arch/cpu/nrf52840/lib/nrf52-sdk` is only used by the nrf52840 platform and is hosted under the contiki-ng organisation.  
It's a stripped down version of the nRF5 SDK, which is no longer getting new releases. Development now happens on the nrfx HAL, which is used by the nrf platform.

Similarly to contiki-ng/cc26xxware and contiki-ng/cc13xxware, the update policy we set for contiki-ng/nrf-sdk-contiki-ng is:
> New versions will only appear in this repository only if and when Contiki-NG's nRF52840 port needs to use them.

nvt (approved):

nfi (approved):

PR #2216, opened 2022-10-21
msp430: use dbg-io

pjonsson:
This adds support for using %zu in printf
on MSP430.


pjonsson:
There is no support for printing floats in dbg-io, drop this.

PR #2222, opened 2022-10-23
zoul: fix return type of grove_gyro_read_reg

pjonsson:
GROVE_GYRO_ERROR is defined as -1, so use a signed type that can represent the value.

### DIFF
--- a/arch/platform/zoul/dev/grove-gyro.c
+++ b/arch/platform/zoul/dev/grove-gyro.c
@@ -66,7 +66,7 @@ grove_gyro_values_t gyro_values;
 /*---------------------------------------------------------------------------*/
 void (*grove_gyro_int_callback)(uint8_t value);
 /*---------------------------------------------------------------------------*/
-static uint16_t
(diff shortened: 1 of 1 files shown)


arch/platform/zoul/dev/grove-gyro.c
```diff
@@ -66,7 +66,7 @@ grove_gyro_values_t gyro_values;
 /*---------------------------------------------------------------------------*/
 void (*grove_gyro_int_callback)(uint8_t value);
 /*---------------------------------------------------------------------------*/
-static uint16_t
+static int16_t
```
nfi, line 69: Good catch, but what do you think about using `int` instead as this seems to be the return type for the other functions in this file that returns `GROVE_GYRO_SUCCESS` or `GROVE_GYRO_ERROR`?

pjonsson: Fixed.

nfi (approved):

PR #2224, opened 2022-10-23
sixtop: use right field as parameter

pjonsson:


pjonsson:
The code is either passing an enum of the wrong type to the function, or passing the wrong field from the struct  to the function. I'm guessing it's a cut and paste error, so this fix makes the type match.

yatch (approved):

yatch:
Thanks! @pjonsson

PR #2231, opened 2022-10-25
memb: Confine return type of memb_numfree

kkrentz:
Presently, the function `memb_numfree` returns `int`, suggesting that an error might occur. However, this function cannot fail and, in the code base, nobody is checking for errors anyway. Therefore, this PR changes the return type to `size_t`.

### DIFF
--- a/os/net/mac/csma/csma-output.c
+++ b/os/net/mac/csma/csma-output.c
@@ -302,7 +302,7 @@ free_packet(struct neighbor_queue *n, struct packet_queue *p, int status)
     queuebuf_free(p->buf);
     memb_free(&metadata_memb, p->ptr);
     memb_free(&packet_memb, p);
-    LOG_DBG("free_queued_packet, queue length %d, free packets %d\n",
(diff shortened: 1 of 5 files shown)


os/net/mac/csma/csma-output.c
```diff
@@ -302,7 +302,7 @@ free_packet(struct neighbor_queue *n, struct packet_queue *p, int status)
     queuebuf_free(p->buf);
     memb_free(&metadata_memb, p->ptr);
     memb_free(&packet_memb, p);
-    LOG_DBG("free_queued_packet, queue length %d, free packets %d\n",
```
nvt, line 305: Did you mean to write "%zu" in the format here?

nvt (approved):

nvt:
Note that queuebuf uses this return value, and is still declared to return an int. So it could be good to change this as well.

```
int
queuebuf_numfree(void)
{
  return memb_numfree(&bufmem);
}
'''

nfi (approved):

PR #2234, opened 2022-10-25
Dockerfile: install git-lfs

pjonsson:
This is required for the Gecko SDK.


pjonsson:
@Yagoor I did `git grep lfs doc` and didn't see anything. I think it would be nice if you added something about git-lfs as a prerequisite somewhere in the documentation.

nfi (approved):

Yagoor:
@pjonsson thank you for the heads up. I already had it installed on my machine, so I forgot it is not a mandatory tool. I created #2240 with the needed information.

PR #2241, opened 2022-10-28
nrf: make linkaddr-arch.c build with TrustZone

pjonsson:
NRF_FICR is not available in non-secure mode when
building with TrustZone, so use a dummy address
for the non-secure TrustZone build for now.

This restores the behavior for non-secure TrustZone builds to how it was before commit 7d62acf0ee287.

### DIFF
--- a/arch/cpu/nrf/sys/linkaddr-arch.c
+++ b/arch/cpu/nrf/sys/linkaddr-arch.c
@@ -73,8 +73,16 @@ populate_link_address(void)
   device_address[0] = (NORDIC_SEMI_VENDOR_OUI) >> 16 & 0xFF;
   device_address[1] = (NORDIC_SEMI_VENDOR_OUI) >> 8 & 0xFF;
   device_address[2] = NORDIC_SEMI_VENDOR_OUI & 0xFF;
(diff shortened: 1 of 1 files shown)


pjonsson:
Failure cause by doxygen website DDoSed, doxygen ticket linked in #2243.

nfi (approved):

Yagoor:
@pjonsson is it possible to add a test build in the arm ports for this use case? It would avoid a regression in the future. :)

pjonsson:
This PR is basically a way to get things into a "working" state (I'm sure bad things will happen with a node ID of 0, but that is how the code was before your change) until we get the TrustZone support into the tree. That support will also contain build tests that will catch this problem.

Yagoor:
@pjonsson I see. Is there working being done to add TrustZone into Contiki-NG?

pjonsson:
@Yagoor the watchdog patch is unrelated, but the other two nrf patches align our internal TrustZone tree with upstream. The complete tree isn't suitable to bring in in its current state though, I know we have some ugly hacks on nrfx and the build system.

PR #2242, opened 2022-10-28
nrf: add nrfx_ipc.c into the build so we can access IPC funcs

pjonsson:


pjonsson:
Failure cause by doxygen website DDoSed, doxygen ticket linked in #2243.

nfi (approved):

PR #2243, opened 2022-10-28
nrf: init watchdog properly

pjonsson:
The first line in nrfx_wdt_init is NRFX_ASSERT(p_config) to ensure that a null config is not passed in.

Mirror the behavior of the nrf52840 port and pass
in a default config to the watchdog.


pjonsson:
Failure is caused by https://github.com/doxygen/doxygen/issues/9652

nfi (approved):

PR #2247, opened 2022-10-29
MPL: accept function

Issam-lakhlef:
@edrose 
When a packet is received by its source, the seed, it is always discarded. In this case, the seed will send the data packet three times, and receiving this packet will not cause "consistency". 

Sincerely.

### DIFF
--- a/os/net/ipv6/multicast/mpl.c
+++ b/os/net/ipv6/multicast/mpl.c
@@ -1415,11 +1415,6 @@ accept(uint8_t in)
   }
 #endif
 
-  /* if(uip_ds6_is_my_addr(&UIP_IP_BUF->srcipaddr) && in == MPL_DGRAM_IN) {
-    LOG_WARN("Received message from ourselves.\n");
-    return UIP_MCAST6_DROP;
-  } */
-
   /* Check the Next Header field: Must be HBHO */
   if(UIP_IP_BUF->proto != UIP_PROTO_HBHO) {
     LOG_ERR("Mcast I/O, bad proto\n");
--- a/os/net/ipv6/multicast/mpl.c
+++ b/os/net/ipv6/multicast/mpl.c
@@ -1415,10 +1415,10 @@ accept(uint8_t in)
   }
 #endif
 
-  if(uip_ds6_is_my_addr(&UIP_IP_BUF->srcipaddr) && in == MPL_DGRAM_IN) {
+  /* if(uip_ds6_is_my_addr(&UIP_IP_BUF->srcipaddr) && in == MPL_DGRAM_IN) {
     LOG_WARN("Received message from ourselves.\n");
     return UIP_MCAST6_DROP;
-  }
+  } */
 
   /* Check the Next Header field: Must be HBHO */
   if(UIP_IP_BUF->proto != UIP_PROTO_HBHO) {


atiselsts:
Not sure edrose is still active here, can you describe the issue and the proposed solution in more detail so the maintainers of the OS could understand it?

Issam-lakhlef:
First of all, thank you for your answer. To understand the problem, I will give an example. Let's assume in a network where the nodes implement MPL as a multicast protocol and let's assume we have one source and the others are sinks. The source "S" will prepare a multicast packet "M" and put in its header the address of the MPL domain as the destination address and its address as the source address, then it will associate a trickle timer to this message and store it in the buffer. In the end, the source will launch the trickle timer which will manage the transmission of "M".
After the reception of the message "M" by the neighbors, they will also store it in the buffer and associate a trickle to manage its transmission.
The problem is the following: since the source address of  "M" which is in its header is the address of "S"; at the reception of the copies of "M" by the multicast source S, S makes the call to the function **accept** which will initially check the source address and since is its own it will remove M which is not correct; it does not have to make this checking, since in this case, Trickle does not know that this message is received and will never trigger a coherence at the reception of "M".

That's why we have to remove this check in the **accept** function.

  ```
 if(uip_ds6_is_my_addr(&UIP_IP_BUF->srcipaddr) && in == MPL_DGRAM_IN) {
    LOG_WARN("Received message from ourselves.\n");
    return UIP_MCAST6_DROP;
  }
```

edrose:
I'm not particularly active here anymore, but I can chime in to help clear things up. I think I understand the issue you're describing, let me just paraphrase it to help others understand because it's taken me nearly an hour to understand what you're trying to fix.

## What is the issue?
The message expiration trickle timer on an MPL Seed is not managed correctly for messages in the buffered message set that originated from itself.

## What is the exppected behaviour?

When an MPL seed receives a re-transmission of one of it's own messages that it deems "consistent" (as per Section 9.2 of RFC7731), it should increment the consistency count for the associated trickle timer.

## What actually happens?

The message is dropped before it gets processed, so the trickle timer is never touched.

--- 

Is that correct @Issam-lakhlef ?

If that is the case, I'm not sure that the fix you propose is the correct one. Have you tested it with all MPL configurations? My understanding is that it should cause runaway transmissions... Although happy to be proven wrong on that.

I think this might be better opened as an issue to allow people to understand the problem and decide how/if a fix is to be implemented. By following that process, you'll be far more likely to get your merge request noticed and approved.

Issam-lakhlef:
Yes, that is exactly the point. I noticed this when I evaluated this MPL implementation using the cooja simulator and the iot-lab testbed. The SEED nodes never trigger the consistency of their messages and they send them 

> MPL_DATA_MESSAGE_TIMER_EXPIRATIONS

 times. 

The purpose of this verification is unclear to me:

> if(uip_ds6_is_my_addr(&UIP_IP_BUF->srcipaddr) && in == MPL_DGRAM_IN) {
>   LOG_WARN("Received message from ourselves.\n");
>   return UIP_MCAST6_DROP;
> }

By removing this part, I noticed that the behavior of Seeds has become normal, i.e., they trigger consistencies when they receive their re-transmitted data messages.

edrose:
> The purpose of this verification is unclear to me:

I believe that on some (or maybe all) platforms, a layer 2 broadcast frame will be received by the node that transmitted it. So there is no way to distinguish between a data message that is a re-transmission by a neighbour, and a data message that a node is receiving it's own transmission. I think that code was put in to ensure a node wasn't reacting to it's own messages and prematurely terminating transmissions.

> By removing this part, I noticed that the behavior of Seeds has become normal, i.e., they trigger consistencies when they receive their re-transmitted data messages.

Taking the above into account, have you confirmed that the Seed is actually reacting to it's neighbours, and not to itself?

If you simulate it with just a single seed and no neighbours, does it still react correctly? I.e: is the message transmitted `MPL_DATA_MESSAGE_TIMER_EXPIRATIONS` times, or does it stop early?

The code was put in there to solve a problem. I can't remember exactly what it was, but I expect removing it to have some kind of undesirable side-effect.

Issam-lakhlef:
I used the Z1 platform for simulations, I found that the Seed node reacts to the re-transmissions of its neighbors and not to its transmissions.

>  So there is no way to distinguish between a data message that is a re-transmission by a neighbour, and a data message that a node is receiving it's own transmission.
Why not do address checking at the MAC level, that is when a node receives a message, at the MAC level if the link-local address is its address, it removes the message, in this case, we avoid processing a message by the node that transmitted it.

edrose:
> Why not do address checking at the MAC level, that is when a node receives a message, at the MAC level if the source link-local address is its address, it removes the message, in this case, we avoid processing a message by the node that transmitted it.

Because that would make the network layer dependent on a specific MAC layer, breaking the layered networking model. It's not the right way forward and I don't even think that information is available within the contiki-ng multicast framework.

--

I'm wondering whether this points to a larger issue around how we deal with source addresses in MPL in general. However a merge request is not the place to be discussing this, so please open this as an issue so we can discuss the issue more generally.

simonduq:
Hi!

PR #2260, opened 2022-11-07
Ip64 bounds checks

Scepticz:
I added bounds checks to the DNS parsing of ip64.

I opted for leaving the current logic as it is and for inserting checks which are easy to understand and verify manually.

For two potential functional issues I added comments in a separate commit (marked with `TODO (functionality)`):
1. One update of the DNS type between `DNS_TYPE_AAAA` <-> `DNS_TYPE_A` seems to be a NO-OP right now. This may have originally been a copy-paste issue.
2. A missing copy of the size field of DNS names

I did not apply the functional changes as I do not know whether anyone relied on the respective behavior.

### DIFF
--- a/os/services/ip64/ip64-dns64.c
+++ b/os/services/ip64/ip64-dns64.c
@@ -195,6 +195,7 @@ ip64_dns64_4to6(const uint8_t *ipv4data, int ipv4datalen,
       LOG_WARN("ip64_dns64_6to4: packet ended while parsing\n");
       return ipv6datalen;
     }
+    /* TODO (functionality): Should replace q[DNS_QUESTION_TYPE1] == DNS_TYPE_A ? */
     if(q[DNS_QUESTION_CLASS0] == 0 && q[DNS_QUESTION_CLASS1] == DNS_CLASS_IN &&
        q[DNS_QUESTION_TYPE0] == 0 && q[DNS_QUESTION_TYPE1] == DNS_TYPE_AAAA) {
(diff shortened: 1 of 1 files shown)


os/services/ip64/ip64-dns64.c
```diff
@@ -244,18 +306,43 @@ ip64_dns64_4to6(const uint8_t *ipv4data, int ipv4datalen,
         ipv6datalen += 12;
 
       } else {
+        if(adata + len > ipv4data + ipv4datalen ) {
```
pjonsson, line 311: Can't this check be done before the if-statement to reduce the code duplication?

All conditions in the if-statements seem to have a trailing whitespace, which none of the existing code seems to have.

Scepticz: Thank you for the comments. Yes, the check can be done before. I had it in this de-duplicated version first, but then duplicated the code to move the check closer to the use of the pointer. However, I will change this back to de-duplicating the check.

I will also remove the trailing whitespaces

Scepticz: I applied the changes

Scepticz:
Is there an update on this - does someone have this in mind? Applying fixes here would be great. :-)

Scepticz:
I updated the commits for them to apply after https://github.com/contiki-ng/contiki-ng/commit/2b8ce65df399838d66b2ccc8cbfe01f27b1fa8b3

simonduq:
Hi!

PR #2281, opened 2022-11-17
Replace int with bool in comparisons of link layer addresses

kkrentz:
Small improvement to use the bool datatype in comparisons of link layer addresses.

### DIFF
--- a/os/net/packetbuf.h
+++ b/os/net/packetbuf.h
@@ -265,10 +265,10 @@ int               packetbuf_set_addr(uint8_t type, const linkaddr_t *addr);
 const linkaddr_t *packetbuf_addr(uint8_t type);
 
 /**
- * \brief      Checks whether the current packet is a broadcast.
- * \retval 0   iff current packet is not a broadcast
+ * \brief       Checks whether the current packet is a broadcast.
+ * \retval true iff the current packet is a broadcast
  */
-int               packetbuf_holds_broadcast(void);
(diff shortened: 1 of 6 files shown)


os/net/packetbuf.h
```diff
@@ -268,7 +268,7 @@ const linkaddr_t *packetbuf_addr(uint8_t type);
  * \brief      Checks whether the current packet is a broadcast.
  * \retval 0   iff current packet is not a broadcast
```
pjonsson, line 269: Documentation should also be updated if the return types are changed.

kkrentz: good proposal

os/net/packetbuf.h
```diff
@@ -265,10 +265,10 @@ int               packetbuf_set_addr(uint8_t type, const linkaddr_t *addr);
 const linkaddr_t *packetbuf_addr(uint8_t type);
 
 /**
- * \brief      Checks whether the current packet is a broadcast.
- * \retval 0   iff current packet is not a broadcast
+ * \brief       Checks whether the current packet is a broadcast.
+ * \retval true iff the current packet is not a broadcast
```
nvt, line 269: Now the comment says that the function returns <code>true</code> "iff the current packet is _not_ a broadcast".

kkrentz: Fixed comment

nfi (approved):

PR #1399, opened 2020-09-30
propose radio API application handlers, and TSCH redesign on radio callbacks

alexrayne:
Here proposes solution for issue [#447](https://github.com/contiki-ng/contiki-ng/issues/447) - Busy wait-free TSCH, and related issue [#220](https://github.com/contiki-ng/contiki-ng/issues/220)

this is done via extending radio driver API with assing application handler (i.e. callback) for next/current radio-operation complete.  ( see radio.h:RADIO_ARM_HANDLE_RX/TX ). That handler executes by driver thread.

TSCH stack have TSCH_CONF_RADIO_APPHANDLES - to enable use this new API. And if radio-driver suports setup RADIO_ARM_HANDLE_xx objects, this will use it to wait packet transmit and receive completion. and so it release CPU from busy-wait in RTC ISR to application code.

this PR provides app handles extention for cc13x0 prop RF, and cooja radio. 

By default cc13x0 target use new extention, and TSCH use it for busy-free. Cooja by default not use it in TSCH - since old busy-wait behaviour is a bit faster for it, and to compatibility with old code issues. To simulate busy-free mode, cooja project should enable it explicitly.

* cc13x0 target could disable RF app-handling extention by RF_CORE_CONF_APP_HANDLING,  for old-code compatibily or code-size effort.  It is enabled by default.

* RADIO_ARM_HANDLE_xx objects is active for single operation - once activated, it is clears. RT/TX operations are mutualy exclusive - so transmit clears rx handler, and on/off power clears both. This is done for hope - it can held made code more robust. TODO - maybe permanent handlers are more robust and less code complex?

this PR tested on cooja at TSCH-simple exeample, and on cc1352 target.

this provided in hope, that somebody finish it, and infer to master.

TODO: make handle for power on() operation.

Issue #447, opened 2018-04-22
Busy wait-free TSCH


simonduq:
Currently TSCH has a number of busy waits, resulting in ISR that run up to a few milliseconds. This is obviously undesirable. Places that need fixing:
* [ ] CCA check: not enabled by default, but if enabled, the CCA check is done as a busy wait
* [ ] Tx and ACK Tx: until we have async RF operations (https://github.com/contiki-ng/contiki-ng/issues/220), `transmit` busy waits for the whole duration of the Tx
* [ ] Rx and ACK Rx: busy wait until packet detected, followed another one until end of reception

atiselsts:
Some thoughts: this is not difficult to do, if the goal is to reduce the time spent in ISR. If the goal is to save energy - I think that could be difficult, at least on some platforms. I'll talk about CC2650, because I have used it.
1) Sleep is not very energy efficient compared to the active mode; deep sleep is, but only works for longer durations. Numbers: Rx + CPU active current on CC2650 is ~7.2 mA; Rx + sleep mode current is ~5.8 mA. Meaning that getting the system out of Rx mode just 1 tick faster saves as much energy as spending 5 ticks in sleep mode instead of active mode.
2) I think that removing some busywaits may mean having to spent more time in Rx mode. Take the packet reception example - AFAIK on CC2650 there is no interrupt that could signal that the radio has finished reception. So the alternative to constant polling is to sleep until the max packet length, which is actually much more costly for short packets. 
3) Getting out of the ISR additional times increases the probability that some other interrupt (e.g. sensor reading) has to be serviced during the TSCH timeslot, which is not nice for the protocol and again can delay the time when Rx mode is turned off.
4) Scheduling the sleep itself requires some small CPU work, and the wakeup can only happen on rtimer tick boundary - so the system may lose some energy on both ends of the sleep.

simonduq:
Yes the first goal here is to reduce time spend in ISR, to remove side effects on other system operations such as serial line or application stuff. Let's try to get this as energy-efficient as possible, and if needed, preserve the busy-wait option through a compile flag.

If you take a closer look, CCA check is straightforward (it's a fixed duration), Tx also comes with no energy trade-off. The discussion really is only about Rx.

[...]

PR #2282, opened 2022-11-18
Asynchronous Radio API

kkrentz:
As noted in #1399, #447, and #220, it would be nice to have a busy wait-free radio API. I actually used this one here in my implementations of IEEE 802.15.4 CSL and ContikiMAC. It enables the MAC layer to react to these events:
- SHR: synchronization header received or transmitted
- FIFOP: a configurable number of bytes was received (useful for frame filtering in software, as well as for getting informed of a complete frame reception)
- TXDONE: a complete frame was transmitted

The rtimer commits are unrelated and only included to avoid conflicts.


arch/platform/cooja/dev/cooja-radio.c
```diff
@@ -78,19 +79,32 @@ enum {
 };
 
 /* COOJA */
-char simReceiving = 0;
-char simInDataBuffer[COOJA_RADIO_BUFSIZE];
-int simInSize = 0;
+int simReceiving = 0;
```
pjonsson, line 82: Changing these types requires updates in the Cooja module as well, so the build system version must be bumped. Does the new code require these types to change, or is it a cleanup?

os/sys/rtimer.h
```diff
@@ -163,6 +164,16 @@ enum {
...
 
+/**
+ * \brief      Schedules a real-time task for a given instant
+ * \param task metadata about the real-time task
+ *
+ *             This variant does not auto-delay real-time tasks that
+ *             lie less than RTIMER_GUARD_TIME ticks in the future,
+ *             but returns an error code in such cases.
+ */
+int rtimer_set_precise(struct rtimer *task);
```
pjonsson, line 175: I only had a quick look, can the pointer parameters be const?

arch/platform/cooja/dev/cooja-radio.c
kkrentz: This is probably why the builds fail. I adapted Cooja to issue those interrupts (https://github.com/kkrentz/cooja/).

os/sys/rtimer.h
kkrentz: Thank you for pointing this out. You can, however, ignore the rtimer commits. I will separate them if there is interest in finalizing this PR.

kkrentz: I have removed the rtimer commits from this PR.

PR #2284, opened 2022-11-18
Cooja: Getters for channel information

kkrentz:
This PR adds some missing getters to the `get_value` function of the Cooja radio driver.

### DIFF
--- a/arch/platform/cooja/dev/cooja-radio.c
+++ b/arch/platform/cooja/dev/cooja-radio.c
@@ -54,6 +54,8 @@
 #define COOJA_RADIO_BUFSIZE 125
 #endif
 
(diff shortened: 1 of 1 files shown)


arch/platform/cooja/dev/cooja-radio.c
```diff
@@ -54,6 +54,8 @@
 #define COOJA_RADIO_BUFSIZE 125
 #endif
 
+#define MIN_CHANNEL 15
+#define MAX_CHANNEL 26
```
nfi, line 58: Instead of starting on channel 15, what do you think about using channels 11 - 26 to match IEEE 802.15.4 2.4GHz radio?

nfi (approved):

PR #2294, opened 2022-11-23
6LoWPAN: Assorted Refactorings

kkrentz:
This PR does some improvements to `sicslowpan.c`.

### DIFF
--- a/os/net/ipv6/sicslowpan.c
+++ b/os/net/ipv6/sicslowpan.c
@@ -1547,11 +1547,6 @@ packet_sent(void *ptr, int status, int transmissions)
 static void
 send_packet(void)
 {
-#if NETSTACK_CONF_BRIDGE_MODE
-  /* This needs to be explicitly set here for bridge mode to work */
-  packetbuf_set_addr(PACKETBUF_ADDR_SENDER,(void*)&uip_lladdr);
-#endif
-
   /* Provide a callback function to receive the result of
      a packet transmission. */
   NETSTACK_MAC.send(&packet_sent, NULL);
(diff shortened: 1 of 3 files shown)


os/net/ipv6/sicslowpan.c
```diff
@@ -689,6 +686,8 @@ compress_hdr_iphc(void)
   uint8_t tmp, iphc0, iphc1, *next_hdr, *next_nhc;
   int ext_hdr_len;
   struct uip_udp_hdr *udp_buf;
+  struct sicslowpan_addr_context *source_context;
+  struct sicslowpan_addr_context *destination_context;
```
pjonsson, line 688: Every platform supports C99 now, so declare variables closer to their initialization.

os/net/ipv6/sicslowpan.c
```diff
@@ -893,15 +892,16 @@ compress_hdr_iphc(linkaddr_t *link_destaddr)
       /* compession compare with link adress (destination) */
 
       iphc1 |= compress_addr_64(SICSLOWPAN_IPHC_DAM_BIT,
-                                &UIP_IP_BUF->destipaddr,
-                                (uip_lladdr_t *)link_destaddr);
+          &UIP_IP_BUF->destipaddr,
+          (const uip_lladdr_t *)packetbuf_addr(PACKETBUF_ADDR_RECEIVER));
```
pjonsson, line 891: Shouldn't need to cast to const in these places.

pjonsson:
Unless there are dependencies between these commits:

I'm guessing the NETSTACK_CONF_BRIDGE_MODE commit, along with the last two commits are trivial to review (the parentheses for macros are obviously an improvement, the two other commits are also correct so it's a question of whether they should be merged or not). Putting them in a separate PR should get them merged quicker.

The uncomp_hdr_len initialization should be fairly easy to review, so might be worth putting that in a separate PR too.

os/net/ipv6/sicslowpan.c
kkrentz: Was necessary because uip uses a different type ...

PR #2295, opened 2022-11-24
6LoWPAN: Put parentheses around macro parameter

kkrentz:


os/net/ipv6/sicslowpan.c
```diff
@@ -83,8 +83,8 @@
 
 #define GET16(ptr,index) (((uint16_t)((ptr)[index] << 8)) | ((ptr)[(index) + 1]))
```
nvt, line 84: It would be good to have brackets around the index parameter in the GET16 macro as well.

nvt: The commit message should say "parentheses" rather than "brackets".

kkrentz: Seems unnecessary to put further brackets around the index parameter.

nfi (approved):

os/net/ipv6/sicslowpan.c
nvt: There are no parentheses around the first occurrence of the index parameter in the GET16 macro definition. Just for consistency (e.g., with line 86), it can be added there too, even if it is not strictly necessary.

PR #2298, opened 2022-11-24
6LoWPAN: Cut unnecessary copying of the destination MAC address

kkrentz:


os/net/ipv6/sicslowpan.c
```diff
@@ -1639,25 +1630,16 @@ output(const linkaddr_t *localdest)
...
 
   /* copy over the retransmission count from uipbuf attributes */
   packetbuf_set_attr(PACKETBUF_ATTR_MAX_MAC_TRANSMISSIONS,
                      uipbuf_get_attr(UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS));
 
-/* Calculate NETSTACK_FRAMER's header length, that will be added in the NETSTACK_MAC */
-  packetbuf_set_addr(PACKETBUF_ADDR_RECEIVER, &dest);
+  /* Copy destination address to packetbuf */
+  packetbuf_set_addr(PACKETBUF_ADDR_RECEIVER,
+      localdest ? localdest : &linkaddr_null);
```
arurke, line 1641: Should we be more defensive and rather use `localdest == NULL`? Avoiding any future- or portability issues.

arurke (approved):

os/net/ipv6/sicslowpan.c
pjonsson: I think that is a style issue, the definition of `NULL` is part of the C standard so a compiler breaking that will not work for a lot of code out there.

arurke: Thx! I did not realize it was in the standard.

kkrentz: I could actually use the elvis operator here (`localdest ?: &linkaddr_null`)

arurke: Had to google that one :) I don't have any strong opinion either way.

os/net/ipv6/sicslowpan.c
```diff
@@ -1632,25 +1623,15 @@ output(const linkaddr_t *localdest)
...
   LOG_INFO("output: sending IPv6 packet with len %d\n", uip_len);
 
   /* copy over the retransmission count from uipbuf attributes */
   packetbuf_set_attr(PACKETBUF_ATTR_MAX_MAC_TRANSMISSIONS,
                      uipbuf_get_attr(UIPBUF_ATTR_MAX_MAC_TRANSMISSIONS));
 
-/* Calculate NETSTACK_FRAMER's header length, that will be added in the NETSTACK_MAC */
-  packetbuf_set_addr(PACKETBUF_ADDR_RECEIVER, &dest);
+  /* Copy destination address to packetbuf */
+  packetbuf_set_addr(PACKETBUF_ADDR_RECEIVER, localdest ?: &linkaddr_null);
```
nvt, line 1633: This ternary operator shortcut is not part of standard C, as far as I know. It would be preferable to avoid compiler-specific extensions in the core modules.

PR #2304, opened 2022-11-28
cooja: allow building outside Cooja

pjonsson:
Removing JNI removed many failure cases from
the build system, so it is now safe to build
binaries targeting Cooja from the shell.
The binaries built from the shell are named
in the same way that they are for other targets.
Cooja will not use these libraries though,
and instead generate a random name that has
not previously been used.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -556,6 +556,12 @@ ifndef LD
   LD = $(CC)
 endif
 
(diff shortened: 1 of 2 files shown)


simonduq:
Hi!

nfi (approved):

PR #2305, opened 2022-11-29
mac-sequence: Set destination sequence number (DSN) in mac-sequence.c

kkrentz:
The motivation behind this PR is to eliminate duplicate code and to avoid special handling for DSN=0.

### DIFF
--- a/os/net/mac/mac-sequence.h
+++ b/os/net/mac/mac-sequence.h
@@ -46,6 +46,16 @@
 #define MAC_SEQUENCE_H
 
 /**
+ * brief       Initializes the destination sequence number to a random value.
+ */
+void mac_sequence_init(void);
+
+/**
+ * \brief      Sets and increments the destination sequence number.
+ */
+void mac_sequence_set_dsn(void);
+
+/**
  * \brief      Tell whether the packetbuf is a duplicate packet
  * \return     Non-zero if the packetbuf is a duplicate packet, zero otherwise
  *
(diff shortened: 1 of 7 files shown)


arurke:
@kkrentz In framer-802154.c there was efforts to ensure the sequence number was not zero - which this PR removes. I can't really tell if it was just to check if a new sequence number was needed, or if there is actually some other parts of the mac or lower layers that rely on this behavior. Any thoughts?

kkrentz:
@arurke That code in `framer-802154.c` was necessary because there was no clear responsibility for setting the destination sequence number (DSN). If DSN=0, `framer-802154.c` took over. This PR makes the MAC driver responsible for setting the DSN, obviating the special handling for DSN=0. Also, in line with IEEE 802.15.4, outgoing frames no longer skip over DSN=0.

os/net/mac/mac-sequence.h
```diff
@@ -45,6 +45,16 @@
 #ifndef MAC_SEQUENCE_H
 #define MAC_SEQUENCE_H
 
+/**
+ * brief       Initializes.
```
arurke, line 49: * `\brief`
* How about "Initializes the MAC sequence number to a random value"?  Making it clear that calling it twice actually influences the value.

arurke (changes requested):

arurke (approved):

arurke:
Thanks for the explanation and fixes @kkrentz

nfi (approved):

PR #2306, opened 2022-11-29
mac-sequence: Fix vulnerability

kkrentz:
Currently, the `mac-sequence` module does not ensure to keep the most recent destination sequence number received from a neighbor. Instead, it has a sort of ring buffer, which basically overwrites the oldest destination sequence number. This is problematic for TSCH's replay protection. Assume an acknowledgment frame from a node A to node B is missed and A receives a lot of frames from other neighbors than B. Then, A's ring buffer gets filled with other sequence numbers and by the time B retransmits, B's sequence number may no longer reside in A's ring buffer. This PR fixes this issue by keeping track of the last destination sequence number per neighbor.

### DIFF
--- a/os/net/mac/mac-sequence.c
+++ b/os/net/mac/mac-sequence.c
@@ -45,6 +45,7 @@
 #include <string.h>
 
 #include "contiki-net.h"
+#include "lib/random.h"
 #include "net/mac/mac-sequence.h"
 #include "net/packetbuf.h"
 
@@ -67,6 +68,20 @@ struct seqno {
 #endif /* NETSTACK_CONF_MAC_SEQNO_HISTORY */
 static struct seqno received_seqnos[MAX_SEQNOS];
 
+static uint8_t mac_dsn;
+
+/*---------------------------------------------------------------------------*/
+void
+mac_sequence_init(void)
+{
+  mac_dsn = random_rand();
+}
+/*---------------------------------------------------------------------------*/
+void
+mac_sequence_set_dsn(void)
+{
+  packetbuf_set_attr(PACKETBUF_ATTR_MAC_SEQNO, mac_dsn++);
+}
 /*---------------------------------------------------------------------------*/
 int
 mac_sequence_is_duplicate(void)
--- a/os/net/mac/mac-sequence.c
+++ b/os/net/mac/mac-sequence.c
@@ -55,6 +55,8 @@ struct seqno {
   uint8_t seqno;
 };
 
+static void mac_sequence_register_seqno(void);
+
 #ifdef NETSTACK_CONF_MAC_SEQNO_MAX_AGE
 #define SEQNO_MAX_AGE NETSTACK_CONF_MAC_SEQNO_MAX_AGE
 #else /* NETSTACK_CONF_MAC_SEQNO_MAX_AGE */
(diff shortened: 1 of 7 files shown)


os/net/mac/mac-sequence.h
```diff
@@ -45,22 +45,31 @@
 #ifndef MAC_SEQUENCE_H
 #define MAC_SEQUENCE_H
 
+enum mac_sequence_result {
+  MAC_SEQUENCE_NO_DUPLICATE = 0,
+  MAC_SEQUENCE_DUPLICATE,
+  MAC_SEQUENCE_ERROR
```
pjonsson, line 51: I think C99 allows a trailing comma after the last enum-member (allows extending enums without affecting git blame).

os/net/mac/mac-sequence.c
```diff
@@ -45,79 +45,89 @@
...
+mac_sequence_set_dsn(void)
 {
-  int i, j;
+  packetbuf_set_attr(PACKETBUF_ATTR_MAC_SEQNO, mac_dsn++);
+}
+/*---------------------------------------------------------------------------*/
+enum mac_sequence_result
+mac_sequence_is_duplicate(void)
+{
+  uint16_t now = clock_seconds();
```
pjonsson, line 92: Isn't `clock_seconds()` an unsigned long?

os/net/mac/mac-sequence.c
```diff
@@ -45,79 +45,89 @@
...
-#ifdef NETSTACK_CONF_MAC_SEQNO_HISTORY
-#define MAX_SEQNOS NETSTACK_CONF_MAC_SEQNO_HISTORY
-#else /* NETSTACK_CONF_MAC_SEQNO_HISTORY */
-#define MAX_SEQNOS 16
-#endif /* NETSTACK_CONF_MAC_SEQNO_HISTORY */
-static struct seqno received_seqnos[MAX_SEQNOS];
+#define BEFORE(a, b) ((int16_t)((a) - (b)) < 0)
+
+struct deduplication_record {
+  uint8_t last_broadcast_seqno;
```
pjonsson, line 66: I think this layout pretty much guarantees padding, starting with the largest members and finishing with the smallest should pad less.

I don't know if the decreased readability is worth it though, some maintainer should chime in on this.

kkrentz: Things will become more simple if we agree on dropping duplicate detection for broadcast frames.

kkrentz: I adjusted the order in the struct and added some #ifdef's for eliminating the broadcast deduplication. Am not sure if we need that and if it should be on or off by default.

os/net/mac/mac-sequence.c
```diff
@@ -45,79 +45,115 @@
...
+mac_sequence_set_dsn(void)
 {
-  int i, j;
+  packetbuf_set_attr(PACKETBUF_ATTR_MAC_SEQNO, mac_dsn++);
+}
+/*---------------------------------------------------------------------------*/
+enum mac_sequence_result
+mac_sequence_is_duplicate(void)
+{
+  uint16_t now;
```
pjonsson, line 102: Everything is built as C99, put the variable declarations at their initialization.

PR #2310, opened 2022-11-30
Allow overriding NETSTACK_CONF_RADIO for simplelink platforms

arurke:
As in title. Mimics cc26x0-cc13x0.


atiselsts (approved):

nvt:
Is it intended that the NETSTACK_CONF_RADIO setting for CC26xx IEEE-mode should not have the same flexibility to override (line 228)?

arurke:
@nvt good catch! I did not notice it had a separate section for cc26xx. Fixed and rebased.

PR #2318, opened 2022-12-07
Update for Cooja long parameter names

pjonsson:
Cooja uses the parameter --[no-]gui to decide
whether to use headless mode. This makes it possible to pass simulation files directly to Cooja without passing -nogui/-quickstart.

Parameter names also use double dashes.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -452,7 +452,7 @@ $(GENDIR)/%: $(GENDIR)/%.diffupdate | $(GENDIR)
 
 # Start cooja through "make file.csc", where the real target is defined through the csc.
 %.csc %.csc.gz: FORCE
-	$(Q)$(GRADLE) --no-watch-fs --parallel --build-cache -p $(COOJA_PATH) run --args="-quickstart=$(addprefix $(CURDIR)/,$@) -contiki=$(realpath $(CONTIKI)) -logdir=$(CURDIR)"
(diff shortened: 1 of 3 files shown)


Makefile.include
```diff
@@ -452,7 +452,7 @@ $(GENDIR)/%: $(GENDIR)/%.diffupdate | $(GENDIR)
 
 # Start cooja through "make file.csc", where the real target is defined through the csc.
 %.csc %.csc.gz: FORCE
-	$(Q)$(GRADLE) --no-watch-fs --parallel --build-cache -p $(COOJA_PATH) run --args="-quickstart=$(addprefix $(CURDIR)/,$@) -contiki=$(realpath $(CONTIKI)) -logdir=$(CURDIR)"
+	$(Q)$(GRADLE) --no-watch-fs --parallel --build-cache -p $(COOJA_PATH) run --args="--contiki=$(realpath $(CONTIKI)) --logdir=$(CURDIR) $(addprefix $(CURDIR)/,$@)"
```
nfi, line 455: Since we moved to long parameter names - what do you think about skipping the equal sign entirely?

```shell
$(Q)$(GRADLE) --no-watch-fs --parallel --build-cache -p $(COOJA_PATH) run --args="--contiki $(realpath $(CONTIKI)) --logdir $(CURDIR) $(addprefix $(CURDIR)/,$@)"
```

pjonsson: The equals sign is how the parameters are described in `--help`, so I think there's a value in Contiki-NG following that.

nfi (approved):

PR #2321, opened 2022-12-07
Correct const placement in remaining lists

arurke:
In #2012, @pjonsson describes and fixes an issue with the `const` placement in the list library. This PR fixes the same issue for the remaining lists: circular-list, dbl-circ-list, and dbl-list.


atiselsts:
Makes sense to apply the same change, for consistency!

atiselsts (approved):

PR #2332, opened 2022-12-11
Common random-arch PRNG for simpelink and cc26xx-cc13xx

arurke:
Currently the cc26xx-cc13xx platform uses the hardware True Random Number Generator (TRNG) in its random-arch implementation, while simplelink uses a pseudo-RNG (PRNG) with the TRNG only for the seed. The random.h API states `random_rand()` should return a pseudo-random number. Thus cc26xx-cc13xx is not giving the expected balance of entropy and resources spent (energy, time). This PR makes the two platforms consistent:
1. Takes the PRNG algorithm from simplelink and moves it into a new smallprng-library
2. Use the smallprng-library in random-arch for both simplelink and cc26xx-cc13xx
3. Seed random-arch in cc26xx-cc13xx using the TRNG

Tested using a launchpad/cc1310 for both platforms.

### DIFF
--- a/arch/cpu/simplelink-cc13xx-cc26xx/dev/random.c
+++ b/arch/cpu/simplelink-cc13xx-cc26xx/dev/random.c
@@ -27,15 +27,12 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
  * OF THE POSSIBILITY OF SUCH DAMAGE.
  */
- /**
+/**
  * \addtogroup cc13xx-cc26xx-cpu
  * @{
  *
  * \defgroup cc13xx-cc26xx-prng Pseudo Random Number Generator (PRNG) for CC13xx/CC26xx.
  * @{
- *
- * Implementation based on Bob Jenkins' small noncryptographic PRNG.
- * - http://burtleburtle.net/bob/rand/smallprng.html
  *
  * This file overrides os/lib/random.c. Note that the file name must
  * match the original file for the override to work.
@@ -44,22 +41,13 @@
  *        Implementation of Pseudo Random Number Generator for CC13xx/CC26xx.
  * \author
  *        Edvard Pettersen <e.pettersen@ti.com>
+ *        Andreas Urke
  */
 /*---------------------------------------------------------------------------*/
-#include <contiki.h>
+#include "contiki.h"
+#include "lib/smallprng.h"
 /*---------------------------------------------------------------------------*/
-#include <stdint.h>
-/*---------------------------------------------------------------------------*/
-typedef struct {
-  uint32_t a;
-  uint32_t b;
-  uint32_t c;
-  uint32_t d;
-} ranctx_t;
-
-static ranctx_t ranctx;
-/*---------------------------------------------------------------------------*/
-#define rot32(x, k) (((x) << (k)) | ((x) >> (32 - (k))))
+static smallprng_t smallprng;
 /*---------------------------------------------------------------------------*/
 /**
  * \brief   Generates a new random number using the PRNG.
@@ -68,15 +56,7 @@ static ranctx_t ranctx;
 unsigned short
 random_rand(void)
 {
-  uint32_t e;
-
-  e        = ranctx.a - rot32(ranctx.b, 27);
-  ranctx.a = ranctx.b ^ rot32(ranctx.c, 17);
-  ranctx.b = ranctx.c + ranctx.d;
-  ranctx.c = ranctx.d + e;
-  ranctx.d = e        + ranctx.a;
-
-  return (unsigned short)ranctx.d;
+  return (unsigned short)smallprng_rand(&smallprng);
 }
 /*---------------------------------------------------------------------------*/
 /**
@@ -86,13 +66,7 @@ random_rand(void)
 void
 random_init(unsigned short seed)
 {
-  uint32_t i;
-
-  ranctx.a = 0xf1ea5eed;
-  ranctx.b = ranctx.c = ranctx.d = (uint32_t)seed;
-  for(i = 0; i < 20; ++i) {
-    (void)random_rand();
-  }
+  smallprng_init(&smallprng, seed);
 }
 /*---------------------------------------------------------------------------*/
 /**
--- a/arch/platform/cc26x0-cc13x0/platform.c
+++ b/arch/platform/cc26x0-cc13x0/platform.c
@@ -57,6 +57,7 @@
 #include "dev/cc26xx-uart.h"
 #include "dev/soc-rtc.h"
 #include "dev/serial-line.h"
+#include "dev/soc-trng.h"
 #include "rf-core/rf-core.h"
 #include "sys_ctrl.h"
 #include "uart.h"
@@ -172,7 +173,10 @@ platform_init_stage_one()
 void
 platform_init_stage_two()
 {
-  random_init(0x1234);
+  soc_trng_init();
+
+  /* Use TRNG to seed random lib. */
+  random_init((uint16_t)soc_trng_rand_synchronous());
 
   /* Character I/O Initialisation */
 #if TI_UART_CONF_ENABLE
--- a/os/lib/smallprng.c
+++ b/os/lib/smallprng.c
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2022, Andreas Urke
+ * Copyright (c) 2018, Texas Instruments Incorporated - http://www.ti.com/
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ */
+
+/** \addtogroup smallprng
+ * @{
+ */
+
+/**
+ * \file
+ *        Implementation of small non-cryptographic random number generator
+ *        based on Edvard Pettersens random.c for simplelink platform,
+ *        which was based on Bob Jenkins' small non-cryptographic PRNG:
+ *        http://burtleburtle.net/bob/rand/smallprng.html
+ *
+ * \author
+ *        Andreas Urke
+ */
+
+#include "lib/smallprng.h"
+/*---------------------------------------------------------------------------*/
+#define rot32(x, k) (((x) << (k)) | ((x) >> (32 - (k))))
+/*---------------------------------------------------------------------------*/
+uint32_t
+smallprng_rand(smallprng_t *smallprng)
+{
+  uint32_t e = smallprng->a - rot32(smallprng->b, 27);
+  smallprng->a = smallprng->b ^ rot32(smallprng->c, 17);
+  smallprng->b = smallprng->c + smallprng->d;
+  smallprng->c = smallprng->d + e;
+  smallprng->d = e + smallprng->a;
+  return smallprng->d;
+}
+/*---------------------------------------------------------------------------*/
+void
+smallprng_init(smallprng_t *smallprng, uint32_t seed)
+{
+  smallprng->a = 0xf1ea5eed;
+  smallprng->b = smallprng->c = smallprng->d = seed;
+  for(uint8_t i = 0; i < 20; ++i) {
+    (void)smallprng_rand(smallprng);
+  }
+}
+/*---------------------------------------------------------------------------*/
+/** @} */
--- a/tests/08-native-runs/14-smallprng.sh
+++ b/tests/08-native-runs/14-smallprng.sh
@@ -1,0 +1,3 @@
+#!/bin/sh -e
+
+./run-one.sh 14-smallprng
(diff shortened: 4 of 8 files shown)


arch/platform/cc26x0-cc13x0/platform.c
```diff
@@ -172,7 +173,11 @@ platform_init_stage_one()
 void
 platform_init_stage_two()
 {
-  random_init(0x1234);
+  soc_trng_init();
+
+  /* Use TRNG to seed random lib. */
+  uint16_t seed = (uint16_t)soc_trng_rand_synchronous();
```
pjonsson, line 179: Can this variable be inlined, or does that make the next line too long?

os/lib/smallprng.c
```diff
@@ -0,0 +1,72 @@
...
+ * \author
+ *        Andreas Urke
+ */
+
+#include "lib/smallprng.h"
+/*---------------------------------------------------------------------------*/
+#define rot32(x, k) (((x) << (k)) | ((x) >> (32 - (k))))
+/*---------------------------------------------------------------------------*/
+uint32_t
+smallprng_rand(smallprng_t* smallprng)
```
pjonsson, line 52: I think most other code in Contiki-NG puts the star on the parameter name instead of the type.

tests/08-native-runs/14-smallprng/Makefile
```diff
@@ -0,0 +1,9 @@
+CONTIKI_PROJECT = test-smallprng
+all: $(CONTIKI_PROJECT)
+
+TARGET = native
+
+MODULES += os/services/unit-test
+
+CONTIKI = ../../..
```
pjonsson, line 8: You don't need to define `CONTIKI` nowadays, the build system will figure that out itself. (So `include ../../../Makefile.include` saves a line and avoids a definition.)

tests/08-native-runs/14-smallprng/test-smallprng.c
```diff
@@ -0,0 +1,149 @@
...
+AUTOSTART_PROCESSES(&test_smallprng_process);
+/*---------------------------------------------------------------------------*/
+#define RAND_BUF_SIZE 128
+/*---------------------------------------------------------------------------*/
+UNIT_TEST_REGISTER(output_sanity, "Basic sanity of the output");
+UNIT_TEST(output_sanity)
+{
+  UNIT_TEST_BEGIN();
+
+  smallprng_t smallprng1 = {0};
```
pjonsson, line 55: This is immediately passed to `smallprng_init`, is the zero initialization required?

pjonsson:
How does this relate to #522?

arch/cpu/simplelink-cc13xx-cc26xx/dev/random.c
```diff
@@ -68,15 +56,7 @@ static ranctx_t ranctx;
...
-  uint32_t e;
-
-  e        = ranctx.a - rot32(ranctx.b, 27);
-  ranctx.a = ranctx.b ^ rot32(ranctx.c, 17);
-  ranctx.b = ranctx.c + ranctx.d;
-  ranctx.c = ranctx.d + e;
-  ranctx.d = e        + ranctx.a;
-
-  return (unsigned short)ranctx.d;
+  return (unsigned short) smallprng_rand(&smallprng);
```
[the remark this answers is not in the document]
arurke: 👍 🤦 I ran uncrustify on the files now and found a few other

arch/platform/cc26x0-cc13x0/platform.c
arurke: I was actually intentional doing it this way to make it clearer. But now with some daylight, and the comment above, I think it is more than clear enough when inlined 👍

os/lib/smallprng.c
arurke: :+1:

tests/08-native-runs/14-smallprng/Makefile
arurke: Neat!

tests/08-native-runs/14-smallprng/test-smallprng.c
arurke: I have the habit of doing this to avoid any future devs. using `smallprng1` in a different way without realizing it has not been initialized. And with the thinking that the compiler will avoid any performance penalty, or any penalty will be worth the reduced risk. But no strong opinion, happy to hear other thoughts.

arurke:
Thanks for the review @pjonsson! There is no overlap or conflict with https://github.com/contiki-ng/contiki-ng/pull/522 functionality-wise, but I did glance at it to ensure the APIs were somewhat consistent. I am not a security-person, so I would be happy to hear if you @kkrentz have any opinions on this PR.

os/lib/smallprng.c
```diff
@@ -0,0 +1,72 @@
...
+  smallprng->c = smallprng->d + e;
+  smallprng->d = e + smallprng->a;
+  return smallprng->d;
+}
+/*---------------------------------------------------------------------------*/
+void
+smallprng_init(smallprng_t *smallprng, uint32_t seed)
+{
+  smallprng->a = 0xf1ea5eed;
+  smallprng->b = smallprng->c = smallprng->d = (uint32_t)seed;
```
pjonsson, line 66: The cast of seed should not be required here.

tests/08-native-runs/14-smallprng.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
```
pjonsson, line 1: I think `#!/bin/sh -e` should be sufficient for this (bash  takes a long time to start compared to /bin/sh on most systems).

arurke: :+1:

os/lib/smallprng.c
arurke: :+1:

tests/08-native-runs/14-smallprng/test-smallprng.c
pjonsson: Sorry for the late response, but initializing the data will make the problem invisible for tools like Valgrind/Memory Sanitizer/etc, otherwise they would give warnings when data is used without being initialized.

Without entering the risk/reward analysis judgement, I want to mention that my experience is that only the most basic "sequential" double initializations will be reliably optimized away even by "modern" compilers (Clang/LLVM, ~2019). In this case, it seems unlikely the compiler will look into `smallprng_init` and deduce the zero-initialization isn't required.

arurke:
Rebased to pass CI

arch/cpu/simplelink-cc13xx-cc26xx/dev/random.c
```diff
@@ -86,13 +66,7 @@ random_rand(void)
...
 random_init(unsigned short seed)
 {
-  uint32_t i;
-
-  ranctx.a = 0xf1ea5eed;
-  ranctx.b = ranctx.c = ranctx.d = (uint32_t)seed;
-  for(i = 0; i < 20; ++i) {
-    (void)random_rand();
-  }
+  smallprng_init(&smallprng, seed);
```
kkrentz, line 69: Is it safe to initialize smallprng with a 16-bit seed?

arurke: Assuming you mean functionality-wise: Yes I believe so. Looking at the [origin of the PRNG algorithm](http://burtleburtle.net/bob/rand/smallprng.html) I see no assumptions regarding the seed, and indeed the author tested seeds sequentially starting from 0 (third paragraph). Random.h also makes no promise as to the strength besides it being pseudo-random.

kkrentz:
I would prefer your `lib/smallprng.c` to replace the default `lib/random.c` implementation. The current `lib/random.c` is based on stdlib, which can lead to problems anywy - see [here](https://dunkels.com/adam/rand-may-call-malloc/).

arch/cpu/simplelink-cc13xx-cc26xx/dev/random.c
kkrentz: There is also a 16-bit version of this algorithm on this [page](https://filterpaper.github.io/prng.html#jsf16). To keep the current interface and avoid using some limited range of seeds (which may perform worse in terms of average cycle lengths), we could switch to that version.

arurke: Thanks for the tip! I will check it out.

arurke:
> I would prefer your lib/smallprng.c to replace the default lib/random.c implementation. The current lib/random.c is based on stdlib, which can lead to problems anywy - see [here](https://dunkels.com/adam/rand-may-call-malloc/).

I might be inclined to agree, but I would prefer doing that in a separate PR

arch/cpu/simplelink-cc13xx-cc26xx/dev/random.c
arurke: @kkrentz One worry: My original implementation was based of Bob Jenkins algorithm which was "vetted" by Texas Instruments. I am not able to verify the 16-bit version by "M.E. O'Neill". I am getting the sense these subjects are in your field - I am happy to use it if you gave your OK.

kkrentz: After testing O'Neill's 16-bit version with [PractRand](https://pracrand.sourceforge.net/), I no longer recommend it:

```
./RNG_test jsf16 -multithreaded
RNG_test using PractRand version 0.94
RNG = jsf16, seed = 0xfbddf8ba
test set = core, folding = standard (16 bit)

rng=jsf16, seed=0xfbddf8ba
length= 512 megabytes (2^29 bytes), time= 2.0 seconds
  no anomalies in 183 test result(s)

rng=jsf16, seed=0xfbddf8ba
length= 1 gigabyte (2^30 bytes), time= 3.9 seconds
  no anomalies in 196 test result(s)

rng=jsf16, seed=0xfbddf8ba
length= 2 gigabytes (2^31 bytes), time= 7.2 seconds
  no anomalies in 209 test result(s)

rng=jsf16, seed=0xfbddf8ba
length= 4 gigabytes (2^32 bytes), time= 13.6 seconds
  no anomalies in 221 test result(s)

rng=jsf16, seed=0xfbddf8ba
length= 8 gigabytes (2^33 bytes), time= 26.7 seconds
  Test Name                         Raw       Processed     Evaluation
  [Low1/16]BCFN(2+0,13-1,T)         R=  -7.5  p =1-5.9e-4   unusual          
  ...and 232 test result(s) without anomalies

rng=jsf16, seed=0xfbddf8ba
length= 16 gigabytes (2^34 bytes), time= 52.9 seconds
  Test Name                         Raw       Processed     Evaluation
  mod3n(5):(1,9-0)                  R= +10.3  p =  8.6e-6   mildly suspicious
  ...and 243 test result(s) without anomalies

rng=jsf16, seed=0xfbddf8ba
length= 32 gigabytes (2^35 bytes), time= 105 seconds
  Test Name                         Raw       Processed     Evaluation
  mod3n(5):(0,9-0)                  R=  +9.7  p =  1.9e-5   mildly suspicious
  mod3n(5):(1,9-0)                  R= +20.7  p =  2.0e-11   VERY SUSPICIOUS 
  ...and 253 test result(s) without anomalies
```

Even the original 32-bit version suffers from some anomalies:

```
./RNG_test jsf32 -multithreaded
RNG_test using PractRand version 0.94
RNG = jsf32, seed = 0x35312dbd
test set = core, folding = standard (32 bit)

rng=jsf32, seed=0x35312dbd
length= 1 gigabyte (2^30 bytes), time= 2.6 seconds
  Test Name                         Raw       Processed     Evaluation
  [Low8/32]DC6-9x1Bytes-1           R=  -5.2  p =1-1.3e-3   unusual          
  ...and 191 test result(s) without anomalies

rng=jsf32, seed=0x35312dbd
length= 2 gigabytes (2^31 bytes), time= 5.1 seconds
  no anomalies in 204 test result(s)

rng=jsf32, seed=0x35312dbd
length= 4 gigabytes (2^32 bytes), time= 10.0 seconds
  Test Name                         Raw       Processed     Evaluation
  BCFN(2+0,13-0,T)                  R=  -7.9  p =1-4.2e-4   unusual          
  [Low1/32]DC6-9x1Bytes-1           R=  -6.4  p =1-2.0e-4   mildly suspicious
  ...and 214 test result(s) without anomalies

rng=jsf32, seed=0x35312dbd
length= 8 gigabytes (2^33 bytes), time= 20.1 seconds
  no anomalies in 229 test result(s)

rng=jsf32, seed=0x35312dbd
length= 16 gigabytes (2^34 bytes), time= 40.0 seconds
  no anomalies in 240 test result(s)

rng=jsf32, seed=0x35312dbd
length= 32 gigabytes (2^35 bytes), time= 77.6 seconds
  no anomalies in 251 test result(s)

rng=jsf32, seed=0x35312dbd
length= 64 gigabytes (2^36 bytes), time= 158 seconds
  Test Name                         Raw       Processed     Evaluation
  [Low8/32]FPF-14+6/16:cross        R=  -2.3  p =1-5.4e-4   unusual          
  ...and 262 test result(s) without anomalies

rng=jsf32, seed=0x35312dbd
length= 128 gigabytes (2^37 bytes), time= 316 seconds
  no anomalies in 273 test result(s)

rng=jsf32, seed=0x35312dbd
length= 256 gigabytes (2^38 bytes), time= 618 seconds
  no anomalies in 284 test result(s)
```

By contrast, the 16-bit version of the Small Fast Chaotic (SFC) PRNG passes the same tests:

```
./RNG_test sfc16 -multithreaded
RNG_test using PractRand version 0.94
RNG = sfc16, seed = 0x2a9bd6e9
test set = core, folding = standard (16 bit)

rng=sfc16, seed=0x2a9bd6e9
length= 1 gigabyte (2^30 bytes), time= 3.5 seconds
  no anomalies in 196 test result(s)

rng=sfc16, seed=0x2a9bd6e9
length= 2 gigabytes (2^31 bytes), time= 6.8 seconds
  no anomalies in 209 test result(s)

rng=sfc16, seed=0x2a9bd6e9
length= 4 gigabytes (2^32 bytes), time= 13.4 seconds
  no anomalies in 221 test result(s)

rng=sfc16, seed=0x2a9bd6e9
length= 8 gigabytes (2^33 bytes), time= 26.6 seconds
  no anomalies in 233 test result(s)

rng=sfc16, seed=0x2a9bd6e9
length= 16 gigabytes (2^34 bytes), time= 52.8 seconds
  no anomalies in 244 test result(s)

rng=sfc16, seed=0x2a9bd6e9
length= 32 gigabytes (2^35 bytes), time= 103 seconds
  no anomalies in 255 test result(s)

rng=sfc16, seed=0x2a9bd6e9
length= 64 gigabytes (2^36 bytes), time= 209 seconds
  no anomalies in 266 test result(s)

rng=sfc16, seed=0x2a9bd6e9
length= 128 gigabytes (2^37 bytes), time= 417 seconds
  no anomalies in 277 test result(s)

rng=sfc16, seed=0x2a9bd6e9
length= 256 gigabytes (2^38 bytes), time= 821 seconds
  no anomalies in 288 test result(s)
```

So, SFC-16 looks like a better choice if we want to keep the current interface for the time being.

arurke: Thanks for looking into this. Reading some of the [docs. in RractRand](https://pracrand.sourceforge.net/RNG_engines.txt), SFC-16 does indeed look promising (and it also seems popular as I found misc. implementations out there). However, from what I can see in the [original implementation](https://github.com/MartyMacGyver/PractRand/blob/master/src/RNGs/sfc.cpp#L41), it expects an 8-byte seed - am I missing something?

kkrentz: scf64 seeds like this:

```c
void PractRand::RNGs::Raw::sfc64::seed(Uint64 s) {
	a = b = c = s;
	counter = 1;
	for (int i = 0; i < 12; i++) raw64();
}
```

The tests still pass when I apply this method to scf16.

I just worry about the magic number `12`. The comments suggest that it was fine-tuned, but I do not know how.

kkrentz: I discovered that PractRand offers a test called `ttseed64` for testing the randomness of the first random number after seeding. Unfortunately, this test fails when I seed sfc16 like described above, even if I crank up that magic number. Interestingly, jsf16 passes that test.

arurke: It sounds like you found some issues with all the alternatives: sfc16(+sc64 seed-strategy), jsf16, and the original jsf32. Are we able to conclude @kkrentz ? If we are split in the middle on the alternatives my two-cents would be that jsf32 was proposed by TI in its current manner.

kkrentz: When seeded with 16 bits, sfc32 and jsf32 fail `ttseed64`. Therefore, I am contemplating revising the `random_init` function to take a 64-bit seed. That seed may come from the CSPRNG to speed things up.

PR #2338, opened 2022-12-13
nrf: import nrf5340 nrfx config

pjonsson:
This is the file from the nrfx templates,
except the addition+use of the NRF_PERIPH
macro.

Cherry-picked from a commit by John Kanwar.

### DIFF
--- a/arch/cpu/nrf/nrfx_config_nrf5340_application.h
+++ b/arch/cpu/nrf/nrfx_config_nrf5340_application.h
@@ -1,0 +1,2173 @@
+/*
+ * Copyright (c) 2019 - 2021, Nordic Semiconductor ASA
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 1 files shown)


Yagoor:
@pjonsson would it make sense to remove the `nrfx/templates` from the include path and copy all the templates that we use? I used the templates folder because I was using the headers vanilla, so I saw no point in copying them. I fear that having the same file listed in two different paths that are includes adds a risk since the file to be included will be determined by the includes order.

pjonsson:
@Yagoor I've also thought about the risk, but isn't that already an issue with `nrfx_glue.h`?

I'm not opposed to the idea on a technical level, but I would like to get the basic TrustZone support in the tree before the nrf target is re-architected.

Yagoor:
@pjonsson That is correct. All files from the template folder should not have been used the way I did it.
My reasoning behind it was that it was easier to maintain the files like this as long as we are using them without modification else for every nrfx update, we would need to update our local copies. I think that if we start to modify them, we should just copy them. Of course, it can be done in a different PR or I can do it in a different PR, it is not a problem for me to do it. :) 
Just let me know what you think. If you agree, I can do it in a subsequent PR.

pjonsson:
@Yagoor our TrustZone tree contains some modifications to the linker scripts too, and I think similar modifications are part of more recent nrfx releases. If we can update nrfx instead of adding more special stuff to Contiki-NG, I think that would be a better way forward. The build system needs some sorting out before that though.

joakimeriksson (approved):

PR #2345, opened 2022-12-16
nrf: import trusted firmware SPU driver

pjonsson:
Import the SPU driver from the trusted firmware
project. The source directory for these files
is:

root/platform/ext/target/nordic_nrf/common/core/native_drivers

The commit also contains the makefile changes
for building this driver when TrustZone support
is enabled.


pjonsson:
These files should be kept in sync with upstream, so the things to review is whether the driver should be in the `dev` directory, and the changes to the Makefile.

PR #2346, opened 2022-12-16
nrf5340: import TrustZone target config

pjonsson:
Import the TrustZone target config
from the trusted firmware project.
The source directory for these files
is:

root/platform/ext/target/nordic_nrf/common/nrf5340

These files are currently not connected
to the build system, they will be when
the TrustZone build support is added.


pjonsson:
These files should be kept in sync with upstream, so the thing to review is whether the files should be in `arch/cpu/nrf/nrf5340`.

PR #2351, opened 2022-12-19
TSCH: use RSSI data from received EB packets in the link stats module

atiselsts:
Currently on RSSI data from network-layer packets is used. I see no reason to not to use RSSI data from EB packets as well. Could help in some situations, when the other types of packets are not sufficiently frequent, or have not started being received yet.

### DIFF
--- a/os/net/mac/tsch/tsch.c
+++ b/os/net/mac/tsch/tsch.c
@@ -505,17 +505,21 @@ tsch_rx_process_pending()
       && frame.fcf.frame_type == FRAME802154_BEACONFRAME;
 
     if(is_data) {
-      /* Skip EBs and other control messages */
-      /* Copy to packetbuf for processing */
+      /* Copy payload to packetbuf for processing */
       packetbuf_copyfrom(current_input->payload, current_input->len);
       packetbuf_set_attr(PACKETBUF_ATTR_RSSI, current_input->rssi);
       packetbuf_set_attr(PACKETBUF_ATTR_CHANNEL, current_input->channel);
-    }
 
-    if(is_data) {
       /* Pass to upper layers */
       packet_input();
+
     } else if(is_eb) {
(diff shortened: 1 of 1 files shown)


os/net/mac/tsch/tsch.c
```diff
@@ -516,6 +516,12 @@ tsch_rx_process_pending()
       /* Pass to upper layers */
       packet_input();
     } else if(is_eb) {
+
+      /* Don't pass to upper layers, but still count it in link stats */
+      packetbuf_set_attr(PACKETBUF_ATTR_RSSI, current_input->rssi);
+      packetbuf_set_attr(PACKETBUF_ATTR_CHANNEL, current_input->channel);
+      link_stats_input_callback((const linkaddr_t *)frame.src_addr);
+
```
arurke, line 524: Makes very much sense to me.

Maybe do a slight refactor, something like
```
   if(is_data) {
      /* Copy to packetbuf for processing, pass to upper layers */
      packetbuf_copyfrom(current_input->payload, current_input->len);
      packetbuf_set_attr(PACKETBUF_ATTR_RSSI, current_input->rssi);
      packetbuf_set_attr(PACKETBUF_ATTR_CHANNEL, current_input->channel);
      packet_input();
    }
    else if(is_eb) {
      /* Don't pass to upper layers, but still count it in link stats */
      packetbuf_set_attr(PACKETBUF_ATTR_RSSI, current_input->rssi);
      packetbuf_set_attr(PACKETBUF_ATTR_CHANNEL, current_input->channel);
      link_stats_input_callback((const linkaddr_t *)frame.src_addr);
      eb_input(current_input);
    }
```
or the `packetbuf_set_attr()`s in a `if(is_eb || is_data)` first

atiselsts: I did some refactoring and updated the comments

arurke (approved):

PR #2367, opened 2023-01-02
Improve ETX guesstimation from RSSI

atiselsts:
The current ETX-from-RSSI guesstimation is useful for faster forming of the RPL tree but has some problems:

1) The values for "good" and "bad" RSSI are hard-coded to -60 and -90, which are not sensible defaults for Sub-GHz radios. This PR allows to configure them.

2) The ETX value at "good" RSSI is 1.0, the best ETX possible, while in reality the link may be unusable due to asymmetry for example. If a new neighbor is discovered with a good RSSI, then RPL might switch to this neighbor even though there is a better neighbor with an actually working link already present. This PR proposes to set `ETX_DEFAULT` value as the best possible initial ETX value, and penalize this ETX value further for RSSI values below the "good" RSSI.

3) The comment in link-stats.c about ETX being a linear function from RSSI is misleading, the old function is not linear. The new logic makes the function linear again for RSSI values between the "good" and "bad".

The PR also cleans up some other things in the system headers - FYI the old `CTASSERT` macro was not working for me.

I'm interested in particular in the opinion of whether using `ETX_DEFAULT` as the starting value is sensible, and whether the new RSSI-to-ETX mapping is acceptable for other maintainers and users.

For the reference this was the old output of RSSI-to-ETX (using scaled-down ETX values):

```
-60: 1.000
-61: 1.034
-62: 1.071
-63: 1.111
-64: 1.154
-65: 1.200
-66: 1.250
-67: 1.304
-68: 1.364
-69: 1.429
-70: 1.500
-71: 1.579
-72: 1.667
-73: 1.765
-74: 1.875
-75: 2.000
-76: 2.143
-77: 2.308
-78: 2.500
-79: 2.727
-80: 3.000
-81: 3.000
-82: 3.000
-83: 3.000
-84: 3.000
-85: 3.000
-86: 3.000
-87: 3.000
-88: 3.000
-89: 3.000
-90: 3.000

```
This is the new output, with the default config options:

```
-60: 2.000
-61: 2.033
-62: 2.067
-63: 2.100
-64: 2.133
-65: 2.167
-66: 2.200
-67: 2.233
-68: 2.267
-69: 2.300
-70: 2.333
-71: 2.367
-72: 2.400
-73: 2.433
-74: 2.467
-75: 2.500
-76: 2.533
-77: 2.567
-78: 2.600
-79: 2.633
-80: 2.667
-81: 2.700
-82: 2.733
-83: 2.767
-84: 2.800
-85: 2.833
-86: 2.867
-87: 2.900
-88: 2.933
-89: 2.967
-90: 3.000

```


os/lib/assert.h
```diff
@@ -52,9 +52,7 @@ void _xassert(const char *, int) CC_NORETURN;
 #endif /* NDEBUG */
 
 #ifndef CTASSERT                /* Allow lint to override */
-#define CTASSERT(x)             _CTASSERT(x, __LINE__)
-#define _CTASSERT(x, y)         __CTASSERT(x, y)
-#define __CTASSERT(x, y)        typedef char __assert ## y[(x) ? 1 : -1]
+#define CTASSERT(x) _Static_assert(x, #x)
```
pjonsson, line 55: This is C11. Both arm and msp430 have build flags to use older standards than that.

atiselsts: Should we aim to continue to support them? The build checks, with the default compiler options, are passing.

atiselsts: Alternatively, can you suggest our own implementation of CTASSERT than would compile with the current build options and a recent compiler?

pjonsson: I haven't tried, but Arm could probably be switched to C11.

Somewhat surprisingly, it seems msp430-gcc accepts this:
```
$ cat test.c 
_Static_assert(3 < 4, "yope");
_Static_assert(3 > 4, "nope");
$ msp430-gcc -Wall test.c -c -std=gnu11
test.c:2:1: error: static assertion failed: "nope"
```
It doesn't have an `assert.h` include though, so the C11 compliance is limited.

pjonsson: It seems msp430-gcc supports `_Static_assert` even with `-std=gnu99`, which I guess is fair since it might be a GNU extension.

What is more surprising to me is that arm-none-eabi-gcc supports it with `-std=c99`. Do you care about non-gcc-ARM-compilers? They might be more strict with enforcing c99.

atiselsts: > Do you care about non-gcc-ARM-compilers? 
IMO we already use GCC specific extensions. For example, the statement-as-expression idiom in the `RTIMER_BUSYWAIT_UNTIL_ABS` implementation. So, probably no.

arurke: Could it be an idea to move the CTASSERT bit in a separate PR so that it is not blocking the ETX changes? You could use the preprocessor for now?

os/net/link-stats.c
```diff
@@ -98,30 +99,30 @@ link_stats_is_fresh(const struct link_stats *stats)
...
-#define RSSI_DIFF (RSSI_HIGH - RSSI_LOW)
-      uint16_t etx;
-      int16_t bounded_rssi = stats->rssi;
-      bounded_rssi = MIN(bounded_rssi, RSSI_HIGH);
-      bounded_rssi = MAX(bounded_rssi, RSSI_LOW + 1);
-      etx = RSSI_DIFF * ETX_DIVISOR / (bounded_rssi - RSSI_LOW);
-      return MIN(etx, ETX_INIT_MAX * ETX_DIVISOR);
+#define RSSI_DIFF (LINK_STATS_RSSI_HIGH - LINK_STATS_RSSI_LOW)
+      CTASSERT(RSSI_DIFF > 0 && "RSSI_HIGH must be greater then RSSI_LOW");
+      CTASSERT(ETX_DIVISOR * RSSI_DIFF < 0x10000 && "RSSI math overflow");
```
arurke, line 117: This is subjective but, in my humble opinion, the defines within the function greatly reduces the readability of the code. I would much prefer it being at the top of the file or in the header-file.

atiselsts: yeah, I don't disagree with this comment. will update the PR

os/lib/assert.h
atiselsts: @arurke done!

arurke (approved):

arurke:
@atiselsts You want to get this merged now, or wait for some more input

atiselsts:
Done from my side, feel free to merge!

os/sys/cc.h
```diff
@@ -109,6 +109,10 @@
 #define ABS(n)      (((n) < 0) ? -(n) : (n))
 #endif
 
+#ifndef BOUND
+#define BOUND(a, minimum, maximum)   MIN(MAX(a, minimum), maximum)
+#endif
```
kkrentz, line 114: Such macros may cause unintended side effects (think of `i++`). A safer option would be an inline function.

atiselsts: We already have MIN, MAX, ABS macros in the same header that have the same behavior. Compared to inline functions, macros have the benefit that they can be used for compile-time computations, including assertions. And I'd hope that contributors to Contiki and reviewers are proficient enough in C programming to be aware of this and catch unintentional bugs.

If we decide to do switch to inline functions, best to do it in another PR that handles all of them at once.

os/net/link-stats.c
```diff
@@ -98,30 +110,26 @@ link_stats_is_fresh(const struct link_stats *stats)
 }
 /*---------------------------------------------------------------------------*/
 #if LINK_STATS_INIT_ETX_FROM_RSSI
+/*
+ * Returns initial ETX value from an RSSI value.
+ *    RSSI >= RSSI_HIGH           -> use default ETX
+ *    RSSI_LOW < RSSI < RSSI_HIGH -> ETX is a linear function of RSSI
+ *    RSSI <= RSSI_LOW            -> use minimal initial ETX
+ **/
 uint16_t
```
kkrentz, line 119: This function could be `static`, couldn't it?

atiselsts: changed and pushed

os/sys/cc.h
arurke: I agree with @atiselsts here

PR #2387, opened 2023-01-19
Use CCFG_CONF_* for both TI platforms

arurke:
The two TI platforms are not aligned when it comes to configuring the CCFG: cc26x0-cc13x0 platform uses `CCXXWARE_CONF` prefix, while simplelink platform uses `CCFG_CONF` prefix. This PR refactors such that they both use `CCFG_CONF`, making it easier to move an application between the two platforms.

Using same deprecation-strategy as in #2311

### DIFF
--- a/doc/platforms/srf06-cc26xx.md
+++ b/doc/platforms/srf06-cc26xx.md
@@ -219,7 +219,7 @@ For more information on the serial bootloader, see its README under the `tools/c
 For deployment/production images, it is _strongly_ recommended to:
 
 * Disable the ROM bootloader by defining `ROM_BOOTLOADER_ENABLE` as 0. In doing so, it is impossible to enter bootloader mode unless one first erases the device's flash.
-* Disable the JTAG interface, by defining `CCXXWARE_CONF_JTAG_INTERFACE_ENABLE` as 0. In doing so, the only JTAG operation available will be a device forced mass erase (using SmartRF Flash Programmer or UniFlash).
+* Disable the JTAG interface, by defining `CCFG_CONF_JTAG_INTERFACE_DISABLE` as 1. In doing so, the only JTAG operation available will be a device forced mass erase (using SmartRF Flash Programmer or UniFlash).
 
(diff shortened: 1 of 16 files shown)


atiselsts:
What about `cc26x0-cc13x0` based custom boards, their config would break after this PR is applied?
If we want to unify at all, perhaps using `CCXXWARE_CONF` defines is better, I'm sure there are far more user-defined boards using the `cc26x0-cc13x0` config as basis than `simplelink` ones.

arurke:
I opted for `CCFG_CONF` because its naming is agnostic to the platform used (both have CCFG). `CCXXWARE_CONF` indicates to users they are configuring the ccxxxxware driver suite - which is not in use in on simplelink. Secondly I think it is fair to assume that simplelink will be the preferred choice in the long term, leaving `CCXXWARE_CONF` increasingly misleading.

Existing use of `CCXXWARE_CONF` will break, but gracefully with a error message telling how to fix. We could consider other deprecation strategies, such as silently (or with warning) moving any `CCXXWARE_CONF` defines values to `CCFG_CONF`.

arurke:
I suggest this for v4.9 since #2311 has been merged and it would be nice to get both these similar changes in the same release. I added WIP label as I want to add a note for the release-notes.

arurke:
I added a note to the release-note.

@atiselsts are your concerns addressed by my comment above?

atiselsts:
I understand you explanation and agree with depreciation strategy. Do you want me to do more in-depth reivew of the code?

arurke:
An extra pair of eyes would be great if you have time

doc/platforms/srf06-cc26xx.md
```diff
@@ -219,7 +219,7 @@ For more information on the serial bootloader, see its README under the `tools/c
 For deployment/production images, it is _strongly_ recommended to:
 
 * Disable the ROM bootloader by defining `ROM_BOOTLOADER_ENABLE` as 0. In doing so, it is impossible to enter bootloader mode unless one first erases the device's flash.
```
atiselsts, line 221: I'm thinking that references to ROM_BOOTLOADER_ENABLE (the backwards-compatible name) probably should be changed to `CCFG_CONF_ROM_BOOTLOADER_ENABLE`, for consistency. But that's maybe outside the scope of this PR

atiselsts (approved):

PR #2389, opened 2023-01-19
Fix include of random-lib in rpl-icmp6

arurke:
As in title, gave me trouble when making binaries for the FIT IoT-lab.

### DIFF
--- a/os/net/routing/rpl-classic/rpl-icmp6.c
+++ b/os/net/routing/rpl-classic/rpl-icmp6.c
@@ -51,7 +51,7 @@
 #include "net/routing/rpl-classic/rpl-private.h"
 #include "net/packetbuf.h"
 #include "net/ipv6/multicast/uip-mcast6.h"
-#include "random.h"
+#include "lib/random.h"
 
 #include "sys/log.h"
 


pjonsson:
Is there something special with this file, or is this always the wrong path to include?

```text
$ git grep -l '#include "random.h"'
arch/platform/nrf/platform.c
examples/coap/coap-plugtest-server/resources/res-plugtest-test.c
examples/coap/coap-plugtest-server/resources/res-plugtest-validate.c
examples/libs/ipv6-hooks/ipv6-hooks.c
examples/libs/ipv6-uipbuf/udp-client.c
examples/libs/stack-check/example-stack-check.c
examples/platform-specific/cc26x0-cc13x0/base-demo/base-demo.c
examples/rpl-udp/udp-client.c
os/net/routing/rpl-classic/rpl-icmp6.c
```

arurke:
To be honest, I don't know. It was an old commit I had lying from long time ago when I did some work on FIT IoT-lab. I noticed all files in `os` used that path. Can't spare time to dig deeper for now unfortunately, but I would argue it is worth a merge given that others may be meeting the same issue with FIT IoT-lab.

PR #2390, opened 2023-01-19
Minor code-style, logging, comments in RPL, TSCH, uIP

arurke:
A small set of minor fixes in RPL, TSCH, uip6.

### DIFF
--- a/os/net/ipv6/uip-ds6-route.c
+++ b/os/net/ipv6/uip-ds6-route.c
@@ -310,7 +310,7 @@ uip_ds6_route_lookup(const uip_ipaddr_t *addr)
     LOG_INFO_6ADDR(uip_ds6_route_nexthop(found_route));
     LOG_INFO_("\n");
   } else {
-    LOG_WARN("No route found\n");
+    LOG_INFO("No route found\n");
   }
 
   if(found_route != NULL && found_route != list_head(routelist)) {
--- a/os/net/routing/routing.h
+++ b/os/net/routing/routing.h
@@ -94,7 +94,7 @@ struct routing_driver {
    * \param node The source routing node
    * \return 1 if the global node address was copied, 0 otherwise
   */
-  int (* get_sr_node_ipaddr)(uip_ipaddr_t *addr, const uip_sr_node_t *node);
+  int (* get_sr_node_ipaddr)(uip_ipaddr_t *ipaddr, const uip_sr_node_t *node);
   /**
    * Leave the network the node is part of
    *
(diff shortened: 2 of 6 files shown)


os/net/routing/routing.h
```diff
@@ -171,9 +171,7 @@ struct routing_driver {
   /**
    * Called by uIP to notify addition/removal of IPv6 neighbor entries
    *
-   * \param addr The link-layer addrress of the packet destination
-   * \param status The transmission status (see os/net/mac/mac.h)
-   * \param numtx The total number of transmission attempts
+   * \param nbr The neighbor entry
```
pjonsson, line 174: Line 93 in this file is also faulty, is this code not part of the doxygen build?

arurke: Unfortunately I am not familiar with Doxygen - it should react to such mismatches?
In any case, I fixed line 93 as well now.

pjonsson: The only thing I know about Doxygen is that it sometimes does, but it sometimes doesn't. And the next revision (1.9.2->1.9.3) might warn for completely different things. When the next revision doesn't break the support for preprocessor defines, the warnings one gets tend to be valid though.

Clang has a warning that also helps (`-Wdocumentation` or similar?).

All that is outside the scope of this patch though, thanks for fixing the extra issue!

os/net/ipv6/uip-ds6-route.c
```diff
@@ -310,7 +310,7 @@ uip_ds6_route_lookup(const uip_ipaddr_t *addr)
     LOG_INFO_6ADDR(uip_ds6_route_nexthop(found_route));
     LOG_INFO_("\n");
   } else {
-    LOG_WARN("No route found\n");
+    LOG_INFO("No route found\n");
```
pjonsson, line 313: From what I understand, taking this path means using the default route. Is that something the user should be worried about, or should the message include ", using default route"?

arurke: Although this function is indeed used for routing decisions (where the default route would be used if route not found), the function is a generic lookup and is also for other purposes - so it would be inappropriate to add a "using default route" here. For routing decisions there is actual such a log (at INFO level) in tcpip.c:529.

I cannot see the need for WARN-level for the lookup not finding a route. It makes the WARN-level verbose since it is printed every time the default route is used - which imo is not something for the user to be worried about.

os/net/routing/routing.h
```diff
@@ -90,7 +90,7 @@ struct routing_driver {
   /**
    * Returns the global IPv6 address of a source routing node
    *
-   * \param ipaddr A pointer where to copy the IP address of the node
+   * \param addr A pointer where to copy the IP address of the node
```
atiselsts, line 93: for consistency, would be better to change the parameter name, not the docs

os/net/ipv6/uip-ds6-route.c
atiselsts: Agree with Andreas here

os/net/routing/routing.h
arurke: 🤦‍♂️ Fixed!

nvt (approved):

nfi (approved):

PR #2393, opened 2023-01-27
TSCH: avoid potential overflow in the backoff window counter

atiselsts:
The TSCH `backoff_window` variable is uint8_t, meaning that it will overflow for large backoff exponent values and create  unexpected results, with nodes backing off less than they should.

This PR changes the variable to 16-bit uint, removes unused last_backoff_window variable, and adds protection against 16-bit overflow to the `tsch_queue_backoff_inc` logic.


os/net/mac/tsch/tsch-queue.c
```diff
@@ -504,13 +504,19 @@ tsch_queue_backoff_inc(struct tsch_neighbor *n)
 {
   /* Increment exponent */
   n->backoff_exponent = MIN(n->backoff_exponent + 1, TSCH_MAC_MAX_BE);
+  /* Avoid potential 16-bit integer overflow */
+  if(n->backoff_exponent > 16) {
+    n->backoff_exponent = 16;
+  }
```
arurke, line 510: Would it be clearer for users to move this compile-time with something like:

```
#if TSCH_MAC_MAX_BE > 16
#error TSCH_MAC_MAX_BE must be 16 or lower
#endif
```

With the run-time check, abnormal large `TSCH_MAC_CONF_MAX_BE` will be masked at run-time and it may be difficult for users to understand the results.

atiselsts: Updated and rebased the PR, please take a look!

arurke: :+1:

arurke (approved):

PR #2398, opened 2023-01-30
Prevent buffer overflows due to misconfigured BLE MTU.

nvt:
Thanks to @SWW13 and @Scepticz for reporting the issue.

### DIFF
--- a/os/net/mac/ble/ble-l2cap.h
+++ b/os/net/mac/ble/ble-l2cap.h
@@ -76,6 +76,10 @@
 #define BLE_L2CAP_NODE_MTU      PACKETBUF_SIZE
 #endif
 
+#if BLE_L2CAP_NODE_MTU > PACKETBUF_SIZE
+#error BLE_L2CAP_NODE_MTU exceeds the PACKETBUF_SIZE
+#endif
+
 /* the max. supported L2CAP fragment length */
 #ifdef BLE_L2CAP_CONF_NODE_FRAG_LEN
 #define BLE_L2CAP_NODE_FRAG_LEN   BLE_L2CAP_CONF_NODE_FRAG_LEN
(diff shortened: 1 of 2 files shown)


os/net/mac/ble/ble-l2cap.h
```diff
@@ -73,7 +73,11 @@
 #ifdef BLE_L2CAP_CONF_NODE_MTU
 #define BLE_L2CAP_NODE_MTU      BLE_L2CAP_CONF_NODE_MTU
 #else
-#define BLE_L2CAP_NODE_MTU              1280
+#define BLE_L2CAP_NODE_MTU      PACKETBUF_SIZE
+#endif
+
+#if BLE_L2CAP_NODE_MTU != PACKETBUF_SIZE
+#error Invalid BLE node MTU setting
 #endif
```
arurke, line 81: I know nothing about BLE MTU, but are you not allowed to have a smaller MTU than the packetbuf? If not, and `BLE_L2CAP_NODE_MTU` must indeed be equal to `PACKETBUF_SIZE`, would it not be cleaner to simply remove `BLE_L2CAP_CONF_NODE_MTU`, as it would be moot?

nvt: Thanks for the comment. The PR has been updated and this restriction has been removed. Moreover, the default value is now PACKETBUF_SIZE instead of 1280.

arurke: Don't you want to keep a check for for `BLE_L2CAP_NODE_MTU > PACKETBUF_SIZE` to provide a compile-time warning for misconfigurations? (if I understood the issue correctly)

nvt: Yes, I added a check for this now.

nfi (approved):

PR #2424, opened 2023-04-14
Fix stack address escape

szsam:
Fix storing the address of a local variable in non-local memory:
https://github.com/contiki-ng/contiki-ng/blob/8518cbaff6a717d02444b686a75cdb240648fbf4/arch/cpu/cc2538/dev/aes.c#L116


szsam:
Okay. If we avoid static allocation here, shall we assign NULL to `REG(AES_DMAC_CH0_EXTADDR)` after the DMA transfer finishes? Because it's undesirable to keep a dangling pointer in MMIO region (AES registers).

joakimeriksson:
Yes, that would be better. Cleaning up traces of where we had the key is a better option (e.g. setting the address to NULL).

szsam:
@joakimeriksson The AES register that saves the key is cleared after DMA transfer. PTAL.

joakimeriksson (approved):

PR #2425, opened 2023-04-14
antelope,storage-cfs: Fix length of filename array

szsam:
https://github.com/contiki-ng/contiki-ng/blob/406ae7da30fb5fadacf0d646a4dc4ac0513a2dec/os/storage/antelope/storage.h#L47-L49
The macro `INDEX_NAME_LENGTH` does not count the trailing null char.


Machiry:
This looks like a off-by-one-byte error. Should there be a CVE for this?

nvt:
@Machiry Yes, it might get a CVE, but it is difficult to exploit the bug without access to the DBMS; e.g., by being able to inject shell commands into the IoT node.

PR #2429, opened 2023-04-19
Disable APPROTECT for nRF52840 rev 3 devices

thvdveld:
This PR disables the APPROTECT for the nRF52840 revision 3 devices. With the third revision, the APPROTECT needs to explicitly be disabled by the firmware. Disabling the APPROTECT allows us to debug the firmware.

### DIFF
--- a/arch/platform/nrf52840/platform.c
+++ b/arch/platform/nrf52840/platform.c
@@ -85,10 +85,58 @@ populate_link_address(void)
   memcpy(&linkaddr_node_addr, &device_address[8 - LINKADDR_SIZE],
          LINKADDR_SIZE);
 }
+
+void
+approtect_device(bool lock)
+{
+  uint32_t PALL;
+  if (lock)
+  {
+    PALL = UICR_APPROTECT_PALL_Enabled;
+  }
+  else
+  {
+    PALL = UICR_APPROTECT_PALL_Disabled;
+  }
+  
+  uint32_t variant = NRF_FICR->INFO.VARIANT;
+  if ((uint8_t)(variant >> 8) >= (uint8_t)'F')
+  {
+  	if ((NRF_UICR->APPROTECT & UICR_APPROTECT_PALL_Msk) !=
+        (PALL << UICR_APPROTECT_PALL_Pos))
+    {
+        NRF_NVMC->CONFIG = NVMC_CONFIG_WEN_Wen;
+        while (NRF_NVMC->READY == NVMC_READY_READY_Busy) {}
+
+        NRF_UICR->APPROTECT = 
+          ((NRF_UICR->APPROTECT & ~((uint32_t)UICR_APPROTECT_PALL_Msk)) |
+          (PALL << UICR_APPROTECT_PALL_Pos));
+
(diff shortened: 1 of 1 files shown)


arch/platform/nrf52840/platform.c
```diff
@@ -85,10 +85,70 @@ populate_link_address(void)
   memcpy(&linkaddr_node_addr, &device_address[8 - LINKADDR_SIZE],
          LINKADDR_SIZE);
 }
+
+void
+lock_device(void)
```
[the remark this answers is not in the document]
thvdveld: You're right! Changed it.

arch/platform/nrf52840/platform.c
```diff
@@ -85,10 +85,58 @@ populate_link_address(void)
...
+        while (NRF_NVMC->READY == NVMC_READY_READY_Busy) {}
+    }
+  }
+}
+
 /*---------------------------------------------------------------------------*/
 void
 platform_init_stage_one(void)
 {
+#ifdef NRF52840_ENABLE_APPROTECT
```
thvdveld, line 125: Does it make sense to have this? Or should I remove the `#ifdef`?

pjonsson: I can't say if the ifdef should exist, but it's probably better to have the ifdef inside the function instead, so `PALL` is defined based on that.

arch/platform/nrf52840/platform.c
```diff
@@ -85,10 +85,58 @@ populate_link_address(void)
   memcpy(&linkaddr_node_addr, &device_address[8 - LINKADDR_SIZE],
          LINKADDR_SIZE);
 }
+
+void
+approtect_device(bool lock)
+{
+  uint32_t PALL;
```
pjonsson, line 92: Put `PALL` inside the rev 3 block since that is where it is used.

arch/platform/nrf52840/platform.c
```diff
@@ -85,10 +85,58 @@ populate_link_address(void)
...
+    }
+  }
+}
+
 /*---------------------------------------------------------------------------*/
 void
 platform_init_stage_one(void)
 {
+#ifdef NRF52840_ENABLE_APPROTECT
+  /*
```
pjonsson, line 126: The block comment is great, but only important when looking at approtect, so put the comment above the rev 3 block inside the function.

alexstanoev:
Thanks for working on this!

Writing to `NRF_UICR->APPROTECT` only covers half of the requirements to open up APPROTECT, as the firmware also needs to write to NRF_APPROTECT->DISABLE after every reset:
https://github.com/NordicSemiconductor/nrfx/blob/master/mdk/system_nrf52_approtect.h#L55
https://infocenter.nordicsemi.com/index.jsp?topic=%2Fps_nrf52840%2Fdif.html&cp=5_0_0_3_7_1&anchor=register.DISABLE

The MDK as part of the old nRF5 SDK used in the nrf52840 port doesn't have the types defined for NRF_APPROTECT unfortunately.

simonduq:
Hi!

thvdveld:
Going to close this as I don't have the capacity to work on this.

PR #2431, opened 2023-04-27
add PLATFORM_HAS_LEDS to coojamote header files

asereq:
Cooja motes implement the LED HAL API (thanks to https://github.com/contiki-ng/contiki-ng/pull/1044), but do not have the right defines in place, so it end up being unused.


arurke (approved):
LGTM

PR #2432, opened 2023-04-27
Fix minor errors on cooja led definitions

asereq:


arch/platform/cooja/contiki-conf.h
```diff
@@ -153,8 +153,8 @@ typedef unsigned long clock_time_t;
 /*---------------------------------------------------------------------------*/
 /* Virtual LED colors */
 #define LEDS_CONF_COUNT                  3
-#define LEDS_CONF_GREEEN                 1
-#define LEDS_CONF_RED                    2
-#define LEDS_CONF_YELLOW                 4
+#define LEDS_CONF_GREEN                 1
+#define LEDS_CONF_RED                    4
+#define LEDS_CONF_YELLOW                 2
```
arurke, line 158: For code-style, maintain the indentation of the value? (and may I suggest moving `LEDS_CONF_RED` to the bottom for better readability

arurke:
@asereq Can you clean the history so we only have one commit?

Also, are you seeing blue color in Cooja as opposed to yellow...or is it just me 🤯

arurke:
Hijacking this to query @pjonsson: I was getting a blue color from my Cooja mote when I lit the "yellow" LED. Do you know if this is an issue with the define-naming in contiki, or the color-selection in cooja GUI? If it is the first I can open a PR to fix, if it is the second I can open an issue in Cooja.

pjonsson:
@arurke It's java/org/contikios/cooja/contikimote/interfaces/ContikiLED.java that contains the LED-implementation for Contiki motes in Cooja. There's no mention of blue in that code, so that sounds like a Cooja-issue. (The colors on line 68-83 are not blue according to the color preview next to the line number in IntelliJ.)

pjonsson:
I think the bug for this issue is actually in Cooja rather than Contiki-NG, so I made a PR for that: https://github.com/contiki-ng/cooja/pull/1298.

@arurke Niclas said the blue color is from the first generation tmote sky having a LED of that color. Don't think we want to touch that before the release.

Issue #2401, opened 2023-02-08
Out of Bound Read when processing TCP MSS option


AmPaschal:
There is a possible out of bound read when a TCP packet containing MSS options is being processed.

The function snippet below shows the code that processes the MSS option.

```
/* Parse the TCP MSS option, if present. */
  if((UIP_TCP_BUF->tcpoffset & 0xf0) > 0x50) {
    for(c = 0; c < ((UIP_TCP_BUF->tcpoffset >> 4) - 5) << 2 ;) {
      opt = uip_buf[UIP_IPTCPH_LEN + c];
      if(opt == TCP_OPT_END) {
        /* End of options. */
        break;
      } else if(opt == TCP_OPT_NOOP) {
        ++c;
        /* NOP option. */
      } else if(opt == TCP_OPT_MSS &&
                uip_buf[UIP_IPTCPH_LEN + 1 + c] == TCP_OPT_MSS_LEN) {
        /* An MSS option with the right option length. */
        tmp16 = ((uint16_t)uip_buf[UIP_IPTCPH_LEN + 2 + c] << 8) |
          (uint16_t)uip_buf[UIP_IPTCPH_LEN + 3 + c];
        uip_connr->initialmss = uip_connr->mss =
          tmp16 > UIP_TCP_MSS? UIP_TCP_MSS: tmp16;

        /* And we are done processing options. */
        break;
      } else {
        /* All other options have a length field, so that we easily
           can skip past them. */
        if(uip_buf[UIP_IPTCPH_LEN + 1 + c] == 0) {
          /* If the length field is zero, the options are malformed
             and we don't process them further. */
          break;
        }
        c += uip_buf[UIP_IPTCPH_LEN + 1 + c];
      }
    }
  }
```

Each iteration of the for loop only checks that at the start of the iteration, c is within the bounds of TCP option. Within the iteration, c is increased and used to access the buffer without any further checks. Hence, the MSS value can be computed using bytes read out of bound.

For example, if a 4 byte option (ox01010204) is added to a packet, the first 2 bytes are skipped (TCP_OPT_NOOP), the 3rd and 4th byte indicates there is an MSS option with 4 byte of length. Hence, this code would read 2 bytes outside the bounds of the buffer.

PR #2434, opened 2023-05-05
Add boundary checks in TCP option processing.

nvt:
This PR adds boundary checks in the processing of TCP options. Without these checks, it is possible to inject a TCP packet that causes out-of-bounds memory access.

In addition, the PR refactors duplicate TCP option processing code into a separate function.

### DIFF
--- a/os/net/ipv6/uip6.c
+++ b/os/net/ipv6/uip6.c
@@ -919,6 +919,61 @@ ext_hdr_options_process(uint8_t *ext_buf)
   return 0;
 }
 /*---------------------------------------------------------------------------*/
+#if UIP_TCP
+static void
+process_tcp_options(struct uip_conn *conn)
+{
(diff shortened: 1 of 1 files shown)


MartinNPN:
It looks like this is related to https://github.com/contiki-ng/contiki-ng/issues/2401

os/net/ipv6/uip6.c
```diff
@@ -919,6 +919,61 @@ ext_hdr_options_process(uint8_t *ext_buf)
   return 0;
 }
 /*---------------------------------------------------------------------------*/
+#if UIP_TCP
+static void
+process_tcp_options(struct uip_conn *conn)
```
pjonsson, line 924: Can the pointer be `const`?

joakimeriksson (approved):

nfi (approved):

Issue #2285, opened 2022-11-18
Bug in IPv6 extension header processing


tinstructor:
In `os/net/ipv6/uip6.c` the function `ext_hdr_options_process()` is called twice for every incoming packet (I'm using RPL-Classic in Storing Mode if that matters):
https://github.com/contiki-ng/contiki-ng/blob/0c2a7d18c50e72e49ac0790e84bf7d2aa9350215/os/net/ipv6/uip6.c#L1203-L1214
https://github.com/contiki-ng/contiki-ng/blob/0c2a7d18c50e72e49ac0790e84bf7d2aa9350215/os/net/ipv6/uip6.c#L1295-L1316
This means that when a rank-error is detected, the first call tolerates the single rank error (as per RPL spec) and sets the corresponding bit in the RPL Options. However, the immediate second call to `ext_hdr_options_process()` then causes the packet to be dropped immediately. Effectively, this means that the single-rank-error tolerating mechanism of RPL is currently broken.

arurke:
@tinstructor I might be meeting this issue in my work now. Did you get any further on this, e.g. towards a solution? Does it match your understanding that this issue is only happening if the packet is addressed to the current node?

I got to say, debugging that 1000+ lines `uip_process()` function riddled with goto's and defines is not good for mental health.

arurke:
Resolved by #2862

PR #2446, opened 2023-05-19
Workaround for RPL forwarding inconsistency leading to dropped packet

arurke:
This is a workaround for issue #2285, where received packets that are addressed to ourselves and are forwarded inconsistently (according to [RFC6550 Section 11.2.2.2](https://www.rfc-editor.org/rfc/rfc6550#section-11.2.2.2)), will be dropped. This happens as the packet goes through the RPL extension-header processing twice: In the first pass it will be marked with the Rank-Error bit, and in the second pass the bit is seen and the packet is dropped.

This PR
1. Adds a workaround to the issue, ensuring the packet is not dropped.
2. ~~Adds a simulation test-case that showcases the issue. The test fails without the workaround, and passes with it.~~. The test for the issue was added separately in #2548 . This PR simply reverts the logic in the test such that it passes only if the issue is resolved.

I am calling this a workaround because it only mitigates the dropping of the packet (the issue still causes reset of DIO timer and erroneous RPL statistics). I have taken this conservative low-risk approach because I am no RPL/uIP expert and I can't really judge what is the proper solution - e.g. if we should rather avoid the double processing of extension-headers in such cases (and if so how to safely implement). As such I will have no trouble if we don't merge this and rather leave it for others to work onward from.

For further details see test-description in #2548 and issue #2285.

### DIFF
--- a/tests/15-rpl-classic/js/12-rpl-rank-error.js
+++ b/tests/15-rpl-classic/js/12-rpl-rank-error.js
@@ -5,13 +5,11 @@ while(true) {
 
   if(msg.contains("RPL Option Error: Dropping Packet")) {
     log.log(time + " " + "node-" + id + " "+ msg + "\n");
-    // FIXME Logic inverted to showcase issue. Revert when implementing fix.
-    log.testOK();
+    log.testFailed();
   }
 
   if(msg.contains("Received all packets")) {
     log.log(time + " " + "node-" + id + " "+ msg + "\n");
-    // FIXME Logic inverted to showcase issue. Revert when implementing fix.
-    log.testFailed();
+    log.testOK();
   }
 }
--- a/os/net/routing/rpl-classic/rpl-ext-header.c
+++ b/os/net/routing/rpl-classic/rpl-ext-header.c
@@ -159,6 +159,10 @@ rpl_ext_header_hbh_update(uint8_t *ext_buf, int opt_offset)
       /* Packet must be dropped and dio trickle timer reset, see
          RFC6550 - 11.2.2.2 */
       rpl_reset_dio_timer(instance);
+      /* Don't drop packet if it was for us. Workaround for issue #2285 */
+      if(uip_ds6_is_my_addr(&UIP_IP_BUF->destipaddr)) {
+        return 1;
+      }
       return 0;
     }
 


arurke:
@tinstructor Does this match your understanding and does the workaround work for you?

tests/15-rpl-classic/code-rank-error/Makefile
```diff
@@ -0,0 +1,8 @@
+CONTIKI_PROJECT = simple-node
+
+all: $(CONTIKI_PROJECT)
+
+MAKE_ROUTING = MAKE_ROUTING_RPL_CLASSIC
+
+CONTIKI = ../../..
```
pjonsson, line 7: The top Makefile defines the top directory, so I don't think `$CONTIKI` is required to be defined by test cases anymore.

tests/15-rpl-classic/js/12-rpl-rank-error-mobility.dat
```diff
@@ -0,0 +1,7 @@
+0 0.0 0 0
+1 0.0 30 0
+2 0.0 60 0
+3 0.0 90 0
+3 180.0 30 20
+3 3600.0 30 20
+
```
pjonsson, line 7: Is the blank line required?

tests/15-rpl-classic/code-rank-error/simple-node.c
```diff
@@ -0,0 +1,103 @@
...
+
+  if(rx_count == SEND_COUNT_MAX) {
+    LOG_INFO("Received all packets\n");
+  }
+}
+/*---------------------------------------------------------------------------*/
+PROCESS_THREAD(udp_process, ev, data)
+{
+  static struct etimer periodic_timer;
+  uip_ipaddr_t dest_addr;
```
pjonsson, line 54: Can this be declared in the while-loop?

tests/15-rpl-classic/js/12-rpl-rank-error-mobility.dat
arurke: I have this as a habit to ensure that the last line has terminating symbols. The idea is that some systems/tools/etc. out there defines a line as chars ending with terminators, and may misbehave if not present. Can't say I have done extensive research on this, but quick googling seems to say it is reasonable - happy to be convinced otherwise.

arurke:
Thanks for the review @pjonsson!

tests/15-rpl-classic/js/12-rpl-rank-error-mobility.dat
pjonsson: This particular file is read by Cooja which isn't really portable at all (Linux/OS X and maybe Windows/FreeBSD). Even if Cooja was portable, one would want a fast host to run simulations, and to run things at all the host still needs a few gigs of RAM and a JVM. I think this narrows down the host platforms to Linux-like/OS X/Windows.

I'm not sure what to search for based on your description, isn't EOF a terminator?

Without being sure what class of bugs you are trying to avoid with the empty line, my gut feeling for the general case is that pipelining basic unix tools (awk/sed/grep/perl) becomes harder with files that contain spurious blank lines since the blank lines need to be filtered out from the main loop to not end up with an empty variable for the line (`$_` in perl, I have no idea what it is called in other languages).

arurke: Sorry @pjonsson, I see now that I have added an actual blank line (as you of course wrote in the first comment). So there is now two LFs at the end of the file. This was not my intention, I wanted to ensure the last line had an LF. I got confused because in my local editor, if the last line has a LF, a blank line is shown beneath it. When I looked at the file here in github.com I saw a blank line and thought everything was as I wanted, but did not realize github does not show a separate blank line when the last line has LF - it was actually a second LF there.

> I'm not sure what to search for based on your description, isn't EOF a terminator?

I don't believe it is no, leaning on SO: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

arurke: I have removed the stray line

tests/15-rpl-classic/js/12-rpl-rank-error.js
```diff
@@ -0,0 +1,26 @@
...
+  YIELD();
+
+  if(msg.contains("RPL Option Error: Dropping Packet")) {
+    log.log(time + " " + "node-" + id + " "+ msg + "\n");
+    log.testFailed();
+  }
+
+  else if(msg.contains("Received all packets")) {
+    log.log(time + " " + "node-" + id + " "+ msg + "\n");
+    all_transmissions_received = true
```
pjonsson, line 24: Does this test always terminate with a timeout? Is there any reason to not let it terminate with testOK() here?

arurke: This was intentional, but after review I can't really see the rationale - there is plenty of time after the nodes move to catch any issues. I have simplified the script.

tests/15-rpl-classic/js/12-rpl-rank-error.js
```diff
@@ -0,0 +1,26 @@
+all_transmissions_received = false;
+
+function check_success() {
```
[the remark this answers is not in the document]
arurke: No indeed, some kind of artifact, removed 👍

arurke:
I have now rebased and removed the tests that were added via #2548.

As mentioned I believe there is more work to be done on the issue (ref. my explanation above of why calling this a workaround) - at the same time I believe this PR is conservative and is unlikely to do any harm.

joakimeriksson:
@arurke - we did pull in Nicolas fix for this. So I do assume that we can close this? #2862

arurke:
@joakimeriksson Indeed! Replaced by #2862 

@nvt In case you aren't aware, there is one more similar issue+workaround in https://github.com/contiki-ng/contiki-ng/pull/2545

nvt:
@arurke Yes, I have seen it, and plan to take a closer look soon.

PR #2452, opened 2023-06-01
arm: silence some GCC 12 warnings

pjonsson:
GCC 12 gives -Warray-bounds warnings
when writing to an address, and
Jakub Jelinek points out this breaks
a lot of existing code. GCC 13 is
supposed to be better in this
respect.

The binutils bundled with ARM GCC 12.1
also gives warnings about RWX-segments.
Disable the warning for now so the code
can build, and put in a FIXME with a
link to the blog post describing the rationale
and solution.


joakimeriksson (approved):
More to do but this moves us forward

PR #2473, opened 2023-06-24
Makefile: stop putting target in .

pjonsson:
This change simplifies the distclean rule, and
avoids overwriting the binary in . when
building the same firmware for two different boards.

Long term, this change will enable more
parallelism in the build system since
the same binary can be built for two different
boards at the same time.

This also improves build performance slightly
since it avoids one copy of the binary for
each target.

This commit requires removing old binaries
from the tree for tests to pass. A dry run
of git clean will show the files:

git clean -n tests examples

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,33 +28,6 @@ tools/coffee-manager/build/
 tools/coffee-manager/coffee.jar
 COOJA.testlog
 
-# platform build artifacts
-*.cc2538dk
-*.cooja
-*.native
-*.nrf52840
-*.openmote
-*.simplelink
-*.sky
-*.firmware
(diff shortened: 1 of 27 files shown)


simonduq:
Hi!

joakimeriksson (approved):

nfi (approved):

arurke:
@pjonsson Would you mind rebasing

pjonsson:
@arurke fixed!

kkrentz:
The new implementation of distclean produces this error on my Ubuntu:
```
make distclean
TARGET not defined, using target 'native'
/bin/sh: 1: [: =: unexpected operator
```

PR #2483, opened 2023-07-03
arm: remove OUT_ELF

pjonsson:
The ARM port assumes the linker produces
an elf file, and makes a copy of that.
Remove the copy, and make the linker produce
the target directly instead.

This allows removing the customrules for
cortex-m, since they become identical to
the rules in Makefile.include.

The output for 02-compile-arm-ports for

cimake NUMCORES=1 Q=''

is identical (modulo reordering and removing
the file copy) before/after this change.


nfi (approved):

simonduq:
Hi!

PR #2484, opened 2023-07-03
Added boundary checks for DIO and DAO input processing in RPL-Lite.

nvt:


os/net/routing/rpl-lite/rpl-icmp6.c
```diff
@@ -246,8 +257,17 @@ dio_input(void)
         if(dio.mc.type == RPL_DAG_MC_NONE) {
           /* No metric container: do nothing */
         } else if(dio.mc.type == RPL_DAG_MC_ETX) {
+	  if(len < 7) {
```
atiselsts, line 260: in the other branches you check `i + const` against the `buffer_length`, why hot here?

nvt: Thanks for the comment. I switched checking `i + len` against `buffer_length` before processing options, and then checking that `len makes sense for the particular option being processed.

nfi (approved):

PR #2490, opened 2023-07-07
nrf: fix typo in name for SLIP arch configuration

nfi:


nfi:
> If I have a custom nrf board and have set this flag using the old naming, would I get some hard-to-understand behavior as I updated the OS? If so, perhaps add a compile-time warning if the old-variable is defined? Or/and add to the next-version-release notes?

Good point. I have added a compile-time error if the old name is used.

arurke (approved):

PR #2503, opened 2023-07-09
Dockerfile: update to nrf-tools 10.22.0

pjonsson:


tools/docker/Dockerfile
```diff
@@ -93,12 +93,10 @@ RUN wget -nv http://simonduq.github.io/resources/mspgcc-4.7.2-compiled.tar.bz2 &
...
-  wget -nv https://nsscprodmedia.blob.core.windows.net/prod/software-and-other-downloads/desktop-software/nrf-command-line-tools/sw/versions-10-x-x/10-17-3/nrf-command-line-tools-10.17.3_linux-amd64.tar.gz && \
-  apt-get -qq -y --no-install-recommends install ./nrf-command-line-tools_10.17.3_amd64.deb > /dev/null && \
-  tar zxf nrf-command-line-tools-10.17.3_linux-amd64.tar.gz ./JLink_Linux_V766a_x86_64.deb && \
-  apt-get -qq -y --no-install-recommends install ./JLink_Linux_V766a_x86_64.deb > /dev/null && \
-  rm -f *.deb *.tar.gz && \
+RUN wget -nv https://nsscprodmedia.blob.core.windows.net/prod/software-and-other-downloads/desktop-software/nrf-command-line-tools/sw/versions-10-x-x/10-22-0/nrf-command-line-tools_10.22.0_amd64.deb && \
+  apt-get -qq -y --no-install-recommends install ./nrf-command-line-tools_10.22.0_amd64.deb > /dev/null && \
+  apt-get -qq -y --no-install-recommends install /opt/nrf-command-line-tools/share/JLink_Linux_V780c_x86_64.deb > /dev/null && \
+  rm -f *.deb *.tar.gz /opt/nrf-command-line-tools/share/JLink_*deb && \
```
arurke, line 99: Ref. discussion in #2506 Have you considered adding the version numbering to ARG or ENV variables? Since the formatting is different (10-22-0, 10.22.0, 10-x-x), separate variables for each number might make sense, e.g. :
```
ARG nrf-clt-major-version=10
ARG nrf-clt-minor-version=22
ARG nrf-clt-patch-version=0
```

An added bonus is that that users can easily override using the CLI with `--build-arg nrf-clt-minor-version=17`

arurke: > I don't think the (current) intention is that users can roll their own image with whatever versions they want, and the project still guarantees that everything will work.

Indeed, but isn't it more than fair to assume that users who build their own images with such args will not consider any guarantees to be valid? A comment above the variables could emphasize the point, but I can't see it being necessary.

arurke: I just noticed we use `ARG` for `RENODE_VERSION`. Is there a principal difference? We don't have a dependency on any specific version perhaps?

PR #2506, opened 2023-07-10
tests: build stack-check library on arm

pjonsson:
The library requires extra symbols from the linker, so ensure that the linking works.

This increases the size of ccache with ~10Mb, so
bump it 20Mb to give some headroom for the future.


.github/workflows/build.yml
```diff
@@ -149,6 +149,6 @@ jobs:
...
+        docker run --privileged --sysctl net.ipv6.conf.all.disable_ipv6=0 -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_RETENTION_DAYS -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e GITHUB_ACTION_REPOSITORY -e GITHUB_ACTION_REF -e GITHUB_PATH -e GITHUB_ENV -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -e GRADLE_USER_HOME=/home/user/contiki-ng/tools/cooja/.gradle_home -e LOCAL_UID=$(id -u $USER) -e LOCAL_GID=$(id -g $USER) $DOCKER_ARGS -v `pwd`:/home/user/contiki-ng -v $GITHUB_WORKSPACE/.ccache:/home/user/.ccache $DOCKER_IMG bash --login -c "source ../.bash_aliases && ccache --set-config=max_size='160M' && cimake -C tests/??-${{ matrix.test }}; ccache -c"
```
arurke, line 152: I am not versed in Github CI but is it possible to define the max-size as a [environment variable](https://docs.github.com/en/github-ae@latest/actions/learn-github-actions/variables#defining-environment-variables) at the start of the Contiki-NG job, and use the variable in these two places instead? That would make it DRYer and easier to maintain.

pjonsson: I have other changes stacked on top of this, so I would prefer to make that in a separate PR later.

On a technical level, I think it's a good idea, but my experience is that there are often exceptions to the exceptions with these "low code" systems, so I'm not sure it will work even if it is the obvious thing to do for someone who knows programming.

arurke: > I have other changes stacked on top of this, so I would prefer to make that in a separate PR later.

:+1:

> On a technical level, I think it's a good idea, but my experience is that there are often exceptions to the exceptions with these "low code" systems, so I'm not sure it will work even if it is the obvious thing to do for someone who knows programming.

:+1: My experience is quite good in Gitlab CI, and also in Dockerfiles, e.g. for specifying versions of tools, libs. etc.

arurke (approved):

PR #2512, opened 2023-07-10
Doc.: Use platform name cc26x0-cc13x0 instead of srf-cc26xx

arurke:
As in title. Long overdue, I suggest it goes into the release since it is only changes in doc.


doc/platforms/cc26x0-cc13x0.md
```diff
@@ -4,18 +4,18 @@
...
   * Platform code under `arch/platform/cc26x0-cc13x0`
   * CPU code under `arch/cpu/cc26x0-cc13x0`
   * Platform-specific examples under `examples/platform-specific/cc26x0-cc13x0`
+* Legacy name `srf06-cc26xx`: Used in releases v4.0, v4.1 and v4.2. Used in the pre-release `release-4.2` branch. Used in branch `develop` before 24 October 2018.
+  * Platform code under `arch/platform/srf06-cc26xx`
+  * CPU code under `arch/cpu/cc26xx-cc13xx`
+  * Platform-specific examples under `examples/platform-specific/cc26xx`
 
-While reading the guide below, you may have to use a different platform name depending on the version of Contiki-NG that you are using. The name `srf06-cc26xx` will be obsoleted in Release v4.3.
+While reading the guide below, you may have to use a different platform name depending on the version of Contiki-NG that you are using.
```
pjonsson, line 18: I think the text will be easier to read if the documentation is for the current version, and the history of how the port was named is omitted. If someone cares deeply about a 5 year old release, they can read the documentation for that release (which is available on readthedocs).

joakimeriksson (approved):

joakimeriksson:
merging as all relevant tests passed.

doc/platforms/cc26x0-cc13x0.md
arurke: I agree with you on this.

However - I tried now to find doc. for such old releases but it wasn't immediately obvious. We did not have readthedocs like we do today, so those pages are mostly empty. Seems like one would need to navigate the commits in the wiki repo and based on dates find the correct-ish one (since there are no tags). Or am I missing something?

PR #2550, opened 2023-07-24
Centralize clock_time_t definition

pjonsson:
Use a similar construct as rtimer
for defining the size of clock_time_t.


arurke (approved):

pjonsson:
Updated to also include the right `CLOCK_LT` macro for each size.

nfi (approved):

PR #2567, opened 2023-07-29
cooja: fix printf for floats

### LINKED ISSUE #2566: printf of double and Contiki logging does not work in Cooja
All of the following print statements would not print the number in Contiki 4.9 
```
printf("Double f = %f\n", 3.1);
printf("Double e = %e\n", 3.1);
printf("Double g = %e\n", 3.1);
LOG_INFO("Double f = %f\n", 7.1);
// and so on
```

pjonsson:
There is no support for printing floats
in dbg-io, so stop using that.

The issues that f45a8c5327 tried to
avoid are not present on my computer,
which could be explained by 64-bit support
in Cooja, FFI instead of JNI, or using
-Wl,--wrap for wrapping the printf-family
of functions.

Fixes #2566

### DIFF
--- a/arch/platform/cooja/Makefile.cooja
+++ b/arch/platform/cooja/Makefile.cooja
@@ -31,9 +31,7 @@ ifneq ($(MAKECMDGOALS),clean)
   endif
 endif
 
-# Use dbg-io for IO functions like printf()
-MODULES += os/lib/dbg-io
-WRAPPED_FUNS = printf putchar puts snprintf sprintf vsnprintf
+WRAPPED_FUNS = printf putchar puts
 
 CC = gcc
 
@@ -46,7 +44,7 @@ else
   LDFLAGS += -shared -Wl,-zdefs
   LDFLAGS += -Wl,--warn-common
   LDFLAGS += -Wl,-T$(CONTIKI_NG_RELOC_PLATFORM_DIR)/cooja/cooja.ld
-  # Use the printf-family replacement functions in dbg-io.
+  # Use the printf-family replacement functions in cooja-log.c.
   LDFLAGS += $(addprefix -Wl$(COMMA)--wrap$(COMMA), $(WRAPPED_FUNS))
 endif
 
--- a/arch/platform/cooja/dev/rs232.c
+++ b/arch/platform/cooja/dev/rs232.c
@@ -46,7 +46,7 @@ char simSerialReceivingFlag;
 
 static int (* input_handler)(unsigned char) = NULL;
 
-void simlog_char(char c);
+int simlog_char(char c);
 /*-----------------------------------------------------------------------------------*/
 void rs232_init(void) { }
 /*-----------------------------------------------------------------------------------*/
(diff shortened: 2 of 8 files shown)


rahav:
Does this fix Contiki LOG_ statements as well?

pjonsson:
@rahav there are a lot of intricate dependencies within Cooja, it's better that you try this PR on your machine and test the feature you have in mind.

rahav:
This is working well so far. Thank you for doing this. I have long research workloads to run and I'll report here.

rahav:
The only issue I saw thus far is that if an `unsigned char` is `0` then a new line will not be printed when the format specifier is `%c`. I do not think this is critical to fix, unless it is easy.

```
unsigned char device_type = 0;

LOG_INFO("Device type = %c\n", device_type);
LOG_INFO("Second line\n");    

LOG_INFO("Device type = %u\n", device_type);
LOG_INFO("Second line\n");
```

What will printed is
```
18:05.382	ID:1	[INFO: Test      ] Device type = [INFO: Test      ] Second line
18:05.382	ID:1	[INFO: Test      ] Device type = 0
18:05.382	ID:1	[INFO: Test      ] Second line
```

By the way, I have thousands of hours of log data that I can run regression tests with. The only challenge is that Cooja random number generator, although it starts with the same seed with which the data was generated, follows a different sequence. I suspect this is because I'm running Contiki 4.9 on a different computer than I ran Contiki 4.7 (that generated the data). If I'm wrong and there is a way to make the new Cooja follow the same random generation sequence, then I will be able to test quickly and much more.

Let me know what you think.

pjonsson:
Had a look at this, and the lack of newline was a bit surprising to me, but the behavior of printf/LOG_* is the same as asprintf in libc on the platform Cooja is running on.

That isn't perfect, but it is at least easy to explain to someone using Cooja: if you replace `<INSERT STRING HERE>` with the thing you want to print, the following program will show what the output should be in Cooja:
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>

int main(void) {
  char *buf;
  if (asprintf(&buf, <INSERT STRING HERE>) > 0) {
    printf("'%s'\n", buf);
    free(buf);
  }
}
```

It wouldn't be that hard to make `SafeRandom.java` read a file with a sequence of numbers and return those instead, but the mote initialization has been simplified after 4.7, so the order the numbers are consumed is also different, as you mention.

rahav:
After more than 5,000 hours in total, running a handful of Cooja simulations, I can attest that Cooja 4.9 is superior to 4.7. Importantly, **Cooja never crashed** during any of these sessions (this was not one long run, but discrete invocations of Cooja). It was not possible to do an exact diff but the data was compared to regression data generated by Cooja 4.7 and it appears to be correct.

simonduq:
Hi!

rahav:
No worries @simonduq I can now say that I have a Simon number of 1, as in Erdős number.

nfi (approved):

nvt (approved):

PR #2568, opened 2023-07-31
tsch: fix GCC 12/13 warning

pjonsson:
GCC 12/13 will warn on most platforms
since this check is always false. Zoul
does however have a dynamic size, so
put the check under ifdef ZOUL.

### DIFF
--- a/arch/platform/zoul/contiki-conf.h
+++ b/arch/platform/zoul/contiki-conf.h
@@ -72,6 +72,7 @@ uint16_t *radio_tsch_timeslot_timing(void);
 #define RADIO_DELAY_BEFORE_RX     radio_delay_before_rx()
 #define RADIO_DELAY_BEFORE_DETECT radio_delay_before_detect()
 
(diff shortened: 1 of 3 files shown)


simonduq:
Hi!

arurke:
Would it not be difficult for any new or altered platform with dynamic sizing to find/know about this and add its own platform to the ifdef? Although I can't really think of a solution.

pjonsson:
> Would it not be difficult for any new or altered platform with dynamic sizing to find/know about this and add its own platform to the ifdef? Although I can't really think of a solution.

Yeah, I don't have a solution either. We could invert the condition and test for `!x86 && !cooja && !nrf && ..`, but that would still trigger the warning for external ports.

arurke:
How about putting a define, something like `TSCH_CONF_DYNAMIC_TIMESLOT_TEMPLATE`, in the hardware-specific section of tsch-conf.h, that defaults to 0 and is set to 1 in arch/platform/zoul/contiki-conf.h?

joakimeriksson (approved):

PR #2573, opened 2023-08-03
Makefile: add detection of holes in structs

pjonsson:
This will give a compilation error when
the compiler inserts padding for alignment
in structs.

This functionality should be enabled in
the compilation tests, but there are still
numerous holes in the structs in Contiki-NG,
so only add the infrastructure to find the holes
for now.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -181,6 +181,20 @@ endif
 
 CP ?= cp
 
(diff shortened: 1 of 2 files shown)


pjonsson:
Updated to not error out when pahole gives no output.

nvt (approved):

nfi (approved):

PR #2594, opened 2023-08-10
sicslowpan.c: fix -fanalyzer warning

pjonsson:
Add an assert so gcc -fanalyzer understands
that prefix is only null when prefcount is 0.


pjonsson:
Adding an assert adds 200 bytes of text on MSP430, and I'm not completely happy with that.

I think being able to compile with `-fanalyzer` adds significant value though, so we should fix it in some way.

pjonsson:
Fixed the increased size issue by adding the ASSUME macro in #2597. This should be on top of that PR.

nfi (approved):

nfi (approved):

PR #2597, opened 2023-08-13
Add support for static analysis

pjonsson:
GCC can perform static analysis on code.
Add support to the build system for
using that.

Run with:

make ANALYZE=1

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -125,6 +125,8 @@ include $(target_makefile)
 
 ifneq ($(MAKECMDGOALS),clean)
   ifeq ($(CLANG),1)
+    # Clang Static Analyzer ignores -Werror, run through clang-tidy instead.
+    override ANALYZER = 0
     # Use GCC sysroot unless user specified something else.
     ifndef SYSROOT
       SYSROOT := $(shell $(CC) $(CFLAGS) --print-sysroot)
@@ -188,6 +190,15 @@ CFLAGS_DEBUG ?= -ggdb
 
 ifeq ($(DEBUG),1)
   CFLAGS += $(CFLAGS_DEBUG)
+endif
+
+# Default to not use static analysis.
+ANALYZE ?= 0
+ifeq ($(ANALYZE),1)
+  # Define something corresponding to __clang_analyzer__ for GCC as well.
+  CFLAGS += -fanalyzer -D__gcc_analyzer__ $(CFLAGS_FANALYZER)
+  # FIXME: poor -fanalyzer results with GCC 13 according to documentation.
+  #        Add ANALYZE support to CXXFLAGS when GCC supports it better.
 endif
 
 # Configure link-time-optimization (LTO).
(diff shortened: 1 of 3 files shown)


arch/cpu/msp430/Makefile.msp430
```diff
@@ -44,6 +44,8 @@ CONTIKI_SOURCEFILES        += $(CONTIKI_TARGET_SOURCEFILES)
 
 # LTO is available in GCC 4.7.2, but crashes for MSP430.
 NO_LTO_TARGET = 1
+# No -fanalyzer before GCC 10.
+CFLAGS_FANALYZE =
```
arurke, line 48: Is it possible to rather set `ANALYZE` to 0 as opposed to emptying the flags? As is, wouldn't a build for msp430 show `ANALYZE=1` but there would be no static-analysis?

pjonsson: This is addressed, but it's now a platform action skip instead. For some reason, the output when running clang --analyze isn't binary code, it's the report.

pjonsson: Reworked again, Clang Static Analyzer ignores -Werror so that needs to run through clang-tidy. This is now just an addition to CFLAGS for platforms that has a GCC that supports it.

Makefile.include
```diff
@@ -193,6 +195,15 @@ ifeq ($(DEBUG),1)
   CFLAGS += $(CFLAGS_DEBUG)
 endif
 
+# Default to not use static analysis.
+ANALYZE ?= 0
+ifeq ($(ANALYZE),1)
+  # Define something corresponding to __clang_analyzer__ for GCC as well.
+  CFLAGS += -fanalyzer -D__gcc_analyzer__ $(CFLAGS_FANALYZER)
+  # FIXME: poor -fanalyzer results with GCC 13 according to documentation.
```
arurke, line 203: You mean gcc 10 here?

pjonsson: I mean GCC 13, but I'm sure the statement is true for GCC 10 as well.

arurke: Ah, sorry I found what you are referring to now

arurke:
Is the commit adding `ASSUME` intentional?

pjonsson:
> Is the commit adding `ASSUME` intentional?

Yes, I don't think there's much other use for it besides guiding static analysis tools, #2594 that fixes an fanalyzer warning needs it.

arurke (approved):

os/sys/cc.h
```diff
@@ -49,6 +49,16 @@
 
 #ifdef __GNUC__
 
+#ifndef CC_CONF_ASSUME
+  #ifdef __clang__
+    #define ASSUME(c) do { __builtin_assume((c)); } while(0)
+  #elif __GNUC__ >= 13
+    #define ASSUME(c) __attribute__((__assume__(c)))
+  #else
+    #define ASSUME(c) do { if (!(c)) __builtin_unreachable(); } while(0)
```
nfi, line 58: Should these not set `CC_CONF_ASSUME` in the same style as the other definitions in this file?

pjonsson: Fixed.

nfi (approved):

PR #2617, opened 2023-08-18
Code-style: Declare variables as locally as possible

arurke:
The current code-style states variables should always be declared at the start of the function. We have moved passed C89 so this is not required to compile, the style has not been enforced for a long time, and my reading of the community opinion is to declare variables as locally as possible. Update the code-style accordingly.

### DIFF
--- a/doc/project/Code-style.md
+++ b/doc/project/Code-style.md
@@ -134,21 +134,22 @@ int
 code_style_example_function(char c)
 {
   /*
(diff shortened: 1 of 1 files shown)


pjonsson (approved):
Reduced scopes greatly help out static analysis and other bug finding tools too, so I think this is a great change.

nfi (approved):

PR #2621, opened 2023-08-19
autostart: make functions static inline

pjonsson:
There are no in-tree users of autostart_exit,
and a single use of autostart_start (contiki-main.c). Make the functions static inline in the header instead.

This saves 12 bytes of text on the MSP430 tests.


pjonsson:
Removing 42 bytes of text (autostart.o), and adding 10 bytes of text to contiki-main.o somehow makes simple-node overflow the memory by 2 bytes on Z1.

Guessing this is some kind of bin-packing effect with the linker and alignment, will rebase this on latest develop after the other space saving PRs are merged.

os/sys/autostart.h
```diff
@@ -52,7 +54,18 @@ struct process * const autostart_processes[] = {__VA_ARGS__, NULL}
 
 extern struct process * const autostart_processes[];
 
-void autostart_start(struct process * const processes[]);
-void autostart_exit(struct process * const processes[]);
+static inline void autostart_start(struct process * const processes[])
```
nfi, line 57: Please add a new line before `autostart_start`. Same for `autostart_exit`.

pjonsson: Fixed.

nfi (approved):

PR #2625, opened 2023-08-19
timers: change *_expired() to return bool

pjonsson:
If a timer has expired is clearly a boolean
question, so change the return type to reflect
that.

As a bonus, this saves 10-20 bytes of
text on the MSP430 tests.

### DIFF
--- a/os/sys/etimer.h
+++ b/os/sys/etimer.h
@@ -65,6 +65,8 @@
 
 #include "contiki.h"
 
+#include <stdbool.h>
+
 /**
  * A timer.
  *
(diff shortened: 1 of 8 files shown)


os/sys/etimer.h
```diff
@@ -189,12 +189,12 @@ clock_time_t etimer_start_time(struct etimer *et);
...
  * \brief      Check if an event timer has expired.
  * \param et   A pointer to the event timer
- * \return     Non-zero if the timer has expired, zero otherwise.
+ * \return     True if the timer has expired.
  *
  *             This function tests if an event timer has expired and
  *             returns true or false depending on its status.
  */
-int etimer_expired(struct etimer *et);
+bool etimer_expired(struct etimer *et);
```
arurke, line 197: Missing `#include <stdbool.h>` here and in the other files?

pjonsson: Fixed.

nfi (approved):

PR #2634, opened 2023-08-20
CI: track binary size changes over time

pjonsson:
Add a step in the CI that shows the size
difference for each PR.

This requires keeping track of previous
sizes, so also add a step that pushes
sizes.log to a data repository when
merging PRs to develop.

There is a small window where two jobs
can push at the same time. Instead of adding
some complicated locking, just let one of
the jobs fail with the push, but make the job
green.

This requires a new repository to be created,
and an access token with write permissions
for that repository. The access token also needs
to be added as a secret to this CI with the name
`DATA_PUSH_TOKEN`.

Here's a screenshot of what it looks like when saving the output from "date":

image


pjonsson:
Use the contiki-ng.github.io repository to store data.

This PR needs a key with read/write permissions on repository on contiki-ng.github.io, and the key put in a secret called `DATA_PUSH_TOKEN` here.

PR #2639, opened 2023-08-25
SHA-256: Disable unnecessary interrupt

kkrentz:
This PR disables the `DMA_IN_DONE` when computing SHA-256 hashes. We only need the `RESULT_AV`, which is issued when the result is available.


kkrentz:
A quote from Section 22.2.2.3 of the user guide:

> Another interrupt is activated when only the input DMA is finished and is intended for debugging purpose.

nfi (approved):

PR #2655, opened 2023-09-05
cooja: allow disabling devices

pjonsson:
This makes it possible to build ContikiMotes
without code that handles pir/vib sensors,
rs232, and so on.

This mechanism is not intended for manual use,
Cooja will call the build system with the right
parameters based on configured mote interfaces
for the mote type.


alexrayne:
the penalry of vib, pir, button, leds, radio - is negligible. it takes a few bytes.
your 8k penalty goes from rs232, cooja-log, eeprom, cfs. And as for them - rs232 + cooja-log cant't be disabled for most cases - since it gives log messages for your simulation or interconnect with applications.
So, most reasonable disable of eeprom, cfs.  All other disabling stuff, imho, useless trash

arurke (approved):

PR #2666, opened 2023-09-09
Cooja: use $(MAKE) in simulation files

pjonsson:
This will expand to the make configured either by the user or the make configured for the platform.

### DIFF
--- a/examples/6tisch/6p-packet/6p-packet.csc
+++ b/examples/6tisch/6p-packet/6p-packet.csc
@@ -19,7 +19,7 @@
       <identifier>mtype639</identifier>
       <description>6P node</description>
       <source>[CONTIKI_DIR]/examples/6tisch/6p-packet/sixp-node.c</source>
-      <commands>make -j$(CPUS) sixp-node.cooja TARGET=cooja</commands>
+      <commands>$(MAKE) -j$(CPUS) sixp-node.cooja TARGET=cooja</commands>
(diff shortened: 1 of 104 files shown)


nfi (approved):

nvt:
From what I recall, we have not had this dependency on MAKE being set before, and I do not see it mentioned in the documentation. If it is not set, we will with this change get this type of error:

org.contikios.cooja.MoteType$MoteTypeCreationException: Compilation error: Cannot run program "$(MAKE)" (in directory "/home/nvt/contiki-ng/examples/rpl-udp"): error=2, No such file or directory

pjonsson:
> From what I recall, we have not had this dependency on MAKE being set before, and I do not see it mentioned in the documentation. If it is not set, we will with this change get this type of error:
> 
> org.contikios.cooja.MoteType$MoteTypeCreationException: Compilation error: Cannot run program "$(MAKE)" (in directory "/home/nvt/contiki-ng/examples/rpl-udp"): error=2, No such file or directory

Is that macOS or Linux? What does PATH_MAKE in Setings->External tools paths show for you?

nvt:
Ubuntu Linux 23.04, native toolchain installation.

PATH_MAKE=make in Settings->External tools

pjonsson:
It is Cooja that should do the expansion, does `git submodule update` make the problem disappear?

nvt:
Thanks, the example builds now with the updated submodule.

Issue #1102, opened 2019-10-21
Feature request: OpenMote-B AT86RF215 support


tinstructor:
Since #1080 now added support for the OpenMote-B node, I think it's a good idea to start working on  support for the AT86RF215 radio chip. Recently, some [interesting work](https://www.mdpi.com/2079-9292/8/9/1045/pdf) has been done with the end-goal of an even more robust TSCH implementation based on OFDM. OpenWSN and RIOT already include drivers or are close to driver adoption. Maybe it is possible to start from the OpenWSN or [RIOT driver](https://github.com/RIOT-OS/RIOT/pull/12128).

edgaralb:
Hello everyone, please help me with the openmote-b drivers for the at86rf215 radio chip for Contiki-ng

PR #2672, opened 2023-09-12
Support for AT86RF215 radio on OpenMoteB board

### LINKED ISSUE #1215: Driver support for openmote-b - at86rf215
Hello everyone. 
Please, help me with the openmote-b drivers for the at86rf215 chip for contiki-ng
Thanks

9morano:
## AT86RF215
The PR adds support for the [AT86RF215](https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-42415-WIRELESS-AT86RF215_Datasheet.pdf) radio available on OpenMoteB devices. 
Support for the radio was previously requested in #1102 .
The radio contains 2 RF cores (2.4 GHz and sub GHz) and supports the following PHY modes: MR-FSK, MR-O-QPSK, and MR-OFDM . 

## Driver
The driver was designed to provide support for only one core at the time (only 2.4 GHz was tested).
The driver currently implements the legacy O-QPSK, but if there is a need, I will be happy to try to help implement other modulations.
The driver is mainly designed to support TSCH MAC mode, but CSMA works as well.

### Usage

`#define OPENMOTEB_CONF_USE_ATMEL_RADIO     1`


arch/platform/openmote/dev/antenna.c
```diff
@@ -73,6 +79,20 @@ antenna_internal(void)
   GPIO_WRITE_PIN(ANTENNA_BSP_RADIO_BASE, ANTENNA_BSP_RADIO_INT, 1);
 }
 /*---------------------------------------------------------------------------*/
+void
+antenna_select_atmel(void)
```
pjonsson, line 83: What is the difference between these two functions?

arch/platform/openmote/openmote-b/Makefile.openmote-b
```diff
@@ -0,0 +1,2 @@
+CONTIKI_TARGET_SOURCEFILES += at86rf215.c at86rf215-arch.c
```
pjonsson, line 1: Isn't `at86rf215.c` part of the module added in the next line?

examples/platform-specific/openmote-b/hello-world/Makefile
```diff
@@ -0,0 +1,8 @@
+CONTIKI_PROJECT = hello-world
+all: $(CONTIKI_PROJECT)
+
+MAKE_MAC = MAKE_MAC_TSCH
+MODULES += os/services/shell
+
+CONTIKI = ../../../..
```
pjonsson, line 7: No need to define CONTIKI in Makefiles any longer.

arch/platform/openmote/dev/antenna.c
9morano: Both OpenMote-CC2538 and OpenMote-B boards contain an antenna switch.

- On OpenMote-CC2538, the switch is used to connect the output of the radio to either the internal (chip antenna) or external antenna (SMA connector).

- For OpenMote-B, on the other hand, the switch selects which output of the radio goes to the SMA connector - either Atmel's AT86RF215 radio or TI CC2538 radio.

On both boards, the RF switch is connected to the same GPIO pins - I have created new functions (which basically do the same thing as the original ones) just to avoid any misunderstandings about function names.

At this point I am not sure if I should create a separate file for each board, or if we should use the originally defined functions and just add some documentation describing the differences between the boards?

examples/platform-specific/openmote-b/hello-world/Makefile
9morano: I didn't know that - just took an example from other examples, all of which contain the definition of CONTIKI.

arch/dev/radio/at86rf215/at86rf215.h
```diff
@@ -0,0 +1,128 @@
...
+    //uint8_t len;
+    //uint16_t crc;
+    uint8_t lqi;
+    int8_t rssi;
+	rtimer_clock_t timestamp;
+} at86rf215_rx_frame_t;
+
+/* Basic radio configuration */
+typedef struct {
+    uint16_t address;
```
pjonsson, line 102: Typedef something for the type of address, otherwise things will be hard to update when the address gets more bits in the future.

arch/dev/radio/at86rf215/at86rf215.h
```diff
@@ -0,0 +1,128 @@
...
+} at86rf215_bbc_irq_t;
+
+/* RX frame struct */
+typedef struct {
+    //uint8_t content[AT86RF215_MAX_PAYLOAD_SIZE];
+    //uint8_t len;
+    //uint16_t crc;
+    uint8_t lqi;
+    int8_t rssi;
+	rtimer_clock_t timestamp;
```
pjonsson, line 97: Unless something in the code mandates the order of the members, put timestamp first to avoid padding between struct elements.

arch/dev/radio/at86rf215/at86rf215.h
```diff
@@ -0,0 +1,128 @@
...
+static radio_result_t at86rf215_get_value(radio_param_t param, radio_value_t *value);
+static radio_result_t at86rf215_set_value(radio_param_t param, radio_value_t value);
+static radio_result_t at86rf215_get_object(radio_param_t param, void *dest, size_t size);
+static radio_result_t at86rf215_set_object(radio_param_t param, const void *src, size_t size);
+
+/*---------------------------------------------------------------------------*/
+/** The NETSTACK data structure for the AT86RF215 driver */
+extern const struct radio_driver at86rf215_driver;
+
+#endif /* AT86RF215_H_ */
```
pjonsson, line 128: I think there's a missing newline at the end of several files, look for the red ring and fix those.

arch/platform/openmote/dev/antenna.c
pjonsson: Initially I thought they set/clear the same pins, but they set/clear the opposite pins, so ignore the comment.

arch/platform/openmote/dev/at86rf215-arch.c
```diff
@@ -0,0 +1,169 @@
...
+}
+/*---------------------------------------------------------------------------*/
+uint8_t 
+at86rf215_arch_spi_txrx(uint8_t b)
+{
+    SPIX_WAITFORTxREADY(AT86RF215_SPI_INSTANCE);
+    SPIX_BUF(AT86RF215_SPI_INSTANCE) = b;
+    SPIX_WAITFOREOTx(AT86RF215_SPI_INSTANCE);
+    SPIX_WAITFOREORx(AT86RF215_SPI_INSTANCE);
+    b = SPIX_BUF(AT86RF215_SPI_INSTANCE);
```
pjonsson, line 114: No need to assign to b, just `return SPIX_BUF(..);` directly instead.

arch/platform/openmote/dev/at86rf215-arch.c
```diff
@@ -0,0 +1,169 @@
...
+    GPIO_SET_OUTPUT(AT86RF215_PWR_PORT_BASE , AT86RF215_PWR_PIN_MASK );
+    
+    /* Initialize reset pin */
+    GPIO_SOFTWARE_CONTROL(AT86RF215_RSTN_PORT_BASE ,AT86RF215_RSTN_PIN_MASK );
+    GPIO_SET_OUTPUT(AT86RF215_RSTN_PORT_BASE ,AT86RF215_RSTN_PIN_MASK );
+
+    /* Turn the radio off */
+    GPIO_CLR_PIN(AT86RF215_PWR_PORT_BASE , AT86RF215_PWR_PIN_MASK );
+    GPIO_CLR_PIN(AT86RF215_RSTN_PORT_BASE ,AT86RF215_RSTN_PIN_MASK );
+    for(uint16_t delay=0; delay<0xA2C2; delay++);
```
pjonsson, line 132: The compiler might optimize these away, there should be some delay function to call if that is what is required.

arch/platform/openmote/dev/antenna.c
9morano: I will update (rebase) this PR when the #2684 is accepted

arurke:
Thanks @9morano! A few comments:
* Add building with the necessary defines in `tests/02-compile-arm-ports/Makefile` to get some test coverage?
* It seems to me the example is quite redundant as it does not showcase anything new besides enabling the radio. Assuming there is nothing special with this radio that needs an example I would suggest you remove the example and update [the docs](https://docs.contiki-ng.org/en/develop/doc/platforms/openmote-cc2538.html) to reflect your changes and explain how to enable the radio.
* Remove unrelated changes (is the changes in platform.c in related? changes of UART and SPI in board.h?)
* You seem to make the sub-ghz radio the default now - I am not familiar with the platform/board but I would advise against changing the default behavior without a thorough rationale

9morano:
@arurke thank you for the comments!

- I have added a compilation test for OpenMote-B platform with AT86RF215 transceiver.
- I agree with the comment. I have removed the redundant example application and updated the documentation on OpenMote platforms (both OpenMote-CC2538 and OpenMote-B boards).
- Unrelated changes in the board.h were removed (changes to the UART). I would leave the changes to the SPI defines, as they emphasize the difference between SPI0 and SPI1. Regarding the platfrom.c, the `board_init()` function was moved into `platform_init_stage_two()` so that the radio's signal path (RF switch) is configured before the radio is initialized.
- The OpenMote-B board provides two radio chips: Texas Instruments' CC2538 and Atmel's AT86RF215. By default, the board operates with the CC2538 radio using the 2.4 GHz RF interface. If you set `OPENMOTEB_CONF_USE_ATMEL_RADIO` to 1, the board will use the AT86RF215 radio with new drivers, otherwise the board will continue using the old implementation. Although the AT86RF215 radio supports both 2.4 GHz and sub-GHz RF interfaces, the current driver only supports the 2.4 GHz interface.

doc/platforms/openmote.md
```diff
@@ -26,28 +20,42 @@ It has the following key features:
...
-## Requirements
+The OpenMote platform supports two boards:
+* The original OpenMote (`openmote-cc2538`),
+* The new OpenMote revision B (`openmote-b`). This board features two radios: TI's CC2538 and Atmel's AT86RF215.
+
+## Port Features
 
-To start using Contiki-NG with the OpenMote-CC2538, the following is required:
+The port is organized as follows:
+* the derivers for CC2538 CPU are located in the `arch/cpu/cc2538` folder,
```
arurke, line 30: typo - derivers

doc/platforms/openmote.md
```diff
@@ -26,28 +20,42 @@ It has the following key features:
...
+* the OpenMote port is located in the `arch/platform/openmote` folder where:
+  * `dev` contains drivers for various sensors,
+  * `openmote-cc2538` includes configuration for the OpenMote-CC2538 board,
+  * `openmote-b` includes configuration for the OpenMote-B board.
+* the driver for AT86RF215 radio (available only on OpenMote B) is located in `arch/dev/radio/at86rf215`.
```
arurke, line 35: nitpicking - OpenMote-B ?

doc/platforms/openmote.md
```diff
@@ -56,34 +64,42 @@ Instructions to flash for different OS are given below.
     * Nodes can also be programmed via the serial boot loader in the cc2538. In `tools/cc2538-bsl/` you can find `cc2538-bsl.py` python script, which can download firmware to your node via a serial connection. If you use this option you just need to make sure you have a working version of python installed. You can read the README in the same directory for more info.
 
 * On Linux:
-    * Nodes can be programmed with TI's [UniFlash] tool. With UniFlash, use the file with `.elf` extension.
-    * Nodes can also be programmed via the serial boot loader in the cc2538. No extra software needs to be installed.
+    * Nodes can be programmed with TI's UniFlash tool. With UniFlash, use the file with `.elf` extension.
```
arurke, line 67: Is there an .elf extension output from the build? Is it not a .openmote or something like that?

arch/platform/openmote/openmote-b/board.h
```diff
@@ -181,15 +182,62 @@
...
+/** @} */
+/*---------------------------------------------------------------------------*/
+/**
+ * \name Radio configuration - use either AT86RF215 or CC2538
+ *
+ * OpenMoteB can use either the AT86RF215 or the CC2538 radio.
+ * @{
+ */
+#ifndef OPENMOTEB_CONF_USE_ATMEL_RADIO
+#define OPENMOTEB_USE_ATMEL_RADIO               1
```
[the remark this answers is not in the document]
9morano: I overlooked that, thanks!

arurke (approved):

joakimeriksson:
We intend to merge this soon - can you do a rebase? @9morano

joakimeriksson:
One final thing before we merge. Could you @9morano remove the CMSIS commit (the last one) as we already have a newer version now in develop?

joakimeriksson:
Thanks for quick responses @9morano!

PR #2675, opened 2023-09-14
tests: remove unbuildable tests

pjonsson:
Remove most tests that are not buildable
on the platform, but keep one to ensure
the functionality doesn't break in the build
system.

### DIFF
--- a/tests/02-compile-arm-ports/Makefile
+++ b/tests/02-compile-arm-ports/Makefile
@@ -107,8 +107,6 @@ platform-specific/cc26x0-cc13x0/cc26x0-web-demo/cc26x0-cc13x0:BOARD=sensortag/cc
 platform-specific/cc26x0-cc13x0/base-demo/cc26x0-cc13x0:BOARD=sensortag/cc2650 \
 platform-specific/cc26x0-cc13x0/base-demo/simplelink:BOARD=sensortag/cc1350 \
 platform-specific/cc26x0-cc13x0/base-demo/simplelink:BOARD=sensortag/cc1352r1 \
-platform-specific/nrf/start-network-core/nrf:BOARD=nrf52840/dk \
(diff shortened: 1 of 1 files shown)


nfi (approved):
I do not understand the comment but I think the change looks good. The nRF52840 does not have a network core.

tests/02-compile-arm-ports/Makefile
```diff
@@ -107,8 +107,6 @@ platform-specific/cc26x0-cc13x0/cc26x0-web-demo/cc26x0-cc13x0:BOARD=sensortag/cc
 platform-specific/cc26x0-cc13x0/base-demo/cc26x0-cc13x0:BOARD=sensortag/cc2650 \
 platform-specific/cc26x0-cc13x0/base-demo/simplelink:BOARD=sensortag/cc1350 \
 platform-specific/cc26x0-cc13x0/base-demo/simplelink:BOARD=sensortag/cc1352r1 \
-platform-specific/nrf/start-network-core/nrf:BOARD=nrf52840/dk \
-platform-specific/nrf/start-network-core/nrf:BOARD=nrf52840/dongle \
 platform-specific/nrf/start-network-core/nrf:BOARD=nrf5340/dk/application \
 platform-specific/nrf/start-network-core/nrf:BOARD=nrf5340/dk/network \
```
nfi, line 111: Line 113 should also be removed. The network core can not start itself.

pjonsson: Maybe this is the comment part you don't understand, removing that line removes all the testing of `TARGET_ONLY`/`BOARD_ONLY` functionality.

nfi: Yes, I did not understand this was the intended test. A comment about this would be nice but probably not easy to add close to the relevant lines.

pjonsson: Yes, a comment will end up ~100 lines earlier.

PR #2652, opened 2023-09-03
Update coap-constants.h  to use iana registered ids

mofosyne:
Porting this PR https://github.com/contiki-os/contiki/pull/2475/files

PR #2680, opened 2023-09-16
Autosynced iana coap content format list

mofosyne:
Continued from this closed PR https://github.com/contiki-ng/contiki-ng/pull/2652

Thanks to @nvt suggestion, I decided to try and give a shot at a python tool that would grab the csv values directly from the IANA CoAP standard authority and use that as a basis for autogenerating the ID for this enum list.

This tool is also smart enough to not rename the enum names if it's already defined in the interest of not breaking other existing code in use in the field.

There may be further changes needed to better fit your workflow, so maybe let me know if this needs to be integrated in some CI/CD actions to be triggered periodically to synchronise this enum list.


os/net/app-layer/coap/coap-constants.h
```diff
@@ -145,29 +145,140 @@ typedef enum {
...
+  /* application/voucher-cose+cbor; Ref: [draft-ietf-anima-constrained-voucher-20] */
+  APPLICATION_VOUCHER_COSE_CBOR = 836,
+  /* application/vnd.ocf+cbor; Ref: [Michael_Koster] */
+  APPLICATION_VND_OCF_CBOR = 10000,
+  /* application/oscore; Ref: [RFC8613] */
+  APPLICATION_OSCORE = 10001,
+  /* application/javascript; Ref: [RFC4329] */
+  APPLICATION_JAVASCRIPT = 10002,
+  /* application/json; deflate; Ref: [RFC8259][RFC9110, Section 8.4.1.2] */
+  APPLICATION_JSON = 11050,
```
pjonsson, line 280: I haven't looked at the script, but the CI logs show it does not generate different names for `application/json` and `application/json; deflate` and a few others.

mofosyne: Yeah good catch, looks like content_coding needs to be taken into account as well not just content_type.

Just pushed my change to update the gen tool and regenerated the header

mofosyne:
Is there other fields that should be auto synced? Also fixed the duplicate key issue.

nvt:
Thanks for the update! Some minor comments:

1. Please add a copyright header to the script, preferably with a 3-clause BSD license as in LICENSE.md in the top directory of the repository.
2. The commit list could be squashed a bit -- perhaps into one commit for the header file and one for the script.
3. The cached CSV file should probably not go into the repository, but just be created on the user's local filesystem.

mofosyne:
Okay I'm happy with what I got for https://github.com/mofosyne/iana-headers so decided to copy it over to this project and adapt the generator script to match as much as I can contiki-ng naming conventions.

-     Please add a copyright header to the script, preferably with a 3-clause BSD license as in LICENSE.md in the top directory of the repository.
    - Code generator now updated to be MIT licensed, header updated to reflect this change
-     The commit list could be squashed a bit -- perhaps into one commit for the header file and one for the script.
   - Commit list squashed. First commit is the unchanged generator script. Second is the changes needed to suit contiki-ng. Last commit is the generated output. 
-     The cached CSV file should probably not go into the repository, but just be created on the user's local filesystem.
    - git ignore file updated to ignore csv files now 
-     A requests.exceptions.ConnectionError was raised sometimes when testing, but it is not handled in the script.
    -  handling for such exception should now be handled

If your project is okay with breaking changes to the header files, you could see if https://github.com/mofosyne/iana-headers/blob/main/coap/c/coap-constants.h the full generator output would be workable for your project.

mofosyne:
Had a look and saw that these constants is not directly present in the code, but is actually used by examples instead. Thus we could support my naming convention https://github.com/mofosyne/iana-headers by perhaps making a new file call iana-coap-constant.h in addition to the original coap-constant.h

nvt:
Yes, I think that it would be good to have backward compatibility with the content format names. One possibility is to add the new values with the new `COAP_CONTENT_FORMAT` prefix (or something shorter), and keep some duplicate values with the old names such as `TEXT_PLAIN`.

Furthermore, there is an extra commit (c20b12cd) that could be removed from this PR.

mofosyne:
Good catch on "Furthermore, there is an extra commit (https://github.com/contiki-ng/contiki-ng/commit/c20b12cd2db707d6c07918f8b26cfeb52cf298ca) that could be removed from this PR." now cleaned up.

And yes my  generator already keeps already defined enums. And yes it already will append COAP_CONTENT_FORMAT

nvt:
Thanks for the update, but I meant that both the old name and the new name should be set to the same value in the enum. For example like the following:

// text/plain; charset=utf-8; Ref: [RFC2046][RFC3676][RFC5147]
COAP_CONTENT_FORMAT_TEXT_PLAIN_UTF_8 = 0,
TEXT_PLAIN = 0, // identifier kept for backwards compatibility

mofosyne:
@nvt okay change made to auto generator to support depreciated enums. Also enabled enum autogen support for `coap_option_t`.

Was wondering if I should also enable coap_code_t autogeneration, but it seems like you split the coap codepoints into `coap_method_t` and `coap_status_t`, but missing signalling coap code. Decided to leave it alone in case there is technical reason for the separation (and the omission of signalling code).

Also disabled option number generation as it's a bit more complicated and not 100% sure on the best representation for it. e.g. `coap_code_7_01_option_number_t`, `coap_code_7_02_option_number_t` and if I should also create a union type that could include all the above subtypes...

mofosyne:
By the way, would like to know about the tool ergonomics. I've adjusted the script to always operate in context to script directory. So users only calls the script directly each time, no arguments used. If you need adjusting to improve modularity, let me know. But aiming to keep things simple

mofosyne:
rebased against latest dev branch.

Also noticed a difference between my naming approach and some of your close but not matching existing naming scheme. I was thining of adding `_PLUS_` as i am not sure it can be excluded, since it has a specific meaning that this is encoded as json.

```c
  // application/td+json; Ref: ["Web of Things (WoT) Thing Description 1.1", April 2022]
  COAP_CONTENT_FORMAT_APPLICATION_TD_AS_JSON = 432,
  COAP_CONTENT_FORMAT_APPLICATION_TD_JSON = 432 /* depreciated but identifier kept for backwards compatibility */,
```

mofosyne:
Any idea why this is failing? https://github.com/contiki-ng/contiki-ng/actions/runs/6451700925/job/17512682698?pr=2680#step:13:3940

tools/iana/makefile
```diff
@@ -0,0 +1,5 @@
+
+.PHONY: generate
+generate:
+	@echo "Generating Headers"
+	cd c; ./c_header_coap.py
```
pjonsson, line 5: This ignores failures from cd, change to `cd c && ./c_header_coap.py`.

tools/iana/makefile
```diff
@@ -0,0 +1,5 @@
+
```
pjonsson, line 1: All other makefiles in the tree are named `Makefile`, I think that makes sense for this too.

tools/iana/makefile
mofosyne: fixed. Also added more to makefile to use venv now for better python package isolation

mofosyne:
@arurke tomllib at the time wasn't available directly in certain python versions, but it seems fixed and is now located in as a different package name now. I've made the adjustment and also have switched to using venv now so hopefully that resolves the issue somewhat.

I'm a bit iffy about automating the updating of the header, as my approach is to use a heuristic to generate the name based on the iana description. So there will always be a chance something strange may happen, so you would periodically have to run this generator and inspect the changes to make sure nothing strange has happened. If there is something strange then it might be something my heuristics description to enum name generator didn't catch in `iana_coap_content_formats_c_enum_name_generate()` and you might need to slightly adjust it as needed.

But overall it should reduce the workload of maintaining the list quite a bit.

I'll also try to keep track in https://github.com/mofosyne/iana-headers so feel free to ping me if you hit something strange.

mofosyne:
Okay simplified the makefile so you can just run `make` now and it will auto install venv and then run the generator script.

This should make it more portable for your docker container. Can you confirm @arurke ?

mofosyne:
Unsure what's the issue with `CI / simulation-base/ubuntu-latest (pull_request) Failing after 6m` where it said it failed at 

```
ERROR [main] [Cooja.java:1186] - Failed tests:
/home/user/contiki-ng/tests/07-simulation-base/21-cooja-rpl-tsch-security.csc seed: 1
```

I have not touched any logic in CBOR.

pjonsson:
The 21 simulation test is intermittently failing, unlikely to be caused by these changes.

mofosyne:
Just did a `git rebase -i develop` on latest develop branch to kick off a new check and hopefully see if any CI drama occurs.

If not then, hopefully this looks all good to you all! Now it's just a `make` command to trigger a resync whenever you feel like it.

arurke:
Hi @mofosyne, I am the latest docker container and when running `make` I am getting:
``` 
user@1434ebf10e75:~/contiki-ng/tools/iana$ make
python3 -m venv venv
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.10-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: /home/user/contiki-ng/tools/iana/venv/bin/python3

make: *** [Makefile:7: venv] Error 1
```

mofosyne:
@arurke which docker are you referring to? All the checks previously passed so I'm rebasing so I can retrigger the CI checks in case there are changes to the docker that is making fail.

Did you also try the `apt install python3.10-venv` it was suggesting? I think it's assuming support for python venv.

arurke:
> which docker are you referring to?

The "official" contiki-ng docker container, see https://docs.contiki-ng.org/en/develop/doc/getting-started/Docker.html?highlight=docker#docker . Ideally, it should work in there without additional changes needed. This is also used in the CI. You mention that you will retrigger the CI checks, but this PR does not add any tests? That might be a good idea to add though (to simply run it and make sure there are no errors), but I am not certain how testing a tool is best done within the framework.

mofosyne:
@arurke I've added a simple build check for the python script and also noticed that the dockerfile does not have `python3-venv` installed. I've included that in now.

FYI, I'm not really sure why you would want to be running the IANA generator in a docker image if it's part of a build process. In my opinion you don't want to always be running the IANA generator on every build. Instead I'm seeing it more as a maintainers tool in allowing for them to more rapidly add new entries to the format list header files.

This is because ultimately the script is applying a heuristic to an external IANA database we don't control and thus there is always a slight chance that my 'auto' naming function will not generate a sensible output. So for accountability we would still need to manually do a PR every time we want to add a new 'canonical' entry into the repository branch.

arurke:
I re-ran the job to fix unrelated error.

----

My thinking is that users (and maintainers) would expect a tool living in the contiki-ng repo. to work within the dev-container. I believe this is the case for all our other tools. Now I could see counter-arguments, e.g. if the tool require bloating the container, but in this case it seems straight forward? (I have not tested yet)

> I'm not really sure why you would want to be running the IANA generator in a docker image if it's part of a build process. In my opinion you don't want to always be running the IANA generator on every build. Instead I'm seeing it more as a maintainers tool in allowing for them to more rapidly add new entries to the format list header files. (...) So for accountability we would still need to manually do a PR every time we want to add a new 'canonical' entry into the repository branch.

Regarding running on every build, I am assuming you mean including it in the CI. My thought was to add test coverage (make sure the script still works, avoid regressions, etc.), not to actually run the tool to update the IANA list in the repo. And I believe this is exactly what your tests are doing now. However, again I could see issues, e.g. prolonging the CI duration since it depends on downloading from third-party website, and also, for the same reason, it may break outside of our control which will cause unrelated pull requests to fail in the CI. Perhaps @pjonsson has opinions here.

tools/iana/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.csv
+venv/
```
arurke, line 2: Missing lineshift?

tools/iana/.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+max-line-length = 125
+ignore = NP100, E1, E2, E3, E4, E5,
+exclude = venv/*
```
arurke, line 4: Missing lineshift?

tools/iana/.gitignore
mofosyne: fixed

tools/iana/.flake8
mofosyne: fixed

arurke:
Thanks for the fixes. Looks good to me now, but the separate workflow in CI is not ideal since we lose coverage (does not test the tool works on the dev-container). As mentioned, I don't know the best way to test using the existing test-framework, but perhaps @pjonsson or others have inputs here.

mofosyne:
@pjonsson @arurke just a friendly followup on the cross review of this PR

arurke:
Sorry @mofosyne I have been away for some time. I you find the motivation to look into if this can be tested using the regular contiki-ng container I am happy to give it a checkmark from my side at least

mofosyne:
@arurke Not very familiar with docker and how to trigger a test, but gave a shot anyway. I did however update the python code to generate the headerfies eventually in a a more consistent manner (and resynced the headerfiles with the new iana entries).

mofosyne:
Cannot figure how to adjust docker :/ anyway added a small bugfix

arurke:
Hi @mofosyne, sorry for the lateness. I had a go at what I meant with the testing, could you have a look at https://github.com/contiki-ng/contiki-ng/pull/3024 and if you like it feel free to cherry-pick? I think that approach is more consistent with the other test, although I am not 100 % on the structure I chose.

Also, is `depreciated` several places a typo or intentional? `deprecated`?

mofosyne:
Looks good. I'll try and get your changes in soon.

Also took the time to rejig in my repo how the generator generates `coap_code_t` and `coap_code_*_option_number_t` to use actual names rather than code numbers in the enum names. I noticed I've disabled it before because I wasn't sure if you would like it, but I think it looks decent enough that I would like you to reconsider.

https://github.com/mofosyne/iana-headers/blob/main/c/src/coap_constants.h

Anyway will give this a shot when I get some extra time to move my changes from https://github.com/mofosyne/iana-headers/ to this PR

mofosyne:
@arurke decided to cherry pick your commits then resync my latest code changes to this commit to fix the typo you spotted and to add the latest changes for generating coap_code_t with more intuitively named enums. Those are still disabled, but can enable it if you like the idea. You can check the link below for an idea of what the full version may look like if you want extra enum for coap code and signaling methods.

https://github.com/mofosyne/iana-headers/blob/main/c/src/coap_constants.h

arurke:
@mofosyne Looks good!

1. Regarding the CI, I notice now that the [previous run failed](https://github.com/contiki-ng/contiki-ng/actions/runs/14733639609/job/41354054951) from what looks like an issue at www.iana.org. I would like to get @pjonsson thoughts about this as it would mean our jobs would depend on third-party uptimes. Perhaps we are better off without a test.
2. Regarding the formatting you mention - I am not following, there is no `coap_code_t` in our repo? And changing e.g. `coap_method_t` would break a lot of things so I don't think that is a good idea.
3. I notice now that you have this tool in a separate repo(?). Did you consider rather adding it as a submodule here?
4. Could you please squash and clean up the commit history?

mofosyne:
1. Okay, happy to hear what he thinks.
2. That's fair enough. Yes these are not present in the current codebase, just proposing extra enums in case it be of use.
3. Yeah I am maintaining it previously on my account, but made a custom copy here to minimize complexity/dependencies. But since you suggested it, I've moved it to https://github.com/iana-tools and updated this PR to use it as a submodule.
4. Alright, will do later if you like the direction I'm going so far.

mofosyne:
> Could you please squash and clean up the commit history?

Done

> I notice now that you have this tool in a separate repo(?). Did you consider rather adding it as a submodule here?

Done

----

@arurke also added you and @pjonsson to the iana-tools github group, so that you can fix any issues if you spot anything in the future.

arurke:
Thanks @mofosyne! ✔️ and 👍  from me now.

Regarding the repo., I am thinking we should make a fork of your repo. in the contiki-ng organization. I can do that as soon as another reviewer OKs this PR (our policies requires two).

mofosyne:
@arurke no probs. Looking forward to what you feel would make this work out in the long run :)

arurke:
@mofosyne We just discussed this PR, and would like to avoid running the tests for every run since it relies on the IANA website. Are you able to modify so that the test is executed only on a release or similar?

mofosyne:
> @mofosyne We just discussed this PR, and would like to avoid running the tests for every run since it relies on the IANA website. Are you able to modify so that the test is executed only on a release or similar?

Something like "[ci: iana-tools test should be for release PRs only](https://github.com/contiki-ng/contiki-ng/pull/2680/commits/f1fadfe34fb81809c897ceab8a73358d6f80ef38)" ?

tools/iana-tools/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.csv
+venv/
```
arurke, line 2: Can you add a lineshift here

arurke:
Thanks. Your new commit probably does what I asked, but the solution adds a lot of duplication, and I am not versed enough in Github actions to address this.

I want to propose a compromise that we keep the test as-is, but remove the execution from the CI altogether. When a future PR updates the tool, it seems fair to assume it will not be merged before the updated tool is executed (which is what the test does). Conversely, I see few possibilities for changes in other parts of contiki-ng that can break the tool, with the exception of moving/renaming coap files. Since the auto-generated files are clearly marked as such, this should be picked up. What do you think @joakimeriksson ?

joakimeriksson:
Let's skip the release-test yaml file for now. We can for now run that manually when the iana file changes and we update the coap-headers. @mofosyne

arurke:
Thank you for your patience @mofosyne ! We want to get this is in for the next release in the upcoming days, could you please:
1. Move the submodule to https://github.com/contiki-ng/iana-headers, which is our fork of your repository
2. Remove testing from the CI altogether (but keep the tests themselves)
3. Squash the last commit containing the minor .gitignore change

.gitmodules
```diff
@@ -43,3 +43,6 @@
 [submodule "os/net/security/mbedtls"]
 	path = os/net/security/mbedtls
 	url = https://github.com/Mbed-TLS/mbedtls.git
+[submodule "tools/iana-tools/iana-headers"]
+	path = tools/iana-tools/iana-headers
+	url = git@github.com:contiki-ng/iana-headers.git
```
arurke, line 48: @mofosyne Can you use https for conformity with all the other submodules?

tools/iana-tools/.gitignore
arurke: Seems there is no lineshift here now

PR #2694, opened 2023-09-19
rtimer: enable optimizing return value check

pjonsson:
The current function always returns
the same value, so make that part
static inline in the header.

This preserves the interface, so future
changes can extend the function with
more return values.

This saves 5 bytes on MSP430.


kkrentz:
This seems to address a similar issue like #556.

pjonsson:
#556 is better, closing this.

PR #2715, opened 2023-09-22
lpm: use RTIMER_CLOCK_DIFF to allow for wrap-arounds

kkrentz:
On CC2538-based platforms, `rtimer` timestamps wrap around after about 36 hours. In most places, `RTIMER_CLOCK_DIFF` is already being used to allow for wrap-arounds when computing durations. This PR fixes an unprotected occurrence in `lpm.c`.

### DIFF
--- a/arch/cpu/cc2538/lpm.c
+++ b/arch/cpu/cc2538/lpm.c
@@ -248,7 +248,7 @@ lpm_enter()
    * Choose the most suitable PM based on anticipated deep sleep duration
    */
   lpm_exit_time = rtimer_arch_next_trigger();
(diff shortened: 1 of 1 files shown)


pjonsson:
@kkrentz Was this deliberately closed? The description makes it sound pretty important.

kkrentz:
I suppose that there is no problem since a future time `t2` is subtracted by the current time `t1`. After executing some examples, it seems that `ABS(RTIMER_CLOCK_DIFF(t2, t1))` produces the same results as `(rtimer_clock_t)(t2 - t1)` even if `t2 < t1`.

PR #2723, opened 2023-09-28
mac-sequence: Allow for wrap-arounds by using CLOCK_LT

kkrentz:
This PR adopts the new macro `CLOCK_LT` in `mac-sequence.c`.

### DIFF
--- a/os/net/mac/mac-sequence.c
+++ b/os/net/mac/mac-sequence.c
@@ -51,7 +51,7 @@
 
 struct seqno {
   linkaddr_t sender;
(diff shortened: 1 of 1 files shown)


joakimeriksson:
This looks good - but are you fixing a specific bug here? (it sounds so in the commit message)? If so - please describe it in more detail.  Also - are we comparing without less than or equal - is that changing the behavior slightly?

But in general - it is nicer to use the macros!

kkrentz:
I did some testing and both methods work fine. I will close this PR.

PR #2730, opened 2023-10-04
CSPRNG: Support for native platform on Linux and macOS

kkrentz:
This PR adds functionality for seeding the CSPRNG on the native platform, at least when run on Linux or macOS.

### DIFF
--- a/arch/platform/native/platform.c
+++ b/arch/platform/native/platform.c
@@ -64,6 +64,15 @@
 #include "net/ipv6/uip-ds6.h"
 #endif /* NETSTACK_CONF_WITH_IPV6 */
 
+#include "lib/assert.h"
+#include "lib/csprng.h"
+#ifdef __APPLE__
+#include <Security/Security.h>
+#include <Security/SecRandom.h>
+#else /* __APPLE__ */
+#include <sys/random.h>
+#endif /* __APPLE__ */
+
 /* Log configuration */
 #include "sys/log.h"
 #define LOG_MODULE "Native"
@@ -238,7 +247,15 @@ platform_init_stage_one()
   gpio_hal_init();
   button_hal_init();
   leds_init();
-  return;
+  struct csprng_seed seed;
+#ifdef __APPLE__
+  if(SecRandomCopyBytes(kSecRandomDefault, CSPRNG_SEED_LEN, seed.u8)
+      == errSecSuccess) {
+#else /* __APPLE__ */
+  if(getrandom(seed.u8, CSPRNG_SEED_LEN, GRND_RANDOM) == CSPRNG_SEED_LEN) {
+#endif /* __APPLE__ */
+    csprng_feed(&seed);
+  }
 }
 /*---------------------------------------------------------------------------*/
 void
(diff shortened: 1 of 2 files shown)


arch/platform/native/platform.c
```diff
@@ -238,7 +247,18 @@ platform_init_stage_one()
...
   leds_init();
-  return;
+#if defined(__linux__) || defined(__MACH__)
+  struct csprng_seed seed;
+#ifdef __linux__
+  if(getrandom(seed.u8, CSPRNG_SEED_LEN, GRND_RANDOM) != CSPRNG_SEED_LEN) {
+#else /* __linux__ */
+  if(SecRandomCopyBytes(kSecRandomDefault, CSPRNG_SEED_LEN, seed.u8)) {
+#endif /* __linux__ */
+    assert(false);
```
pjonsson, line 257: Can this state be recovered, or is it a fatal error? If it is fatal, it sounds like this should be an abort/exit.

That would also allow eliminating the else-block and just call `csprng_feed` without indentation.

arch/platform/native/platform.c
```diff
@@ -64,6 +64,15 @@
 #include "net/ipv6/uip-ds6.h"
 #endif /* NETSTACK_CONF_WITH_IPV6 */
 
+#include "lib/assert.h"
+#include "lib/csprng.h"
+#ifdef __linux__
+#include <sys/random.h>
+#elif defined(__MACH__)
```
pjonsson, line 71: There is a single mention of `__MACH__` in the tree right now, other places use `__APPLE__` to detect compilation for macOS. Can this place use `__APPLE__` too, or is that different from `__MACH__`?

arch/platform/native/platform.c
```diff
@@ -238,7 +247,18 @@ platform_init_stage_one()
   gpio_hal_init();
   button_hal_init();
   leds_init();
-  return;
+#if defined(__linux__) || defined(__MACH__)
+  struct csprng_seed seed;
+#ifdef __linux__
+  if(getrandom(seed.u8, CSPRNG_SEED_LEN, GRND_RANDOM) != CSPRNG_SEED_LEN) {
```
pjonsson, line 253: The Windows support has been removed, and FreeBSD has getrandom, so I think the inner ifdef can be `#ifdef __APPLE__ ... #else ... #endif`, and the outer ifdef can be dropped.

arch/platform/native/platform.c
kkrentz: It is not that fatal as long as the return value of `csprng_rand` is always being checked. I will remove that assert.

arch/platform/native/platform.c
```diff
@@ -238,7 +247,16 @@ platform_init_stage_one()
   gpio_hal_init();
   button_hal_init();
   leds_init();
-  return;
+#if defined(__linux__) || defined(__APPLE__)
```
pjonsson, line 250: There is no Windows support, so I think this ifdef can be dropped.

arch/platform/native/platform.c
```diff
@@ -238,7 +247,16 @@ platform_init_stage_one()
...
+  struct csprng_seed seed;
+#ifdef __linux__
```
pjonsson, line 252: Make this `#ifdef __APPLE__` and flip the branches and the code should work on FreeBSD too (assuming `csprng_feed` works on FreeBSD).

nvt (approved):

nfi (approved):

PR #2736, opened 2023-10-09
nrf: verify image programming

nfi:
Verify image after programming to avoid a warning when programming nrf devices.


pjonsson:
Shouldn't the same flags be used for the nrf52840 platform?

nfi:
Yes, you are correct. I have added this to the nrf52840 platform also.

nvt (approved):

PR #2737, opened 2023-10-10
CI: shorten job name

pjonsson:
The default job name is too long
to be visible in the "Checks" tab.
Set the job name to "$TESTNAME/$OS"
so the test results for each test
on different operating systems
are adjacent.


nfi (approved):

nfi:
Closes this since it has been merged as part of #2890.

PR #2741, opened 2023-10-16
Verify the payload length in the nRF radio driver

nvt:
Thanks to Tobias Scharnowski for reporting the issue.


arch/cpu/nrf/net/nrf-ieee-driver-arch.c
```diff
@@ -711,6 +711,13 @@ read_frame(void *buf, unsigned short bufsize)
     return 0;
   }
 
+  if(payload_len > bufsize) {
```
pjonsson, line 714: Can't the log message on line 708 also print bufsize, and this if-statement be folded into the check on line 707, `if(!phr_is_valid(rx_buf.phr) || payload_len > bufsize) {`?

pjonsson (approved):

nfi (approved):

PR #2748, opened 2023-10-22
tests/compile-arm-ports: switch gecko tests

pjonsson:
The two Gecko boards that are supported only
have two leds, and the gpio-hal example requires
three leds.

Compile the button-hal example instead.

### DIFF
--- a/examples/dev/button-hal/Makefile
+++ b/examples/dev/button-hal/Makefile
@@ -3,7 +3,7 @@ CONTIKI = ../../..
 
 all: $(CONTIKI_PROJECT)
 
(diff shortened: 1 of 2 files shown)


pjonsson:
@Yagoor since you did the Gecko port, can you confirm that there are only two leds on the supported boards?

nfi (approved):

Yagoor:
> @Yagoor since you did the Gecko port, can you confirm that there are only two leds on the supported boards?

Sorry for the late response but you are right.
Well, the brd4166a has 4 additional RGB LEDs but I did not add support for them because they are PWM controllable.

PR #2765, opened 2023-10-30
6LoWPAN: Avoid double context lookups

kkrentz:
This PR solves a TODO in `sicslowpan.c`, namely that the lookup of compression contexts happens twice.

### DIFF
--- a/os/net/ipv6/sicslowpan.c
+++ b/os/net/ipv6/sicslowpan.c
@@ -510,9 +510,6 @@ static struct sicslowpan_addr_context
 addr_contexts[SICSLOWPAN_CONF_MAX_ADDR_CONTEXTS];
 #endif
 
-/** pointer to an address context. */
-static struct sicslowpan_addr_context *context;
-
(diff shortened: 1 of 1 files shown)


os/net/ipv6/sicslowpan.c
```diff
@@ -1179,16 +1176,20 @@ uncompress_hdr_iphc(uint8_t *buf, uint16_t buf_size, uint16_t ip_len)
       PACKETBUF_IPHC_BUF[2] >> 4 : 0;
 
     /* Source address - check context != NULL only if SAM bits are != 0*/
+    struct sicslowpan_addr_context *source_context;
```
pjonsson, line 1179: Doesn't this shadow the previous declaration of source_context that this PR added?

pjonsson: Forget that comment, it's just Github stating everything is in the same function.

nfi (approved):

PR #2767, opened 2023-10-30
Import tinyCBOR 0.6.0 

pjonsson:
These are the files from tinyCBOR 0.6.0
along with Contiki-NG build support
and default test data (input.json and
input.cbor).

The differences compared to upstream are:

* The examples have renamed "main" to "original_main",
  modified to print to stderr instead of stdout,
  and gotten basic unit-test support added.


pjonsson:
This is what I expect the final tinyCBOR import to look like, but Doxygen is putting up a fight, so marking this as a draft for now.

This depends on #2768.

Have also split out the test/build support into #2769 and #2770 so the infrastructure improvements doesn't get blocked on Doxygen.

kkrentz:
To avoid naming conflicts, I would rather like to have this as a service.

pjonsson:
> To avoid naming conflicts, I would rather like to have this as a service.

Ok, that is annoying. Fixed.

pjonsson:
I don't think this is relevant any longer, so closing this PR.

PR #2769, opened 2023-10-30
Makefile.include: allow extra libraries

pjonsson:
This enables the native platform to easily
link against system libraries in examples.


Makefile.include
```diff
@@ -645,7 +645,7 @@ TARGET_LIBEXTRAS = $(LD_START_GROUP) $(TARGET_LIBFILES) $(LD_END_GROUP)
 $(BUILD_DIR_BOARD)/%.$(TARGET): LIBNAME ?= $@
 $(BUILD_DIR_BOARD)/%.$(TARGET): $(OBJECTDIR)/%.o $(LDSCRIPT) $(PROJECT_OBJECTFILES) $(PROJECT_LIBRARIES) $(CONTIKI_OBJECTFILES) $(INTERNAL_DEPS)
 	$(TRACE_LD)
-	$(Q)$(LD) $(LDFLAGS) $(TARGET_STARTFILES) $(sort ${filter-out FORCE $(LDSCRIPT) %.a,$^}) ${filter %.a,$^} $(TARGET_LIBEXTRAS) -o $(LIBNAME)
+	$(Q)$(LD) $(LDFLAGS) $(TARGET_STARTFILES) $(sort ${filter-out FORCE $(LDSCRIPT) %.a,$^}) ${filter %.a,$^} $(TARGET_LIBEXTRAS) $(LDLIBS) -o $(LIBNAME)
```
nfi, line 648: What do you think about using TARGET_LIBFILES instead that ensures the extra libraries are grouped within LD_START_GROUP and LD_END_GROUP?

pjonsson: The intention is for examples to be able to link against external libraries. These libraries are not necessarily target-specific.

On a more practical level, the things within start-group and end-group increase linking time, so we want to avoid adding things that shouldn't be there.

nfi: I agree about the name but I thought LD_START_GROUP/LD_END_GROUP were needed for simplelink-cc13xx-cc26xx ?

pjonsson: Because the libraries within the group on that target depends on each other.

nfi: Ok, thanks for the explanation.

nfi (approved):

PR #2772, opened 2023-11-01
CCM*: Revise sanity checks

kkrentz:
This PR revises both sanity checks in the `aead` function. One of them presumably remained from an older version. The other hinders the reuse of the implementation. The changes build upon #2754.

### DIFF
--- a/os/lib/ccm-star.c
+++ b/os/lib/ccm-star.c
@@ -47,9 +47,6 @@
     | ((((mic_len) - 2u) >> 1) << 3) \
     | 1u)
 #define CCM_STAR_ENCRYPTION_FLAGS 1
-/* The auth data can have a variable length, but this implementation supports
- * only the case of 0 < l(a) < (2^16 - 2^8) */
-#define MAX_A_LEN 0xfeff
 /* Valid values are 4, 6, 8, 10, 12, 14, and 16 octets */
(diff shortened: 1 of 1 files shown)


os/lib/ccm-star.c
```diff
@@ -152,9 +151,7 @@ aead(const uint8_t* nonce,
     uint8_t *result, uint8_t mic_len,
     int forward)
 {
-  if(!MIC_LEN_VALID(mic_len)) {
-    return;
-  }
+  assert(mic_len <= AES_128_BLOCK_SIZE);
```
pjonsson, line 154: Since code might not be compiled with assertions enabled, this changes the semantics. Are there external callers to this function?

kkrentz: As the previous version only worked when passing standards-compliant MIC lengths, nothing should break when we expand the valid range.

pjonsson: What happens with a faulty length in code built without assertions, does Contiki-NG crash?

kkrentz: I am going to create another PR where I change the return type to int so that we can bubble up errors.

nfi (approved):

joakimeriksson (approved):

PR #2778, opened 2023-11-08
CC2538: Faster AES-128 driver

kkrentz:
This is a refactored AES-128 driver for CC2538-based platforms. It is implemented in the same manner as the drivers in #2773 and #1653.

### DIFF
--- a/arch/cpu/cc2538/dev/cc2538-aes-128.c
+++ b/arch/cpu/cc2538/dev/cc2538-aes-128.c
@@ -28,6 +28,7 @@
  *
  * This file is part of the Contiki operating system.
  */
+
 /**
  * \addtogroup cc2538-aes-128
  * @{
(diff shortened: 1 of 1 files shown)


arurke:
@kkrentz I am not familiar with the platform so please bare with me, but the idea is to deprecate the now duplicated code in `cc2538/dev/aes.c/.h`? Or do they have reasons to both exist?

kkrentz:
The original motivation behind this PR was to accelerate the AES-128. For some reason, I saw no improvement when I prepared this PR anymore. Maybe, I accidentally configured the software implementation.

I had just added a speed test to this PR. It invokes `AES_128.set_key` & `AES_128.encrypt` 1000 times. This takes the old driver 49ms on an OpenMote, whereas the refactored one only needs 30ms.

The superseded files might stay or not. I could create a separate PR for this discussion.

arurke:
From the outside it looks like the platform has drivers for AES and ECB (in aes.c/.h and ecb.c/.h), but its AES-128-bit driver now, rather than use existing functionality in those drivers, re-implements better/faster versions themselves. Intuitively, I would think the improvements belonged in aes.c/.h and ecb.c/.h. What am I missing?

kkrentz:
I managed to backport my tricks, yet a remaining performance penalty are the long parameter lists and nested function calls in the existing code. Doing everything in one go seems to help performance a lot. Plus, I could to delete various files in `cpu/cc2538/dev/`. This would also eliminate some already existent functional overlap with `cc2538-ccm-star.c`.

arurke:
hi @kkrentz , is this PR still a draft or ready for review?

kkrentz:
I reserve the "speed test" for a separate PR.

PR #2797, opened 2023-11-29
nullradio: remove no-op call

pjonsson:
The prepare function does not do anything,
so remove the call to it.

### DIFF
--- a/os/dev/nullradio.c
+++ b/os/dev/nullradio.c
@@ -27,8 +27,7 @@ transmit(unsigned short transmit_len)
 static int
 send(const void *payload, unsigned short payload_len)
 {
-  prepare(payload, payload_len);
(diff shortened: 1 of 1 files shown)


os/dev/nullradio.c
```diff
@@ -27,7 +27,6 @@ transmit(unsigned short transmit_len)
 static int
 send(const void *payload, unsigned short payload_len)
 {
-  prepare(payload, payload_len);
   return transmit(payload_len);
```
nfi, line 30: If the call to `prepare` is removed because it does nothing, should not the call to `transmit` also be removed for the same reason?

pjonsson: Fixed.

nfi (approved):

PR #2825, opened 2023-12-27
Update to Doxygen 1.10.0

pjonsson:
Update the Doxyfile to the same version,
and restore the default value of
EXPAND_ONLY_PREDEF to avoid mqtt.h
warnings from Doxygen not understanding
that MQTT_5 is set.


pjonsson:
Failure on macOS job is intermittent and not related to Doxygen.

nvt (approved):

nfi (approved):

PR #2846, opened 2024-01-21
rtimer.h: put signature in comment

pjonsson:
This fixes a CodeQL warning.


nfi (approved):

arurke:
Is this intended as documentation so that arch learns the function signature they should implement?

joakimeriksson:
@pjonsson  - what about moving it up in the previous comment and adding that this is the "signature" of the function?

nfi (approved):

PR #2857, opened 2024-02-22
Checks for NULL parameters in list_add, list_push, list_insert

sandipghosal:
(1) Not checking if (item != NULL) as in line 78 would result "Segmentation fault" for a function call list_add(list, NULL)

(2) Not checking if (item != NULL) as in line 97 would result "Segmentation fault" for a function call list_push(list, NULL)

(3) Not checking if (newitem != NULL) as in line 182 would result "Segmentation fault" for a function call list_insert where newitem = NULL.

### DIFF
--- a/os/lib/list.c
+++ b/os/lib/list.c
@@ -74,15 +74,17 @@ list_add(list_t list, void *item)
 
   /* Make sure not to add the same element twice */
   list_remove(list, item);
+	
+  if(item != NULL){
+  	((struct list *)item)->next = NULL;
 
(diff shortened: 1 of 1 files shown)


alexrayne:
imho this checks should be via asserts.
also GCC have function attributes __attribute__((nonnull X, ... )) - that prevents from passing NULL

nvt:
Closing this PR as it has been inactive for over a year and not gotten an updated code style, as commented in an e-mail conversation.

PR #2861, opened 2024-02-26
(arch-nrf) Switching from TIMER0 to RTC rtimer

marcel-cd:
The current nrf implementation is using the TIMER (and HFXO) as a time source for the rtimer tick. The TIMER with the HFXO is very accurate, but needs even in sleep more than 500uA. 

This PR swtiches to a RTC based implementation, which will reduce the current during sleep to around 7uA (measured on a nRF52840-DK with the Nordic Power Profiler)


arch/cpu/nrf/sys/soc-rtc.c
```diff
@@ -0,0 +1,212 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+#include "sys/log.h"
+#define LOG_MODULE "RTC"
+#define LOG_LEVEL LOG_LEVEL_INFO
+
+/*---------------------------------------------------------------------------*/
+/* Prototype of a function in clock.c. Called every time the handler fires */
+void clock_update(void);
```
Yagoor, line 59: It is more clean if you make a header file to the clock-arch and have this function prototype there, so it is clear that this function is used outside the clock-arch file.

arch/cpu/nrf/sys/soc-rtc.c
```diff
@@ -0,0 +1,212 @@
...
+  nrfx_rtc_enable(&rtc);
+
+  next = (nrfx_rtc_counter_get(&rtc) + COMPARE_INCREMENT) & MULTIPLE_256_MASK;
+  nrfx_rtc_cc_set(&rtc, SOC_RTC_SYSTEM_CH, next, true);
+}
+/*---------------------------------------------------------------------------*/
+void
+soc_rtc_schedule_one_shot(uint32_t channel, rtimer_clock_t ticks)
+{
+  /* clock_time_t next = (clock_time_t)RTC_WRAP(ticks); // % rtc_max_rtimer_ticks; */
```
Yagoor, line 181: You forgot to clean up these lines. :)

Yagoor (commented):
It looks good, just some small comments.
Was this tested on CSMA and TSCH?

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
```diff
@@ -720,10 +719,18 @@ read_frame(void *buf, unsigned short bufsize)
...
+  timestamps.framestart = nrf_timer_cc_get(NRF_RTIMER_TIMER, NRF_TIMER_CC_CHANNEL3);
+  timestamps.end = nrf_timer_cc_get(NRF_RTIMER_TIMER, NRF_TIMER_CC_CHANNEL2);
   timestamps.mpdu_duration = rx_buf.phr * BYTE_DURATION_RTIMER;
 
+
+  /* Use NRF_RTIMER_TIMER CC0 for conversion to rtimer(based on RTC) */
+  /* timestamps.sfd will be used in TSCH Module for slot calculation */
+  NRF_RTIMER_TIMER->TASKS_CAPTURE[0] = 1;
+  uint64_t now = RTIMER_NOW();
+  timestamps.framestart = now - ((NRF_RTIMER_TIMER->CC[0] - timestamps.framestart) / (float)62500 * RTIMER_SECOND);
```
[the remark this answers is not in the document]
marcel-cd: Easiest would be, if I could use the RTC in the PPI, but nRF52840 cant use PPI on RTC (NRF_RTC_TASK_CAPTURE_x is not available) so I have still keep the PPI with TIMER. 

In Line 981, TSCH MAC will ask for the  RADIO_PARAM_LAST_PACKET_TIMESTAMP, which should come as rtimer value, but timestamps.framestart was a CC Value from the TIMER

so I decided to convert in 731 the timestamps.framestart CC from the TIMER to the corresponding rtimer value.  
for this I need at the moment of conversion (731) the current TIMER counter (which I get with a capture on CC[0]) and the current rtimer counter, than I can do the converstion. 

The synchronization of the slots is very good, I did tests with a Logic Analyzer on GPIOs and had an offset between 10-80 us

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
```diff
@@ -720,10 +719,18 @@ read_frame(void *buf, unsigned short bufsize)
...
   /* Latch timestamp values for this most recently received frame */
   timestamps.phr = rx_buf.phr;
-  timestamps.framestart = nrf_timer_cc_get(NRF_TIMER0, NRF_TIMER_CC_CHANNEL3);
-  timestamps.end = nrf_timer_cc_get(NRF_TIMER0, NRF_TIMER_CC_CHANNEL2);
+  timestamps.framestart = nrf_timer_cc_get(NRF_RTIMER_TIMER, NRF_TIMER_CC_CHANNEL3);
+  timestamps.end = nrf_timer_cc_get(NRF_RTIMER_TIMER, NRF_TIMER_CC_CHANNEL2);
   timestamps.mpdu_duration = rx_buf.phr * BYTE_DURATION_RTIMER;
 
+
+  /* Use NRF_RTIMER_TIMER CC0 for conversion to rtimer(based on RTC) */
```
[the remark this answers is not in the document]
marcel-cd: I think no. 
I use CC0 to convert the value in timestamps.framestart to the value which it would have, if the PPI would run with the RTC_Capture.

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
```diff
@@ -720,10 +719,18 @@ read_frame(void *buf, unsigned short bufsize)
...
+  /* timestamps.sfd will be used in TSCH Module for slot calculation */
+  NRF_RTIMER_TIMER->TASKS_CAPTURE[0] = 1;
+  uint64_t now = RTIMER_NOW();
+  timestamps.framestart = now - ((NRF_RTIMER_TIMER->CC[0] - timestamps.framestart) / (float)62500 * RTIMER_SECOND);
+  timestamps.end = now - ((NRF_RTIMER_TIMER->CC[0] - timestamps.end) / (float)62500 * RTIMER_SECOND);
```
pjonsson, line 732: You should be able to get a floating point constant by appending an `f` instead of having to do a cast.

I don't know anything about the code here, but is the floating point division standard practice for timer code?

arch/cpu/nrf/sys/lpm.c
```diff
@@ -0,0 +1,293 @@
...
+ * \addtogroup nrf
+ * @{
+ *
+ * \addtogroup nrf-sys System drivers
+ * @{
+ *
+ * This implementation is based on the cc26x0-cc13x0 platform.
+ *
+ * \file
+ *         Low Power Mode (LPM) driver for the nRF52
```
pjonsson, line 41: The nRF cpu/platform in Contiki-NG is also used for the nRF 5340.

arch/cpu/nrf/sys/soc-rtc.c
```diff
@@ -0,0 +1,212 @@
...
+}
+/*---------------------------------------------------------------------------*/
+/**
+ * @brief Function for handling the RTC<instance> interrupts
+ * @param int_type Type of interrupt to be handled
+ */
+static void
+rtc_handler(nrfx_rtc_int_type_t int_type)
+{
+  uint32_t next;
```
pjonsson, line 102: Push the variable declaration into the branch where it is used, unless this file is something that should be synchronized with an upstream-source regularly.

arch/cpu/nrf/sys/soc-rtc.c
```diff
@@ -0,0 +1,212 @@
...
+#define NRF_CLOCK_RTC_INSTANCE 0
+#endif
+
+/* Prototype of a function in clock.c. Called every time the handler fires */
+void clock_update(void);
+/*---------------------------------------------------------------------------*/
+/**< RTC instance used for platform clock */
+static const nrfx_rtc_t rtc = NRFX_RTC_INSTANCE(NRF_CLOCK_RTC_INSTANCE);
+/*---------------------------------------------------------------------------*/
+static rtimer_clock_t last_isr_time = 0;
```
pjonsson, line 84: Aren't static variables automatically initialized to 0 in C, so these RHSes can be omitted?

arch/cpu/nrf/sys/soc-rtc.c
```diff
@@ -0,0 +1,212 @@
...
+
+/*---------------------------------------------------------------------------*/
+#define COMPARE_INCREMENT (RTIMER_SECOND / CLOCK_SECOND)
+#define MULTIPLE_256_MASK 0xFFFFFF00
+/*---------------------------------------------------------------------------*/
+#if CLOCK_SIZE != 4
+#error CLOCK_CONF_SIZE must be 4 (32 bit)
+#endif
+#if RTIMER_SECOND != 32768
+#error RIMTER_SECOND must be 32768 when using the NRF_RTC
```
pjonsson, line 69: `RTIMER_SECOND`.

arch/cpu/nrf/net/nrf-ieee-driver-arch.c
marcel-cd: This is a conversion and has to be as exact a possible, 
but good point, there's a better way to do this calc without the float cast ;-)
```
timestamps.framestart = now - ((NRF_RTIMER_TIMER->CC[0] - timestamps.framestart) * RTIMER_SECOND / 62500 );
```

arch/cpu/nrf/sys/rtimer-arch.c
```diff
@@ -42,71 +43,74 @@
...
-  NVIC_ClearPendingIRQ(NRF_RTIMER_IRQn);
-  NVIC_EnableIRQ(NRF_RTIMER_IRQn);
   nrf_timer_task_trigger(NRF_RTIMER_TIMER, NRF_TIMER_TASK_START);
 }
 /*---------------------------------------------------------------------------*/
+/**
+ *
+ * This function schedules a one-shot event with the AON RTC.
+ *
+ * This functions converts \e to a value suitable for the AON RTC.
```
pjonsson, line 80: What does `\e` refer to?

arch/cpu/nrf/sys/lpm.h
```diff
@@ -0,0 +1,115 @@
...
+ * have to check board configuration before calling this.
+ */
+void lpm_pin_set_default_state(uint32_t ioid);
+/*---------------------------------------------------------------------------*/
+/* /\*---------------------------------------------------------------------------*\/ */
+/* /\** */
+/*  * @brief Stop and wait for an interrupt */
+/*  *\/ */
+/* static inline void */
+/* lpm_drop(void) */
```
pjonsson, line 99: There are lots of warnings about commented out code in Contiki-NG already, so please don't add new code with commented out code.

arch/cpu/nrf/sys/soc-rtc.c
marcel-cd: this is correct, not necessary.

arch/cpu/nrf/sys/lpm.h
```diff
@@ -0,0 +1,115 @@
...
+/* { */
```
pjonsson, line 100: On second thought: what `lpm_drop()` is called from `arch/platform/nrf/platform.c` if this is removed?

arch/cpu/nrf/sys/rtimer-arch.c
marcel-cd: some garbage, I've used a lot ideas (and code) from the CC26x0 platform.

arch/cpu/nrf/sys/lpm.c
```diff
@@ -0,0 +1,293 @@
...
+    max_pm = pm;
+  }
+  trigger_etimer = RTIMER_CLOCK_MAX;
+  if(max_pm == LPM_MODE_SLEEP) {
+    /* In sleep mode, HFXO is powered up, therefore we do  not have
+     * to wakeup earlier to powering up the HFXO */
+    if(next_etimer_set) {
+      /* Schedule the next system wakeup due to etimer */
+      if(RTIMER_CLOCK_LT(next_etimer, now + MIN_SAFE_SCHEDULE)) {
+        /* Too soon in future, keep the system awake */
```
pjonsson, line 188: We get warnings from empty if-blocks, would be nice if the logic is changed to avoid this branch.

arch/cpu/nrf/sys/lpm.h
marcel-cd: lpm.h was in arch/platform/nrf before, and  did only a __WFI(); 
In the old implementation, the nrf woke up ever clock tick, because there was no logic regarding more intelligent sleep.

as it is only one lpm needed for nrf I moved  this implementation to arch/cpu/nrf 
Please correct me, if this was wrong.

arch/cpu/nrf/sys/lpm.c
```diff
@@ -0,0 +1,293 @@
...
+        /* Too far in future, use MAX_SLEEP_TIME instead */
+        trigger_etimer = now + MAX_SLEEP_TIME;
+      } else {
+        trigger_etimer = next_etimer;
+      }
+    } else {
+      /* No etimers set. Since by default the CH1 RTC fires once every clock tick,
+       * need to explicitly schedule a wakeup in the future to save energy.
+       * But do not stay in this mode for too long, otherwise watchdog will be trigerred. */
+      trigger_etimer = now + MAX_SLEEP_TIME;
```
pjonsson, line 199: When is `trigger_etimer` set to MAX_SLEEP_TIME, `!next_etimer_set || RTIMER_CLOCK_LT(now + MAX_SLEEP_TIME, next_etimer)`?

arch/cpu/nrf/sys/lpm.c
marcel-cd: I normally do no empty if-blocks, but here it made sense to me, because it shows the algorithm better. Let me know, if you need to have this corrected

arch/cpu/nrf/sys/lpm.c
```diff
@@ -0,0 +1,293 @@
...
+    }
+  } else if(max_pm == LPM_MODE_DEEP_SLEEP) {
+    /* Watchdog is not enabled, so deep sleep can continue an arbitrary long time.
+     * On the other hand, when we have to start again the system, it takes a while
```
pjonsson, line 203: start again the system => start the system again

arch/cpu/nrf/sys/soc-rtc.c
marcel-cd: I do not understand your point here?

arch/cpu/nrf/sys/lpm.c
marcel-cd: for me, your code is  harder to read, dont you think?

arch/cpu/nrf/sys/lpm.h
pjonsson: Oh, I see, it's moved to a C file. Well, remove the commented out code then.

arch/cpu/nrf/sys/soc-rtc.c
pjonsson: If this file comes from a third-party source so we should preserve formatting, then do nothing.

Otherwise make line 105 `uint32_t next = ..` since the variable is only used in that branch.

arch/cpu/nrf/sys/soc-rtc.c
```diff
@@ -0,0 +1,212 @@
...
+
+  /*Initialize RTC instance */
+  nrfx_rtc_config_t config = NRFX_RTC_DEFAULT_CONFIG;
+  config.prescaler = RTC_FREQ_TO_PRESCALER(32768); /* full speed... */
+  config.interrupt_priority = 6;
+  config.reliable = 0;
+
+  err_code = nrfx_rtc_init(&rtc, &config, rtc_handler);
+
+  if(err_code != NRFX_SUCCESS) {
```
pjonsson, line 150: This if-statment doesn't make any difference since no other code follows, so it can be removed.

arch/cpu/nrf/sys/lpm.c
pjonsson: "If next etimer is not set, or next etimer is more than MAX_SLEEP_TIME into the future"?

I'm trying to understand the logic in this part of the function, I'm not sure it is equivalent to what I wrote.

pjonsson:
I know TrustZone does stuff with timers, so someone who knows that stuff should check if this change is compatible with how things are currently set up for TrustZone.

arch/cpu/nrf/sys/soc-rtc.c
marcel-cd: ah ok ;-)

arch/cpu/nrf/sys/lpm.c
marcel-cd: let me explain in some words:

1. soc_rtc.c sets in the rtc_handler the next wakeup time in 1/CLOCK_SECOND ms.  (rtimer is not relevant here, as rimter will be woken up exactly to the defined time)
2. setup_sleep_mode in lpm.c can overewrite this next wakeup time.
3. if the next rtimer and etimer is  > (now+STANDBY_MIN_DURATION) (10ms) -> we go into deep_sleep otherwise into sleep

Sleep: 
if an etimer is set, we set the next wakup time to the next  etimer (will be max 10ms) 
if no etimer is set (but eg. a rtimer within then next 10ms) we sleep for MAX_SLEEP_TIME

DeepSleep: 
- if we have an rtimer set, set the wakeup a bit before this rtimer, to fireup the HFXO
- if an etimer is set, and it is before the rtimer, set the next wakeup time to this etimer

makes this sense?

marcel-cd:
Hi Yago,
I did some tests with CSMA yesterday. It works, but only if max LPM mode is reduced
```
#LPM_MODE_MAX_SUPPORTED_CONF LPM_MODE_DEEP_AWAKE
```
I will change lpm.c in the same way as in cc26x0-cc13x0. Modules (like the nrf-ieee-driver-arch) can register a callback and change the max PM (eg. when RX is on)

Yagoor (approved):
It looks good from my side, if TSCH and CSMA are still working. 
This is a great improvement, thanks for that.

arch/cpu/nrf/sys/soc-rtc.c
```diff
@@ -0,0 +1,212 @@
...
+{
+  (void)event;
+}
+/*---------------------------------------------------------------------------*/
+/**
+ * @brief Function for handling the RTC<instance> interrupts
+ * @param int_type Type of interrupt to be handled
+ */
+static void
+rtc_handler(nrfx_rtc_int_type_t int_type)
```
g-lazaridis, line 95: Have you considered what will happen if an overflow irq occurs while the clock_update or rtimer_run_next functions are executed? This is quite possible to happen (rtimer_run_next runs tsch slot operation).

fatihayvenli (commented):
I've tried 6tisch/sixtop example and it works. The drift is around 1 ppm (min/max delta seen: -3/3). 
But there is a caveat. I get deadline miss by -1 tick in tsch-slot-operation after 1024 seconds and it grows by 1 tick in every 512 seconds. This corresponds to the RTIMER overflow. For some reason, one tick gets lost after each RTIMER overflow. And when the deadline miss reaches to -51 ticks, the node restarts.

PR #2866, opened 2024-03-07
sys: prevent event allocations from wrapping

nvt:


nfi (approved):

alexrayne:
imho, there better assert for this failure that spots problem on debug

PR #2874, opened 2024-03-19
sys: modernize the process module

nvt:
This PR updates the C programming style, improves the log output, removes the notion of a zombie process, and adds some defensive checks for NULL and incorrect configuration.

### DIFF
--- a/os/sys/process.c
+++ b/os/sys/process.c
@@ -53,6 +53,19 @@
 #define LOG_LEVEL LOG_LEVEL_SYS
 
 /*
+ * The process_num_events_t type is an uint8_t. It must be able to store
+ * the value of PROCESS_CONF_NUMEVENTS + 1, for the additional
+ * boolean value indicating whether a poll has been requested.
+ */
+static_assert(PROCESS_CONF_NUMEVENTS > 0 && PROCESS_CONF_NUMEVENTS <= 128,
+  "PROCESS_CONF_NUMEVENTS must be a positive value of at most 128.");
+
+/* Require that PROCESS_CONF_NUMEVENTS is a power of 2 to allow
+   optimization of modulo operations. */
+static_assert(!(PROCESS_CONF_NUMEVENTS & (PROCESS_CONF_NUMEVENTS - 1)),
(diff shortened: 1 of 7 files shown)


os/sys/process.h
```diff
@@ -318,11 +334,12 @@ struct process {
...
 #else
   const char *name;
-#define PROCESS_NAME_STRING(process) (process)->name
+#define PROCESS_NAME_STRING(process) ((process) != NULL ? (process)->name : "")
 #endif
   PT_THREAD((* thread)(struct pt *, process_event_t, process_data_t));
   struct pt pt;
-  unsigned char state, needspoll;
+  uint8_t state;
+  bool needspoll;
```
alexrayne, line 341: bool value on ARM platform takes 4byte. therefore new structure takes more RAM

os/sys/process.c
```diff
@@ -301,7 +288,7 @@ do_event(void)
   }
 }
 /*---------------------------------------------------------------------------*/
-int
+process_num_events_t
```
alexrayne, line 291: As i understand C - operator + returns int value.
You down-converts it into u8 - that takes more code and cpu cycles

os/sys/process.c
```diff
@@ -248,26 +243,19 @@ do_poll(void)
...
    */
-
   if(nevents > 0) {
 
     /* There are events that we should deliver. */
-    ev = events[fevent].ev;
-
-    data = events[fevent].data;
-    receiver = events[fevent].p;
+    process_event_t ev = events[fevent].ev;
```
alexrayne, line 256: useless changes for nothig

os/sys/process.c
alexrayne: poll_requeted should be same type as nevents - to avoid this penalty. but you use bool type - that is int sized

os/sys/process.h
nfi: I checked on the Contiki-NG ARM-based platforms and size of bool is 1 byte.

nfi (approved):

PR #2885, opened 2024-04-19
Fix default define for RPL_NBR_POLICY_MAX_NEXTHOP_NEIGHBORS in RPL-CLASSIC

kevin-forest-cysca:
The original development seems to have made the wrong #ifndef.

If RPL_NBR_POLICY_MAX_NEXTHOP_NEIGHBORS was not defined by the contiki-default-conf.h it was defined instead by default by os\net\routing\rpl-classic\rpl-nbr-policy.h, which was NBR_TABLE_CONF_MAX_NEIGHBORS.


arurke:
Is the entry in `contiki-default-conf.h` needed given that the same default value is set in the module? Also the entry in `contik-default-conf.h` will fail if `NBR_TABLE_CONF_MAX_NEIGHBORS` is not set, which seems arbitrary. Am I missing something?

kevin-forest-cysca:
@arurke,
There should not be any issues with NBR_TABLE_CONF_MAX_NEIGHBORS since in the contiki-default-conf.h it is defined before if it was not set (line 71)

And for your first question, I tried to copy the style of the define for NBR_TABLE_CONF_CAN_ACCEPT_NEW  which was redefined in nbr-table.h.
I think I would prefer to remove RPL_NBR_POLICY_CONF_MAX_NEXTHOP_NEIGHBORS from contiki-default-conf.h and just keep the header in the rpl-nbr-policy.h. It`s a duplicate entry for nothing, but I don`t know if it was also setup there to know more easily the tweakable settings of the project.

arurke:
> but I dont know if it was also setup there to know more easily the tweakable settings of the project.

I think you are right, and given that the  `NBR_TABLE_CONF_MAX_NEIGHBORS` are also "duplicated" in both the default-conf and the module, it makes sense for the neighbor policy max neighbor to be so as well.

arurke (approved):

os/contiki-default-conf.h
```diff
@@ -107,8 +107,8 @@
 #define NBR_TABLE_CONF_CAN_ACCEPT_NEW          rpl_nbr_can_accept_new
 #endif /* NBR_TABLE_CONF_CAN_ACCEPT_NEW */
 /* Leave 3 spots for candidate parents incl. default route. */
-#if RPL_NBR_POLICY_MAX_NEXTHOP_NEIGHBORS
-#define RPL_NBR_POLICY_MAX_NEXTHOP_NEIGHBORS  MAX(NBR_TABLE_CONF_MAX_NEIGHBORS - 3, 0)
+#ifndef RPL_NBR_POLICY_CONF_MAX_NEXTHOP_NEIGHBORS
+#define RPL_NBR_POLICY_CONF_MAX_NEXTHOP_NEIGHBORS  MAX(NBR_TABLE_CONF_MAX_NEIGHBORS - 3, 0)
 #endif /* RPL_NBR_POLICY_MAX_NEXTHOP_NEIGHBORS */
```
pjonsson, line 112: Comment needs to be updated to match the new name of the variable.

nvt (approved):

os/contiki-default-conf.h
kevin-forest-cysca: Fixed

nfi (approved):

PR #2888, opened 2024-04-24
Allow disabling CoAP well known resource

arurke:
As in title. The resource is still enabled by default. Disabling saves almost 500 bytes code space on cc13xx-cc26xx.

### DIFF
--- a/os/net/app-layer/coap/coap-engine.c
+++ b/os/net/app-layer/coap/coap-engine.c
@@ -134,8 +134,10 @@ call_service(coap_message_t *request, coap_message_t *response,
 /*- Server Part -------------------------------------------------------------*/
 /*---------------------------------------------------------------------------*/
 
-/* the discover resource is automatically included for CoAP */
+#if COAP_WELL_KNOWN_RESOURCE_ENABLED
+/* The discover resource is included by default for CoAP */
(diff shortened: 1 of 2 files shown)


os/net/app-layer/coap/coap-engine.c
```diff
@@ -134,8 +134,10 @@ call_service(coap_message_t *request, coap_message_t *response,
 /*- Server Part -------------------------------------------------------------*/
 /*---------------------------------------------------------------------------*/
 
-/* the discover resource is automatically included for CoAP */
+#if COAP_ENABLE_WELL_KNOWN_RESOURCE
+/* The discover resource is included by default for CoAP */
```
nfi, line 138: I think it is a nice option to be able to disable this.

Many configurations in Contiki-NG have names like `LWM2M_QUEUE_MODE_ENABLED`, `LLSEC802154_ENABLED`, `STACK_CHECK_ENABLED`, `QUEUEBUF_ENABLED`, `LOG_INFO_ENABLED`, etc. What do you think about changing to a name like `COAP_WELL_KNOWN_RESOURCE_ENABLED` ?

arurke: Agreed and updated

nfi (approved):

PR #2889, opened 2024-04-24
Add missing include of inttypes.h in rf-ble-cmd.c

arurke:
As in title.

### DIFF
--- a/arch/cpu/cc26x0-cc13x0/rf-core/ble-hal/rf-ble-cmd.c
+++ b/arch/cpu/cc26x0-cc13x0/rf-core/ble-hal/rf-ble-cmd.c
@@ -45,6 +45,7 @@
 #include "rf_ble_cmd.h"
 #include "rf-core/rf-core.h"
 #include "rf-core/ble-hal/rf-ble-cmd.h"
+#include <inttypes.h>
 
 /*---------------------------------------------------------------------------*/
 #include "sys/log.h"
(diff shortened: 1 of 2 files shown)


arurke:
@pjonsson Are you familiar with the failures in all the mac os jobs? A quick google did not offer me much.

pjonsson:
@arurke it seems the Github runners got a new macOS version, and that switched them from Intel to M1 silicon. I'm just heading into a meeting, can you please make a PR here that switches from macos-latest to the older version used by the Cooja repository? It's not just the matrix that needs to be updated, there are also checks further down that needs to be updated, brew for example is only executed on macOS.

nfi:
I made a PR to address this issue: #2890

arch/cpu/cc26x0-cc13x0/rf-core/ble-hal/rf-ble-cmd.c
```diff
@@ -45,6 +45,7 @@
 #include "rf_ble_cmd.h"
 #include "rf-core/rf-core.h"
 #include "rf-core/ble-hal/rf-ble-cmd.h"
+#include <inttypes.h>
```
pjonsson, line 48: If this file is missing, shouldn't the include be in `rf_ble_cmd.h` instead? There are plenty of uses of `uint8_t` in the function prototypes.

arurke: iirc, it was the `PRIX32` on line 525 that I got a complaint about. I am guessing `stdint.h` is carried with some of the other includes in `rf_ble_cmd.h` and that is why it passes. But in any case, you are right -  I added `stdint.h` to the header-file (as opposed to `inttypes.h`) to keep it as targeted as possible.

pjonsson: Oh, for `PRIX32` you are right about the header, nevermind my comment then.

pjonsson (approved):

nfi (approved):

PR #2893, opened 2024-04-29
Cooja platform: append to CONTIKI_TARGET_SOURCEFILES rather than set them

atiselsts:
Other Contiki-NG platforms add to `CONTIKI_TARGET_SOURCEFILES`  in their own Makefile.

Cooja simply sets the variable `CONTIKI_TARGET_SOURCEFILES`. 

This breaks user application code that sets `CONTIKI_TARGET_SOURCEFILES` from the application Makefiles.

### DIFF
--- a/arch/platform/cooja/Makefile.cooja
+++ b/arch/platform/cooja/Makefile.cooja
@@ -67,7 +67,7 @@ COOJA_INTFS	= beep.c ip.c leds-arch.c moteid.c \
 COOJA_CORE = platform.c mtype.c random.c sensors.c leds.c gpio-hal-arch.c buttons.c
 
 # (COOJA_SOURCEFILES contains additional sources set from simulator)
-CONTIKI_TARGET_SOURCEFILES = \
+CONTIKI_TARGET_SOURCEFILES += \
 $(COOJA_BASE) $(COOJA_INTFS) $(COOJA_CORE) $(COOJA_NET) $(COOJA_SOURCEFILES)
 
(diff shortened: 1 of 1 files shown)


arch/platform/cooja/Makefile.cooja
```diff
@@ -67,7 +67,7 @@ COOJA_INTFS	= beep.c ip.c leds-arch.c moteid.c \
 COOJA_CORE = platform.c mtype.c random.c sensors.c leds.c gpio-hal-arch.c buttons.c
 
 # (COOJA_SOURCEFILES contains additional sources set from simulator)
-CONTIKI_TARGET_SOURCEFILES = \
+CONTIKI_TARGET_SOURCEFILES += \
```
pjonsson, line 70: `doc/getting-started/The-Contiki-NG-build-system.md` describes this variable as:

>  It contains the list of C files (`CONTIKI_TARGET_SOURCEFILES`) that the platform adds to the Contiki-NG system.

which I interpret as "this is a system variable not meant to be modified by applications".

atiselsts:
We have an inconsistency and I think there are two course of action to fix that inconsistency:
a) Change how `CONTIKI_TARGET_SOURCEFILES` are handled in other platforms and break user applications that rely on the undocumented behavior
b) Change the Cooja platform and the documentation
I made the PR so my opinion on this tradeoff should be pretty clear. Porting an app to a Cooja, or to a new version of Contiki-NG should not require in-depth Makefile changes from the user IMO.

PR #2898, opened 2024-05-06
elliptic curve cryptography: platform-independent interface and two implementations

kkrentz:
This PR introduces a platform-independent interface for ECC operations. Also, this PR includes a hardware-accelerated implementation for the CC2538, and a software implementation based on uECC.


examples/libs/ecc/ecdh.c
```diff
@@ -0,0 +1,148 @@
...
+ * \author
+ *         Konrad Krentz <konrad.krentz@gmail.com>
+ */
+
+#include "contiki.h"
+#include "lib/ecc.h"
+
+/* Log configuration */
+#include "sys/log.h"
+#define LOG_MODULE "ECDSA"
```
nvt: Shouldn't this log module name be "ECDH"?

examples/libs/ecc/ecdh.c
```diff
@@ -0,0 +1,148 @@
...
+  if(result) {
+    LOG_ERR("generate_key_pair failed\n");
+    PROCESS_EXIT();
+  }
+  PROCESS_PT_SPAWN(ECC.get_protothread(),
+                   ECC.generate_key_pair(public_key_b,
+                                         private_key_b,
+                                         &result));
+  if(result) {
+    LOG_ERR("generate_key_pair failed\n");
```
nvt, line 94: To prevent resource leaks, shouldn't also ECC.disable() be called before every PROCESS_EXIT()? This may also apply to the other modules in this directory.

kkrentz: Indeed. The only exception is `ECC.enable()`, which releases the ECC mutex automatically upon errors. Thanks for spotting this.

kkrentz:
I turned the `struct ecc_driver` into regular functions to help the compiler exclude unused functions from the firmware.

nvt (approved):

nfi (approved):

PR #2936, opened 2024-07-11
Added snmp_packet->used validation check in snmp_ber_decode_string_len_buffer

OwenCochell:
The `snmp_ber_decode_string_len_buffer` function does not properly verify the the packet size when de-referencing the `in` pointer, allowing for an out of bounds read of 1 byte in some scenarios. I originally disclosed this issue via email and was instructed to file a PR.

This PR fixes the issue by adding a check that will return if `snmp_packet->used == 0` before the de-reference, which will prevent this out of bounds read from occurring.

### DIFF
--- a/os/net/app-layer/snmp/snmp-ber.c
+++ b/os/net/app-layer/snmp/snmp-ber.c
@@ -372,19 +372,15 @@ snmp_ber_decode_string_len_buffer(snmp_packet_t *snmp_packet, const char **str, 
     return 0;
   }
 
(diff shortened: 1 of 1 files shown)


Yagoor:
Good catch. I did perform the check every time before accessing the `in` pointer and moving to the pointer, as in lines 381 and 385, but I overlooked the fact that when peeking the in pointer, I should also perform the boundary check.

Yagoor (approved):

nvt (approved):

PR #2937, opened 2024-07-11
Added additional snmp_packet->used verification check in snmp_message_decode

OwenCochell:
The `snmp_message_decode` function does not properly verify the packet size when de-referencing the `in` pointer, allowing for an out-of-bounds read of 1 byte to occur in some scenarios. I originally disclosed this issue via email and was instructed to file a PR.

This PR fixes the issue by adding a check that will return if `snmp_packet->used == 0` before the de-reference, which will prevent this out of bounds read from occurring.


Yagoor:
Good catch, same explanation as #2936, I overlooked boundary checks when peeking into the `in` pointer.

Yagoor (approved):

nvt (approved):

PR #2952, opened 2024-09-04
replaced leshan jar for better DTLS testing support

joakimeriksson:
Just a quick test for updating the leshan jar file to the latest one - supporting DTLS and the latest features.


joakimeriksson:
Will fail due to missing python execution environment. Will fix that later (after 5.0 release).

PR #2961, opened 2024-09-18
Add ctimer API to read start and expiration time, and reset with new interval

nfi:


nvt (approved):

os/sys/ctimer.h
```diff
@@ -151,6 +164,33 @@ ctimer_set(struct ctimer *c, clock_time_t t, void (*f)(void *), void *ptr)
...
 
+/**
+ * \brief      Get the expiration time for the callback timer.
+ * \param c    A pointer to the callback timer
+ * \return     The expiration time for the callback timer.
+ *
+ *             This function returns the expiration time for a callback timer.
+ */
+static inline clock_time_t
+ctimer_expiration_time(struct ctimer *c)
```
pjonsson, line 175: Can the arguments to the getter functions be const?

nfi: Yes, but would require that also the `etimer` API is updated to be const.

PR #2967, opened 2024-09-24
Remove unused functions uip_log() and log_message()

nfi:


arch/platform/sky/platform.c
```diff
@@ -85,9 +85,6 @@ force_float_inclusion()
   return __fixsfsi + __floatsisf + __mulsf3 + __subsf3;
 }
 #endif
-/*---------------------------------------------------------------------------*/
-void uip_log(char *msg) { puts(msg); }
```
pjonsson, line 89: This function is still declared in `uipopt.h`, so should probably be removed from that file too.

nvt (approved):

arch/platform/sky/platform.c
nfi: This was removed in #2966.

PR #2970, opened 2024-10-02
Fix CoAP option boundary checking.

nvt:
The CoAP message parsing code is currently checking the boundary of options even when an option will not be read, causing an invalid decision to report that some messages are too small.

This PR introduces a macro for option boundary checking, and puts it only immediately before the statement that reads the option byte.

### DIFF
--- a/os/net/app-layer/coap/coap.c
+++ b/os/net/app-layer/coap/coap.c
@@ -420,6 +420,16 @@ coap_serialize_message(coap_message_t *coap_pkt, uint8_t *buffer)
 coap_status_t
 coap_parse_message(coap_message_t *coap_pkt, uint8_t *data, uint16_t data_len)
 {
+#define CHECK_OPTION_BOUNDARY(byte_count)                               \
+  do {                                                                  \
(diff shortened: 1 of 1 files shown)


os/net/app-layer/coap/coap.c
```diff
@@ -497,57 +502,34 @@ coap_parse_message(coap_message_t *coap_pkt, uint8_t *data, uint16_t data_len)
...
-      return BAD_REQUEST_4_00;
-    }
 
     if(option_delta == 13) {
+      CHECK_OPTION_BOUNDARY(1);
       option_delta += current_option[0];
       ++current_option;
     } else if(option_delta == 14) {
       option_delta += 255;
+      CHECK_OPTION_BOUNDARY(2);
```
pjonsson, line 512: Isn't it better to consistently check the boundary as soon as it is clear that will be required? Will make it easier to review since boundary check comes first, and also happens to save a little bit of CPU when the check fails.

nfi (approved):

PR #2971, opened 2024-10-03
Add support for macOS utun for the native platform and the native border-router

nfi:
- Combine TUN handling for the native platform and the native rpl-border-router
- Import support for macOS utun and MTU size from tools/serial-io/tunslip6
- Add arguments to specify tun device, subnet prefix, and MTU for the native platform


os/services/rpl-border-router/native/tun-bridge.c
```diff
@@ -33,166 +33,71 @@
...
-    /* Error message handled by caller */
-    return -1;
+  /* Optional delay between outgoing packets */
+  /* Base delay times number of 6lowpan fragments to be sent */
+  /* delaymsec = 10; */
+  if(delaymsec) {
+    struct timeval tv;
+    int dmsec;
+    gettimeofday(&tv, NULL);
+    dmsec = (tv.tv_sec - delaystartsec) * 1000 + tv.tv_usec / 1000 - delaystartmsec;
```
pjonsson, line 63: Can `int` be squeezed in at the start of the line without forcing line-wrapping?

arch/cpu/native/net/tun6-net.c
```diff
@@ -95,35 +104,144 @@ ssystem(const char *fmt, ...)
...
+/*---------------------------------------------------------------------------*/
+static bool
+set_default_prefix(const char *prefix)
+{
+  char *ipaddr = strdup(prefix);
+  if(!ipaddr) {
+    return false;
+  }
+
+  uip_ip6addr_t prefix_addr;
```
pjonsson, line 116: I think this line can be pushed inside the if-statement a few lines below.

arch/cpu/native/net/tun6-net.c
```diff
@@ -95,35 +104,144 @@ ssystem(const char *fmt, ...)
...
+  return config_ipaddr;
+}
+/*---------------------------------------------------------------------------*/
+void
+tun6_net_set_prefix(const char *prefix)
+{
+  if(!set_default_prefix(prefix)) {
+    LOG_WARN("Failed to set default prefix %s\n", prefix);
+  }
+  config_ipaddr = prefix;
```
pjonsson, line 142: The variable is called prefix, should this assignment happen if the default prefix cannot be set?

arch/cpu/native/net/tun6-net.c
```diff
@@ -95,35 +104,144 @@ ssystem(const char *fmt, ...)
...
+CONTIKI_OPTION(TUN_PRIO + 2, { "mtu", required_argument, NULL, 0 },
+               mtu_callback, "interface MTU size\n");
+/*---------------------------------------------------------------------------*/
+static void
+cleanup(void)
 {
 #define TMPBUFSIZE 128
   /* Called from signal handler, avoid unsafe functions. */
   char buf[TMPBUFSIZE];
+#ifndef __APPLE__
```
pjonsson, line 210: This duplicates the entire strncat-block, but the start of the block is identical, and the "ifconfig device down" command is also identical. It would be nice to not duplicate that code.

arch/cpu/native/net/tun6-net.c
```diff
@@ -183,41 +298,110 @@ tun_alloc(char *dev, uint16_t devsize)
...
+  unsigned int tunif;
+
+  if(sscanf(config_tundev, "utun%u", &tunif) != 1 || tunif >= UINT8_MAX) {
+    fprintf(stderr, "tun_alloc: invalid utun interface specified: %s\n", config_tundev);
+    return -1;
+  }
+
+  LOG_INFO("Opening tun interface %s\n", config_tundev);
+
+  memset(&ctl_info, 0, sizeof(ctl_info));
```
pjonsson, line 342: Saves a line to initialize at declaration instead: `struct ctl_info ctl_info = { 0 };`.

arch/cpu/native/net/tun6-net.c
nfi: The native rpl-border-router should still work as long as the external network commands accept the configuration, and I tried to keep the behaviour close to the earlier behaviour. We could exit the application with an error instead. What do you think?

arch/cpu/native/net/tun6-net.c
```diff
@@ -183,41 +298,110 @@ tun_alloc(char *dev, uint16_t devsize)
...
 }
 #else
-/*---------------------------------------------------------------------------*/
 static int
-tun_alloc(char *dev, uint16_t devsize)
+tun_alloc(void)
 {
-  LOG_INFO("Opening: %s\n", dev);
-  return devopen(dev, O_RDWR);
+  char t[8 + sizeof(config_tundev)] = "/dev/";
```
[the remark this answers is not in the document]
nfi: Not exactly, but same functionality. I added, for example, explicit null termination.

arch/cpu/native/net/tun6-net.c
```diff
@@ -183,41 +298,110 @@ tun_alloc(char *dev, uint16_t devsize)
...
+
+/*
+ * Reference for utun on macOS:
+ * http://newosxbook.com/src.jl?tree=listings&file=17-15-utun.c
+ */
+static int
+tun_alloc(void)
+{
+  struct sockaddr_ctl sc;
+  struct ctl_info ctl_info;
```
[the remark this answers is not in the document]
nfi: I tried to keep most of the code unmodified from earlier, but I can fix this.

os/services/rpl-border-router/native/tun-bridge.c
nfi: No, the line was already a little too long. I broke it into two lines.

arch/cpu/native/net/tun6-net.c
nfi: Fixed.

arch/cpu/native/net/tun6-net.c
nfi: I kept the OS blocks separated because I thought it looked messy to mix them, but I agree that it is nice to avoid code duplication. I updated this.

arch/cpu/native/net/tun6-net.c
pjonsson: If this is the old behavior, I suggest we keep it, too close to the release.

nfi: Trying to set the default prefix at all is a new behaviour. There were no arguments to the native platform to set the prefix and tun device name earlier, but this allows multiple native applications with a network to run at the same time with their own sub-net. I leave it as is for now.

nvt (approved):

PR #1349, opened 2020-08-05
EDHOC module implementation [draft-ietf-lake-edhoc-05]

LPFraile:
The EDHOC module is an implementation of the [draft-ietf-lake-edhoc-01] which specifies Ephemeral Diffie-Hellman Over COSE (EDHOC), a very compact, and lightweight authenticated Diffie-Hellman key exchange with ephemeral keys that provide mutual authentication, perfect forward secrecy, and identity protection. The module includes the asymmetric key authentication feature by using static Diffie-Hellman keys and uses the EDHOC cypher suite 2 that consists of AES-CCM-16-64-128 as an AEAD algorithm, SHA-256 as a Hash algorithm, and the P-256 as ECDH curve.

The EDHOC module uses the MicroECC (https://github.com/kmackay/micro-ecc/tree/601bd11062c551b108adbb43ba99f199b840777c) and the "SHA" (Secure Hash Algorithm) Contiki implementation of Oriol Piniol (https://github.com/oriolpinol/contiki/tree/master/apps/ecc). It also, uses the "cbor" Contiki-NG implementation from "group-oscore" (https://github.com/Gunzter/contiki-ng/tree/group_oscore/os/net/app-layer/coap/oscore-suppor) where the "cbor_put_bytes_identifier()" function has been added to cover the EDHOC protocol requirements. Additionally, the EDHOC module uses the COSE_Encrypt0 struct encryption from [RFC8152] for cryptography as well as the COSE_key format. The required COSE functionality has been implemented in a lightweight module under "os/net/security/cose".

The EDHOC implementation has been tested using "example\edhoc-test\edhoc-test-client.c" and "example\edhoc-test\edhoc-test-server.c" on Cooja and Zoletia RE-Mote platforms. In the latter case, it is mandatory to either set the 32Mhz Clock or else disable the watchdog so that the [MicroECC] library runs without modifications.

PR #2993, opened 2024-12-20
Add support for the EDHOC protocol

nvt:
This PR adds support for EDHOC, an implementation of Ephemeral Diffie-Hellman Over COSE (EDHOC) (IETF RFC9528).

It is based on work by Lidia Pocero (original implementation), Christos Koulamas, Peter A Jonsson, Rikard Höglund, Marco Tiloca, Joel Höglund, Niclas Finne, and myself.

It updates and replaces PR #1349.


rikard-sics:
I wanted to write a short comment in this PR about recent interoperability testing I did with this implementation of EDHOC. This was done during the IETF 122 Hackathon.

I successfully tested the EDHOC client from this codebase towards the following two EDHOC implementations:
- [Python implementation](https://github.com/chrysn/aiocoap) with demo server available at demo.coap.amsuess.com
- [Java implementation](https://github.com/rikard-sics/contiki-ng/tree/edhoc)

The configuration used for the testing was the following:
```
Credential type: CCS
ID cred type: KID
Cipher suite: 2
Method: 3 (Java & Python)
Method: 0 (Java)
```

In order to test successfully with the Python implementation I submitted a small fix to this codebase, which has been incorporated.

nvt:
Thanks for doing the interoperability testing at the IETF meeting, @rikard-sics! It is good to know that the tests were successful with method 0 and 3.

I can also mention that the next steps for this work-in-progress PR is to switch to the new Contiki-NG CBOR implementation and to do some general source code improvements before it is ready for review.

kkrentz:
On constrained devices, the elliptic curve cryptography (ECC) will block other protothreads for a long time. This may cause the watchdog to reboot or the radio to miss duty cycles. The ECC API suggested in #2898 allows other prothreads to make progress while executing ECC operations.

os/lib/cbor.c
```diff
@@ -465,5 +465,17 @@ cbor_read_map(cbor_reader_state_t *state)
   return read_array_or_map(state);
 }
 /*---------------------------------------------------------------------------*/
+const uint8_t *
+cbor_get_position(cbor_reader_state_t *state)
+{
+  return state ? state->cbor : NULL;
```
kkrentz: Allowing `state` to be NULL is inconsistent with the other functions.

os/lib/cbor.c
```diff
@@ -477,5 +477,76 @@ cbor_get_remaining(cbor_reader_state_t *state)
   return state ? state->cbor_size : 0;
 }
 /*---------------------------------------------------------------------------*/
+bool
+cbor_skip_next(cbor_reader_state_t *state)
+{
+  cbor_major_type_t major_type;
```
kkrentz: These declarations could be moved closer to where they are used.

os/net/security/edhoc/ecc-uecc.h
```diff
@@ -0,0 +1,96 @@
...
+ * \param private_key The private key used for ECC shared secret calculation
+ * \param ikm Output buffer where the generated IKM will be stored
+ * \param crv The ECC curve information used for the operation
+ * \return true for success or false for failure
+ *
+ * This function generates the IKM by performing ECC point multiplication
+ * using the public point coordinates (gx, gy) and the private key. The public point is first uncompressed,
+ * and the ECC shared secret is computed using the uECC library. The result is stored in the output buffer `ikm`.
+ */
+bool uecc_generate_ikm(const uint8_t *gx_in, const uint8_t *gy_in,
```
kkrentz: It would be nicer to use descriptive variable names instead of gx_in, gy_in.

os/net/security/edhoc/ecc-uecc.c
```diff
@@ -0,0 +1,139 @@
...
+/*----------------------------------------------------------------------------*/
+uint8_t
+uecc_generate_key(ecc_key_t *key, ecc_curve_t curve)
+{
+  int err = 0;
+  watchdog_periodic();
+  uECC_set_rng(&RNG);
+  uint8_t public_key[64];
+  watchdog_periodic();
+  err = uECC_make_key(public_key, key->priv, curve.curve);
```
kkrentz: Couldn't you reuse `lib/ecc.h` here, at least by busy-waiting for the protothread to exit?

os/net/security/cose/cose.c
```diff
@@ -0,0 +1,451 @@
...
+#include "lib/cbor.h"
+#include "cose.h"
+#include "uECC.h"
+#include <string.h>
+
+#include "sys/log.h"
+#define LOG_MODULE "COSE"
+#define LOG_LEVEL LOG_LEVEL_EDHOC
+
+MEMB(encrypt0_storage, cose_encrypt0_t, 1);
```
kkrentz: Is it really necessary for COSE to be stateful. For my purposes, I, e.g., defined this interface:

```
/**
 * Implements AES-CCM-16-64-128 as specified in RFC 8152.
 *
 * @param aad     Data to authenticate.
 * @param aad_len Length of @p aad in bytes.
 * @param m       Data to encrypt.
 * @param m_len   Length of @p m in bytes.
 * @param key     Symmetric key to use.
 * @param nonce   Nonce to use.
 * @param forward @c true when securing and @c false when unsecuring.
 *
 * @return        0 on error and non-zero on success.
 */
int cose_encrypt0_aes_ccm_16_64_128(
    const uint8_t *aad, size_t aad_len,
    uint8_t *m, size_t m_len,
    const uint8_t key[COSE_ALGORITHM_AES_CCM_16_64_128_KEY_LEN],
    const uint8_t nonce[COSE_ALGORITHM_AES_CCM_16_64_128_IV_LEN],
    bool forward);
```

nvt:
Thanks for the comments, @kkrentz -- they should now be addressed. Morever, CBOR changes were split into a separate PR.

nfi (approved):

PR #3000, opened 2025-01-13
CBOR library

kkrentz:
This PR adds a CBOR library. Unlike the CBOR library suggested in #2993, this one here supports parsing, checks for buffer overflows, and does not ask for the number of elements/entries when opening an array/map. Like in QCBOR, it is unnecessary to check for errors after each write operation. This is because once an error state is entered, subsequent write operations are being ignored. CBOR is written in reverse in order to avoid asking for the number of elements/entries, while also avoiding rearranging already written data.

### DIFF
--- a/os/lib/cbor.c
+++ b/os/lib/cbor.c
@@ -1,0 +1,428 @@
+/*
+ * Copyright (c) 2018, SICS, RISE AB
+ * Copyright (c) 2023, Uppsala universitet
+ * Copyright (c) 2024, Siemens AG
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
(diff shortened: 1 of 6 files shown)


nvt (approved):

tests/08-native-runs/16-cbor/test-cbor.c
```diff
@@ -0,0 +1,111 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * This file is part of the Contiki operating system.
+ *
+ */
+
+/**
+ * \file
+ *         Testing CCM*-MICs
```
nvt: This comment seems to be a left-over from some other file.

tests/08-native-runs/16-cbor/test-cbor.c
```diff
@@ -0,0 +1,111 @@
...
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * This file is part of the Contiki operating system.
```
nvt: This line is not needed in the copyright header, especially since we are now Contiki-NG.

kkrentz:
Thank you @nvt for your review. I made some more minor changes to match Contiki-NG's style. I am currently thinking about another slight adjustment to `cbor_stop_reader`.

nvt:
We have discussed this a little bit more internally, and thought that it would be more intuitive to have an append order for insertions. Otherwise it will be some extra work to port applications that use other CBOR libraries with append order. For example, there has been some work done to switch the EDHOC PR to use zcbor instead. Would it be viable to add append-order to this PR?

kkrentz:
There could be a separate set of functions for writing in append-order. A simple implementation is to ask for the number of elements/entries when opening an array/map. However, that simple implementation is error-prone. Think, e.g., of optional elements/entries being written in if statements. A hard implementation is to insert the actual numbers at the correct locations once arrays and maps are being closed like in QCBOR. While simple from a user's perspective, this consumes more stack memory, requires `memmove`s, and complicates the library code.

kkrentz:
Implemented "append-mode"

joakimeriksson:
Looks good, but what about doing a minimal change to the API naming (similar to QCBOR) init and finish/end (instead of init/stop) for the reader/writer, and open/close (instead of open/wrap) for the arrays and maps.

os/lib/cbor.c
```diff
@@ -0,0 +1,434 @@
...
+/*---------------------------------------------------------------------------*/
+void
+cbor_break_writer(cbor_writer_state_t *state)
+{
+  state->buffer = NULL;
+  state->buffer_size = 0;
+}
+/*---------------------------------------------------------------------------*/
+const uint8_t *
+cbor_peek_output(cbor_writer_state_t *state)
```
nfi: I understand what this function does but I am unsure of what it can be used for. Can you give an example of when someone would like to use this function?

os/lib/cbor.c
```diff
@@ -0,0 +1,434 @@
...
+void
+cbor_init_reader(cbor_reader_state_t *state,
+                 const uint8_t *cbor, size_t cbor_size)
+{
+  state->cbor = cbor;
+  state->cbor_size = cbor_size;
+}
+/*---------------------------------------------------------------------------*/
+cbor_major_type_t
+cbor_read_next(cbor_reader_state_t *state)
```
nfi: All read functions except for this consume data. What do you think about naming it `cbor_peek_next()` or similar?

nfi (approved):

os/lib/cbor.c
kkrentz: Done

os/lib/cbor.c
kkrentz: I had to compute a hash over a map while writing. For the time being, I removed this function.

PR #3017, opened 2025-03-17
Initial Implementation of RFC9009 DCO for RPL

dianadeac:
This pull request implements main elements of  [RFC 9009 ](https://www.rfc-editor.org/rfc/rfc9009.html): Efficient Route Invalidation into RPL classic. 

RFC9009 introduces a new route invalidation mechanism for RPL via Destination Cleanup Object (DCO) and DCO-ACK messages. Unlike the No-Path DAO, which originates from the node that changes parents and travels upward, the DCO originates from the node that detects the mismatch in the route (the common ancestor of the old and new parent). The DAO sent by the switching node to its new parent contains a flag also defined in RFC9009, the I (Invalidate previous route) in the Transit Information Option. This flag, if set, signals the common ancestor to issue a DCO. The DAO is propagated to the common ancestor of the old and new parent, which will send a DCO message downwards on the old path. 

The PR includes:
- support for the I flag in the transit information option
- the code for building, sending and handling DCO and DCO-ACK
- retransmission logic in case DCO-ACK is not received
- a simulation test case on a simple network to validate logic.


joakimeriksson:
Hi, are you planning to update this anytime soon? It would be great to see some tests also (if possible).

dianadeac:
Hello, I was on a longer personal leave but I will resume work on the DCO invalidation. Will include tests too

dianadeac:
@joakimeriksson I finished the logic that was missing and added a simple simulation test. Can you please have a look at it?

PR #3020, opened 2025-03-25
CBOR: add API to read/write signed values

nfi:


os/lib/cbor.c
```diff
@@ -349,6 +360,31 @@ cbor_read_unsigned(cbor_reader_state_t *state, uint64_t *value)
...
 /*---------------------------------------------------------------------------*/
+cbor_size_t
+cbor_read_signed(cbor_reader_state_t *state, int64_t *value)
+{
+  if(!state->cbor_size) {
+    return CBOR_SIZE_NONE;
+  }
+
+  uint64_t unsigned_value;
+  uint8_t type = *state->cbor & 0xE0;
```
kkrentz: To be more descriptive, I would suggest declaring `type` as `cbor_major_type_t`

os/lib/cbor.c
```diff
@@ -349,6 +360,31 @@ cbor_read_unsigned(cbor_reader_state_t *state, uint64_t *value)
...
+  cbor_size_t size = cbor_read_unsigned(state, &unsigned_value);
+  if(size == CBOR_SIZE_NONE) {
+    return CBOR_SIZE_NONE;
+  }
+
+  if(type == 0x00) {
```
kkrentz: To avoid magic numbers, I would suggest using the existing constants `CBOR_MAJOR_TYPE_xxxxxx`. It may also make sense to use a `switch` here since this typically compiles to a more compact binary.

os/lib/cbor.c
```diff
@@ -349,6 +360,31 @@ cbor_read_unsigned(cbor_reader_state_t *state, uint64_t *value)
...
+    *value = (int64_t)unsigned_value;
```
kkrentz, line 375: These casts seem to fail when the unsigned value is larger than INT64_MAX.

kkrentz (changes requested):

os/lib/cbor.c
nfi: Good suggestion!

os/lib/cbor.c
nfi: Yes, you are right. I add a check and return `CBOR_SIZE_NONE` if the value is too large.

kkrentz: Analogously, I suppose that the negative value can get smaller than INT64_MIN.

tests/08-native-runs/16-cbor/test-cbor.c
```diff
@@ -38,48 +38,141 @@
...
     cbor_size = cbor_end_writer(&writer);
   }
 
+  LOG_DBG("CBOR (%zu bytes): ", cbor_size);
+  LOG_DBG_BYTES(buffer, cbor_size);
+  LOG_DBG_("\n");
+
   UNIT_TEST_ASSERT(cbor_size);
 
+  if(1) {
```
kkrentz: `if(1)` could be removed

tests/08-native-runs/16-cbor/test-cbor.c
```diff
@@ -38,48 +38,141 @@
...
 
+/* Log configuration */
+#include "sys/log.h"
+#define LOG_MODULE "test-cbor"
+#define LOG_LEVEL LOG_LEVEL_NONE
+
 PROCESS(test_process, "test");
 AUTOSTART_PROCESSES(&test_process);
 
+#define TEXT_DATA "Hello!"
```
kkrentz: This string could become a `const` like the other example values.

kkrentz (approved):

nvt (approved):

PR #3041, opened 2025-06-01
CC2538: Fix compilation with Arm GNU Toolchain 14.2.Rel1

kkrentz:
This PR reorganizes `rom-util.ch` to resolve this compilation error:

```
rom-util.h:124:16: error: array subscript 0 is outside array bounds of 'const struct rom_util_api[0]'
```

This should also resolve [discussions/1567](https://github.com/contiki-ng/contiki-ng/discussions/1567).

### DIFF
--- a/arch/cpu/cc2538/dev/rom-util.h
+++ b/arch/cpu/cc2538/dev/rom-util.h
@@ -94,32 +94,32 @@ struct rom_util_api {
 /** \name Pointer to the ROM utility function library API table
  * @{
  */
-#define ROM_UTIL_API            ((struct rom_util_api *)0x00000048)
+extern const struct rom_util_api *const rom_util_api;
 /** @} */
 /*---------------------------------------------------------------------------*/
 /** \name ROM utility function library API accessor macros
  * @{
  */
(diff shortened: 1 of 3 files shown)


nvt (approved):

arch/cpu/cc2538/dev/rom-util.h
```diff
@@ -57,7 +57,7 @@
 /** \name ROM utility function library API access table structure
  * @{
  */
-struct rom_util_api {
+typedef struct rom_util_api_t {
```
nvt: The base struct should be named "rom_util_api" to distinguish it from the typedef name "rom_util_api_t".

nfi (approved):

arch/cpu/cc2538/dev/rom-util.h
kkrentz: Using equal names is also recommended sometimes, e.g., by the Indian Hill Style Guide. A benefit is the reservation of that struct name. Other styles suggest leaving structs unnamed (just `typedef struct { ...`) or never typedefing structs. Meanwhile we have all flavours in the code base.

PR #3054, opened 2025-07-07
Bitrev on cc13xx for standard compliant 802.15.4g

joakimeriksson:
Due to 802.15.4g having one bit-order in the header and another in the payload, we have had a non-compliant radio driver. This PR will allow configuration of bit order for the payload by setting RADIO_PAYLOAD_BIT_REVERSE for the radio. 

NOTE: It is a bit of a reverse library included, so it should be easy to add to other radios.

### DIFF
--- a/os/contiki-default-conf.h
+++ b/os/contiki-default-conf.h
@@ -58,6 +58,14 @@
 #ifndef QUEUEBUF_CONF_NUM
 #define QUEUEBUF_CONF_NUM 8
 #endif /* QUEUEBUF_CONF_NUM */
+
+/* Enable 802.15.4g payload bit-reversal (LSB⇄MSB) on TX/RX */
+#ifdef RADIO_CONF_PAYLOAD_BIT_REVERSE
+#define RADIO_PAYLOAD_BIT_REVERSE RADIO_CONF_PAYLOAD_BIT_REVERSE
+#else
+#define RADIO_PAYLOAD_BIT_REVERSE 0
+#endif
+
 /*---------------------------------------------------------------------------*/
 /* uIPv6 configuration options.
  *
--- a/os/lib/bitrev.c
+++ b/os/lib/bitrev.c
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2023, RISE Research Institutes of Sweden AB
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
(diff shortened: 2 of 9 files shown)


os/lib/bitrev.c
```diff
@@ -0,0 +1,90 @@
...
+ * \file
+ *         Bit reversal library implementation
+ * \author
+ *         Joakim Eriksson <joakim.eriksson@ri.se>
+ */
+
+#include "bitrev.h"
+#include <string.h>
+/*---------------------------------------------------------------------------*/
+/**
```
nfi: This is an internal constant that Doxygen does not need to include in the documentation.
```suggestion
/*
```

tests/08-native-runs/18-bitrev-test.sh
```diff
@@ -0,0 +1,40 @@
...
+CONTIKI=$1
+
+# Example code directory
+CODE_DIR=$CONTIKI/tests/08-native-runs/18-bitrev/
+CODE=test-bitrev
+
+# Starting Contiki-NG native node
+echo "Starting native node for bit reversal test"
+make -C $CODE_DIR TARGET=native > make.log 2> make.err
+$CODE_DIR/$CODE.native > $CODE.log 2> $CODE.err &
```
nfi: Contiki-NG no longer copies generated artifacts outside build directory by default.
```suggestion
$CODE_DIR/build/native/$CODE.native > $CODE.log 2> $CODE.err &
```

tests/08-native-runs/18-bitrev/test-bitrev.c
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * \file
+ *         Unit tests for bit reversal library
+ * \author
+ *         Joakim Eriksson <joakim.eriksson@ri.se>
+ */
+
+#include "contiki.h"
+#include "services/unit-test/unit-test.h"
```
nfi: No need to use full path when module is included in build.
```suggestion
#include "unit-test.h"
```

os/contiki-default-conf.h
```diff
@@ -214,4 +214,15 @@
 #define SICSLOWPAN_CONF_COMPRESSION SICSLOWPAN_COMPRESSION_IPHC
 #endif /* SICSLOWPAN_CONF_COMPRESSION */
 
+/*---------------------------------------------------------------------------*/
+/* Library configurations
```
nfi: Sounds a little odd with  `Library configurations` for `RADIO_CONF_PAYLOAD_BIT_REVERSE` ?

nfi (approved):

PR #3058, opened 2025-07-12
Consolidate instances of random.c

kkrentz:
This PR adopts non-cryptographic PRNGs from PractRand. `srand()` and `rand()` from `<stdlib.h>` could also serve this purpose, but may raise an issue explained [here](https://dunkels.com/adam/rand-may-call-malloc/) and generally provide randomness of lower quality. This PR obviates serveral platform-specific versions of `random.c`. In particular,  CC26x0-CC13x0 no longer uses its TRNG, thus fixing #2332. The CC2538, by contrast, continues to use a hardware-accelerated PRNG by default.


arurke:
Lovely cleanup @kkrentz.

One question: By tying `random_init()` into CSPRNG, the CSPRNG becomes mandatory on the platforms. I am assuming this drags with it a non-negligible increase in, at least, code-size. Are there any significant benefits to that approach that would outweigh this drawback, as opposed to initializing the two modules independently as before? In my mind, one of the strengths of Contiki-NG is the minimal resources needed.

kkrentz:
Platforms do not have to feed the CSPRNG and can continue to call `random_init` by themselves. This is useful when starting off with a platform. When the CSPRNG is being fed, the CSPRNG now takes care of generating a seed for `random_init`, which obviates much code. Conversely, if we allowed for disabling the CSPRNG, it would still be necessary to generate a seed for `random_init` in configurations where the CSPRNG is disabled. I think that maintaining both of these configurations is not worth the effort since a CSPRNG or the underlying block cipher is usually required anyway.

kkrentz:
That being said, I made the following compromise between code reuse and minimum footprint: I revised `random_init` to automatically generate a seed with the CSPRNG if enabled and else fall back on using the MAC address as seed.

kkrentz:
contiki-ng/example-lwm2m-standalone#6 would have to be merged first in order for `tests/18-coap-lwm2m` to pass.

.gitmodules
```diff
@@ -15,7 +15,7 @@
 	url = https://github.com/contiki-ng/cooja.git
 [submodule "tests/18-coap-lwm2m/example-lwm2m-standalone"]
 	path = tests/18-coap-lwm2m/example-lwm2m-standalone
-	url = https://github.com/contiki-ng/example-lwm2m-standalone.git
+	url = https://github.com/kkrentz/example-lwm2m-standalone.git
```
nfi: https://github.com/contiki-ng/example-lwm2m-standalone/pull/6 has been merged so please restore the original submodule.

nvt (approved):

PR #3064, opened 2025-07-14
nrf52840: Feed CSPRNG with the built-in TRNG

kkrentz:
This PR is to feed the CSPRNG on nrf52840 targets. The implementation is analogous to [random.c](https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/nrf52840/dev/random.c). As I do not have an nrf52840 target at hand, someone else should test this PR, e.g., by running the [CSPRNG example](https://github.com/contiki-ng/contiki-ng/tree/develop/examples/libs/csprng).

### DIFF
--- a/arch/platform/nrf52840/platform.c
+++ b/arch/platform/nrf52840/platform.c
@@ -37,10 +37,12 @@
 
 #include "sdk_config.h"
 #include "nrfx_gpiote.h"
+#include <nrfx_rng.h>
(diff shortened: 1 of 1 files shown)


nvt:
The CSPRNG example works on the nrf52840 target.

nvt (approved):

nfi (approved):

PR #3065, opened 2025-07-14
nrf: Feed CSPRNG with the built-in TRNG

kkrentz:
This PR is to feed the CSPRNG on nrf targets. The implementation is analogous to [random-arch.c](https://github.com/contiki-ng/contiki-ng/blob/develop/arch/cpu/nrf/os/random-arch.c). As I do not have an nrf target at hand, someone else should test this PR, e.g., by running the [CSPRNG example](https://github.com/contiki-ng/contiki-ng/tree/develop/examples/libs/csprng).

### DIFF
--- a/arch/platform/nrf/platform.c
+++ b/arch/platform/nrf/platform.c
@@ -44,6 +44,7 @@
 #include "dev/button-hal.h"
 #include "dev/leds.h"
 #include "dev/serial-line.h"
+#include "lib/csprng.h"
 
 #include "random.h"
 #include "int-master.h"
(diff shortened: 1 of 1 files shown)


nvt:
The CSPRNG example works on the nrf target with BOARD=nrf52840/dongle. When using BOARD=nrf5340/dk/application, the NRF_RNG parameter is not defined.

nvt (approved):

kkrentz:
At least, including `nrfx_config.h` instead of `nrfx.h` results in `NRF_RNG` being defined on `BOARD=nrf5340/dk/network`, too. On `BOARD=nrf5340/dk/application`, however, `NRF_RNG` is still not defined because that core cannot access the RNG.

PR #3084, opened 2025-10-31
Coffee: Avoid invalid header causing endless loops

arurke:
We had experiences where a corrupt file header could cause endless loops in coffee. In our cases it was caused by the header `max_pages` field that was either negative or extremely large. This in turn caused scans through the flash to jump backwards when meeting this header, thus scanning indefinitely.

Without being too familiar with the internal workings of coffee I added some handling I that I believe is benign, but would love to get some feedback one it.

### DIFF
--- a/os/storage/cfs/cfs-coffee.c
+++ b/os/storage/cfs/cfs-coffee.c
@@ -410,6 +410,18 @@ next_file(coffee_page_t page, struct file_header *hdr)
   } else if(HDR_ISOLATED(*hdr)) {
     return page + 1;
   }
+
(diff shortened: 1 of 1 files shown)


nvt:
It is indeed a step in the right direction with such a check, but the next page after a corrupt header might contain regular file contents.

For a while during the summer, I was working on a branch to upgrade/modernize Coffee, where the header validation was enhanced, but it never got ready for a PR. I can perhaps make a PR with these changes soon.

arurke:
Replaced by #3091

PR #3090, opened 2025-11-10
Update TSCH-and-6TiSCH.md

msinis:
Fix broken link to TSCH tutorial


arurke:
Link works fine when browsing on the documentation webpage: https://docs.contiki-ng.org/en/develop/doc/programming/TSCH-and-6TiSCH.html Yet, it does not work when browsing the markdown files directly in the repo. I am not familiar with readthedocs, is this PR the way to go to make it work in both worlds? (if so, there are probably hundreds of places we can update).

PR #3096, opened 2025-11-22
lib: Expose ARRAY_LENGTH macro to other modules

kkrentz:
This PR exposes the `ARRAY_LENGTH` macro to avoid duplicating it. Like in the Linux kernel, I added a compile time check, which ensures that the provided parameter is an array.

### DIFF
--- a/os/dev/button-hal.h
+++ b/os/dev/button-hal.h
@@ -81,6 +81,7 @@
 /*---------------------------------------------------------------------------*/
 #include "contiki.h"
 #include "dev/gpio-hal.h"
+#include "sys/array-length.h"
 #include "sys/clock.h"
 #include "sys/ctimer.h"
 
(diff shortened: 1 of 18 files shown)


arurke:
:+1: LGTM, and a handy addition

arurke (approved):

joakimeriksson:
What about moving the macro into cc.h so that we avoid a new file? (with a check if not already defined around it).

kkrentz:
This is what I had initially. My concern was that `cc.h` is intended for handling compiler particularities. The `CC_MUST_BE_ARRAY` fits perfectly in `cc.h`. Besides, the compliant name would then be `CC_ARRAY_LENGTH`.

nfi:
The macro `ARRAY_LENGTH` (or possibly `CC_ARRAY_LENGTH`) is a nice addition. Although I also think `cc.h` was originally intended for compiler-specific particulars, it already includes macros such as `MIN`, `MAX`, `ABS`, `BOUND`, etc., and IMHO it would be nice to avoid adding a new header file for a single macro if possible.

nfi (approved):

PR #3097, opened 2025-11-22
fix(platform): fix LED_CONF for cooja plaftorm and nrf52840 dongle board

andimatteo:
**Description**
This PR updates the LED pin definitions for the nRF52840 Dongle and Cooja mote to match the actual hardware layout.

**Changes**
- Corrected LED pin mappings in `nrf52840-board-def.h`.
- Corrected LED pin mappings in `arch/platform/cooja/contiki-conf.h`.

### DIFF
--- a/arch/platform/cooja/contiki-conf.h
+++ b/arch/platform/cooja/contiki-conf.h
@@ -149,8 +149,8 @@ typedef unsigned short uip_stats_t;
 /*---------------------------------------------------------------------------*/
 /* Virtual LED colors */
 #define LEDS_CONF_COUNT                  3
-#define LEDS_CONF_GREEN                  1
-#define LEDS_CONF_RED                    2
-#define LEDS_CONF_YELLOW                 4
+#define LEDS_CONF_GREEN                  0
+#define LEDS_CONF_RED                    1
+#define LEDS_CONF_YELLOW                 2
 /*---------------------------------------------------------------------------*/
 #endif /* CONTIKI_CONF_H_ */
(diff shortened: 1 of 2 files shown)


kkrentz:
I tested `leds_toggle(LEDS_YELLOW);` in Cooja. With your changes, the red LED toggles, whereas the current version works.

andimatteo:
`leds_toggle(LEDS_YELLOW)` shouldn't be the right way to toggle yellow LED otherwise comments and `os/dev/leds.h` and `.c` should be changed accordingly.

From the definition of leds colors: (`os/dev/leds.h:125`)
```
/* LED colour to number mappings. Applicable to both APIs */
#ifdef LEDS_CONF_RED
#define LEDS_RED LEDS_CONF_RED
#else
#define LEDS_RED LEDS_COLOUR_NONE
#endif
```

So LED_RED, YELLOW and so on should be LEDs **numbers**.

From the declaration of `leds_toggle` (`os/dev/leds.h:299`)
```
/**
 * \brief Toggle multiple LEDs
 * \param leds The leds to be toggled as an OR mask
 *
 * The \e led argument should be a bitwise mask of the LEDs to be changed.
 * For example, to turn on LEDs 1 and 3, you should pass
 * LED_NUM_TO_MASK(LED_L1) | LED_NUM_TO_MASK(LED_L3) = 1 | 4 = 5
 *
 * This function will not change the state of other LEDs.
 */
void leds_toggle(leds_mask_t leds);
```

So it takes an LED **mask** as input. The conversion from LED **number to mask** is done via:
```
/**
 * \brief Convert a LED number to a mask representation
 * \param l The pin number (normally a variable of type leds_num_t)
 * \return An OR mask of type leds_mask_t
 */
#define LEDS_NUM_TO_MASK(l) (1 << (l))
```

So the right way of toggling yellow LED should be `leds_toggle(LEDS_NUM_TO_MASK(LEDS_YELLOW));`

Please let me know if something isn't right.

kkrentz:
The documentation is misleading. Judging by how the `LEDS_<COLOUR>` macros are defined and used, they are bit masks already.

PR #3118, opened 2026-02-01
nfr52: always flash with `--recover`

rmburg:
Flashing my nRF52840 devkit with `make node.upload` always used to fail until I manually ran `nrfutil device recover`.
This had to be repeated at least after every power cycle.

With this PR, this is no longer necessary. It works for me, but I'm not sure if this change has any unintended side-effects.
I'm leaving it here for the maintainers to decide.

### DIFF
--- a/arch/cpu/nrf52840/Makefile.nrf52840
+++ b/arch/cpu/nrf52840/Makefile.nrf52840
@@ -137,7 +137,7 @@ endif
 ifeq (, $(shell which $(NRFJPROG)))
 	$(error Could not find nrfjprog "$(NRFJPROG)", please install it)
 else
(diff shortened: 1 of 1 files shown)


nfi:
Recover causes a full flash wipe, which takes longer and erases any application data stored in flash. What do you think about adding another make rule, for example, `upload-recover` that uploads using recover, but leaves the old 'upload' rule as is? Users can then select which to use.

rmburg:
Thanks for the input! A separate rule sounds like a good idea.

nfi:
#3183 adds support for this, but via a variable instead of a separate rule: `make app.upload NRF_RECOVER=1`.

If you want to always use recover, you can add `NRF_RECOVER=1` to your project Makefile, and all uploads in your project will use recover.

PR #3119, opened 2026-02-17
adding linkaddr u8 8 byte support macro function to construct linkaddr_t

DeepPG:
Added linkadr_u8 so we can now construct linkaddr_t, that makes code cleaner and more understandable.

It's my first contribution, I hope this is good enough to be added and I really appreciate any comments, feedbacks or suggestions to improve.

### DIFF
--- a/os/net/linkaddr.h
+++ b/os/net/linkaddr.h
@@ -71,6 +71,31 @@ typedef union {
 } linkaddr_t;
 
 /**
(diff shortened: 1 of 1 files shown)


joakimeriksson (approved):
Before we merge, can you squash this into a single commit?

DeepPG:
Recent commits passed, I had just squashed the commits and it seem there is 128 error, maybe it's not realted to my commits and it need rereview.

PR #3125, opened 2026-03-13
nrf5340: support communication in the app core using IPC with the net core

nvt:
This PR adds an IPC-based radio driver that enables networking on the nRF5340 application core by forwarding 802.15.4 radio operations to the network core over shared memory.

Previously, the nRF5340 application core could only build with nullradio_driver since the radio peripheral is exclusive to the network core. This PR adds a shared-memory IPC protocol that lets the application core run the complete network stack while the network core acts as a dedicated radio service.

Note that TSCH is not supported.


arch/cpu/nrf/net/nrf-ipc-mac.c
```diff
@@ -0,0 +1,204 @@
...
+  /* Build ACK: [Frame Control = ACK type, 0x00, Sequence Number]. */
+  ack[0] = FRAME802154_ACKFRAME;
+  ack[1] = 0;
+  ack[2] = frame[2];  /* DSN is byte 2 of the received frame. */
+
+  /* Disable CCA for ACK -- 802.15.4 requires ACKs sent without CCA. */
+  NETSTACK_RADIO.set_value(RADIO_PARAM_TX_MODE, 0);
+  NETSTACK_RADIO.send(ack, ACK_FRAME_LEN);
+  /* Re-enable CCA for normal transmissions. */
+  NETSTACK_RADIO.set_value(RADIO_PARAM_TX_MODE, RADIO_TX_MODE_SEND_ON_CCA);
```
nfi: I think this should restore the previous configuration if the radio already had send-on-cca disabled, and only disable using the `RADIO_TX_MODE_SEND_ON_CCA` flag in case RADIO_PARAM_TX_MODE supports more modes in the future.

arch/cpu/nrf/net/nrf-ipc-radio.c
```diff
@@ -0,0 +1,625 @@
...
+        LOG_ERR("Network core init timeout\n");
+        return 0;
+      }
+    }
+  }
+#else
+  {
+    clock_time_t start = clock_time();
+    while(!shm->net_ready) {
+      watchdog_periodic();
```
nfi: Is watchdog_periodic() inside this loop really needed?

arch/cpu/nrf/net/nrf-ipc-radio.c
```diff
@@ -0,0 +1,625 @@
...
+
+  shm->log.tail = tail;
+}
+/*---------------------------------------------------------------------------*/
+PROCESS_THREAD(ipc_radio_process, ev, data)
+{
+  static struct etimer rx_poll_timer;
+#ifndef TRUSTZONE_SECURE
+  int len;
+#endif
```
nfi: `len` can be declared where it is used to avoid this extra `#ifndef`.

arch/cpu/arm/cortex-m/trustzone/normal/tz-radio-normal.c
```diff
@@ -0,0 +1,217 @@
...
+  nr_off,
+  nr_get_value,
+  nr_set_value,
+  nr_get_object,
+  nr_set_object
+};
+/*---------------------------------------------------------------------------*/
+PROCESS_THREAD(tz_radio_process, ev, data)
+{
+  int len;
```
nfi: Why not declare `len` where the other variables are declared?

arch/cpu/arm/cortex-m/trustzone/tz-radio.c
```diff
@@ -0,0 +1,325 @@
...
+
+  memcpy(secure_buf, payload, payload_len);
+
+  return ipc_radio_driver.send(secure_buf, payload_len);
+}
+/*---------------------------------------------------------------------------*/
+CC_TRUSTZONE_SECURE_CALL int
+tz_radio_read(void *buf, unsigned short buf_len)
+{
+  static uint8_t secure_buf[TZ_RADIO_MAX_PAYLOAD];
```
nfi: `tz_radio_prepare()`, `tz_radio_send()`, and `tz_radio_read()` each has their own static buffer of TZ_RADIO_MAX_PAYLOAD bytes. Why are these needed instead of either using the stack or a single shared static buffer?

arch/cpu/arm/cortex-m/trustzone/tz-radio.c
```diff
@@ -0,0 +1,325 @@
...
+    memcpy(dest, secure_buf, size);
+  }
+
+  return result;
+}
+/*---------------------------------------------------------------------------*/
+CC_TRUSTZONE_SECURE_CALL radio_result_t
+tz_radio_set_object(radio_param_t param, const void *src, size_t size)
+{
+  static uint8_t secure_buf[TZ_RADIO_MAX_OBJECT];
```
nfi: Do `tz_radio_get_object()` and `tz_radio_set_object()` need their own static buffer, or could they share or use the stack?

arch/cpu/nrf/net/nrf-ipc-mac.c
```diff
@@ -0,0 +1,204 @@
...
+  /* Do not ACK frames that are themselves ACKs. */
+  if((frame[0] & 0x07) == FRAME802154_ACKFRAME) {
+    return;
+  }
+
+  /* Check the ACK request bit (bit 5 of FCF byte 0). */
+  if(!(frame[0] & FCF_ACK_REQUEST_BIT)) {
+    return;
+  }
+
```
nfi, line 99: I think we should only send ack if `RADIO_RX_MODE_AUTOACK` is set?

arch/cpu/nrf/net/nrf-ipc-radio.c
```diff
@@ -0,0 +1,625 @@
...
+}
+/*---------------------------------------------------------------------------*/
+static int
+ipc_radio_transmit(unsigned short transmit_len)
+{
+  int result;
+
+  LOG_DBG("TX %u bytes\n", transmit_len);
+
+  result = send_command(NRF_IPC_CMD_SEND, tx_frame_buf, transmit_len);
```
nfi, line 285: Do we not need to check that `transmit_len` <= `sizeof(tx_frame_buf)` ?

arch/cpu/nrf/net/nrf-ipc-radio.c
```diff
@@ -0,0 +1,625 @@
...
+   * in case the IPC interrupt path doesn't deliver them. */
+  etimer_set(&rx_poll_timer, CLOCK_SECOND / 20);
+
+  while(1) {
+    PROCESS_WAIT_EVENT();
+
+    drain_net_log();
+
+    if(etimer_expired(&rx_poll_timer)) {
+      etimer_reset(&rx_poll_timer);
```
nfi: Is there any reason to require exact timing, or would it not be better to use `etimer_restart()` instead, in case the CPU has been occupied with other work for a long time?

arch/cpu/nrf/os/dbg-arch.c
```diff
@@ -58,7 +58,51 @@
...
+  while(s && *s != 0) {
+    if(i >= len) {
+      break;
+    }
+    dbg_putchar(*s++);
+    i++;
+  }
+
+  return i;
+}
```
nfi: Could we not use the standard `dbg_send_bytes()` instead of adding another implementation if we define `flush()` as an empty macro, as is done on line 58?

examples/platform-specific/nrf/ipc-radio-service/project-conf.h
```diff
@@ -0,0 +1,53 @@
...
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+/*---------------------------------------------------------------------------*/
+#ifndef PROJECT_CONF_H_
+#define PROJECT_CONF_H_
+/*---------------------------------------------------------------------------*/
+/* Minimal logging for the network core. */
+#define LOG_CONF_LEVEL_MAIN LOG_LEVEL_INFO
```
nfi, line 35: Is `LOG_LEVEL_INFO` the minimal logging for the network core?

nfi (approved):

PR #3127, opened 2026-03-13
trustzone: bug fixes and security improvement

nvt:
nrf5340: fix TrustZone init ordering, atomicity, and SPU diagnostics
                                                                                  
Bug fixes and hardening for the nRF5340 TrustZone infrastructure:
                                                                                  
* Fix TZ API validation, init ordering, and poll signaling.                     
* Make tz_api_poll reentrancy guard and ns_poll_pending test-and-clear atomic.                                                                       
* Fix arr_ram array sizing bug in the SPU driver.                    
* Run system_reset_cfg and nvic_interrupt_enable during TZ setup, target all NVIC interrupts to NS by default, and add an SPU security-violation handler that captures diagnostics across reset.            
* Add a .noinit linker section so violation state survives reset.

### DIFF
--- a/arch/cpu/arm/cortex-m/trustzone/tz-api.c
+++ b/arch/cpu/arm/cortex-m/trustzone/tz-api.c
@@ -145,7 +145,7 @@ tz_api_println(const char *text, size_t len)
 }
 /*---------------------------------------------------------------------------*/
 bool
-tz_api_request_poll_from_ns(void)
+tz_api_request_ns_poll(void)
 {
   if(!initialized) {
     return false;
--- a/arch/cpu/arm/cortex-m/trustzone/tz-api.c
+++ b/arch/cpu/arm/cortex-m/trustzone/tz-api.c
@@ -46,7 +46,7 @@
 
 static struct tz_api tz_api;
 static bool initialized;
-static volatile unsigned poll_wait_counter;
+static volatile bool ns_poll_pending;
 
 /*---------------------------------------------------------------------------*/
 #include "sys/log.h"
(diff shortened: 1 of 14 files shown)


arch/cpu/arm/cortex-m/trustzone/tz-api.c
```diff
@@ -114,10 +114,17 @@ tz_api_poll(void)
...
 CC_TRUSTZONE_SECURE_CALL void
 tz_api_println(const char *text)
 {
-  printf("n> %s\n", text);
+  if(text == NULL ||
+     cmse_check_address_range((void *)text, TZ_API_PRINTLN_MAX_LEN,
+                              CMSE_NONSECURE) == NULL) {
+    return;
+  }
+  printf("n> %.*s\n", TZ_API_PRINTLN_MAX_LEN, text);
```
nfi: `dbg-arch.c` has a default buffer size of 256 bytes when calling `tz_api_println()` and should be updated to match this max length. However, what do you think about changing the API to provide the size instead of adding a max length constant, for example, `tz_api_println(const char *text, uint16_t length)`?

nfi (changes requested):

arch/cpu/arm/cortex-m/trustzone/tz-api.c
nvt: Yes, adding a max length argument would be better. More changes are coming also.

arch/cpu/arm/cortex-m/trustzone/tz-api.c
```diff
@@ -109,29 +119,47 @@ tz_api_poll(void)
...
+  bool pending = ns_poll_pending;
+  ns_poll_pending = false;
+  int_master_status_set(pending_primask);
+
+  if(pending) {
+    tz_api.request_poll();
+  }
 
   return process_nevents() > 0;
 }
```
nfi, line 131: Since the normal world will do a request poll if `tz_api_poll()` returns `true`, could this not be simplified as:

```
CC_TRUSTZONE_SECURE_CALL bool
tz_api_poll(void)
{
  static bool is_poll_running;

  /* [...] */
  if(__get_IPSR() != 0) {
    return false;
  }
  if(!initialized || is_poll_running) {
    return false;
  }
  is_poll_running = true;
  ns_poll_pending = false;

  process_num_events_t event_count = process_nevents();

  if(event_count > 0) {
    LOG_DBG("Processing %u event(s) at %" CLOCK_PRI "\n", event_count, clock_time());
  }

  while(event_count-- > 0) {
    process_run();
    watchdog_periodic();
  }

  is_poll_running = false;

  return ns_poll_pending || process_nevents() > 0;
}
```
?

nvt: Yes, that looks like a good simplification.

nfi (approved):

PR #3132, opened 2026-04-07
Add built-in NAT64 gateway to the native border router

nvt:
Adds a socket-based NAT64 gateway to the native border router, letting Contiki-NG IoT nodes reach the IPv4 Internet without an external tool setup. Enable with --nat64 on border-router.native.

* UDP and TCP translation between IPv6 (64:ff9b::/96) and IPv4, using host sockets on the BR side.
* DNS64 synthesis: A records returned by upstream resolvers are rewritten to AAAA under 64:ff9b::/96.
* TCP proxy with per-session seqstate, ACK flushing from the main-loop select, and chunked sends to survive 6LoWPAN fragmentation.
* New example examples/nat64/: a DNS client + UDP probe + HTTP GET via http-socket, plus a Cooja simulation (nat64-dns-cooja.csc) and run-nat64-br.sh helper so the whole path can be tested with no hardware.
* Documentation rewritten to cover the built-in gateway, configuration, and limitations.

The default NAT64 prefix is the well-known 64:ff9b::/96 (RFC 6052); to use a non-default prefix, call ip64_addr_set_prefix() at startup.


joakimeriksson:
Great contribution!

What about making it very clear that this remove the need for a Jool or similar NAT64 software? This is a bit of a "trick" to use the regular sockets instead of just doing NAT64 translation so I think we should make that more clear also - it will consume socket resources from the OS while normal NAT64 is not doing that. I think it might be worth mentioning these details (might be there already but I did not see it).

doc/getting-started/NAT64-for-Contiki-NG.md
```diff
@@ -1,35 +1,235 @@
...
+NAT64 allows IPv6-only IoT devices to communicate with IPv4 servers on
+the Internet. The border router translates between the two protocols so
+that any 6LoWPAN node can reach IPv4 hosts without running a dual stack.
 
-### Installing Jool
-Jool consists of two parts - Kernel modules, and User-space tools - both need to be installed to make use of the NAT64 Jool.
+Contiki-NG includes a built-in NAT64 gateway that runs inside the native
+border router. It requires no external software, kernel modules, or
+separate DNS servers — just pass `--nat64` when starting the border
+router.
```
joakimeriksson: Here is a good spot to add description of limitations and details of this implementation.

joakimeriksson (changes requested):
Looks good in general - I will test but adding some more details on docs (and given that my tests works) I think this is good to go.

joakimeriksson (approved):

arurke:
Cool addition!
Is it the intention to replace the jool-approach altogether (I am not up-to-date on its working state)? If yes, there is at least [one place](https://github.com/contiki-ng/contiki-ng/blob/c1e1b97a9821246445414dd161cce45542862527/doc/tutorials/LWM2M-and-IPSO-Objects.md?plain=1#L62) where Jool is still referenced. If no, how about retaining the jool-documentation in some way?

nvt:
Yes, the idea is to replace the other solutions with an easy-to-use service that runs in the border router.

Thanks for the pointer to the remaining jool reference. I can remove that and the tools/ip64 directory also.

PR #3138, opened 2026-04-22
Feature/nrf54l15

joakimeriksson:
This PR adds a working nRF54L15 application-core platform with:

  - Nordic nrf_802154 802.15.4 radio integration
  - board support for the nRF54L15 DK and Seeed XIAO nRF54L15
  - GPIO, LEDs, buttons, UART, reset, clock, rtimer, and watchdog stub support
  - GRTC-based system timing for the platform
  - RF switch support on the XIAO
  - OpenOCD flashing support for the XIAO
  - a radio test app and working rpl-udp example coverage

  At the network/radio level, the branch includes:

  - the initial nRF54L15 radio driver port
  - CSMA integration so radio ACK / no-ACK results feed the MAC correctly
  - XIAO-specific RF bring-up fixes
  - TX/RX stabilization work for the Nordic 802.15.4 stack
  - the latest low-power timer backend rework, replacing the fragile GRTC-based SL timer path with a Zephyr-style local timer backend in arch/cpu/nrf/nrf54l15/
    nrf_802154_platform_sl_lptimer.c
  - safer TX-timeout handling in arch/cpu/nrf/nrf54l15/nrf-ieee-driver-nrf54l15.c so unknown completion-path failures do not poison RPL metrics


Yagoor:
Great work, I am glad to see that you managed to integrate the official nrf 802154 driver from nordic. 
Is there a reason why you didn't use the https://github.com/nrfconnect/sdk-nrfxlib/tree/main/nrf_802154 as a submodule instead of copying the files? Do you have local changes?

joakimeriksson:
the reason was only for heavy instrumentation due to issues with the ack-timing, etc. But not that should be resolved and instrumentation no longer needed. Will switch to submodule.

joakimeriksson:
Needed some 5430 files also due to updated nrfx library.

arch/cpu/nrf/nrf54l15/Makefile.nrf54l15
```diff
@@ -0,0 +1,270 @@
...
+CFLAGS += -Wno-unused-variable
+CFLAGS += -Wno-unused-function
+CFLAGS += -Wno-unused-but-set-variable
+# Compatibility defines: nrfx v3 HAL uses TIMER_CC_NUM but nRF54L15 MDK
+# defines TIMER_CC_NUM_SIZE instead. Add aliases needed by nrfx drivers.
+CFLAGS += -DTIMER10_CC_NUM=TIMER10_CC_NUM_SIZE
+CFLAGS += -DTIMER00_CC_NUM=TIMER00_CC_NUM_SIZE
+CFLAGS += -DTIMER20_CC_NUM=TIMER20_CC_NUM_SIZE
+# Override nrf_802154 assert to use BKPT (triggers hardfault handler with PC dump)
+# instead of abort() which loops forever on bare-metal newlib-nano.
```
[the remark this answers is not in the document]
joakimeriksson: Should be fixed.

arch/platform/nrf/nrf54l15/xiao/README.md
```diff
@@ -0,0 +1,293 @@
...
+- UART console (115200 baud)
+- LEDs (user LED on P0.00)
+- Buttons (user button on P0.18)
+- OpenOCD flashing via CMSIS-DAP
+- hello-world example
+
+### ⏳ Pending Features:
+- 802.15.4 radio driver (stub in place)
+- IPv6 networking stack (disabled until radio works)
+- Low-power modes
```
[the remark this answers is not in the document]
joakimeriksson: Readme updated.

arch/platform/nrf/lpm.h
```diff
@@ -56,10 +62,13 @@ lpm_drop(void)
   abort = process_nevents();
   if(!abort) {
     ENERGEST_SWITCH(ENERGEST_TYPE_CPU, ENERGEST_TYPE_LPM);
-    __WFI();
+    /* NOP loop instead of WFI: after soft-reset, GRTC interrupts may not
+     * wake the CPU from WFI.  TODO: investigate WFI + GRTC after SREQ. */
+    for(volatile int _i = 0; _i < 1000; _i++) { __NOP(); }
     ENERGEST_SWITCH(ENERGEST_TYPE_LPM, ENERGEST_TYPE_CPU);
```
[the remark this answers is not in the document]
joakimeriksson: Fixed.

arch/cpu/arm/CMSIS
nfi, line 1: The CMSIS submodule is for all architectures and not only the `nrf` platform, as the commit message indicates.

I am assuming a newer version of CMSIS is needed for this platform? It would be good to include the CMSIS version in the commit message.

os/net/mac/csma/csma-output.c
```diff
@@ -208,6 +208,8 @@ send_one_packet(struct neighbor_queue *n, struct packet_queue *q)
       case RADIO_TX_OK:
         if(is_broadcast) {
           ret = MAC_TX_OK;
+        } else if(CSMA_TX_OK_ACKED) {
```
nfi: I think `CSMA_TX_OK_ACKED` for this configuration is a little confusing name, since it is so close to `RADIO_TX_OK` and `MAC_TX_OK`. We have `CSMA_CONF_SEND_SOFT_ACK` to configure whether CSMA should send soft ack. What do you think about something like `CSMA_CONF_CHECK_ACK`, `CSMA_CONF_SOFT_ACK_RX`, `CSMA_CONF_HANDLE_ACK_RX`, `CSMA_CONF_USE_RADIO_ACK`, or something like that? Just a suggestion, and I will not argue if you want to keep the current name.

os/lib/dbg-io/printf.c
```diff
@@ -39,8 +39,9 @@
 static strformat_result
 write_str(void *user_data, const char *data, unsigned int len)
 {
-  if(len > 0) {
-    dbg_send_bytes((unsigned char *)data, len);
+  unsigned int i;
+  for(i = 0; i < len; i++) {
+    dbg_putchar(data[i]);
   }
```
nfi, line 44: Why change this? Is it to print `\0` characters?

arch/cpu/nrf/arm/hardfault-handler.c
```diff
@@ -80,80 +188,62 @@ HardFault_process()
...
-    [22] = NULL,
-    [23] = NULL,
-    [24] = "The processor has made an unaligned memory access",
-    [25] = "The processor has executed an SDIV or UDIV instruction with a divisor of 0",
-  };
-
-  uint32_t cfsr = SCB->CFSR;
+  /* Immediately print "HF!" using the working uarte_write */
+  uarte_write('\n');
+  uarte_write('H'); uarte_write('F'); uarte_write('!'); uarte_write(' ');
```
nfi: I do not think all nRF platforms, such as nRF52840, use `uarte` for output? Perhaps this should be a configuration since `dbg_putchar()` can not be used safely from interrupts?

arch/cpu/nrf/nrfx_glue.h
```diff
@@ -158,6 +201,18 @@ _NVIC_ClearPendingIRQ(IRQn_Type irq_number)
...
+/*------------------------------------------------------------------------------ */
+
+/* Standard MIN/MAX macros needed by nrf_802154_core.c (normally provided by
+ * the Zephyr SDK sys/util.h; we supply them here for bare-metal builds). */
+#ifndef MAX
+#define MAX(a, b) (((a) > (b)) ? (a) : (b))
+#endif
+#ifndef MIN
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+#endif
```
nfi: These are already defined by `sys/cc.h`. Is it not possible to use those?

arch/cpu/nrf/Makefile.nrf5340_application
```diff
@@ -42,3 +42,8 @@ LDGENFLAGS += -x c -P -E
 include $(CONTIKI)/$(CONTIKI_NG_CM33_DIR)/Makefile.cm33
 
 include $(CONTIKI_CPU)/Makefile.nrf
+
+ifeq ($(TRUSTZONE_SECURE_BUILD),1)
+  CONTIKI_CPU_SOURCEFILES := $(filter-out hardfault-handler.c,$(CONTIKI_CPU_SOURCEFILES))
+  CONTIKI_SOURCEFILES := $(filter-out hardfault-handler.c,$(CONTIKI_SOURCEFILES))
+endif
```
nfi: Why exclude `hardfault-handler` from secure build?

arch/cpu/nrf/nrf54l15/nrf54l15-conf.h
```diff
@@ -0,0 +1,38 @@
...
+#define QUEUEBUF_CONF_NUM 4
+#endif
+
+/* nrf_802154 resolves ACK status in the radio driver. */
+#ifndef CSMA_CONF_TX_OK_ACKED
+#define CSMA_CONF_TX_OK_ACKED 1
+#endif
+
+#ifndef LOG_CONF_LEVEL_RADIO
+#define LOG_CONF_LEVEL_RADIO LOG_LEVEL_WARN
```
nfi: Should the radio logger be enabled by default?

arch/cpu/nrf/nrf54l15/rtimer-arch.c
```diff
@@ -0,0 +1,90 @@
...
+void
+rtimer_arch_schedule(rtimer_clock_t t)
+{
+  if(!rtimer_channel_active) {
+    return;
+  }
+
+  uint64_t target = (uint64_t)t;
+  uint64_t now;
+  nrfx_grtc_syscounter_get(&now);
```
nfi, line 89: Does this not assume 64-bit `rtimer_time_t`? I think it is set to 32-bit now?

arch/cpu/arm/CMSIS
joakimeriksson: Good catch. The CMSIS bump in this branch was actually a downgrade (5.9.0 → 5.8.0) there's no nRF54L15-specific reason to touch CMSIS. Dropped the commit; CMSIS now stays at the version develop already has (5.9.0).

os/net/mac/csma/csma-output.c
joakimeriksson: Renamed to CSMA_USE_RADIO_ACK (config knob CSMA_CONF_USE_RADIO_ACK). Reads more clearly as "trust the radio's own ACK verdict". Updated csma.h, csma-output.c, and the nrf54l15 default in nrf54l15-conf.h.

os/lib/dbg-io/printf.c
joakimeriksson: Yes, exactly — to allow embedded '\0' bytes (e.g., binary blobs through the formatter) and not silently truncate. The root cause is actually in nrf's dbg_send_bytes, which has while(s && *s != 0) even though len is provided. Reverted the printf.c change and fixed the bug locally in arch/cpu/nrf/os/dbg-arch.c instead narrower change, doesn't touch portable code.

arch/cpu/nrf/arm/hardfault-handler.c
joakimeriksson: Right, the bare uarte_write only exists on the boards using a uarte console — nRF52840/nRF5340 may not. Switched all four call sites in the HardFault path to dbg_putchar, which already abstracts both the uarte and TrustZone-secure (tz_api_println) routes.

arch/cpu/nrf/nrfx_glue.h
joakimeriksson: Good point — they were left over from when sys/cc.h wasn't being pulled in here. Removed the local defines and added #include "sys/cc.h" at the top so we use the existing ones.

arch/cpu/nrf/Makefile.nrf5340_application
joakimeriksson: Added a comment in the Makefile explaining: in TRUSTZONE_SECURE_BUILD, the secure firmware/SPM provides its own HardFault_Handler (and the related fault handlers), so we filter our shared hardfault-handler.c out to avoid duplicate-symbol link errors. The non-secure side keeps the regular handler.

arch/cpu/nrf/nrf54l15/nrf54l15-conf.h
joakimeriksson: Agreed, no reason to enable it by default. Lowered to LOG_LEVEL_NONE.

arch/cpu/nrf/nrf54l15/rtimer-arch.c
joakimeriksson: Yep — the GRTC syscounter is 64-bit, so the schedule arithmetic only works if rtimer_clock_t is 64-bit. Added #define RTIMER_CONF_CLOCK_SIZE 8 to nrf54l15-conf.h so the type matches the underlying counter.

arch/platform/nrf/nrf54l15/xiao/board-buttons.c
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+#include "contiki.h"
+#include "dev/button-hal.h"
+
+/* Stub implementations to avoid nrfx v2.x API dependency */
+#include <stddef.h>
+
+/* Provide empty button list */
+void *button_hal_buttons[] = { NULL };
```
nfi: Why use `void *` instead of the type (`button_hal_button_t *button_hal_buttons[] = { NULL };`)?

arch/cpu/nrf/dev/gpio-hal-arch.c
```diff
@@ -74,26 +137,102 @@ pin_event_handler(nrfx_gpiote_pin_t pin, nrf_gpiote_polarity_t action)
...
+    return;
+  }
+
+  tmp = cfg & GPIO_HAL_PIN_CFG_INT_MASK;
+  if(tmp == GPIO_HAL_PIN_CFG_INT_ENABLE) {
+    gpio_hal_arch_interrupt_enable_nrfx_v3(port, pin);
+  } else {
+    gpio_hal_arch_interrupt_disable_nrfx_v3(port, pin);
+  }
+#else
```
nfi: Perhaps comment to what this refers to (`#else /* NRF_API_VER_AT_LEAST(3, 2, 0) */`) ?

arch/cpu/nrf/dev/gpio-hal-arch.c
```diff
@@ -122,6 +261,7 @@ gpio_hal_arch_port_pin_cfg_set(gpio_hal_port_t port, gpio_hal_pin_t pin, gpio_ha
   } else if(tmp == GPIO_HAL_PIN_CFG_INT_ENABLE) {
     nrfx_gpiote_in_event_enable(pin_number, true);
   }
+#endif
```
nfi: Perhaps comment to what this refers to (`#endif /* NRF_API_VER_AT_LEAST(3, 2, 0) */`) ?

arch/cpu/nrf/nrf54l15/clock-arch.c
```diff
@@ -0,0 +1,330 @@
+/*
+ * Minimal GRTC-backed clock implementation for nRF54L15.
+ *
+ * NOTE: This ignores advanced low-power tuning and only provides the
+ *       tick functionality required for basic Contiki timers.
+ */
+
```
nfi, line 14: Missing copyright header

arch/cpu/nrf/nrf54l15/nrf-ieee-driver-stub.c
```diff
@@ -0,0 +1,122 @@
...
+  channel_clear,
+  receiving_packet,
+  pending_packet,
+  on,
+  off,
+  get_value,
+  set_value,
+  get_object,
+  set_object
+};
```
nfi: Is this file needed anymore? And if so, could not `os/dev/nullradio.c` be used instead?

arch/cpu/nrf/nrf54l15/nrf54l15-conf.h
```diff
@@ -0,0 +1,44 @@
+/*
+ * Minimal configuration header for the nRF54L15 Cortex-M33 application core.
+ *
+ * Mirrors the existing nRF52/nRF53 style so platform code can include a
+ * CPU-specific configuration header if needed. Extend as drivers mature.
+ */
+#ifndef NRF54L15_CONF_H_
```
nfi, line 14: Missing copyright header

arch/cpu/nrf/nrf54l15/nrf54l15-def.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * CPU capabilities description for the nRF54L15 application core.
+ * Keep the structure aligned with existing nRF ports.
+ */
+#ifndef NRF54L15_DEF_H_
```
nfi, line 12: Missing copyright header

arch/cpu/nrf/nrf54l15/nrf_802154_platform_assert.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Platform assert override for nrf_802154 on nRF54L15.
+ * Prints file:line via UART and resets via NVIC_SystemReset().
+ *
+ * IMPORTANT: We must NOT use BKPT on the XIAO nRF54L15 because the onboard
+ * CMSIS-DAP debugger keeps CoreDebug->DHCSR.C_DEBUGEN set.  When C_DEBUGEN
+ * is active, BKPT halts the CPU in Debug state instead of escalating to
+ * HardFault — the HardFault handler never fires, and the system hangs
+ * silently until the debugger chip issues an SREQ reset.
+ */
```
nfi, line 17: Missing copyright header

arch/cpu/nrf/nrf54l15/nrf_802154_platform_sl_lptimer.c
```diff
@@ -0,0 +1,856 @@
...
+      timer_compare_int_enable(ALARM_CC);
+      if(timer_event_check_cc(ALARM_CC) || ((force_isr_mask & FORCE_MASK_ALARM) != 0U)) {
+        NVIC_SetPendingIRQ(LP_TIMER_IRQn);
+      }
+    }
+  }
+
+  irq_unlock_local(primask);
+}
+nrf_802154_sl_lptimer_platform_result_t
```
nfi, line 593: Missing separator

arch/cpu/nrf/nrf54l15/rtimer-arch.c
```diff
@@ -0,0 +1,91 @@
+/*
+ * Minimal rtimer implementation backed by the nRF54L15 GRTC syscounter.
+ */
+
```
nfi, line 11: Missing copyright header

arch/platform/nrf/nrf54l15/xiao/xiao-nrf54l15-conf.h
```diff
@@ -0,0 +1,16 @@
+/*
+ * Placeholder configuration header for the XIAO nRF54L15 board.
+ * Extend with board-specific overrides as the port matures.
+ */
+#ifndef XIAO_NRF54L15_CONF_H_
```
nfi, line 12: Missing copyright header

arch/platform/nrf/nrf54l15/xiao/xiao-nrf54l15-def.h
```diff
@@ -0,0 +1,36 @@
+/*
+ * Board definition header for the Seeed Studio XIAO nRF54L15.
+ * Mirrors the structure used by other nRF boards.
+ */
+#ifndef XIAO_NRF54L15_DEF_H_
```
nfi, line 12: Missing copyright header

examples/platform-specific/nrf/radio-test/radio-test.c
```diff
@@ -0,0 +1,911 @@
...
+    state.tx_collision++;
+    break;
+  default:
+    state.tx_err++;
+    break;
+  }
+
+  process_poll(&radio_test_process);
+}
+static void
```
nfi, line 514: Missing separator

examples/platform-specific/nrf/radio-test/radio-test.c
```diff
@@ -0,0 +1,911 @@
...
+                     state.max_transmissions);
+
+  state.tx_started++;
+  state.last_tx_seq = hdr->seq;
+  state.pending_tx--;
+  state.tx_busy = true;
+
+  NETSTACK_MAC.send(tx_done, NULL);
+/*---------------------------------------------------------------------------*/
+}
```
nfi, line 607: Separator should be outside the function

examples/platform-specific/nrf/radio-test/radio-test.c
```diff
@@ -0,0 +1,911 @@
...
+    PT_EXIT(pt);
+  }
+
+  SHELL_OUTPUT(output,
+               "Usage: radio-test [status|status-brief|target|clear-target|start|run|stop|once|interval|len|txmax|channel|power|verbose|reset]\n");
+
+  PT_END(pt);
+/*---------------------------------------------------------------------------*/
+}
+
```
nfi: Separator should be between functions, not inside.

examples/platform-specific/nrf/radio-test/radio-test.c
```diff
@@ -0,0 +1,911 @@
...
+  if(!strcmp(args, "reset")) {
+    reset_stats();
+    nrf54l15_radio_debug_reset();
+    SHELL_OUTPUT(output, "Statistics reset\n");
+    PT_EXIT(pt);
+  }
+
+  SHELL_OUTPUT(output,
+               "Usage: radio-test [status|status-brief|target|clear-target|start|run|stop|once|interval|len|txmax|channel|power|verbose|reset]\n");
```
nfi: Long line. Perhaps break it up on multiple rows?

arch/cpu/nrf/nrf54l15/nrf-ieee-driver-stub.c
joakimeriksson: You're right — removed in 8b7101557. The stub was never compiled by any build path, and `os/dev/nullradio.c` (`nullradio_driver`) is the standard fallback for no-radio builds.

joakimeriksson:
Pushed 24f4f9478 — addresses the remaining nits in one commit: `button_hal_button_t *` in board-buttons.c, `/* NRFX_API_VER_AT_LEAST(3, 2, 0) */` on the long `#else`/`#endif` pair in gpio-hal-arch.c, the two missing separators (sl_lptimer.c:592, radio-test.c:513), wrapped the long Usage string, and dropped the stale stub paragraph in xiao/README.md.

Also force-pushed the branch to scrub `nrf-ieee-driver-stub.c` out of history entirely — it was never wired into any build path, so it only added review noise. `nullradio_driver` (`os/dev/nullradio.c`) is the proper fallback for no-radio builds anyway.

joakimeriksson:
Heads-up on a regression I hit while hardware-testing on an nRF5340 DK: this PR breaks the **nrf5340 TrustZone secure/normal-world examples**. Non-TZ nrf5340 (hello-world etc.) is unaffected, and the PR's actual target (nrf54l15) works.

Bisect points at the nrfx submodule bump (2.10.0 → 3.12.1) interacting with the existing `arch/cpu/nrf/nrf5340/tz-target-cfg.c`:
- on `c1e1b97a9` (develop): TZ works fully — secure boots, hands off to normal world, `[INFO: TZNormalWorld] Initialize TrustZone API: SUCCESS`, secure timer fires
- on `cfee2785d` (TZ build-fix + filter `hardfault-handler.c`): secure firmware hangs in the default `NMI_Handler` (`b .`)
- on `1e2010fa0` (restore TZ fault handlers): same root cause, but now resets in a loop (handlers reset instead of hang)
- on PR tip: same as 1e2010fa0

Root cause: the new nrfx's `system_nrf5340_application.c` SystemInit does additional peripheral access (FICR-NS RAM copy, errata 160 workaround writing to 0x50101110 / 0x502012A0 / 0x5000470C, etc.) before our `tz-target-cfg.c` configures the SPU. On nRF5340 the SPU's peripheral-access error routes to NMI by default, hence the silent hang/reset loop.

`-DNRF_SKIP_FICR_NS_COPY_TO_RAM=1` + `-DNRF53_ERRATA_160_ENABLE_WORKAROUND=0` under `TRUSTZONE_SECURE_BUILD=1` clear the original reset loop, but the firmware still doesn't reach `main()` — some later peripheral access is still tripped. A real fix needs either a custom SystemInit override (mirroring what the working nrf54l15 TZ port already does on `trustzone-port`) or a TF-M-style rewrite of tz-target-cfg.c that configures the SPU up front (my older nrf5340 TZ work in a separate tree did that, but on nrfx 2.10.0).

Either of those is a meaningful chunk of work and reviewer iteration, so I'd suggest landing this PR as-is for the nrf54l15 target and tracking **nrf5340 TZ adaptation to new nrfx as a follow-up**. Happy to open a separate issue/PR for that.

joakimeriksson:
Fixed in c9901bcfb — nrf5340 TZ now boots and runs on hardware.

The clean fix turned out to be: vendor the nrfx 2.10.0 application-core MDK (`system_nrf5340_application.c` + `gcc_startup_nrf5340_application.S`) under `arch/cpu/nrf/nrf5340/legacy-mdk/` and point `Makefile.nrf5340_application` at the local copies. The new 3.12.1 SystemInit's extra peripheral writes were the actual trigger; the 2.10.0 version doesn't do them and is the one that worked with our existing TZ setup. Driver/HAL code still comes from the new nrfx submodule, so nrf54l15 keeps the new nrfx it needs.

One small patch in the vendored file: `TRIMCNF[].ADDR` is `uint32_t` in nrfx 3.x headers (was `uint32_t *` in 2.10), so I dropped the cast on that comparison.

Verified on an nRF5340 DK:

```
[INFO: TZSecureWorld] Initializing TrustZone
[INFO: TZSecureWorld] Enabling fault handlers
[INFO: TZSecureWorld] Non-secure image has correct permissions
[INFO: TZSecureWorld] Preparing to jump to the normal world
n> [INFO: TZNormalWorld] Initialize TrustZone API: SUCCESS
[INFO: SecureProcess] Ready to run
n>   Iteration 1 at 1288
n>   Iteration 2 at 2568
...
```

Non-TZ nRF5340 (hello-world) still works unchanged.

nfi (approved):

PR #3140, opened 2026-05-05
cc1200: throttle pending_packet() to avoid SPI starvation of RX IRQ

joakimeriksson:
CSMA's RTIMER_BUSYWAIT_UNTIL after a unicast TX polls NETSTACK_RADIO.pending_packet() in a tight loop. Unthrottled, the rapid LOCK_SPI/single_read(NUM_RXBYTES)/RELEASE_SPI cycle starves the cc1200 RX IRQ chain that needs the same SPI bus to drain the incoming ACK from the radio's RX FIFO into rx_pkt[]. The net effect: the ACK is received over the air, but never delivered to MAC, so every unicast retransmits to the MAC retry cap.

300 us is the empirical minimum that resolves the race on the CC2538 SoC at 32 MHz (the threshold sits between 200 us and 300 us in sweep tests on Zolertia Firefly).

### DIFF
--- a/arch/dev/radio/cc1200/cc1200.c
+++ b/arch/dev/radio/cc1200/cc1200.c
@@ -92,6 +92,20 @@ static rtimer_clock_t sfd_timestamp = 0;
  */
 #define CC1200_WITH_TX_BUF (!MAC_CONF_WITH_TSCH)
 /*
+ * Minimum spacing between two SPI reads of the RX FIFO byte count in
+ * pending_packet(). Tight pollers (e.g. CSMA's RTIMER_BUSYWAIT_UNTIL after a
+ * unicast TX) would otherwise hammer the SPI bus and starve the RX IRQ chain
+ * that needs the same bus to drain the incoming ACK. 300 us is the empirical
+ * minimum that resolves the race on the CC2538 SoC at 32 MHz (the threshold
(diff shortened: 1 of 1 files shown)


arch/dev/radio/cc1200/cc1200.c
```diff
@@ -1087,6 +1087,18 @@ pending_packet(void)
...
+   * single_read cycle starves the RX IRQ chain that needs the same SPI
+   * bus to drain the incoming ACK frame from the radio's RX FIFO into
+   * rx_pkt[]. The net effect is that the ACK is received over the air
+   * but never delivered to MAC, so every unicast retransmits to the
+   * MAC retry cap. 300 us is the empirical minimum that resolves this
+   * on the CC2538 SoC at 32 MHz (the race threshold sits between 200
+   * and 300 us). Until DEBUG_LEVEL >= 3 was disabled this was masked
+   * by the printf above, which provided the same throttle as a side
+   * effect of UART blocking. */
+  clock_delay_usec(300);
```
nfi: Any reason to delay at all if `rx_pkt_len != 0` or `SPI_IS_LOCKED()` when this function is called? Also, would it not be better to delay before checking whether the radio has received a packet rather than delaying afterwards, to allow a little more time for a packet to be received?

nfi (approved):

nfi (approved):

Issue #2991, opened 2024-12-10
etimer bugs？


yanxxx08cn:
I encountered some problems when using etimer to drive tasks. I am not sure if it is a bug. The phenomenon is that after contiki-ng runs for a period of time, the etimer pointers of some random tasks will disappear from the timerlist, causing the corresponding tasks to not run. I did not call process_exit to exit the process. Is there anything else that may cause the process to exit?

alexrayne:
as i remember, last year was acepted optimisation of etimer, you can try take early source of etimer from history and check with it.
former etimer was a bit simpler.

yanxxx08cn:
> as i remember, last year was acepted optimisation of etimer, you can try take early source of etimer from history and check with it. former etimer was a bit simpler.



> as i remember, last year was acepted optimisation of etimer, you can try take early source of etimer from history and check with it. former etimer was a bit simpler.
I checked the code of etimer_process. When it happened before, I forgot to check whether the process in process_list still existed. At present, it is judged that after timer_expired in etimer_process, the process_post function is used to send the event, but the target task is not executed, resulting in the etimer_set in the target task unable to add its own etimer to timerlist, thus causing the task to fail to run. Now I am still simulating and waiting for the situation to reappear.

yanxxx08cn:
I checked the code of etimer_process. When it happened before, I forgot to check whether the process in process_list still existed. At present, it is judged that after timer_expired in etimer_process, the process_post function is used to send the event, but the target task is not executed, resulting in the etimer_set in the target task unable to add its own etimer to timerlist, thus causing the task to fail to run. Now I am still simulating and waiting for the situation to reappear.

yanxxx08cn:
> as i remember, last year was acepted optimisation of etimer, you can try take early source of etimer from history and check with it. former etimer was a bit simpler.

I checked the code of etimer_process. When it happened before, I forgot to check whether the process in process_list still existed. At present, it is judged that after timer_expired in etimer_process, the process_post function is used to send the event, but the target task is not executed, resulting in the etimer_set in the target task unable to add its own etimer to timerlist, thus causing the task to fail to run. Now I am still simulating and waiting for the situation to reappear.

fozc:
Hello, I am currently experiencing a very similar issue. Although there are many tasks in the system, one of them stops running after a while. This task is polled using an etimer event. I haven't had any problems with tasks triggered by other events so far, but I am encountering this issue with tasks driven by etimer in two different projects.

[...]

PR #3155, opened 2026-06-10
CC2538: Fix timer issues

kkrentz:
I encountered situations, where `clock_time()` jumped and `etimer_expired()` returned true right after having called `etimer_reset()`, similar to what was reported in #2991 and #186. There seem to be two problem areas:
* **clock.c:** Disabling the SysTick interrupt via `SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;` might happen only after that interrupt occurs. Besides, `clock.c` concurrently reads and writes `rt_ticks_startup`.
* **rtimer-arch.c**: When an ISR interrupts a process that is inside of `rtimer_arch_now()` and also calls this function, I suspect there is a chance for reading a garbage timer value.

### DIFF
--- a/arch/cpu/cc2538/clock.c
+++ b/arch/cpu/cc2538/clock.c
@@ -54,7 +54,7 @@
 #include "cpu.h"
 #include "dev/gptimer.h"
 #include "dev/sys-ctrl.h"
-
+#include "sys/atomic.h"
 #include "sys/energest.h"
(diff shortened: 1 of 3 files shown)


joakimeriksson:
Any plans for finalising this soon?

kkrentz:
I have not seen timer issues again for a while.

PR #3156, opened 2026-06-10
Allow for clock_time() to wrap around zero in timer.c and clock(-arch).c

kkrentz:
This PR replaces comparisons of clock_time_t values with wraparound-safe equivalents in `timer_expired()`, `timer_remaining`, and `clock_wait()`.


kkrentz:
I am closing this PR because I realized that the current version works fine, even when time values wrap around zero.

Issue #889, opened 2019-03-31
Orchestra improvement proposal


atiselsts:
The way I see it, currently Orchestra suffers from two problems (please correct if my understanding is not right):
1) With the non-storing mode rule (`unicast_per_neighbor_rpl_ns`) the scheduler wakes up the system on every slot. This may severely affect the system's energy consumption or even cause it to avoid deep LPM altogether.
2) With the storing mode rule (`unicast_per_neighbor_rpl_storing`) dedicated links are added only to the timesource node and nodes to which routes are present. This is more efficient. However, RPL does not add routes to peer nodes. These peer nodes also happen to be good candidates for alternative parents and are used as RPL probing targets. As there are no dedicated slots to these nodes, the probing packets are sent on shared slots and suffer from collisions. This may cause RPL parent probing to produce invalid results, long-term suboptimal routing and/or failure to switch to an alternative parent when the link to the main parent becomes bad.

It may be better to add links to all RPL candidate parents. (We should also add functionality for timing out inactive RPL parents.) This way only relevant links will be activated, and there will be less collisions in the setup phase.

This is relevant to the Orchestra receiver-based mode only.

simonduq:
I agree with your understanding of Orchestra's current main limitations (next to these is the fact we have just one choffset per slotframe, unlike MSF).

Your suggested fix would be an improvement but what would you think about taking this one step further, with fully dynamic allocation, not limited to any set (e.g. candidate parents) of neighbors? The idea is to add link whenever queueing a packet to a given neighbor. Only thing missing to make this happen is the ability to set the slot operation rtimer between slots, while still accounting for drift compensation.

Mano2Code:
Hi , 
Is there any workaround for this .. (With the non-storing mode rule (unicast_per_neighbor_rpl_ns) the scheduler wakes up the system on every slot)

Mano2Code:
> The way I see it, currently Orchestra suffers from two problems (please correct if my understanding is not right):
> 
> 1. With the non-storing mode rule (`unicast_per_neighbor_rpl_ns`) the scheduler wakes up the system on every slot. This may severely affect the system's energy consumption or even cause it to avoid deep LPM altogether.
[...]

PR #3160, opened 2026-06-16
Orchestra: Fix memory leak

### LINKED ISSUE #3158: Memory leak with TSCH/Orchestra
I tried TSCH/Orchestra and experienced very low packet delivery ratios. When enabling logging, I get many of such error messages:

```
33:37.536	ID:2	[ERR : TSCH Queue] ! add packet failed: 0 0x778832dfd3e0 20 (nil) (nil)
33:37.536	ID:2	[ERR : TSCH      ] ! can't send packet to 0003.0003.0003.0003 with seqno 117, queue 20/32 20/20
```

Apparently, the issue is specific to Orchestra and independent of which RPL implementation is used.

I created a minimal example on this branch: https://github.com/kkrentz/contiki-ng/commits/orchestra-leak/. The issue occurs half an hour into the simulation for the first time. It looks like a call to `tsch_queue_free_packet` is missing somewhere.

kkrentz:
This fixes #3158. Orchestra does not currently schedule slots for communicating with siblings, which is why frames for siblings remain in the TSCH queue forever.

### DIFF
--- a/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-ns.c
+++ b/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-ns.c
@@ -130,17 +130,6 @@ select_packet(uint16_t *slotframe, uint16_t *timeslot, uint16_t *channel_offset)
 }
 /*---------------------------------------------------------------------------*/
 static void
-new_time_source(const struct tsch_neighbor *old, const struct tsch_neighbor *new)
-{
-  if(new != old) {
-    const linkaddr_t *old_addr = tsch_queue_get_nbr_address(old);
-    const linkaddr_t *new_addr = tsch_queue_get_nbr_address(new);
-    remove_uc_link(old_addr);
-    add_uc_link(new_addr);
-  }
-}
-/*---------------------------------------------------------------------------*/
-static void
 init(uint16_t sf_handle)
 {
   uint16_t rx_timeslot;
@@ -159,7 +148,7 @@ init(uint16_t sf_handle)
 /*---------------------------------------------------------------------------*/
 struct orchestra_rule unicast_per_neighbor_rpl_ns = {
   init,
-  new_time_source,
+  NULL,
   select_packet,
   NULL,
   NULL,
--- a/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-storing.c
+++ b/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-storing.c
@@ -92,8 +92,8 @@ neighbor_has_uc_link(const linkaddr_t *linkaddr)
     return 0;
   }
 
-  if(!ORCHESTRA_UNICAST_SENDER_BASED) {
-    /* With the receiver-based Orchestra,
+  if(ORCHESTRA_UNICAST_SENDER_BASED) {
+    /* With the sender-based Orchestra,
      * all nodes have a link installed at their own timeslot */
     return 1;
   }


joakimeriksson:
Do you have any timeline for this? It is still tagged as draft? (we are planning next release fairly soon).

kkrentz:
The common issue of the predefined Orchestra rules is that they presume RPL to only route along the tree, i.e., to children or to the root. However, RPL may also send packages directly to neighboring nodes. I now added communication slots to avoid the resulting memory leak. The packet delivery ratios also jumped up and "ALICE" (classic RPL + Orchestra + Link-based rules) even achieves 99% (yet at a duty cycle above 10%). An alternative fix could be to add a configuration option to deny RPL to take shortcuts if Orchestra is enabled. This would probably reduce the duty cycle, but I do not know the RPL implementations well enough to do this quickly.

arurke:
Ref. discussions in #3158 , how about splitting this into changes for the non-storing rule, and the other rules? At least from my perspective, the proposed changes for the other rules need careful considerations as-is. The non-storing is more straightforward.

kkrentz:
I only retained the fix to the non-storing rules. In storing mode with link-based rules, the memory leak does not seem to occur. Maybe I messed up my configuration or just saw error messages that were caused by an overload situation. In storing mode with node-based rules, the memory leak is still there, but I have no simplistic simulation, yet.

kkrentz:
I added a commit, which seems to fix the memory leak in storing mode. I have a simplistic simulation of the issue on this [branch](https://github.com/kkrentz/contiki-ng/commits/orchestra-leak/).

arurke:
I looked at the issue in storing mode and reproduced it in your simulation. I do not quite agree with the solution in your commit.

I believe this issue was also introduced in #1520, specifically in this [commit](https://github.com/contiki-ng/contiki-ng/pull/1520/changes/e62ded185e7695bec65151513104d44f20e9e0d9), which changed when the unicast-rule is selected or not. Two important points first: 1) The unicast-rule does not add TX links to every neighbor (as the non-storing does), and 2) When select_packet() selects a link (timeslot, channel) for a packet, it does not check if that link is actually in the schedule (same as discussed in #3158). Thus, it is crucial that the unicast-rule does not select packets for all neigbors (only childs, parents, etc.), and rather let the common rule pick them up. Before #1520, this was true, perhaps by chance, as neighbor_has_uc_link() (used by select_packet()) would return false in these cases. In #1520, the function was updated to return true for all neighbors if Orchestra is configured in receiver-based mode. This makes the function more semantically correct since all nodes have a RX link. However, it removed the last guard against select_packet() selecting packets for which there is no TX link.

Below is a diff which reverts the offending commit in #1520. Feel free to test and see if it solves your issue.

An alternative fix might be to rather have the storing-mode rule also add TX links to all neighbors as in non-storing mode. @atiselsts discusses this in #889, and there seems to be consesus, but then it seems to have slipped the cracks as it does not appear in the list of improvements. Although, I do believe there are drawbacks, especially in dense networks, where the RPL probing increases contention on the unicast link.

```
diff --git a/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-storing.c b/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-storing.c
index ded3e6840..a30e75633 100644
--- a/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-storing.c
+++ b/os/services/orchestra/orchestra-rule-unicast-per-neighbor-rpl-storing.c
@@ -92,15 +92,9 @@ neighbor_has_uc_link(const linkaddr_t *linkaddr)
     return 0;
   }

-  if(!ORCHESTRA_UNICAST_SENDER_BASED) {
-    /* With the receiver-based Orchestra,
-     * all nodes have a link installed at their own timeslot */
-    return 1;
-  }
-
   if(linkaddr_cmp(&orchestra_parent_linkaddr, linkaddr)) {
-    /* The node is our parent */
-    return orchestra_parent_knows_us ? 1 : 0;
+    /* Our parent node. It has a link if it knows us, or is receiver-based */
+    return orchestra_parent_knows_us || !ORCHESTRA_UNICAST_SENDER_BASED ? 1 : 0;
   }

   if(nbr_table_get_from_lladdr(nbr_routes, (linkaddr_t *)linkaddr) != NULL) {
```

kkrentz:
Thank you @arurke for the fix. I do not understand all the intricacies, but the memory leak has gone.

arurke:
@kkrentz I think we should separate the two issues into two PRs. I am fairly confident in the storing-mode fix, not so much in the non-storing mode so that might need some more brewing time. Would you like me to open them?

kkrentz:
@arurke Go ahead please. I leave my fixes here for comparison.

arurke:
I opened #3212 and #3214 on the two issues and fixes. I would suggest you close this PR to avoid confusion as to what is our "up-to-date" proposed fixes.

PR #3162, opened 2026-06-21
serialradio: serial-radio example + native border-router integration

joakimeriksson:
Adds the serialradio example (radio control, RSSI/spectrum scan, sniffing, jamming, CBOR-over-SLIP+CRC16 protocol, Python tools + web UI) and teaches the native RPL border router to drive a serialradio node as its 802.15.4 radio — replacing the legacy ASCII !S/!R/?M slip-radio protocol with the CBOR protocol. The BR queries the radio's EUI-64 (GET_ADDR64), adopts it, sets PAN/channel, and enables a router mode (hardware address-filter + auto-ACK). 

Validated on two Zoul Fireflies (2.4 GHz): host pings a wireless node end-to-end over the serial radio, and the spectrum web UI streams live.

(Added read-the-docs fix also to get it green).


joakimeriksson:
Thanks for the review! Pushed a follow-up addressing all the feedback:

Copilot comments
- sniff now sends RX_ON/RX_OFF to the device (so it actually emits frames) and accepts start/stop synonyms
- Corrected the webserver help text (single HTTP port; WebSocket = port + 1)
- webserver stop no longer hangs — the HTTP server now uses a timeout and closes its socket
- Web UI sniff/jam buttons now send cli_command text (the only command the server executes)
- Escaped the hex-dump ASCII column to prevent HTML injection from packet payloads
- GET_ADDR64 now guards LINKADDR_SIZE == 8 so it always returns an 8-byte EUI-64

Also addressed
- Made the native border router's serialradio/CBOR support opt-in (BORDER_ROUTER_SERIAL_RADIO=1); the default keeps the legacy slip-radio protocol, so existing deployments and the tun-rpl-br test are unaffected
- Bumped the Read the Docs build image to ubuntu-22.04 (the pinned ubuntu-20.04 was retired by RTD)

Verified end-to-end on two Zoul Fireflies (2.4 GHz): host pings a node over the serial radio. Native border router builds clean in both default and BORDER_ROUTER_SERIAL_RADIO=1 configurations.

joakimeriksson:
replace the pixi config with uv for python version management and clean up the mix of python/c.

examples/serialradio/Makefile
```diff
@@ -0,0 +1,16 @@
+CONTIKI_PROJECT = serial-radio-main
+all: $(CONTIKI_PROJECT)
+
+# Project source files
+PROJECT_SOURCEFILES += serial-radio.c
+PROJECT_SOURCEFILES += sniffer-mac.c
+
+# Include CRC16 from Contiki-NG lib
+MODULES += os/lib
```
nfi: `os/lib` is always included.

examples/serialradio/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+# Include nullnet and CSMA (for TX support)
+MODULES += os/net/nullnet
```
nfi, line 13: Any reason not to use `MAKE_NET` instead?

examples/serialradio/Makefile
```diff
@@ -0,0 +1,16 @@
...
+MODULES += os/net/mac/csma
```
nfi: Any reason not to use `MAKE_MAC` instead?

examples/serialradio/serial-radio-main.c
```diff
@@ -0,0 +1,83 @@
...
+
+  /* The serial_radio_process handles everything */
+
+  while(1) {
+    PROCESS_WAIT_EVENT();
+  }
+
+  PROCESS_END();
+}
+/*---------------------------------------------------------------------------*/
```
nfi: What is the reason for having this in a separate file instead of in `serial-radio.c`?

examples/serialradio/Makefile
joakimeriksson: Right, removed the redundant MODULES += os/lib.

examples/serialradio/serial-radio-main.c
joakimeriksson: No good reason — it was just a thin entry point whose main_process printed a banner and then idled. I've folded the AUTOSTART_PROCESSES and the startup banner into serial-radio.c and removed serial-radio-main.c (and its idle process) entirely. CONTIKI_PROJECT is now serial-radio. Thanks!

examples/serialradio/Makefile
joakimeriksson: None — switched to MAKE_MAC = MAKE_MAC_CSMA. (The actual RX MAC is still the sniffer_mac_driver selected in project-conf.h; CSMA is pulled in for TX.)

examples/serialradio/Makefile
joakimeriksson: I tried MAKE_NET = MAKE_NET_NULLNET but it breaks the link: the host framing uses os/dev/slip.c, which stores received bytes in uip_buf/uip_len. Those come from the IPv6 stack that the default MAKE_NET compiles, so selecting nullnet drops slip.c and the uip_buf symbols (undefined slip_write, slip_process, uip_len, uip_aligned_buf).

For now I've kept nullnet as a module with the default net and added a comment explaining why. If you'd prefer the clean MAKE_NET = MAKE_NET_NULLNET form, I can refactor the SLIP path to use its own RX buffer instead of uip_buf — happy to do that in a follow-up if you think it's worth it. Let me know which you'd rather have.

nfi (approved):

PR #3177, opened 2026-06-29
Adding missing nrf52840 cc310 crypto hw acc support

highlunder:
### Contribution

An opt-in nRF52840 CryptoCell-310 backend for the `lib/ecc.h` ECC service, mirroring the existing `cc2538-ecc.c` precedent. Enabled with `ECC_HW_BACKEND=cc310`; default builds are unchanged. secp256r1 by default, Curve25519/Ed25519 behind `CC310_ECC_25519=1`.

The Nordic `nrf_crypto` frontend + cc310 backend sources come from the in-tree `nrf52-sdk` submodule. The closed CryptoCell runtime blob and CRYS headers are referenced from nRF5 SDK 17.1.0 (via `NRF_CC310_SDK`), not vendored - the build emits a clear error if the SDK isn't found. A patched `nrf_common.ld` (shipped in `arch/cpu/nrf52840/ld/`, used only for CC310 builds) supplies the `crypto_data` linker section `nrf_crypto_init()` walks, leaving the submodule untouched.

Adds a small `ecc_session_begin()/ecc_session_end()` ECC-service API so a caller can hold accelerator power and perform the one-time CC310 TRNG instantiation across a handshake; both are no-ops on the software backend.

### Testing

Builds cleanly for `nrf52840/dk` in both the CC310 and default software configurations. Validated on nRF52840-DK hardware: secp256r1 keygen/sign/verify/ECDH run correctly on CC310, ~19 ms per operation.

### Note for reviewers

The binary blob is licensed under Nordic's 5-clause SLA and cannot be redistributed in-tree - hence the reference-by-path. The `INTERRUPTS_ENABLED=1` / BASEPRI-isolated TRNG instantiation is required: the blob's TRNG wait depends on the CRYPTOCELL NVIC interrupt, and entropy collection must be isolated from the RTC tick (tested on the nrf hardware).

### DIFF
--- a/arch/cpu/nrf52840/dev/nrf52840-cc310-ecc.c
+++ b/arch/cpu/nrf52840/dev/nrf52840-cc310-ecc.c
@@ -1,0 +1,527 @@
+/*
+ * Copyright (c) 2026, RISE Research Institutes of Sweden.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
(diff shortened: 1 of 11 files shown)


os/services/ecc/Makefile.ecc
```diff
@@ -1,5 +1,15 @@
...
+# uECC_decompress for EC point decompression (nrf_crypto has none).
 MODULES += os/net/security/micro-ecc
 
+# ECC_HW_BACKEND=cc310 routes lib/ecc.h onto the nRF52840 CryptoCell-310
+# (arch/cpu/nrf52840/dev/nrf52840-cc310-ecc.c, wired via
+# arch/cpu/nrf52840/Makefile.cc310-crypto).  Exclude the software lib/ecc.h
+# backend so its ecc_* symbols don't clash with the hardware backend's.
+ifeq ($(ECC_HW_BACKEND),cc310)
+  CONTIKI_SOURCES_EXCLUDES += ecc.c
+endif
```
kkrentz, line 12: I would prefer to preserve the current behavior that including this module configures the software-based ECC implementation. If a utility from uECC is needed, one could include uECC directly. Besides, the current style is to configure the hardware-accelerated drivers by default. This might obviate the need for the ECC_HW_BACKEND flag.

arch/cpu/nrf52840/dev/nrf52840-cc310-ecc.c
```diff
@@ -0,0 +1,527 @@
...
+ecc_enable(const ecc_curve_t *c)
+{
+  if(map_curve(c) == NULL) {
+    LOG_WARN("curve '%s' not supported by CC310 backend\n",
+             c ? c->name : "(null)");
+    process_mutex_unlock(&mutex);
+    return -1;     /* caller falls back to software backend */
+  }
+  curve = c;
+  return 0;
```
kkrentz, line 185: To preserve the current semantic, one could start a session automatically if not done, yet. Once, `ecc_disable` is called, the session could automatically be terminated again. This would be similar to the "was_crypto_enabled" business for the CC2538.

arch/cpu/nrf52840/dev/nrf52840-cc310-ecc.c
```diff
@@ -0,0 +1,527 @@
...
+#define LOG_LEVEL LOG_LEVEL_NONE
+
+/*
+ * Raw key/element sizes for secp256r1.  nrf_crypto's "raw" encodings are
+ * big-endian and prefix-less, matching the lib/ecc.h convention exactly:
+ *   - private key / shared secret / hash : |CURVE|      = 32 bytes
+ *   - public key (X||Y, no 0x04)         : 2|CURVE|     = 64 bytes
+ *   - ECDSA signature (R||S)             : 2|CURVE|     = 64 bytes
+ */
+#define P256_ELEMENT_BYTES   32
```
kkrentz, line 114: This definition could be replaced with `ECC_CURVE_P_256_SIZE`.

PR #3180, opened 2026-07-01
link-stats: Use wraparound-safe comparison of clock_time_t values

kkrentz:
This PR replaces a comparison between `clock_time_t` values with the wraparound-safe `CLOCK_LT` macro. I also ran uncrustify over this snippet.


arurke (approved):

kkrentz:
I am closing this PR because I realized that the current version works fine, even when time values wrap around zero.

Issue #2402, opened 2023-02-08
Uip_len is incorrectly calculated which can lead to an Out of Bound Read


AmPaschal:
The Contiki-ng stack accepts IPv6 packets whose payload field is up to 40 bytes more than the actual length of the IPv6 payload. This leads to an out of bound read when processing IPv6 options.

Network Interface drivers deliver packets to the IPv6 stack by calling the tcpip_input() function. The packet is copied to UIP_IP_BUF while uip_len is set to the length of the packet.

When uip_process() is called in uip6.c, the function checks that the length of the packet (uip_len) is not less than the length returned by uipbuf_get_len_field().

```
if(uip_len < uipbuf_get_len_field(UIP_IP_BUF)) {
    UIP_STAT(++uip_stat.ip.drop);
    LOG_ERR("packet shorter than reported in IP header\n");
    goto drop;
  }
```

But in this function, uip_len corresponds to the IPv6 header length + the IPv6 payload length, while uipbuf_get_len_field() returns only the IPv6 payload length.

Hence, the check above passes even if the IPv6 payload length in the IPv6 header is up to 40 bytes greater than the actual payload length.

In the next line, uip_len is set again

`uip_len = uipbuf_get_len_field(UIP_IP_BUF) + UIP_IPH_LEN;`

At this point, the value of uip_len becomes greater than the actual length of the packet copied into UIP_IP_BUF

The next check verifies that the newly calculated length is not greater than the size of uip_buf

```
/* Check that the packet length is acceptable given our IP buffer size. */
  if(uip_len > sizeof(uip_buf)) {
    UIP_STAT(++uip_stat.ip.drop);
    LOG_WARN("dropping packet with length %d > %d\n",
	     (int)uip_len, (int)sizeof(uip_buf));
    goto drop;
  }
```
But uip_buf is statically allocated and sizeof(uip_buf) return a value that can be as high as the MTU or configured buffer size (1280)

The next function that is called tries to get the last header by iterating through the various extension headers in the packet and using the uip_len which was incorrectly set as the terminating condition. Hence, all derefencing and buffer reads in the function could be potentially read out of bounds.

PR #3193, opened 2026-07-13
uip6: fix uip_len validation to account for the IPv6 header size

94xhn:
## Summary

`uip_process()` validates that the genuinely-received `uip_len` is not
smaller than the packet size claimed by the IPv6 header before trusting
that header. The check omits `UIP_IPH_LEN` (the 40-byte fixed IPv6
header size):

```c
if(uip_len < uipbuf_get_len_field(UIP_IP_BUF)) {
  ...
  goto drop;
}
```

`uipbuf_get_len_field()` returns only the IPv6 "Payload Length" field
(bytes *after* the 40-byte header), while `uip_len` at this point is the
number of bytes actually received, header included. Two lines later the
same payload-length field is used again, this time correctly adjusted:

```c
uip_len = uipbuf_get_len_field(UIP_IP_BUF) + UIP_IPH_LEN;
```

Because the earlier guard is missing the same `+ UIP_IPH_LEN` term, a
header whose Payload-Length field overstates the true payload by up to
40 bytes passes the check, and `uip_len` is then set 40 bytes past what
was actually received. Downstream code (e.g. the extension-header walk
in `uipbuf_get_last_header()`) trusts `uip_len` as the boundary of valid
data, so up to 40 bytes of stale/adjacent buffer memory can be treated
as this packet's payload.

This addresses #2402.

## Fix

Compare the payload length against `uip_len - UIP_IPH_LEN`, avoiding an
addition that could overflow on 16-bit targets while checking the same
quantity used two lines later:

```c
if(uipbuf_get_len_field(UIP_IP_BUF) > uip_len - UIP_IPH_LEN) {
```

This is a minimal, one-line change with no behavioral effect other than
correctly rejecting packets that claim more payload than was actually
received. Legitimate packets (exact-length, and link-layer padded, as
already described by the function's own comment) are accepted
identically to before.

## Testing

- Built `examples/hello-world` for `TARGET=native` with the fix applied;
  compiles and links cleanly.
- Wrote and ran a standalone C harness modeling the exact guard/recompute
  sequence with the real struct layout and macro values
  (`UIP_IPH_LEN=40`, `UIP_BUFSIZE=1280`):
  - A forged header (genuinely received 100 bytes = 40-byte header + 60
    real payload bytes, but header claims Payload-Length=90) is accepted
    by the current check and produces `uip_len=130`, i.e. 30 bytes past
    what was actually received. With the fix, the same forged packet is
    dropped.
  - A second forged-header scenario at a smaller size reproduces the
    same pattern and is likewise dropped by the fix.
  - Two regression scenarios: an exact-length legitimate packet, and a
    link-layer-padded legitimate packet (the case the existing code
    comment explicitly describes) are accepted identically before and
    after the fix, confirming no regression.

## Disclosure

Generative AI (Claude, Anthropic) was used to help investigate this
issue and implement the fix. All changes were reviewed by the human
submitter (94xhn) before submission.

### DIFF
--- a/os/net/ipv6/uip6.c
+++ b/os/net/ipv6/uip6.c
@@ -1237,7 +1237,7 @@ uip_process(uint8_t flag)
    * packet header, the packet has been padded, and we set uip_len to
    * the correct value.
    */
-  if(uip_len < uipbuf_get_len_field(UIP_IP_BUF)) {
+  if(uipbuf_get_len_field(UIP_IP_BUF) > uip_len - UIP_IPH_LEN) {
     UIP_STAT(++uip_stat.ip.drop);
     LOG_ERR("packet shorter than reported in IP header\n");
     goto drop;


nvt:
This looks like a valid problem, but I think it would be better to remove the cast to uint16_t because it introduces the possibility of an overflow when uipbuf_get_len_field() returns large values.

94xhn:
Thanks, good point. Commit f92ebca3a removes the uint16_t cast and avoids the potentially overflowing addition altogether by comparing the payload length against uip_len - UIP_IPH_LEN. The earlier guard already ensures uip_len is at least UIP_IPH_LEN, so the subtraction is safe. I also updated the stale Fix example in the PR body. All 24 Actions checks and ReadTheDocs pass, including packet-parsing on Ubuntu and macOS.

nvt:
Looks good, but I think that the commits could be squashed now, with an update of the remaining commit message.

94xhn:
Squashed the branch to a single commit (c9f3320d1) and updated the commit message to describe both the received-length validation and the 16-bit overflow constraint. Thanks.

94xhn:
The squashed branch has one remaining CI failure: simulation-base/ubuntu-latest timed out in 21-cooja-rpl-tsch-security.csc at its built-in 660000 ms simulation limit. The same final source tree passed this job before the squash, and unrelated PR #3198 hit the same scenario and timeout today, so this appears to be a Cooja flake rather than a result of this one-line change.

I do not have permission to rerun upstream Actions. Could a maintainer rerun the failed job?

nvt (approved):

PR #3208, opened 2026-07-27
jsonparse: accept all JSON whitespace after literals

### LINKED ISSUE #3194: jsonparse: true/false/null literals fail to parse when followed by a newline
In `os/lib/json/jsonparse.c`, function `atomic()`, the `JSON_TYPE_NULL` / `JSON_TYPE_TRUE` / `JSON_TYPE_FALSE` branch, when scanning an atomic `boolean`/`null` literal, the value-end scan stops only at `' '`, `','`, `']'`, `'}'`:

https://github.com/contiki-ng/contiki-ng/blob/69a7e26ea4f9448f82bdd36e2ac409c51f3dd2d8/os/lib/json/jsonparse.c#L116

It does not treat `'\n'` (nor `'\r'`/`'\t'`) as a terminator. When a `boolean`/`null` is immediately followed by a newline, the scan over-reads into the next line, inflating vlen, so the subsequent `strncmp` validation fails and `jsonparse_next()` returns `JSON_TYPE_ERROR`. This makes valid, pretty-printed JSON fail whenever a `boolean`/`null` is the last value on its line.

Minimal repro: parsing
```json
{
  "flag": true
}
```
returns a syntax error, whereas `{ "flag": true }` and `{"flag": true,...}` parse fine. Numbers and strings are unaffected (the number scanner stops at any non-digit; strings terminate on the closing quote).

Suggested fix: include the remaining JSON whitespace characters in the terminator set, consistent with skip_ws():

```c
while ((c = state->json[state->pos]) && c != ' ' && c != '\n' && c != '\r' && c != '\t' && c != ',' && c != ']' && c != '}') {
    state->pos++;
}
```

giri256:
Fixes #3194.

## Summary

`jsonparse` stopped scanning `true`, `false`, and `null` only at a space or a structural delimiter. A literal followed by line feed was therefore read together with the next line and rejected. Carriage return and tab were also not handled consistently because `skip_ws()` recognized only space and line feed.

This adds one shared JSON-whitespace predicate for the four characters defined by RFC 8259: space, line feed, carriage return, and tab. Both atomic literal scanning and whitespace skipping now use it.

The new native unit test covers `true`, `false`, and `null` followed by each of those four whitespace characters and is registered in the existing `08-native-runs` suite.

## Advantages

- Valid pretty-printed JSON parses consistently across all legal JSON whitespace.
- Literal scanning and whitespace skipping use the same definition, preventing a terminator from being accepted by one parser stage but rejected by the next.
- Invalid literal suffixes remain rejected.

## Disadvantages

- The native test adds a small amount of CI build time.

## Validation

- Pre-fix native test failed on the line-feed case after the space case passed.
- Post-fix direct native run passed all 12 literal/whitespace combinations.
- `tests/08-native-runs/26-jsonparse.sh` completed with `TEST OK`.
- A clean rebuild and integrated run passed on top of `release-5.2`.
- `git diff --check` passed.

## AI-assisted contribution disclosure

AI assistance was used to inspect the parser flow, identify that the issue's suggested scanner-only change would still leave carriage return and tab unsupported by `skip_ws()`, and prepare the focused source and test changes. I manually reproduced the failure before editing, reviewed the final five-file diff, ran the repository-native test path, rebased onto the required `release-5.2` branch, and repeated a clean build and test run.

### DIFF
--- a/arch/dev/radio/cc1200/cc1200.c
+++ b/arch/dev/radio/cc1200/cc1200.c
@@ -95,15 +95,21 @@ static rtimer_clock_t sfd_timestamp = 0;
  * Minimum spacing between two SPI reads of the RX FIFO byte count in
  * pending_packet(). Tight pollers (e.g. CSMA's RTIMER_BUSYWAIT_UNTIL after a
  * unicast TX) would otherwise hammer the SPI bus and starve the RX IRQ chain
- * that needs the same bus to drain the incoming ACK. 300 us is the empirical
- * minimum that resolves the race on the CC2538 SoC at 32 MHz (the threshold
- * sits between 200 and 300 us on Zolertia Firefly); other host SoCs / SPI
- * clocks may need a different value.
+ * that needs the same bus to drain the incoming ACK. The threshold was
+ * originally measured between 200 and 300 us on the CC2538 SoC at 32 MHz, but
+ * 300 us is that measured minimum with no margin, and has since been observed
+ * to fail on Zolertia Firefly: nodes receive broadcast DIOs, but no unicast is
+ * ever acknowledged, so they never join a RPL DAG. Default to 500 us, which
+ * restores convergence on the same hardware. A larger window costs no
+ * latency: a delivered packet is reported from rx_pkt_len on every call, and
+ * only the fallback SPI peek at the FIFO byte count is rate-limited. Other
+ * host SoCs / SPI clocks may need a different value; override with
+ * CC1200_CONF_PENDING_POLL_THROTTLE_US.
  */
 #ifdef CC1200_CONF_PENDING_POLL_THROTTLE_US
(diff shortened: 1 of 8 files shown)


joakimeriksson:
Thanks for this, and sorry for the slow response.

I retargeted this from `release-5.2` to `develop`: the release branch gets deleted when 5.2 ships, which would have auto-closed your PR. Your branch was cut from `release-5.2`, so the diff temporarily shows four unrelated release commits — that clears by itself once `release-5.2` is merged back into `develop`, leaving just your commit. No action needed from you.

Verified locally on `develop`: cherry-picks cleanly, builds, and `26-jsonparse` reports `TEST OK`. Nice to see the new test — `jsonparse` had no coverage at all before this.

nvt:
Closing because the commit of this PR was integrated in PR #3246, which has now been merged.

PR #3209, opened 2026-07-27
jsonparse: accept JSON whitespace after literals

### LINKED ISSUE #3194: jsonparse: true/false/null literals fail to parse when followed by a newline

giri256:
## Summary

- accept all JSON whitespace characters after `true`, `false`, and `null`
- use the same whitespace predicate for literal scanning and `skip_ws()`
- add a native regression covering each literal before space, LF, CR, and tab

## Problem

`jsonparse_next()` only treated a plain space as whitespace while scanning JSON literals. Valid pretty-printed JSON such as a `true` value followed by a newline was therefore scanned past the literal and rejected.

## Validation

- reproduced the failure before the change; the LF case failed on the second assertion
- ran `tests/08-native-runs/26-jsonparse.sh`; all 12 literal/whitespace combinations pass
- confirmed the native runner is registered in the existing `08-native-runs` suite
- `git diff --check`

Fixes #3194.

AI assistance was used to inspect the parser, implement the focused change, and develop the regression test. The resulting diff and test output were reviewed locally.

### DIFF
--- a/arch/dev/radio/cc1200/cc1200.c
+++ b/arch/dev/radio/cc1200/cc1200.c
@@ -95,15 +95,21 @@ static rtimer_clock_t sfd_timestamp = 0;
  * Minimum spacing between two SPI reads of the RX FIFO byte count in
  * pending_packet(). Tight pollers (e.g. CSMA's RTIMER_BUSYWAIT_UNTIL after a
  * unicast TX) would otherwise hammer the SPI bus and starve the RX IRQ chain
- * that needs the same bus to drain the incoming ACK. 300 us is the empirical
- * minimum that resolves the race on the CC2538 SoC at 32 MHz (the threshold
- * sits between 200 and 300 us on Zolertia Firefly); other host SoCs / SPI
- * clocks may need a different value.
+ * that needs the same bus to drain the incoming ACK. The threshold was
+ * originally measured between 200 and 300 us on the CC2538 SoC at 32 MHz, but
+ * 300 us is that measured minimum with no margin, and has since been observed
+ * to fail on Zolertia Firefly: nodes receive broadcast DIOs, but no unicast is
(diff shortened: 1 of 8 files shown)


giri256:
Closing this as an accidental duplicate of #3208. #3208 is the canonical PR and targets the repository-required release-5.2 branch.

Issue #3190, opened 2026-07-10
Release 5.2 - task list before release


joakimeriksson:
# Preparations for Release 5.2
For a release, some additional testing beyond the regular automatic tests needs to be done, mainly to validate functionality on the different platforms. This extended testing phase plays a crucial role in ensuring that the software performs optimally across a variety of devices that Contiki-NG supports.

For each platform test at least the following examples:
- hello-world
- rpl-udp (the most basic communication)
- leds-example
- 6tisch/simple-node - TSCH on relevant platforms
- buttons/gpio

## Documentation
While testing platform - check documentation consistency

## Platforms - arch/platforms
- [x] zoul (Firefly or Remote) - also CC1200 (CSMA)
- [x] cc2538dk (not tested on hardware; CC2538 SoC covered by zoul — see comment)
- [x] cc26x0-cc13x0
- [x] simplelink
- [x] nrf52840
- [x] nrf
    - [x] nrf52840
    - [x] nrf5340
    - [x] nrf54L15
- [x] nrf-vpr (RISC-V on nrf54L15)
- [x] Tmote Sky
- [x]  msp-exp430fr5969
- [x] Gecko / EFR32
- [x] native (without communication)
- [x] Cooja (just to test that it works outside the test framework)

## Other modules
- [x] EDHOC (RFC 9528, new in 5.2)
- [x] CoAP
- [x] LwM2M
- [x] RPL-Border-router + slip-radio & serial-radio
- [x] MQTT
- [x] NAT64 (border-router + native)

joakimeriksson:
Release testing for 5.2 is now open!

- `develop/v5.2` tagged and branch `release-5.2` created at 7f844118d
- Per the development-cycle process: bug fixes and pre-release polishing go to `release-5.2`; normal development continues on `develop`

**Results so far** (macOS/arm64 host, arm-gcc 15.2.1 / Apple clang 21, on release-5.2 + #3205):

- Build matrix: 8 examples x 12 target/board combos all build (includes the new EDHOC client/server after #3204). MSP430 and nrf-vpr/RISC-V still need Docker/CI coverage.
- native: boots and runs (without communication) — ticked
- Cooja standalone on Java 25: interactive GUI session with the prepared 6tisch/simple-node TSCH simulation — load, in-GUI firmware compile, 28 simulated minutes of RPL+TSCH, radio logger with pcap export, clean exit — ticked
[...]

PR #3218, opened 2026-07-31
CoAP: fix stale example client address and a no-op test comparison

joakimeriksson:
Two small, independent CoAP fixes found while testing the CoAP row of #3190 on hardware. Neither is urgent — CoAP itself works, verified end-to-end on hardware (9/9 requests over a real 802.15.4 path). These are a test that can't fail and an example that can't connect.

### 1. `09-native-coap.sh` cannot fail

```sh
if [ $SUCCESS > 0 ]; then
```

That is a redirection, not a comparison: it runs `test $SUCCESS` with stdout sent to a file named `0`. `grep -c` always prints a number, so the operand is never empty and the branch is taken for a count of 0 exactly as for 1. `OKCOUNT` then always reaches `TESTCOUNT`, the only `exit 1` is unreachable, and there is no `set -e` — so the script always exits 0, even if `coap-client` is missing or the server never starts. It also leaves a stray empty file named `0` behind.

Fixed with `-gt`.

**Nothing was masked by this.** The two endpoints the test checks both answer 2.05 today, and the `Makefile.compile-test` rule around it still caught build failures, so the row had quietly become a build-only test for `coap-example-server`. The cost was lost signal, not a live defect.

It is also a regression rather than a day-one bug — the 2017 original (01d06c30d) used `if [ $SUCCESS == 1 ]`, which works; cbe7082c1 (2019-10-12) loosened it to "greater than zero" with the wrong operator. Dead since October 2019.

I swept the rest of the suite: this is the only occurrence. Every other test asserts on an exit status, where the mistake isn't expressible, and the Cooja `.csc` scripts are JavaScript, where `>` is a real comparison.

### 2. `coap-example-client` points at a server that cannot exist

`SERVER_EP` held `fe80::212:7402:0002:0202`, whose EUI-64 is
`00:12:74:02:00:02:02:02` — the Moteiv OUI `00:12:74`, i.e. a Tmote Sky DS2411
serial. The `FIXME` above it says "hard-coded for Cooja", and that was exactly
right: this is the address Cooja gives an **emulated Sky mote 2**.

That addressing is still live today. An emulated Sky mote 1 in current
Contiki-NG boots as:

```
[INFO: Main      ] Node ID: 1
[INFO: Main      ] Link-layer address: 0012.7401.0001.0101
[INFO: Main      ] Tentative link-local IPv6 address: fe80::212:7401:1:101
```

so mote 2 is `fe80::212:7402:2:202` — the hard-coded value. The same scheme is
asserted by `tests/17-tun-rpl-br/03-border-router-sky.csc`, which pings
`fd00::0212:7404:0004:0404` (mote 4).

**So why is it broken?** Because the server can no longer *be* a Sky mote:
`examples/coap/coap-example-server/Makefile` (and the client's) carry
`PLATFORMS_EXCLUDE = sky z1`, on code-size grounds. The address is therefore
unreachable in any buildable configuration — not because no mote can hold it,
but because no mote that can run `coap-example-server` can.

The mote type these examples *can* use is the native `cooja` one, and it
addresses differently: `arch/platform/cooja/platform.c:set_lladdr()` repeats
the mote ID across the link-layer address, so mote 2 is `0002.0002.0002.0002`,
link-local `fe80::202:2:2:2`. That is what this patch uses.

Verified in Cooja with two native motes: with the old address the client gets
no responses and times out; with the new one every request is answered
(`.well-known/core` returns the full link-format listing). Builds clean for
`native`, and `strings` confirms the new address is the one compiled in. The
`FIXME` stays — the address is still hard-coded, just no longer pointing at a
mote type these examples cannot be built for.

Nothing in the tree executes this example (compile-only in CI, and no
simulation uses it), which is why it went unnoticed. The test in item 1
exercises `coap-example-server` with libcoap as the client, not this example.

Targets `release-5.2`: both are small fixes to things that ship broken today.

### DIFF
--- a/examples/coap/coap-example-client/coap-example-client.c
+++ b/examples/coap/coap-example-client/coap-example-client.c
@@ -55,7 +55,9 @@
 #define LOG_LEVEL  LOG_LEVEL_APP
 
 /* FIXME: This server address is hard-coded for Cooja and link-local for unconnected border router. */
-#define SERVER_EP "coap://[fe80::212:7402:0002:0202]"
+/* Cooja mote 2: arch/platform/cooja/platform.c repeats the mote ID across the
+   link-layer address, giving 0002.0002.0002.0002 for mote 2. */
+#define SERVER_EP "coap://[fe80::202:2:2:2]"
 
 #define TOGGLE_INTERVAL 10
 
--- a/tests/08-native-runs/09-native-coap.sh
+++ b/tests/08-native-runs/09-native-coap.sh
@@ -27,7 +27,7 @@ for TARGET in .well-known/core test/push; do
   cat coap.log >> $BASENAME.log
   # Fetch coap status code (not $? because this is piped)
   SUCCESS=`grep -c '2.05' coap.log`
-  if [ $SUCCESS -gt 0 ]; then
+  if [ "${SUCCESS:-0}" -gt 0 ]; then
     printf "> OK\n"
     OKCOUNT+=1
   else
--- a/tests/08-native-runs/09-native-coap.sh
+++ b/tests/08-native-runs/09-native-coap.sh
@@ -27,7 +27,7 @@ for TARGET in .well-known/core test/push; do
   cat coap.log >> $BASENAME.log
   # Fetch coap status code (not $? because this is piped)
   SUCCESS=`grep -c '2.05' coap.log`
-  if [ $SUCCESS > 0 ]; then
+  if [ $SUCCESS -gt 0 ]; then
     printf "> OK\n"
     OKCOUNT+=1
   else


joakimeriksson:
Thanks — taking one of the two review comments, and explaining why not the other.

**Taken: the empty-operand guard on `09-native-coap.sh:30`.** Pushed as
596c1d3, `[ $SUCCESS -gt 0 ]` -> `[ "${SUCCESS:-0}" -gt 0 ]`.

To be precise about what it does and does not fix: this is noise rather than
a correctness hole, and it cannot fire today, since `tee` always creates
`coap.log`. Even with an empty operand the test still *fails*, it just prints
an error first:

```
  SUCCESS=""    without guard -> "unary operator expected", then FALSE
  SUCCESS=""    with guard    -> FALSE
  SUCCESS="0"   both          -> FALSE
  SUCCESS="1"   both          -> TRUE
```

Worth doing anyway: a comparison should be well-defined for every operand it
can receive, which is exactly the point of the preceding commit.

**Not taken: rewording the FIXME.** The suggestion is to change it from
"hard-coded for Cooja and link-local for unconnected border router" to
"hard-coded to Cooja mote 2", but both halves of the existing text are still
accurate. `fe80::202:2:2:2` *is* link-local, so that clause holds; and
"for Cooja" is correct without being narrower than the truth — the constraint
is the mote type's addressing scheme, not one specific mote ID. Naming mote 2
in the FIXME would make it more specific than the code warrants. The
derivation is spelled out in the PR description instead.

**One correction to the overview,** since it is the version left visible in
the thread. It says the old address "no longer matches how Contiki-NG assigns
Cooja mote addresses". That is not right, and I had the same thing wrong in
an earlier revision of this PR body before correcting it: `fe80::212:7402:2:202`
is still exactly how Cooja addresses an emulated **Sky** mote. Verified on
current `release-5.2` — an emulated Sky mote 1 boots as:

```
[INFO: Main      ] Node ID: 1
[INFO: Main      ] Link-layer address: 0012.7401.0001.0101
[INFO: Main      ] Tentative link-local IPv6 address: fe80::212:7401:1:101
```

so mote 2 is the hard-coded value, and `tests/17-tun-rpl-br/03-border-router-sky.csc`
asserts the same form for mote 4.

The actual reason the address is unreachable is narrower: both CoAP examples
carry `PLATFORMS_EXCLUDE = sky z1` (code size), so the server can never *be* a
Sky mote. The mote type these examples can use is the native `cooja` one, which
addresses differently — `arch/platform/cooja/platform.c:set_lladdr()` repeats
the mote ID across the link-layer address, giving `fe80::202:2:2:2` for mote 2.
The fix is unaffected either way; only the justification changed.

nvt (approved):

PR #3220, opened 2026-08-04
nrf: enable the instruction cache at boot on the nRF54L series

nvt:
The nRF54L series executes in place from wait-stated RRAM, with a standalone ICACHE peripheral in front of it that is disabled at reset. Nothing enabled it, so every application ran with the CPU stalling on raw RRAM code fetches.

Enable it in platform_init_stage_one(), invalidating first: the cache memory is undefined at power-on and stale after a warm reset over rewritten code, such as a firmware update. The code is gated on the NRF_ICACHE peripheral macro, which only the nRF54L series defines, and on a new NRF_CONF_ICACHE option (default on; set 0 to run uncached, e.g. for cycle-exact profiling of code-fetch behavior).

Measured on an nRF54L15 at 128 MHz (DWT cycle counter, deterministic across runs):

  trial-division prime count   10359465 -> 4879210 cycles   2.1x
  byte-wise checksum loop       3694455 ->  718717 cycles   5.1x

### DIFF
--- a/arch/cpu/nrf/nrf54l15/nrf54l15-conf.h
+++ b/arch/cpu/nrf/nrf54l15/nrf54l15-conf.h
@@ -43,6 +43,18 @@
 #define WATCHDOG_CONF_ENABLE 0
 #endif
 
+/*
+ * Enable the instruction cache at boot. The SoC executes from wait-stated
+ * RRAM, so code-fetch-bound work runs several times slower uncached.
+ * Set 0 to run uncached, e.g. for cycle-exact profiling of code-fetch
+ * behavior.
+ */
+#ifdef NRF_CONF_ICACHE
+#define NRF_ICACHE_ENABLED NRF_CONF_ICACHE
+#else
+#define NRF_ICACHE_ENABLED 1
+#endif
+
 /* Enable extended HardFault handler with register dump. */
 #ifndef NRF_CONF_HARDFAULT_HANDLER_EXTENDED
 #define NRF_CONF_HARDFAULT_HANDLER_EXTENDED 1
--- a/arch/cpu/nrf/nrf54l15/nrf54l15-conf.h
+++ b/arch/cpu/nrf/nrf54l15/nrf54l15-conf.h
@@ -43,6 +43,16 @@
 #define WATCHDOG_CONF_ENABLE 0
 #endif
 
+/*
+ * Enable the instruction cache at boot. The SoC executes from wait-stated
+ * RRAM, so code-fetch-bound work runs several times slower uncached.
+ * Set 0 to run uncached, e.g. for cycle-exact profiling of code-fetch
+ * behavior.
+ */
+#ifndef NRF_CONF_ICACHE_ENABLE
+#define NRF_CONF_ICACHE_ENABLE 1
+#endif
+
 /* Enable extended HardFault handler with register dump. */
 #ifndef NRF_CONF_HARDFAULT_HANDLER_EXTENDED
 #define NRF_CONF_HARDFAULT_HANDLER_EXTENDED 1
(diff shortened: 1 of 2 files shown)


arch/platform/nrf/platform.c
```diff
@@ -76,9 +76,26 @@ platform_init_board_stage_two(void)
...
+   * Enable the instruction cache, on SoCs that have the ICACHE peripheral
+   * and have not opted out (see NRF_CONF_ICACHE). Invalidate before
+   * enabling: ENABLE performs no invalidation of its own, and the cache
+   * memory is undefined at power-on and stale after a warm reset over
+   * rewritten code, such as a firmware update.
+   */
+  NRF_ICACHE->TASKS_INVALIDATECACHE = 1;
+  NRF_ICACHE->ENABLE = CACHE_ENABLE_ENABLE_Enabled;
+#endif /* defined(NRF_ICACHE) && NRF_ICACHE_ENABLED */
+}
```
nfi, line 110: Cache maintenance tasks are asynchronous; the peripheral exposes `STATUS.READY` (Busy = 1) for that reason. The code writes `TASKS_INVALIDATECACHE` and `ENABLE` back to back, so the cache can be enabled while the invalidate is still running. Please poll `STATUS` before enabling. The bundled nrfx HAL (`hal/nrf_cache.h`) covers all of this, so raw register writes should not be needed:

```c
#ifdef NRF_ICACHE
#include <hal/nrf_cache.h>
#endif
static void
icache_init(void)
{
#if defined(NRF_ICACHE) && NRF_CONF_ICACHE_ENABLE
  /* Invalidate and wait for the task to finish, then enable. */
  nrf_cache_invalidate(NRF_ICACHE);
#if NRF_CACHE_HAS_STATUS
  while(nrf_cache_busy_check(NRF_ICACHE)) {
  }
#endif
  nrf_cache_enable(NRF_ICACHE);
#endif /* defined(NRF_ICACHE) && NRF_CONF_ICACHE_ENABLE */
}
```

Regarding the commit message, I did not find anything supporting "undefined at power-on and stale after a warm reset over rewritten code". ENABLE resets to Disabled on every reset, and Nordic's own MDK boot code does not invalidate. Keeping the invalidate is fine, but the message should say that rather than present it as a requirement.

arch/cpu/nrf/nrf54l15/nrf54l15-conf.h
```diff
@@ -43,6 +43,18 @@
...
+ * Enable the instruction cache at boot. The SoC executes from wait-stated
+ * RRAM, so code-fetch-bound work runs several times slower uncached.
+ * Set 0 to run uncached, e.g. for cycle-exact profiling of code-fetch
+ * behavior.
+ */
+#ifdef NRF_CONF_ICACHE
+#define NRF_ICACHE_ENABLED NRF_CONF_ICACHE
+#else
+#define NRF_ICACHE_ENABLED 1
+#endif
```
nfi, line 54: `NRF_CONF_ICACHE` reads as if it configures the MDK's `NRF_ICACHE` peripheral symbol. I think something like `NRF_CONF_ICACHE_ENABLE` (matching `WATCHDOG_CONF_ENABLE`) would be clearer.

```c
#ifndef NRF_CONF_ICACHE_ENABLE
#define NRF_CONF_ICACHE_ENABLE 1
#endif
```

arch/platform/nrf/platform.c
```diff
@@ -76,9 +76,26 @@ platform_init_board_stage_two(void)
 {
 }
 /*---------------------------------------------------------------------------*/
```
nfi, line 90: `NRF_CONF_CACHE_ENABLED` is only defined in `nrf54l15-conf.h` but tested in the shared platform.c, so on nRF52/nRF5340 it is an undefined identifier under `#if`.

```c
#ifndef NRF_CONF_ICACHE_ENABLE
#define NRF_CONF_ICACHE_ENABLE 0
#endif
```

nfi (approved):

PR #3254, opened 2026-09-03
ip64: bound the DNS64 parser and correct its answer rewriting

nvt:
ip64_dns64_4to6() copied every answer record using the length field from the wire without checking it against the message, so a single UDP packet from port 53 can overrun ip64_packet_buffer on any build with NAT64 enabled, and the Type A branch of that copy is CVE-2020-24336. A second defect sits in the question loop in both directions, where the guard is skipped entirely for a zero-length label.

The first two commits come from #2260. They bound the walks against the length of the message that came in, which is not the room available for the one going out, so on their own they stop the module from translating answers at all. The third gives the function an output capacity instead, and fixes the 6to4 caller, which passed a buffer capacity where a message length belongs. The fourth applies the two functional fixes #2260 left as TODOs. The reply kept the question type that 6to4 had rewritten on the way out, and record names were not copied when a rewrite moved them.

The remaining commits came out of review. A parser that gave up partway still returned a length, and ip64_4to6() sent out the half-rewritten reply, so it now drops the packet instead. The walk also stopped at the answers, leaving an EDNS OPT record where it was while the answer before it grew. The bounds themselves were right, but formed a pointer past the end of the buffer before comparing it, so the walks work in offsets now, which also picks up a name that ends in a compression pointer after some labels.

Each fix has a vector in the native ip64 test that fails without it.

Acknowledgements: @Scepticz reported the answer-record copy and the question-loop guard bypass, wrote the first two commits, and flagged the two rewriting defects that the fourth one fixes.

Closes #2260.

### DIFF
--- a/os/services/ip64/ip64-dns64.c
+++ b/os/services/ip64/ip64-dns64.c
@@ -155,6 +155,11 @@ ip64_dns64_4to6(const uint8_t *ipv4data, int ipv4datalen,
   uint8_t *q;
   struct dns_hdr *hdr;
 
+  if(ipv4datalen < sizeof(struct dns_hdr)) {
+    LOG_WARN("ip64_dns64_4to6: packet ended while parsing header (in)\n");
+    return ipv6datalen;
+  }
+
   hdr = (struct dns_hdr *)ipv4data;
   LOG_DBG("dns64_4to6 id: %02x%02x\n", hdr->id[0], hdr->id[1]);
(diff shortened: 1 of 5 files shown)


nfi (approved):

Scepticz:
Hi @nvt,

thank you for merging the PR! Do you track these changes under a CVE? Is it CVE-2020-24336?

Best,
Tobi

PR #3258, opened 2026-09-04
roll-tm: bound the sequence lists in an ICMPv6 message

nvt:
A Trickle Multicast control message carries lists of sequence values, each list headed by a count of the 2-octet entries that follow it. The walk entered a list once the header began inside the payload and took the count from there, so a header near the end of a message, or a count larger than the message holds, moved the reads past uip_buf.

Take the end of the message once, require a whole header before reading one, and check that the entries a header announces arrived before walking them. The comparisons are distances, so a message too short for a header never enters the loop.

Thanks to Francesco La Spina and Stanislav Dashevskyi for reporting this problem.

### DIFF
--- a/os/net/ipv6/multicast/roll-tm.c
+++ b/os/net/ipv6/multicast/roll-tm.c
@@ -1092,6 +1092,7 @@ icmp_input()
   uint16_t *seq_ptr;
   uint16_t *end_ptr;
   uint16_t val;
+  const uint8_t *payload_end;
 
 #if UIP_CONF_IPV6_CHECKS
   if(!uip_is_addr_linklocal(&UIP_IP_BUF->srcipaddr)) {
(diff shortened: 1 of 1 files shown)


joakimeriksson:
The bounds checks look like a useful fix. Before approval, could we explicitly reject a truncated sequence-list header rather than let the loop terminate as though parsing completed successfully? As written, a partial header (including one after a valid list) falls through to the normal buffer-consistency processing.

Could we also add regression tests covering an incomplete header, a sequence count larger than the remaining payload, an exactly fitting list, and a valid list followed by a truncated header?

One follow-up observation: the new oversized-list error uses `goto drop`, which still updates Trickle timers/counters. That follows the existing error-handling pattern, so I would not block this bounds fix on changing it, but malformed-input handling may deserve a separate look.

nvt:
Fixed the truncated header. The walk advances by exactly the header plus `2 * seq_len`, so a well-formed message lands on the end of the payload, and anything left over is now dropped instead of falling through to the buffer-consistency pass. An empty payload still passes.

The four cases are the right ones, but they would not detect this bug as the suite stands. roll-tm is not in the `tests/20-packet-parsing` injector build, and that harness only catches a crash or a hang. Since `uip_buf` is a global, the over-read stays inside the data segment and the injector exits 0 with or without the fix. Making those packets meaningful takes a sanitizer build, which is a larger change than this one.

Agreed on `goto drop`. Going to `discard` would be worse here, since the loop may already have set `inconsistency` on a trickle parameter and only the `drop` block clears it, so a malformed packet would still reach a later timer reset. What that needs is for the error paths to clear the per-message state before they leave, which is a different fix from bounding the reads.

